PR #19, opened 2020-04-02
Add common files for instrumentors

### LINKED ISSUE #18: Add base files for all instrumentors
Add files like `tox.ini` and similar to test all the instrumentors.

ocelotl:
Adds common files for instrumentors. Fixes #18

### DIFF
--- a/instrumentors/tox.ini
+++ b/instrumentors/tox.ini
@@ -1,0 +1,41 @@
+[tox]
+skipsdist = True
+skip_missing_interpreters = True
+envlist =
+; Environments are organized by individual package, allowing
+; for specifying supported Python versions per package.
+
+  py3{4,5,6,7,8}-test-flask
+  pypy3-test-flask
+
+  py3{4,5,6,7,8}-test-django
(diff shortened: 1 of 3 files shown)


instrumentors/tox.ini
```diff
@@ -0,0 +1,41 @@
...
+  django: django/tests
+
+commands_pre =
+; Install without -e to test the actual installation
+  python -m pip install -U pip setuptools wheel
+;  test: pip install {toxinidir}/flask
+  test: pip install {toxinidir}/utils/opentelemetry-auto-instrumentation
+  test: pip install {toxinidir}/utils/opentelemetry-ext-testutil
+
+  flask: pip install {toxinidir}/flask
```
mauriciovasquezbernal, line 34: We don't have those packages now. Once they are present the path should be `instrumentors/`, right?

ocelotl: Yes, but I don't think we will at all, I'll remove them from the tox.ini file.

PR #29, opened 2020-04-08
Add eachdist and move tox to root folder

mauriciovasquezbernal:
Move tox and other configuration files to root folder. Add eachdist to make it
easier to handle operations with different packages.

The eachdist is taken from opentelemetry but slighty modified to avoid ignoring
a given path.


scripts/coverage.sh
```diff
@@ -0,0 +1,29 @@
...
+        --cov-append \
+        --cov-branch \
+        --cov-report='' \
+        ${1}
+}
+
+
+coverage erase
+
+cov opentelemetry-api
```
codeboten, line 18: this will likely need changing here

pyproject.toml
```diff
@@ -0,0 +1,11 @@
+[tool.black]
+line-length = 79
+exclude = '''
+(
+  /(  # generated files
+    docs/examples/opentelemetry-example-app/src/opentelemetry_example_app/grpc/gen|
+    ext/opentelemetry-ext-jaeger/src/opentelemetry/ext/jaeger/gen|
```
codeboten, line 7: probably dont need the docs and ext lines in here

codeboten (changes requested):
just a couple of changes to not look for directories that are specific to the opentelemetry-python repo. note: this could be a good opportunity to make the scripts a bit more generic

codeboten (approved):
LGTM

PR #39, opened 2020-04-15
Draft for the depedency handling on the opentelemetry-python repo

mauriciovasquezbernal:


toumorokoshi:
@codeboten @lzchen should we close this one out?

I am actually a little curious about how things like versioning are intended to work. Is there documentation around that? Or should that be proposed?

lzchen:
Closing this issue.
I believe @NathanielRN will include some instructions and information on how the two repos interact with each other based off PRs,releases,etc.

NathanielRN:
I have a plan based on what I brought up at the Python SIG meeting and will update the read me with a basic explanation, but the actual implementation is still up for discussion so any opinions are welcome.

Still think closing is the right call.

PR #47, opened 2020-10-17
Move core files to contrib

NathanielRN:
# Description

This PR moves over many files from the Core repo so that they can live in the Contrib repo with a basic CI setup.

## Type of change

# How Has This Been Tested?

Adding CI validates dependency code from Core repos is installed correctly.

# Checklist:


.github/workflows/test.yml
```diff
@@ -60,7 +54,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        tox-environment: [ "docker-tests", "lint", "docs" ]
```
NathanielRN: I removed the `docs` since I thought docs might only live on the Core repo

scripts/eachdist.py
```diff
@@ -263,7 +263,11 @@ def find_targets_unordered(rootpath):
     for subdir in rootpath.iterdir():
         if not subdir.is_dir():
             continue
-        if subdir.name.startswith(".") or subdir.name.startswith("venv"):
+        if (
+            subdir.name.startswith(".") or 
+            subdir.name.startswith("venv") or 
+            subdir.name.startswith("skiptargetinstall-")
```
NathanielRN, line 269: I thought this was a "consistent with what we've done before" way to get `eachdist.py` to not try to install repos it shouldn't. I think a better fall up would be for this for this script to read a "ignore" configuration from the `eachdist.ini` file.

.github/workflows/test.yml
NathanielRN: I removed the `docs` since I thought docs might only live on the Core repo

scripts/eachdist.py
NathanielRN, line 269: I thought this was a "consistent with what we've done before" way to get `eachdist.py` to not try to install repos it shouldn't. I think a better fall up would be for this for this script to read a "ignore" configuration from the `eachdist.ini` file.

tox.ini
```diff
@@ -313,49 +271,38 @@ commands_pre =
...
 
-  mypy: mypy --namespace-packages opentelemetry-api/src/opentelemetry/
-; For test code, we don't want to enforce the full mypy strictness
-  mypy: mypy --namespace-packages --config-file=mypy-relaxed.ini opentelemetry-api/tests/
-
-; Test that mypy can pick up typeinfo from an installed package (otherwise,
-; implicit Any due to unfollowed import would result).
-  mypyinstalled: mypy --namespace-packages opentelemetry-api/tests/mypysmoke.py --strict
-
 [testenv:lint]
```
NathanielRN, line 300: Even with these changes, the lint command fails because the code in the Core repo and the `skiptargetinstall-reference` code (the code that was originally donated to the project) has conflicting pylist issues. We have to either add these to an ignore which may require [a monkey patch due to an unresolved bug](https://github.com/PyCQA/pylint/issues/2686) or override the requirement and merge it in.

requirements.txt
```diff
@@ -0,0 +1,4 @@
+-e git+https://github.com/open-telemetry/opentelemetry-python.git@937863f8294d6347e131eaaedbc1b877598b13d9#egg=skiptargetinstall-opentelemetry-api&subdirectory=opentelemetry-api
```
NathanielRN, line 1: One thing that sucks about this solution is that pip must install __the whole repo___ when it just wants to install a target in a sub directory...

The best solution I’ve found is that we could instead [install from a zip file](https://stackoverflow.com/a/24811490). That would require us to update the release format on the Contrib package to release the sub targets (api, SDK, proto, instrumentation, test/utils) as zip files.

NathanielRN:
I mistakenly added the `test/util/` folder here when I planned to not include it in the contrib repo. I'll wait until some feedback to go back and do the commit merge needed to remove it (especially if that's what we want)

codeboten:
This PR can be closed as it was broken into more small ones

PR #49, opened 2020-10-20
Move instrumentation aiohttp client

NathanielRN:
# Description

Moves the `instrumentation/opentelemetry-instrumentation-aiohttp-client` from the core repo into the contrib repo.

The original code is being copied over from the Core repo here: https://github.com/open-telemetry/opentelemetry-python/tree/master/instrumentation/opentelemetry-instrumentation-aiohttp-client

# How Has This Been Tested?

For example the files in this PR have these import statements:

```
from opentelemetry.instrumentation.utils import (
    http_status_to_canonical_code,
    unwrap,
)
```

Both of which don't exist yet in this repo until I add them, in future PRs.

After the PRs with the packages are merged, I'll create a final PR to add tests. (And I count on it working based on my big #47 PR)

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-aiohttp-client/setup.cfg
+++ b/instrumentation/opentelemetry-instrumentation-aiohttp-client/setup.cfg
@@ -1,0 +1,55 @@
+# Copyright 2020, OpenTelemetry Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+[metadata]
+name = opentelemetry-instrumentation-aiohttp-client
+description = OpenTelemetry aiohttp client instrumentation
+long_description = file: README.rst
+long_description_content_type = text/x-rst
+author = OpenTelemetry Authors
+author_email = cncf-opentelemetry-contributors@lists.cncf.io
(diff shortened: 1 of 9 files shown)


bhautikpip (approved):
Looks good :)

codeboten (approved):
LGTM

instrumentation/opentelemetry-instrumentation-aiohttp-client/setup.cfg
```diff
@@ -0,0 +1,55 @@
...
+# limitations under the License.
+#
+[metadata]
+name = opentelemetry-instrumentation-aiohttp-client
+description = OpenTelemetry aiohttp client instrumentation
+long_description = file: README.rst
+long_description_content_type = text/x-rst
+author = OpenTelemetry Authors
+author_email = cncf-opentelemetry-contributors@lists.cncf.io
+url = https://github.com/open-telemetry/opentelemetry-python/instrumentation/opentelemetry-instrumentation-aiohttp-client
```
NathanielRN, line 22: @codeboten It just occurred to me that this url is out of date. Should we wait to merge in all the PRs and update all these links at the same time? Or would you prefer I make commits on all the PRs to update the links individually?

I would prefer the one commit to change all the links at the end but let me know what you think.

codeboten: Updating them all at once makes sense to me

instrumentation/opentelemetry-instrumentation-aiohttp-client/setup.cfg
```diff
@@ -0,0 +1,55 @@
...
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+
+[options]
+python_requires = >=3.5.3
+package_dir=
+    =src
+packages=find_namespace:
+install_requires =
+    opentelemetry-api == 0.15.dev0
```
codeboten, line 42: should this be updated to the latest released version here?

NathanielRN: Yes it should! This is to make sure we take in the latest from the Core master and not lose any work already merged into the Core repo.

The blame shows it was merged into Core on this commit:

image

PR #59, opened 2020-10-21
Move instrumentation grpc

NathanielRN:
# Description

Moves the `instrumentation/opentelemetry-instrumentation-grpc` from the core repo into the contrib repo.

The original code is being copied over from the Core repo here: https://github.com/open-telemetry/opentelemetry-python/tree/master/instrumentation/opentelemetry-instrumentation-grpc

# How Has This Been Tested?

# Checklist:


toumorokoshi:
It looks like there's some big changes coming in from the original repo: https://github.com/open-telemetry/opentelemetry-python/pull/1308. Should we wait for those to settle first?

NathanielRN:
@toumorokoshi Yes definitely agree with that. Thanks for your quick review on it! I think best case scenario we should iterate quickly on it.

PR #79, opened 2020-10-21
Move exporter opencensus

NathanielRN:
# Description

Moves the `exporter/opentelemetry-exporter-opencensus` from the core repo into the contrib repo.

The original code is being copied over from the Core repo here: https://github.com/open-telemetry/opentelemetry-python/tree/master/exporter/opentelemetry-exporter-opencensus

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/exporter/opentelemetry-exporter-opencensus/CHANGELOG.md
+++ b/exporter/opentelemetry-exporter-opencensus/CHANGELOG.md
@@ -1,0 +1,32 @@
+# Changelog
+
+## Unreleased
+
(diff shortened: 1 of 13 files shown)


codeboten (changes requested):
This exporter will remain in the core repo, see the list of exporters here https://github.com/open-telemetry/opentelemetry-specification/blob/b8b130d6d26e11d00c2b82be3cc3c9431da20fd0/spec-compliance-matrix.md#exporters

PR #85, opened 2020-10-21
Move docker tests

NathanielRN:
# Description

Moves the `tests/opentelemetry-docker-tests` from the core repo into the contrib repo.

The original code is being copied over from the Core repo here: https://github.com/open-telemetry/opentelemetry-python/tree/master/tests/opentelemetry-docker-tests

# How Has This Been Tested?

# Checklist:


tests/opentelemetry-docker-tests/tests/docker-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+      COLLECTOR_ZIPKIN_HTTP_PORT: "9411"
+    ports:
+        - "5775:5775/udp"
+        - "6831:6831/udp"
+        - "6832:6832/udp"
+        - "5778:5778"
+        - "16686:16686"
+        - "14268:14268"
+        - "9411:9411"
+  otopencensus:
```
codeboten, line 42: The otopencensus tests should remain in the core repo as the opencensus exporter will remain there.

tests/opentelemetry-docker-tests/tests/opencensus/test_opencensusexporter_functional.py
```diff
@@ -0,0 +1,60 @@
+# Copyright The OpenTelemetry Authors
```
codeboten, line 1: Same comment as above.

tests/w3c_tracecontext_validation_server.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python3
```
codeboten, line 1: This test should stay in the core repo as well, since this will be testing trace context propagation which is part of the API

codeboten (changes requested):
Just a few changes requested.

tests/opentelemetry-docker-tests/tests/docker-compose.yml
NathanielRN: Sounds good! I've removed all the OpenCensus Tests.

codeboten (approved):
Thanks for the update!

PR #150, opened 2020-11-05
Fixes and Improvements to botocore instrumentation

NathanielRN:
# Description

This PR aims to update the `botocore` instrumentation by doing the following:

Fixes:
- Removed code that modified `span.resource` since resource should be immutable
- Change `SpanKind` from `CONSUMER` to `CLIENT` because `botocore` api calls can be considered _remote outgoing_ and _synchronous requests_ requests as laid out in the [SpanKind specifications](https://github.com/open-telemetry/opentelemetry-specification/blob/06d16d7db88c3e266636f84887de61e3363b5429/specification/trace/api.md#spankind) - Important so that AWS services are treated as separate services in Tracing Backends. (See picture below!)
- Use `instrumentation.utils.unwrap` instead of creating the function again

Simplifications:
- Remove code that searches for nested attributes using `deep_getattr` and instead pulls values from properties of the `botocore.BaseClient` class since they should always be there
- Removed _params_ from attributes. There is [a PR](https://github.com/open-telemetry/opentelemetry-specification/pull/1095/files) to add more specific semantic conventions for which params should be trace in AWS SDKs in the specifications so the plan was to revisit when it gets merged

Improvements:
- More comprehensive logic to parse out `request_id` if it can be found in the headers instead of the usual location. Just like [.NET botocore instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/c8c73749e365f9bcddf9ffb45661f18c2effbee2/src/OpenTelemetry.Contrib.Extensions.AWSXRay/Instrumentation/AWSTracingPipelineHandler.cs#L240-L267)
- Add `aws.table_name` and `aws.queue_url` for DynamoDB and SQS specifically as traced properties (this will be added in the specifications)
- Format tests checking of attributes to look nice!

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) - Not all params are recorded now

# How Has This Been Tested?

- [x] With the change to `SpanKind.CLIENT`, S3 (and other services) are considered their own services

Before:
image

After:
image

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-boto/src/opentelemetry/instrumentation/boto/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-boto/src/opentelemetry/instrumentation/boto/__init__.py
@@ -55,6 +55,8 @@ from opentelemetry.trace import SpanKind, get_tracer
 
 logger = logging.getLogger(__name__)
 
+SERVICE_PARAMS_BLOCK_LIST = {"s3": ["params.Body"]}
+
 
 def _get_instance_region_name(instance):
     region = getattr(instance, "region", None)
@@ -203,7 +205,24 @@ class BotoInstrumentor(BaseInstrumentor):
         )
 
 
-def add_span_arg_tags(span, endpoint_name, args, args_names, args_traced):
+def flatten_dict(dict_, sep=".", prefix=""):
+    """
+    Returns a normalized dict of depth 1 with keys in order of embedding
+    """
+    # NOTE: This should probably be in `opentelemetry.instrumentation.utils`.
+    # adapted from https://stackoverflow.com/a/19647596
+    return (
+        {
+            prefix + sep + k if prefix else k: v
+            for kk, vv in dict_.items()
+            for k, v in flatten_dict(vv, sep, kk).items()
+        }
+        if isinstance(dict_, dict)
+        else {prefix: dict_}
+    )
+
+
+def add_span_arg_tags(span, aws_service, args, args_names, args_traced):
     def truncate_arg_value(value, max_len=1024):
         """Truncate values which are bytes and greater than `max_len`.
         Useful for parameters like "Body" in `put_object` operations.
@@ -213,34 +232,19 @@ def add_span_arg_tags(span, endpoint_name, args, args_names, args_traced):
 
         return value
 
-    def flatten_dict(dict_, sep=".", prefix=""):
-        """
-        Returns a normalized dict of depth 1 with keys in order of embedding
-        """
-        # adapted from https://stackoverflow.com/a/19647596
-        return (
-            {
-                prefix + sep + k if prefix else k: v
-                for kk, vv in dict_.items()
-                for k, v in flatten_dict(vv, sep, kk).items()
-            }
-            if isinstance(dict_, dict)
-            else {prefix: dict_}
-        )
-
     if not span.is_recording():
         return
 
-    if endpoint_name not in {"kms", "sts"}:
+    if aws_service not in {"kms", "sts"}:
         tags = dict(
             (name, value)
             for (name, value) in zip(args_names, args)
             if name in args_traced
         )
         tags = flatten_dict(tags)
-        for key, value in {
-            k: truncate_arg_value(v)
-            for k, v in tags.items()
-            if k not in {"s3": ["params.Body"]}.get(endpoint_name, [])
-        }.items():
-            span.set_attribute(key, value)
+
+        for param_key, value in tags.items():
+            if param_key in SERVICE_PARAMS_BLOCK_LIST.get(aws_service, {}):
+                continue
+
+            span.set_attribute(param_key, truncate_arg_value(value))
--- a/instrumentation/opentelemetry-instrumentation-boto/src/opentelemetry/instrumentation/boto/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-boto/src/opentelemetry/instrumentation/boto/__init__.py
@@ -235,6 +235,8 @@ def add_span_arg_tags(span, aws_service, args, args_names, args_traced):
     if not span.is_recording():
         return
 
+    # Do not trace `Key Management Service` or `Secure Token Service` API calls
+    # over concerns of security leaks.
     if aws_service not in {"kms", "sts"}:
         tags = dict(
             (name, value)
--- a/instrumentation/opentelemetry-instrumentation-boto/src/opentelemetry/instrumentation/boto/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-boto/src/opentelemetry/instrumentation/boto/__init__.py
@@ -48,8 +48,8 @@ from boto.connection import AWSAuthConnection, AWSQueryConnection
 from wrapt import wrap_function_wrapper
 
 from opentelemetry.instrumentation.boto.version import __version__
-from opentelemetry.instrumentation.botocore import add_span_arg_tags, unwrap
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
+from opentelemetry.instrumentation.utils import unwrap
 from opentelemetry.sdk.trace import Resource
 from opentelemetry.trace import SpanKind, get_tracer
 
@@ -201,3 +201,46 @@ class BotoInstrumentor(BaseInstrumentor):
             args,
             kwargs,
         )
+
+
+def add_span_arg_tags(span, endpoint_name, args, args_names, args_traced):
+    def truncate_arg_value(value, max_len=1024):
+        """Truncate values which are bytes and greater than `max_len`.
+        Useful for parameters like "Body" in `put_object` operations.
+        """
+        if isinstance(value, bytes) and len(value) > max_len:
+            return b"..."
+
+        return value
+
+    def flatten_dict(dict_, sep=".", prefix=""):
+        """
+        Returns a normalized dict of depth 1 with keys in order of embedding
+        """
+        # adapted from https://stackoverflow.com/a/19647596
+        return (
+            {
+                prefix + sep + k if prefix else k: v
+                for kk, vv in dict_.items()
+                for k, v in flatten_dict(vv, sep, kk).items()
+            }
+            if isinstance(dict_, dict)
+            else {prefix: dict_}
+        )
+
+    if not span.is_recording():
+        return
+
+    if endpoint_name not in {"kms", "sts"}:
+        tags = dict(
+            (name, value)
+            for (name, value) in zip(args_names, args)
+            if name in args_traced
+        )
+        tags = flatten_dict(tags)
+        for key, value in {
+            k: truncate_arg_value(v)
+            for k, v in tags.items()
+            if k not in {"s3": ["params.Body"]}.get(endpoint_name, [])
+        }.items():
+            span.set_attribute(key, value)
--- a/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
@@ -49,12 +49,14 @@ API
 import logging
 
 from botocore.client import BaseClient
+from botocore.exceptions import ClientError, ParamValidationError
 from wrapt import ObjectProxy, wrap_function_wrapper
 
 from opentelemetry import context as context_api
 from opentelemetry import propagators
 from opentelemetry.instrumentation.botocore.version import __version__
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
+from opentelemetry.instrumentation.utils import unwrap
 from opentelemetry.sdk.trace import Resource
 from opentelemetry.trace import SpanKind, get_tracer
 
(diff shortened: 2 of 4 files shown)


instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_instrumentation.py
```diff
@@ -144,29 +139,42 @@ def test_s3_put(self):
...
+                "aws.operation": "PutObject",
+                "aws.region": "us-west-2",
+                "aws.service": "s3",
+                "http.status_code": 200,
+            },
         )
-        self.assertEqual(spans[1].attributes["params.Key"], str(params["Key"]))
+        self.assertTrue("params.Body" not in spans[1].attributes.keys())
+        get_object_attributes = dict(spans[2].attributes)
         self.assertEqual(
```
NathanielRN, line 169: I'm not sure why `RequestId` doesn't return here, but I raise an issue to ask if my setup is wrong.

https://github.com/spulec/moto/issues/3442

NathanielRN: Got a reply confirming that the `RequestId` is missing because its simply not added to the S3 "mock" of the service in `moto`, so this is the best we can do right now :)

instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_instrumentation.py
```diff
@@ -66,20 +61,17 @@ def test_traced_client(self):
...
-                attributes={
-                    "endpoint": "ec2",
-                    "operation": "describeinstances",
-                }
-            ),
+            span.attributes,
+            {
+                "aws.operation": "DescribeInstances",
+                "aws.region": "us-west-2",
+                "aws.request_id": "fdcdcab1-ae5c-489e-9c33-4637c5dda355",
```
NathanielRN, line 69: Not sure why the IDs are sometimes consistent and sometimes not with `moto`.

Something the ID is consistent like this EC2 call:

```
{
    'Reservations': [],
    'ResponseMetadata': {'RequestId': 'fdcdcab1-ae5c-489e-9c33-4637c5dda355',
    'HTTPStatusCode': 200,
    'HTTPHeaders': {'server': 'amazon.com'},
    'RetryAttempts': 0}
}
```

and other times it's randomly generated like in this SQS call:
```
{
    'ResponseMetadata': {
        'RequestId': 'JWH6CYKFXYT2KXQSAYY50T6FK558XMQ7U9SOWLJ5YEQZOKULQ012',
        'HTTPStatusCode': 200,
        'HTTPHeaders': {'server': 'amazon.com',
        'x-amzn-requestid': 'JWH6CYKFXYT2KXQSAYY50T6FK558XMQ7U9SOWLJ5YEQZOKULQ012',
        'x-amz-crc32': '1416772655'},
        'RetryAttempts': 0
    }
}

```

instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_instrumentation.py
```diff
@@ -134,10 +127,12 @@ def test_s3_client(self):
...
         span = spans[2]
         self.assertEqual(
-            span.resource,
-            Resource(
-                attributes={"endpoint": "s3", "operation": "listobjects"}
-            ),
+            span.attributes,
+            {
+                "aws.operation": "ListObjects",
+                "aws.region": "us-west-2",
```
bhautikpip, line 133: Any reason on why don't we validate status code?

instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_instrumentation.py
```diff
@@ -148,29 +143,42 @@ def test_s3_put(self):
         location = {"LocationConstraint": "us-west-2"}
         s3.create_bucket(Bucket="mybucket", CreateBucketConfiguration=location)
         s3.put_object(**params)
+        s3.get_object(Bucket="mybucket", Key="foo")
```
bhautikpip, line 146: +1 for cleaning up these tests

instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_instrumentation.py
NathanielRN: Because this is a `ParamValidationError` the `botocore` library will never send the HTTP request and so there's no `http.status_code` to check!

But that's a good point! I can add a check to make sure this span was recorded as an error.

```
self.assertIs(
    span.status.status_code, trace_api.status.StatusCode.ERROR,
)
```

instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_instrumentation.py
NathanielRN: Thank you! 😄

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
```diff
@@ -104,137 +104,66 @@ def _instrument(self, **kwargs):
...
-                )
-
-            add_span_arg_tags(
-                span,
-                endpoint_name,
-                args,
-                ("action", "params", "path", "verb"),
-                {"params", "path", "verb"},
-            )
+            error = None
```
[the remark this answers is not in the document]
NathanielRN: Sure can do. I guess I thought it would be better to leave them down here though since they don't need to be in scope at the top?

instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_instrumentation.py
NathanielRN: Yes! I looked into it, and found this code used to generate the request ID in the `moto` repository:

https://github.com/spulec/moto/blob/9db62d32bf70e18f315305b7915d199e3ba1210a/moto/core/utils.py#L238-L244

It was [explained to me](https://github.com/spulec/moto/issues/3442) that this function is used by _some_ of the resources to attach a request ID.

Notably the `amzn_request_id()` decorator is used on SQS and DynamoDB and that's why their request IDs look like this 🙂 

And some services have hardcoded requests like EC2 `DescribeInstances` (https://github.com/spulec/moto/blob/50d71eccbee12ce5cf2c1ecf61bfcbd60317976d/moto/ec2/responses/instances.py#L479)

I'm not sure why it doesn't format the request IDs in the expected way (lower case joined with hyphens) because it [used to be correct](https://github.com/spulec/moto/blame/2bb3e841d12c5f614971492d241629794f35fe85/moto/core/utils.py#L68).

In that case, should I remove the regex check and just make it a check for the key? Let me know what you think!

instrumentation/opentelemetry-instrumentation-boto/src/opentelemetry/instrumentation/boto/__init__.py
```diff
@@ -204,3 +204,46 @@ def _patched_auth_request(self, original_func, instance, args, kwargs):
...
+        tags = dict(
+            (name, value)
+            for (name, value) in zip(args_names, args)
+            if name in args_traced
+        )
+        tags = flatten_dict(tags)
+        for key, value in {
+            k: truncate_arg_value(v)
+            for k, v in tags.items()
+            if k not in {"s3": ["params.Body"]}.get(endpoint_name, [])
```
owais, line 247: This is a bit cryptic and intention might not be very clear. Mind breaking it down or simplifying? Looks we really want to check something like `if endpoint_name == 's3' and k == 'params.Body'`? Personally, I'd write it out as a regular loop clearly self-documenting the intent and separating logic to multiple lines.

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
```diff
@@ -104,137 +104,66 @@ def _instrument(self, **kwargs):
...
         unwrap(BaseClient, "_make_api_call")
 
+    # pylint: disable=too-many-branches
     def _patched_api_call(self, original_func, instance, args, kwargs):
         if context_api.get_value("suppress_instrumentation"):
             return original_func(*args, **kwargs)
 
-        endpoint_name = deep_getattr(instance, "_endpoint._endpoint_prefix")
+        # pylint: disable=protected-access
+        service_name = instance._service_model.service_name
```
owais, line 113: Do we want a defensive check here in case `_service_model` is renamed or removed in future versions?

instrumentation/opentelemetry-instrumentation-boto/src/opentelemetry/instrumentation/boto/__init__.py
```diff
@@ -204,3 +204,46 @@ def _patched_auth_request(self, original_func, instance, args, kwargs):
             args,
             kwargs,
         )
+
+
+def add_span_arg_tags(span, endpoint_name, args, args_names, args_traced):
+    def truncate_arg_value(value, max_len=1024):
+        """Truncate values which are bytes and greater than `max_len`.
```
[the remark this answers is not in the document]
NathanielRN: Sorry this is my bad, I should've mentioned that I just copied and pasted this code into this file because the `botocore` instrumentation no longer needs it.

`boto` and `botocore` both used to share it, but now I removed it out of `botocore` and left it to only be used by `boto`.

[Removing the import in boto](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/150/files#diff-5696731c7113e6a7f09773bb9233a2227cf642a1cb6a0c5aa45c4fb330da735bL54)

[Removing the function from botocore because it doesn't need it](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/150/files#diff-05ef2bbeb79bb2d595bf8a3f7fa2d475d0f34f80804e5e1ba0d9a253d2c0e0e3L176)

I can see us refactoring the `boto` instrumentation in a follow up PR?

I could have made the diff simpler by leaving it in `botocore`, but `botocore` doesn't need it. Do you have an opinion on this?

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
NathanielRN: Do you mean something like the following:

```suggestion
        service_name = instance._service_model.service_name if instance._service_model is not None else None
```

I personally don't think we have to. I'm not saying this is a very convincing reason why, but even though botocore releases minor patches almost every day, this specific property [has not changed in 6 years](https://github.com/boto/botocore/blame/develop/botocore/client.py#L628).

The tests we have will make sure the name is being correctly set, and if `botocore` _does_ update this, we can either

-  lock the version (right now it's `botocore~=1.0`) or 
- add a defensive check.

instrumentation/opentelemetry-instrumentation-boto/src/opentelemetry/instrumentation/boto/__init__.py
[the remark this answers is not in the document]
NathanielRN: It makes sense to address your comments regardless I imagine, I can make it part of another PR if we want.

Created an issue here: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/186

instrumentation/opentelemetry-instrumentation-boto/src/opentelemetry/instrumentation/boto/__init__.py
```diff
@@ -204,3 +204,46 @@ def _patched_auth_request(self, original_func, instance, args, kwargs):
...
+        # adapted from https://stackoverflow.com/a/19647596
+        return (
+            {
+                prefix + sep + k if prefix else k: v
+                for kk, vv in dict_.items()
+                for k, v in flatten_dict(vv, sep, kk).items()
+            }
+            if isinstance(dict_, dict)
+            else {prefix: dict_}
+        )
```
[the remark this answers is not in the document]
NathanielRN: This should probably be moved into `opentelemetry.instrumentation.utils` anyways. I'll keep the Stack Overflow comment and add a comment about it being better suited as a util function for now.

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
owais: I meant using `getattr` instead of directly accessing the attribute to avoid possible AttributeError but if you think it's not a concern with boto 1.0 then I'm fine with it.

instrumentation/opentelemetry-instrumentation-boto/src/opentelemetry/instrumentation/boto/__init__.py
NathanielRN: I really agree with you, thanks for the tip! I've updated it to look cleaner, please let me know what you think!

instrumentation/opentelemetry-instrumentation-boto/src/opentelemetry/instrumentation/boto/__init__.py
```diff
@@ -204,3 +204,46 @@ def _patched_auth_request(self, original_func, instance, args, kwargs):
...
+
+    if not span.is_recording():
+        return
+
+    if endpoint_name not in {"kms", "sts"}:
```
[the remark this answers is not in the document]
NathanielRN: Sorry I just copied this code because I didn't want to change the `boto` functionality while updating `botocore`. I can only guess that it's because these are the `Key Management Service` and `Secure Token Service` services so maybe their API trace calls should not be traced?

I added a comment mentioning that but I can't guess pass that. It doesn't look like I the commit history says who added these in the first place either.

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
NathanielRN: Sounds good! I won't add it then if everyone's alright with it.

bhautikpip (approved):
LGTM :)

codeboten (approved):

Issue #134, opened 2020-11-03
Add hook to set custom span name in flask instrumentation


lzchen:
Similar to [wsgi](https://github.com/open-telemetry/opentelemetry-python/blob/master/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py#L210), accept a callback function in `FlaskInstrumentor().instrument` to be able to pass custom span name.

The [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#name) say "MAY" for this feature, but we should be consistent across our instrumentations.

Part of https://github.com/open-telemetry/opentelemetry-python/issues/1297

PR #152, opened 2020-11-06
Add hook to set custom span name in flask instrumentation

lonewolf3739:
Addresses #134


instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -75,6 +75,15 @@ def get_excluded_urls():
 _excluded_urls = get_excluded_urls()
 
 
+def get_default_span_name():
+    span_name = None
```
lzchen, line 79: Nit: span_name = ""

lzchen:
CHANGELOG entry?

lzchen (approved):
Nice! Just a couple of comments.

lonewolf3739:
> CHANGELOG entry?

I missed that, added now.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -105,43 +114,40 @@ def _start_response(status, response_headers, *args, **kwargs):
...
-            attributes["http.route"] = flask.request.url_rule.rule
-        for key, value in attributes.items():
-            span.set_attribute(key, value)
-
-    activation = tracer.use_span(span, end_on_exit=True)
-    activation.__enter__()
-    environ[_ENVIRON_ACTIVATION_KEY] = activation
-    environ[_ENVIRON_SPAN_KEY] = span
-    environ[_ENVIRON_TOKEN] = token
+        environ = flask.request.environ
```
lzchen, line 122: I think we should pass in `environ` as an argument to the span callback, so users can change the span name based off these parameters.

lzchen (changes requested):
Just a comment.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
lonewolf3739: Yes, I thought about it. The request context `flask.request` is accessible for users just like how we are using it in instrumentation. Passing `environ` wouldn't really give any extra benefit. What do you think?

lzchen: Ahh that is true. Yeah maybe we don't need it in the params then.

lzchen (approved):

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -188,14 +196,18 @@ def _instrument(self, **kwargs):
         self._original_flask = flask.Flask
```
lzchen, line 196: What if user instruments with `instrument()`? We should provide this functionality for that too.

lzchen (changes requested):
a comment

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
lonewolf3739: The current implementation is using `get_default_span_name`. I am wondering how do we hook the name call back in this case?

lzchen: Similar to [requests](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/158/files), use a kwarg.

lonewolf3739: I added it. One thing is we are patching the `flask.Flask` to get `instrument()` to work but if an import of Flask is made before the `FlaskInstrumentor().instrument()` then it doesn't do anything. I don't think if this is documented somewhere.

lzchen: Yes this is known behaviour.

lzchen (approved):

Issue #159, opened 2020-10-28
Make sure database instrumentations follow semantic conventions


lzchen:
From [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md).

lonewolf3739:
Hi @lzchen , I tried to address this [here](https://github.com/open-telemetry/opentelemetry-python/pull/1354).

@toumorokoshi left some comments _I think there's still some reconciliation that needs to happen on the specification and the fields emitted, especially on the MongoDB side_.

Some db specific call level attributes are added which are not mentioned in spec. What are you thoughts on this?

toumorokoshi:
Hey @lonewolf3739! Sorry about the hassle, but would you mind sending your PR again? I think the libraries moved to this repo while this PR was still in discussion.

Regarding:

>  I think there's still some reconciliation that needs to happen on the specification and the fields emitted, especially on the MongoDB side.

[Here's a spec ticket](https://github.com/open-telemetry/opentelemetry-specification/issues/1218) to start a discussion. I think some of this may be follow-up work as well, so the PR as-is would be great to merge in.

lonewolf3739:
@toumorokoshi I don't mind opening PR here again. I just wanted to see if others can chime in and share their opinions.

lzchen:
@lonewolf3739 

> I don't mind opening PR here again. I just wanted to see if others can chime in and share their opinions.

The extra attributes not defined in the specs can be left until the specs are finalized.

awssandra:
Are there any more remaining?

lonewolf3739:
@awssandra We haven't made any changes since last time. I think there should be some gaps between the specification and current implementation.

srikanthccv:
I will create a new issue with the same title and epic containing all the db instrumentations.

PR #175, opened 2020-11-10
Update database instrumentations to follow semantic conventions

lonewolf3739:
Addresses #159

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-pymemcache/src/opentelemetry/instrumentation/pymemcache/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-pymemcache/src/opentelemetry/instrumentation/pymemcache/__init__.py
@@ -55,10 +55,11 @@ logger = logging.getLogger(__name__)
 # https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/span-general.md#general-network-connection-attributes
 _HOST = "net.peer.name"
 _PORT = "net.peer.port"
+_TRANSPORT_PROTOCOL = "net.transport"
 # Database semantic conventions here:
 # https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md
-_DB = "db.type"
-_URL = "db.url"
+_DB = "db.system"
+_URL = "db.connection_string"
 
 _DEFAULT_SERVICE = "memcached"
 _RAWCMD = "db.statement"
@@ -115,7 +116,7 @@ def _with_tracer_wrapper(func):
 @_with_tracer_wrapper
 def _wrap_cmd(tracer, cmd, wrapped, instance, args, kwargs):
     with tracer.start_as_current_span(
-        _CMD, kind=SpanKind.INTERNAL, attributes={}
+        _CMD, kind=SpanKind.CLIENT, attributes={}
     ) as span:
         try:
             if span.is_recording():
@@ -174,8 +175,11 @@ def _get_address_attributes(instance):
             address_attributes[_HOST] = host
             address_attributes[_PORT] = port
             address_attributes[_URL] = "memcached://{}:{}".format(host, port)
+            address_attributes[_TRANSPORT_PROTOCOL] = "IP.TCP"
         elif isinstance(instance.server, str):
             address_attributes[_URL] = "memcached://{}".format(instance.server)
+            address_attributes[_HOST] = instance.server
+            address_attributes[_TRANSPORT_PROTOCOL] = "Unix"
 
     return address_attributes
 
--- a/instrumentation/opentelemetry-instrumentation-pymemcache/src/opentelemetry/instrumentation/pymemcache/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-pymemcache/src/opentelemetry/instrumentation/pymemcache/__init__.py
@@ -59,7 +59,6 @@ _TRANSPORT_PROTOCOL = "net.transport"
 # Database semantic conventions here:
 # https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md
 _DB = "db.system"
-_URL = "db.connection_string"
 
 _DEFAULT_SERVICE = "memcached"
 _RAWCMD = "db.statement"
@@ -174,10 +173,8 @@ def _get_address_attributes(instance):
             host, port = instance.server
             address_attributes[_HOST] = host
             address_attributes[_PORT] = port
-            address_attributes[_URL] = "memcached://{}:{}".format(host, port)
             address_attributes[_TRANSPORT_PROTOCOL] = "IP.TCP"
         elif isinstance(instance.server, str):
-            address_attributes[_URL] = "memcached://{}".format(instance.server)
             address_attributes[_HOST] = instance.server
             address_attributes[_TRANSPORT_PROTOCOL] = "Unix"
 
--- a/instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
@@ -73,21 +73,23 @@ class CommandTracer(monitoring.CommandListener):
             span = self._tracer.start_span(name, kind=SpanKind.CLIENT)
             if span.is_recording():
                 span.set_attribute("component", DATABASE_TYPE)
-                span.set_attribute("db.type", DATABASE_TYPE)
-                span.set_attribute("db.instance", event.database_name)
+                span.set_attribute("db.system", DATABASE_TYPE)
+                span.set_attribute("db.name", event.database_name)
                 span.set_attribute("db.statement", statement)
                 if event.connection_id is not None:
                     span.set_attribute("net.peer.name", event.connection_id[0])
                     span.set_attribute("net.peer.port", event.connection_id[1])
 
                 # pymongo specific, not specified by spec
-                span.set_attribute("db.mongo.operation_id", event.operation_id)
-                span.set_attribute("db.mongo.request_id", event.request_id)
+                span.set_attribute(
+                    "db.mongodb.operation_id", event.operation_id
+                )
+                span.set_attribute("db.mongodb.request_id", event.request_id)
 
                 for attr in COMMAND_ATTRIBUTES:
                     _attr = event.command.get(attr)
                     if _attr is not None:
-                        span.set_attribute("db.mongo." + attr, str(_attr))
+                        span.set_attribute("db.mongodb." + attr, str(_attr))
 
             # Add Span to dictionary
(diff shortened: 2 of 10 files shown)


toumorokoshi (changes requested):
code looks good! could you add changelog entries for each the instrumentations? and we should be good to go.

lonewolf3739:
@toumorokoshi I will add the entries. General question for my reference so that I don't miss it in future, What kind of changes require a CHANGELOG entry? Is it all or is it some based on the chage? Thanks !

toumorokoshi:
> What kind of changes require a CHANGELOG entry?

Pretty much every change that affects user behavior. Something like span attributes changing could be pretty impactful since people author queries and aggregations based on values in specific attributes.

instrumentation/opentelemetry-instrumentation-pymemcache/src/opentelemetry/instrumentation/pymemcache/__init__.py
```diff
@@ -115,7 +115,7 @@ def wrapper(wrapped, instance, args, kwargs):
 @_with_tracer_wrapper
 def _wrap_cmd(tracer, cmd, wrapped, instance, args, kwargs):
     with tracer.start_as_current_span(
-        _CMD, kind=SpanKind.INTERNAL, attributes={}
+        _CMD, kind=SpanKind.CLIENT, attributes={}
```
lzchen, line 118: I believe the span name should be a low-cardinality value representing the executed statement.

lzchen: I would also add `db.name` as a span attribute to be consistent with the other instrumentations.

instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
```diff
@@ -73,21 +73,23 @@ def started(self, event: monitoring.CommandStartedEvent):
             span = self._tracer.start_span(name, kind=SpanKind.CLIENT)
```
lzchen, line 73: In terms of the name, I think we can remove the "DATABASE_TYPE" portion and just use the `command_name`.

instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
```diff
@@ -73,21 +73,23 @@ def started(self, event: monitoring.CommandStartedEvent):
...
                 span.set_attribute("db.statement", statement)
                 if event.connection_id is not None:
                     span.set_attribute("net.peer.name", event.connection_id[0])
                     span.set_attribute("net.peer.port", event.connection_id[1])
 
                 # pymongo specific, not specified by spec
-                span.set_attribute("db.mongo.operation_id", event.operation_id)
-                span.set_attribute("db.mongo.request_id", event.request_id)
+                span.set_attribute(
+                    "db.mongodb.operation_id", event.operation_id
```
lzchen, line 85:
```suggestion
                    "db.operation_id", event.operation_id
```

instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
```diff
@@ -73,21 +73,23 @@ def started(self, event: monitoring.CommandStartedEvent):
...
+                )
+                span.set_attribute("db.mongodb.request_id", event.request_id)
```
lzchen, line 87:
```suggestion
                span.set_attribute("db.request_id", event.request_id)
```

instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
```diff
@@ -73,21 +73,23 @@ def started(self, event: monitoring.CommandStartedEvent):
...
+                span.set_attribute("db.name", event.database_name)
                 span.set_attribute("db.statement", statement)
                 if event.connection_id is not None:
                     span.set_attribute("net.peer.name", event.connection_id[0])
                     span.set_attribute("net.peer.port", event.connection_id[1])
 
                 # pymongo specific, not specified by spec
-                span.set_attribute("db.mongo.operation_id", event.operation_id)
-                span.set_attribute("db.mongo.request_id", event.request_id)
+                span.set_attribute(
```
lzchen, line 84: I would remove these attributes as they are not in the semantic conventions.

instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
```diff
@@ -106,7 +108,7 @@ def succeeded(self, event: monitoring.CommandSucceededEvent):
             return
         if span.is_recording():
             span.set_attribute(
-                "db.mongo.duration_micros", event.duration_micros
+                "db.mongodb.duration_micros", event.duration_micros
```
lzchen, line 111: And these should be removed as well.

instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
lzchen: EDIT: We can leave these in until the specs are finalized.

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py
```diff
@@ -20,15 +20,14 @@
 def _extract_conn_attributes(conn_kwargs):
     """ Transform redis conn info into dict """
     attributes = {
-        "db.type": "redis",
```
lzchen, line 23: Span name for redis needs to be fixed too.

lzchen:
I think there are a couple of other changes that are missing, from `dbapi` and `asyncpg` specifically. 
`db.system`, `db.name` should be span attributes. The span name should be the command executed. If available provide the attributes related to connections as well.

lzchen (changes requested):
Some comments.

instrumentation/opentelemetry-instrumentation-pymemcache/src/opentelemetry/instrumentation/pymemcache/__init__.py
lonewolf3739: The span name for all pymemcache commands is `memcached.command`. Regarding `db.name` there is `db.system` which is set to [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/99d6dfcb4f68705d43120a5cd5a2c7e677ceaed6/instrumentation/opentelemetry-instrumentation-pymemcache/src/opentelemetry/instrumentation/pymemcache/__init__.py#L167). Do you think we need `db.name` set to same as well?

instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
lonewolf3739: @toumorokoshi also expressed the same concern on this attrs. I will remove them.

instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
lonewolf3739: The spec doesn't have these either.

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py
lonewolf3739: This is span attribute. The span name for redis is `redis.command`.

lonewolf3739:
> `db.system`, `db.name` should be span attributes. 

Spec has `db.system` I haven't seen `db.name` mentioned.

> The span name should be the command executed.

I would appreciate some more elaboration on this.

> If available provide the attributes related to connections as well.

redis wasn't setting the connection attributes `net.peer.name` and `net.peer.port` as mentioned in spec. I have added them in this PR. I only came across redis not doing that I will check others also.

instrumentation/opentelemetry-instrumentation-pymemcache/src/opentelemetry/instrumentation/pymemcache/__init__.py
lzchen: > The span name for all pymemcache commands is memcached.command

I'm not too familiar in how to use pymemcache, but does that mean there is no "command" or "statement" that is executed?

> Do you think we need db.name set to same as well?

Yes, we have `db.name` in other instrumentations so we should be consistent.

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py
lzchen: Yes, I only highlighted this because Github didn't let point to the above code.

Simiarly to pymemcache for redis, is there not a concept of "statement" or "command"? The string "redis.command" doesn't really tell me much in terms of what logical operation redis is doing.

lzchen:
@lonewolf3739 
> Spec has db.system I haven't seen db.name mentioned.

`db.name` is mentioned under span name in the specs, in which it should be the span name if the "command executed" is not present.

> I would appreciate some more elaboration on this.

I expect the name to be the "command executed" on the database. For example, [commands](https://docs.mongodb.com/manual/reference/command/) for mongobdb.

instrumentation/opentelemetry-instrumentation-pymemcache/src/opentelemetry/instrumentation/pymemcache/__init__.py
lonewolf3739: It has commands. Yes we should be using them instead of hard-coded `memcached.command` for span name. I don't think `db.name` is applicable for memcached.

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py
lonewolf3739: > The string "redis.command" doesn't really tell me much in terms of what logical operation redis is doing.

Agree.

lonewolf3739:
Hi @lzchen ,

I am thinking of closing this PR in favor of opening multiple small PRs as changes are becoming large. What do you suggest?

lonewolf3739:
This is becoming huge. I will be opening multiple small PRs which will be easier to work and review.

PR #178, opened 2020-11-11
Add requests as a test dependency to AWS SDK extension

NathanielRN:
# Description

Adds `requests` module as a test dependency to the sdk aws extension.

Fixes # (issue)

## Type of change

Minor.

# Checklist:

### DIFF
--- a/sdk-extension/opentelemetry-sdk-extension-aws/setup.cfg
+++ b/sdk-extension/opentelemetry-sdk-extension-aws/setup.cfg
@@ -42,6 +42,8 @@ install_requires =
     opentelemetry-api == 0.16.dev0
 
 [options.entry_points]
+install_requires =
+    requests ~= 2.24.0
 opentelemetry_propagator =
     aws_xray = opentelemetry.sdk.extension.aws.trace.propagation.aws_xray_format:AwsXRayFormat
 


sdk-extension/opentelemetry-sdk-extension-aws/setup.cfg
```diff
@@ -42,6 +42,8 @@ install_requires =
     opentelemetry-api == 0.16.dev0
 
 [options.entry_points]
+install_requires =
+    requests ~= 2.24.0
```
toumorokoshi, line 46: is this correct? this would be adding requests to entry points. I would have expected this be added to the section above.

toumorokoshi (changes requested):
Can you double check the syntax? this may be working today in unit tests because requests is already installed as a test dependency.

toumorokoshi:
> Adds requests module as a test dependency to the sdk aws extension.

Can you clarify how this should work? There's a test_requires directive in setup.py that is used as part of `setup.py test`, but we don't leverage that mechanism to run our unit tests.

NathanielRN:
@toumorokoshi Thanks for commenting! I think I made this PR too hastily, for some reason I thought this was needed for tests, but I forgot that this is handled by the `tox.ini` file here:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/79b519f5cc517b3775112f0bc08e9e3ae4415859/tox.ini#L259

My bad! Will close it now 😄

PR #184, opened 2020-11-16
Update redis instrumentation to follow semantic conventions

lonewolf3739:
# Description
Part of  #159 

# Type of change

# How Has This Been Tested?

- [x] `tox -e test-instrumentation-redis`

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
@@ -90,8 +90,7 @@ def _traced_execute_pipeline(func, instance, args, kwargs):
     cmds = [_format_command_args(c) for c, _ in instance.command_stack]
     resource = "\n".join(cmds)
 
-    pipeline_cmnds = " ".join([args[0] for args, _ in instance.command_stack])
-    span_name = "PIPELINE" + " " + pipeline_cmnds
+    span_name = " ".join([args[0] for args, _ in instance.command_stack])
 
     with tracer.start_as_current_span(
         span_name, kind=trace.SpanKind.CLIENT
--- a/instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
@@ -59,7 +59,6 @@ from opentelemetry.instrumentation.utils import unwrap
 
 _DEFAULT_SERVICE = "redis"
 _RAWCMD = "db.statement"
-_CMD = "redis.command"
 
 
 def _set_connection_attributes(span, conn):
@@ -75,7 +74,7 @@ def _traced_execute_command(func, instance, args, kwargs):
     tracer = getattr(redis, "_opentelemetry_tracer")
     query = _format_command_args(args)
     with tracer.start_as_current_span(
-        _CMD, kind=trace.SpanKind.CLIENT
+        args[0], kind=trace.SpanKind.CLIENT
     ) as span:
         if span.is_recording():
             span.set_attribute("service", tracer.instrumentation_info.name)
@@ -91,8 +90,11 @@ def _traced_execute_pipeline(func, instance, args, kwargs):
     cmds = [_format_command_args(c) for c, _ in instance.command_stack]
     resource = "\n".join(cmds)
 
+    pipeline_cmnds = " ".join([args[0] for args, _ in instance.command_stack])
+    span_name = "PIPELINE" + " " + pipeline_cmnds
+
     with tracer.start_as_current_span(
-        _CMD, kind=trace.SpanKind.CLIENT
+        span_name, kind=trace.SpanKind.CLIENT
     ) as span:
         if span.is_recording():
             span.set_attribute("service", tracer.instrumentation_info.name)
--- a/instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
@@ -70,8 +70,13 @@ def _set_connection_attributes(span, conn):
 def _traced_execute_command(func, instance, args, kwargs):
     tracer = getattr(redis, "_opentelemetry_tracer")
     query = _format_command_args(args)
+    name = ""
+    if len(args) > 0 and args[0]:
+        name = args[0]
+    else:
+        name = instance.connection_pool.connection_kwargs.get("db", 0)
     with tracer.start_as_current_span(
-        args[0], kind=trace.SpanKind.CLIENT
+        name, kind=trace.SpanKind.CLIENT
     ) as span:
         if span.is_recording():
             span.set_attribute("service", tracer.instrumentation_info.name)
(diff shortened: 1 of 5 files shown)


lonewolf3739:
cc @lzchen

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
```diff
@@ -59,7 +59,6 @@
 
 _DEFAULT_SERVICE = "redis"
 _RAWCMD = "db.statement"
-_CMD = "redis.command"
```
lzchen, line 62: Where is `db.system`?

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
```diff
@@ -75,7 +74,7 @@ def _traced_execute_command(func, instance, args, kwargs):
     tracer = getattr(redis, "_opentelemetry_tracer")
     query = _format_command_args(args)
     with tracer.start_as_current_span(
-        _CMD, kind=trace.SpanKind.CLIENT
+        args[0], kind=trace.SpanKind.CLIENT
```
lzchen, line 77: What is in `args[0]`?

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
```diff
@@ -91,8 +90,11 @@ def _traced_execute_pipeline(func, instance, args, kwargs):
     cmds = [_format_command_args(c) for c, _ in instance.command_stack]
     resource = "\n".join(cmds)
 
+    pipeline_cmnds = " ".join([args[0] for args, _ in instance.command_stack])
+    span_name = "PIPELINE" + " " + pipeline_cmnds
```
lzchen, line 94: Remove the `PIPELINE`.

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
```diff
@@ -75,7 +74,7 @@ def _traced_execute_command(func, instance, args, kwargs):
     tracer = getattr(redis, "_opentelemetry_tracer")
     query = _format_command_args(args)
     with tracer.start_as_current_span(
-        _CMD, kind=trace.SpanKind.CLIENT
+        args[0], kind=trace.SpanKind.CLIENT
     ) as span:
```
lzchen, line 78: Fallback `db.name` for span name.

lonewolf3739: After looking at the implementation of client library [here](https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L935) I thought we can safely use it without having to worry about fallback. Do you think we need this?

toumorokoshi: maybe the one concern is that redis changes their argument format in a backwards-incompatible way. But I agree that's quite unlikely as it would break all consumers.

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py
```diff
@@ -20,15 +20,16 @@
 def _extract_conn_attributes(conn_kwargs):
     """ Transform redis conn info into dict """
     attributes = {
-        "db.type": "redis",
-        "db.instance": conn_kwargs.get("db", 0),
+        "db.system": "redis",
+        "db.name": conn_kwargs.get("db", 0),
     }
     try:
```
toumorokoshi, line 26: it looks like this is old code, but an if/else statement checking existence of keys is faster than a try/catch (more work like capture the stack when an exception is thrown)

toumorokoshi (approved):
great, thanks!

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
lzchen: I think it is less likely for other instrumentations as well, I just want to be consistent because there may be some exporters that rely on the existence of `db.name` so we should provide it if we can.

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py
lonewolf3739: https://docs.python.org/3/glossary.html?highlight=eafp#term-eafp, this is the reason I let this be `try`.. `catch`. I will change it if needed.

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
lonewolf3739: `db.name` will be set for redis instrumentation. From official redis docs `If none of these options are specified, the default db=0 is used`. This will be numeral value. While we all agree that it unlikely redis changes their format, but just in case if it changes what do guys think setting this db numeral as span name?

lzchen: I think it is fine.

lzchen (approved):

NathanielRN:
@lonewolf3739 Just curious if there was thought to add `db.redis.database_index` for `redis` as seen on [OTel DB Specifications](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#call-level-attributes-for-specific-technologies)? It looks like it is covered in most cases by `db.name` though.

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py
```diff
@@ -20,15 +20,16 @@
 def _extract_conn_attributes(conn_kwargs):
     """ Transform redis conn info into dict """
     attributes = {
-        "db.type": "redis",
-        "db.instance": conn_kwargs.get("db", 0),
+        "db.system": "redis",
+        "db.name": conn_kwargs.get("db", 0),
```
NathanielRN, line 24: Actually, the footnote on the [OTel Database Specificatiosn](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#call-level-attributes-for-specific-technologies) seems says it is required to use `db.redis.database_index` if the connection is not `0`?

lzchen: Good catch, maybe create a new issue to address this? @lonewolf3739

lonewolf3739: Sure, will do that.

lonewolf3739:
> @lonewolf3739 Just curious if there was thought to add `db.redis.database_index` for `redis` as seen on [OTel DB Specifications](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#call-level-attributes-for-specific-technologies)? It looks like it is covered in most cases by `db.name` though.

Good catch. I will add this.

PR #188, opened 2020-11-17
Update asyncpg instrumentation to follow semantic conventions

lonewolf3739:
# Description
Part of  #159 

# Type of change

# How Has This Been Tested?

- [x] `tox -e test-instrumentation-asyncpg`


instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
```diff
@@ -105,16 +118,19 @@ async def _do_execute(self, func, instance, args, kwargs):
...
+        span_attributes = _hydrate_span_from_args(
+            instance, args[0], args[1:] if self.capture_parameters else None
+        )
+        name = ""
+        if args[0]:
+            name = args[0]
+        elif span_attributes.get("db.name"):
+            name = span_attributes["db.name"]
+        else:
+            name = "postgresql"  # Does it ever happen?
```
lzchen, line 130: Probably not, but it's good just in case.

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
```diff
@@ -105,16 +118,19 @@ async def _do_execute(self, func, instance, args, kwargs):
...
-        with tracer.start_as_current_span(
-            "postgresql", kind=SpanKind.CLIENT
-        ) as span:
+        span_attributes = _hydrate_span_from_args(
+            instance, args[0], args[1:] if self.capture_parameters else None
+        )
+        name = ""
+        if args[0]:
+            name = args[0]
+        elif span_attributes.get("db.name"):
```
lzchen, line 127: Nit: Can you just use `getattr(params, "database", None)` to get the `db.name` instead of calling `_hydrate_span_from_args` before the span is created? Ideally, we don't want to call any unnecessary functions related to span attributes if the span is not recording, which is why originally hydrating the span occurs within the `is_recording()` check.

lzchen (approved):
Some non-blocking comments.

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
lonewolf3739: Good point. I will change.

lzchen (approved):

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
```diff
@@ -105,10 +118,16 @@ async def _do_execute(self, func, instance, args, kwargs):
...
-            "postgresql", kind=SpanKind.CLIENT
-        ) as span:
+        params = getattr(instance, "_params", None)
+        name = ""
+        if args[0]:
+            name = args[0]
+        elif params and params.get("database"):
+            name = params.get("database")
+        else:
+            name = "postgresql"  # Does it ever happen?
```
NathanielRN, line 128: Just a nit

```suggestion
        params = getattr(instance, "_params", {})
        name = args[0] if args[0] else params.get("database", "postgresql")
```

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
```diff
@@ -49,11 +49,24 @@
...
+        span_attributes["db.user"] = user
+
+    addr = getattr(connection, "_addr", None)
+    if isinstance(addr, tuple):
+        span_attributes["net.peer.name"] = addr[0]
+        span_attributes["net.peer.ip"] = addr[1]
+        span_attributes["net.transport"] = "IP.TCP"
+    elif isinstance(addr, str):
+        span_attributes["net.peer.name"] = addr
+        span_attributes["net.transport"] = "Unix"
```
NathanielRN, line 69: Nit: Are there any docs explaining how to parse these out? I think they could be useful to add to the function doc string for `_hydrate_span_from_args`

lonewolf3739: Connection contains `_addr` attribute which is either a host/port tuple, or Unix socket string. May be something like this?

lonewolf3739: Updated in [a467449](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/188/commits/a46744976beb940ab90b5ed867de13c0ecfd496a)

PR #190, opened 2020-11-17
Remove unused properties in exceptions for request instrumentation

lzchen:
Part of https://github.com/open-telemetry/opentelemetry-python/issues/1301


instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -158,9 +154,6 @@ def _instrumented_requests_call(
                     result = call_wrapped()  # *** PROCEED
                 except Exception as exc:  # pylint: disable=W0703
                     exception = exc
-                    setattr(
-                        exception, EXCEPTION_STATUS_FIELD, StatusCode.ERROR,
-                    )
```
codeboten, line 163: Can you explain a bit more about why this is being removed or why this was here in the first place?

lzchen: I believe from this [pr](https://github.com/open-telemetry/opentelemetry-python/pull/1162), the purpose was get the status code that was returned from a failed instrumentation, but now since we don't track statuscanonicalcode anymore, we don't need this.

codeboten (approved):
Thanks for clarifying! 👍

PR #197, opened 2020-11-20
Mark test case as flaky

### LINKED ISSUE #196: test_batch_span_processor_reset_timeout is flaky
The test mentioned above fails unpredictably. If it is not possible to fix the test case at least handle it in a way that does not affect the build process.

ocelotl:
This marks the test case as flaky, making it run at most 3 times. It is
enough for one of this runs to pass to consider this test case passed
and ran no more. If 3 consecutive runs of this test case fail, the test
case will be considered failed. It has been reported that running this
test case again makes it pass, usually. This approach is preferred over
marking it as xfail(strict=False) because most of the times the test
ends up passing after another run, so that in most of the cases we
can still benefit from running this test case (since if it is actually
failing because of a bug it will be reported as such after failing 3
times, making the team aware of an actual issue happening here).

# Description

Fixes #196
## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


tox.ini
```diff
@@ -251,7 +251,7 @@ commands_pre =
 
   aiopg: pip install {toxinidir}/instrumentation/opentelemetry-instrumentation-dbapi pip install {toxinidir}/instrumentation/opentelemetry-instrumentation-aiopg[test]
 
-  datadog: pip install {toxinidir}/exporter/opentelemetry-exporter-datadog
+  datadog: pip install flaky {toxinidir}/exporter/opentelemetry-exporter-datadog
```
ocelotl, line 254: Not sure if this is the right place to put this dependency, but after trying in several places, this is the one that worked. Any suggestions @NathanielRN?

toumorokoshi: this works fine. generally it's nice to put it into a multi-name match above this one:

```
datadog,other-instrumentation-that-needs-it: pip install flaky
datadog: {toxinidir}/..../exporter-datadog
```

toumorokoshi (approved):
I'll approve this for the sake of moving forward, but can you provide more context on #196 or this commit message on why we can't just fix the test?

It's good context for someone to investigate in the future.

tox.ini
NathanielRN: Tthis is the right place to put it (since only `datadog` needs it for now) but I would also like to know why the test sometimes fail 😕

lzchen (approved):

PR #205, opened 2020-11-23
Move Instrumentation Example App to Contrib

NathanielRN:
# Description

Follow up to https://github.com/open-telemetry/opentelemetry-python/pull/1410, move the example app to the Contrib repo since it uses instrumentation.

## Type of change

# How Has This Been Tested?

Tests confirm the moving of the app works well.

Not dependent on a change in Core.

# Checklist:


lzchen:
Did we decide on keeping this? I still am not sure if we need this or not, since each instrumentation already has example code in the doc string. The examples here also don't seem to add much value.

NathanielRN:
@lzchen I also don't think it's all that useful :) I just knew how to make the PR here quickly after seeing it being closed from the Core repo so thought I would put it here for discussion purposes but feel free to close it if we don't want it anymore!

codeboten:
Thanks for opening the PR. I agree with both of you that the example' value here is limited. Happy to close this and re-open in the future if someone feels strongly about it.

PR #206, opened 2020-11-23
Add Prometheus Remote Write Exporter (2/6)

AzfaarQureshi:
# Description
This is PR 2/6 of adding a Prometheus Remote Write Exporter in Python SDK and address Issue https://github.com/open-telemetry/opentelemetry-python/issues/1302

**[Part 1/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/180)**
* Adds class skeleton
* Adds all function signatures

👉 **[Part 2/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/206)**
* Adds validation of exporter constructor commands
* Add unit tests for validation

**[Part 3/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/207)**
* Adds conversion methods from OTel metric types to Prometheus [TimeSeries](https://github.com/prometheus/prometheus/blob/master/prompb/types.proto#L47)
* Add unit tests for conversion

**[Part 4/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/212)**
* Adds methods to export metrics to Remote Write endpoint
* Add unit tests for exporting

**[Part 5/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/216)**
* Add integration tests using sample app and instance of Cortex

**Part 6/6**
* Add README, Design Doc and other necessary documentation.

## Type of change
- [x] New feature (non-breaking change which adds functionality)
# How Has This Been Tested?

- [ ] Added class `TestValidation` in `test_prometheus_remote_write_exporter.py`

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:
- [x] Followed the style guidelines of this project
- [x] Changelogs have been updated
- [x] Unit tests have been added
- [ ] ~~Documentation has been updated~~
cc- @shovnik, @alolita

PR #180, opened 2020-11-12
Add Prometheus Remote Write Exporter (1/6)

shovnik:
# Description
This is PR 1/6 of adding a Prometheus Remote Write Exporter in Python SDK and address Issue https://github.com/open-telemetry/opentelemetry-python/issues/1302

👉 **[Part 1/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/180)**
* Adds class skeleton
* Adds all function signatures

**[Part 2/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/206)**
* Adds validation of exporter constructor commands
* Add unit tests for validation

**[Part 3/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/207)**
* Adds conversion methods from OTel metric types to Prometheus [TimeSeries](https://github.com/prometheus/prometheus/blob/master/prompb/types.proto#L47)
* Add unit tests for conversion

**[Part 4/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/212)**
* Adds methods to export metrics to Remote Write endpoint
* Add unit tests for exporting

**[Part 5/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/216)**
* Add integration tests using sample app and instance of Cortex

**Part 6/6**
* Add README, Design Doc and other necessary documentation.

## Type of change
- [x] New feature (non-breaking change which adds functionality)
# How Has This Been Tested?

- [ ] Added class `TestValidation` in `test_prometheus_remote_write_exporter.py`

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:
- [x] Followed the style guidelines of this project
- [x] Changelogs have been updated
- [x] Unit tests have been added
- [ ] Documentation has been updated (Full documentation in final exporter PR)
cc- @AzfaarQureshi , @alolita

PR #207, opened 2020-11-23
Add Prometheus Remote Write Exporter (3/6)

shovnik:
# Description
This is PR 3/6 of adding a Prometheus Remote Write Exporter in Python SDK and address Issue https://github.com/open-telemetry/opentelemetry-python/issues/1302

* Adds class skeleton

👉 **[Part 3/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/207)**

**[Part 4/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/212)**

**[Part 5/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/216)**
* Add integration tests using sample app and instance of Cortex

**Part 6/6**

## Type of change
# How Has This Been Tested?

- [ ] Added class `TestValidation` in `test_prometheus_remote_write_exporter.py`

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:
- [x] Changelogs have been updated (change logs will be updated when PR 7/7 is merged and the RW exporter is complete)
cc- @AzfaarQureshi , @alolita

### DIFF
--- a/exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
+++ b/exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
@@ -12,8 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
+import re
 from typing import Dict, Sequence
 
+from opentelemetry.exporter.prometheus_remote_write.gen.remote_pb2 import (
+    WriteRequest,
+)
 from opentelemetry.exporter.prometheus_remote_write.gen.types_pb2 import (
     Label,
     Sample,
@@ -24,6 +29,15 @@ from opentelemetry.sdk.metrics.export import (
     MetricsExporter,
     MetricsExportResult,
 )
+from opentelemetry.sdk.metrics.export.aggregate import (
+    HistogramAggregator,
+    LastValueAggregator,
+    MinMaxSumCountAggregator,
+    SumAggregator,
+    ValueObserverAggregator,
+)
+
+logger = logging.getLogger(__name__)
 
 
 class PrometheusRemoteWriteMetricsExporter(MetricsExporter):
@@ -54,6 +68,14 @@ class PrometheusRemoteWriteMetricsExporter(MetricsExporter):
         self.timeout = timeout
         self.tls_config = tls_config
         self.proxies = proxies
+
+        self.converter_map = {
+            MinMaxSumCountAggregator: self._convert_from_min_max_sum_count,
+            SumAggregator: self._convert_from_sum,
+            HistogramAggregator: self._convert_from_histogram,
+            LastValueAggregator: self._convert_from_last_value,
+            ValueObserverAggregator: self._convert_from_value_observer,
+        }
 
     @property
     def endpoint(self):
@@ -142,50 +164,145 @@ class PrometheusRemoteWriteMetricsExporter(MetricsExporter):
     def shutdown(self) -> None:
         raise NotImplementedError()
 
-    def convert_to_timeseries(
+    def _convert_to_timeseries(
         self, export_records: Sequence[ExportRecord]
+    ) -> Sequence[TimeSeries]:
+        timeseries = []
+        for export_record in export_records:
+            aggregator_type = type(export_record.aggregator)
+            converter = self.converter_map.get(aggregator_type)
+            if converter:
+                timeseries.extend(converter(export_record))
+            else:
+                logger.warning(
+                    "%s aggregator is not supported, record dropped",
+                    aggregator_type,
+                )
+        return timeseries
+
+    def _convert_from_sum(
+        self, sum_record: ExportRecord
+    ) -> Sequence[TimeSeries]:
+        return [
+            self._create_timeseries(
+                sum_record,
+                sum_record.instrument.name + "_sum",
+                sum_record.aggregator.checkpoint,
+            )
+        ]
+
+    def _convert_from_min_max_sum_count(
+        self, min_max_sum_count_record: ExportRecord
+    ) -> Sequence[TimeSeries]:
+        timeseries = []
+        for agg_type in ["min", "max", "sum", "count"]:
+            name = min_max_sum_count_record.instrument.name + "_" + agg_type
+            value = getattr(
+                min_max_sum_count_record.aggregator.checkpoint, agg_type
+            )
+            timeseries.append(
+                self._create_timeseries(min_max_sum_count_record, name, value)
+            )
+        return timeseries
+
+    def _convert_from_histogram(
+        self, histogram_record: ExportRecord
+    ) -> Sequence[TimeSeries]:
+        timeseries = []
+        for bound in histogram_record.aggregator.checkpoint.keys():
+            bound_str = "+Inf" if bound == float("inf") else str(bound)
+            value = histogram_record.aggregator.checkpoint[bound]
+            timeseries.append(
+                self._create_timeseries(
+                    histogram_record,
+                    histogram_record.instrument.name + "_histogram",
+                    value,
+                    extra_label=("le", bound_str),
+                )
+            )
+        return timeseries
+
+    def _convert_from_last_value(
+        self, last_value_record: ExportRecord
+    ) -> Sequence[TimeSeries]:
+        return [
+            self._create_timeseries(
+                last_value_record,
+                last_value_record.instrument.name + "_last",
+                last_value_record.aggregator.checkpoint,
+            )
+        ]
+
+    def _convert_from_value_observer(
+        self, value_observer_record: ExportRecord
+    ) -> Sequence[TimeSeries]:
+        timeseries = []
+        for agg_type in ["min", "max", "sum", "count", "last"]:
+            timeseries.append(
+                self._create_timeseries(
+                    value_observer_record,
+                    value_observer_record.instrument.name + "_" + agg_type,
+                    getattr(
+                        value_observer_record.aggregator.checkpoint, agg_type
+                    ),
+                )
+            )
+        return timeseries
+
+    # TODO: Implement convert from quantile once supported by SDK for Prometheus Summaries
+    def _convert_from_quantile(
+        self, summary_record: ExportRecord
     ) -> Sequence[TimeSeries]:
         raise NotImplementedError()
 
-    def convert_from_sum(self, sum_record: ExportRecord) -> TimeSeries:
-        raise NotImplementedError()
-
-    def convert_from_min_max_sum_count(
-        self, min_max_sum_count_record: ExportRecord
-    ) -> TimeSeries:
-        raise NotImplementedError()
-
-    def convert_from_histogram(
-        self, histogram_record: ExportRecord
-    ) -> TimeSeries:
-        raise NotImplementedError()
-
-    def convert_from_last_value(
-        self, last_value_record: ExportRecord
-    ) -> TimeSeries:
-        raise NotImplementedError()
-
-    def convert_from_value_observer(
-        self, value_observer_record: ExportRecord
-    ) -> TimeSeries:
-        raise NotImplementedError()
-
-    def convert_from_quantile(
-        self, summary_record: ExportRecord
-    ) -> TimeSeries:
-        raise NotImplementedError()
-
     # pylint: disable=no-member
-    def create_timeseries(
-        self, export_record: ExportRecord, name, value: float
+    def _create_timeseries(
+        self,
+        export_record: ExportRecord,
+        name: str,
+        value: float,
+        extra_label: (str, str) = None,
     ) -> TimeSeries:
-        raise NotImplementedError()
+        timeseries = TimeSeries()
+        seen = set()
 
-    def create_sample(self, timestamp: int, value: float) -> Sample:
-        raise NotImplementedError()
+        def add_label(label_name: str, label_value: str):
+            # Label name must contain only alphanumeric characters and underscores
+            label_name = re.sub("[^\\w_]", "_", label_name)
+            if label_name not in seen:
+                label = Label()
+                label.name = label_name
+                label.value = label_value
+                timeseries.labels.append(label)
+                seen.add(label_name)
+            else:
+                logger.warning(
+                    "Duplicate label with name %s and value %s",
+                    label_name,
+                    label_value,
+                )
 
-    def create_label(self, name: str, value: str) -> Label:
-        raise NotImplementedError()
+        # The __name__ label is required by PromQL as its value appears as the metric_name
+        add_label("__name__", name)
+        if extra_label:
+            add_label(extra_label[0], extra_label[1])
+        if export_record.resource.attributes:
+            for (
+                label_name,
+                label_value,
+            ) in export_record.resource.attributes.items():
+                add_label(label_name, str(label_value))
+        if export_record.labels:
+            for [label_name, label_value] in export_record.labels:
+                add_label(label_name, label_value)
+
+        sample = Sample()
+        sample.timestamp = int(
+            export_record.aggregator.last_update_timestamp / 1000000
+        )
+        sample.value = value
+        timeseries.samples.append(sample)
+        return timeseries
 
     def build_message(self, timeseries: Sequence[TimeSeries]) -> bytes:
         raise NotImplementedError()
--- a/exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
+++ b/exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
@@ -13,10 +13,28 @@
 # limitations under the License.
 
 import unittest
+from logging import Logger
+from unittest.mock import MagicMock, Mock, patch
 
 from opentelemetry.exporter.prometheus_remote_write import (
     PrometheusRemoteWriteMetricsExporter,
 )
+from opentelemetry.exporter.prometheus_remote_write.gen.types_pb2 import (
+    Label,
+    Sample,
+    TimeSeries,
+)
+from opentelemetry.sdk.metrics import Counter
+from opentelemetry.sdk.metrics.export import ExportRecord, MetricsExportResult
+from opentelemetry.sdk.metrics.export.aggregate import (
+    HistogramAggregator,
+    LastValueAggregator,
+    MinMaxSumCountAggregator,
+    SumAggregator,
+    ValueObserverAggregator,
+)
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.util import get_dict_as_key
 
 
 class TestValidation(unittest.TestCase):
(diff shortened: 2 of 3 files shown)


exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+    # Ensures sum aggregator is correctly converted to timeseries
+    def test_convert_from_sum(self):
+        sum_record = self._generate_record(SumAggregator)
+        sum_record.aggregator.update(3)
+        sum_record.aggregator.update(2)
+        sum_record.aggregator.take_checkpoint()
+
+        self._exporter.create_timeseries = self._converter_mock()
+        timeseries = self._exporter.convert_from_sum(sum_record)
+        self.assertEqual(timeseries[0], (SumAggregator, "testname", 5))
```
ocelotl, line 191: Here you used `SumAggregator` in [this](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/207/files#diff-842c4d8d0cb6c63a88f2c4850fb0162d7f19e1d4b929af3c56a67d09c782f922R184) line. `3` and `2` from [these](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/207/files#diff-842c4d8d0cb6c63a88f2c4850fb0162d7f19e1d4b929af3c56a67d09c782f922R185-R186) lines add up to `5`. It is natural then for the test reader to find `SumAggregator` and `5` in the assertion. But where did `"testname"` was set in this test case? That is not that obvious and is a bit confusing. You may or may not want to refactor this test case to avoid that, just keep that in mind that whoever reads a test case expects to find initial conditions being set, operations on objects under test being done and finally assertions and the assertions should make sense based on the initial conditions set previously.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+            return ExportRecord(
+                self._test_metric, None, aggregator_type(), Resource({}),
+            )
+
+        self._generate_record = generate_record
+
+        def converter_method(record, name, value):
+            return (type(record.aggregator), name, value)
+
+        self._converter_mock = mock.MagicMock(return_value=converter_method)
```
ocelotl, line 150: Why use a [`MagicMock`](https://docs.python.org/3/library/unittest.mock.html#magic-mock) here? Is there a call to any magic method in the tests?

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+
+    # Ensures histogram aggregator is correctly converted to timeseries
+    def test_convert_from_histogram(self):
+        histogram_record = self._generate_record(HistogramAggregator)
+        histogram_record.aggregator.update(5)
+        histogram_record.aggregator.update(2)
+        histogram_record.aggregator.update(-1)
+        histogram_record.aggregator.take_checkpoint()
+
+        self._exporter.create_timeseries = self._converter_mock()
```
ocelotl, line 227: Same here, an attribute of an object under test is being changed into a `Mock` but it is not reset afterwards.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+        self.assertEqual(
+            timeseries[3], (ValueObserverAggregator, "testname_count", 3)
+        )
+        self.assertEqual(
+            timeseries[4], (ValueObserverAggregator, "testname_last", 2)
+        )
+
+    # Ensures quantile aggregator is correctly converted to timeseries
+    # TODO: Add test once method is implemented
+    def test_convert_from_quantile(self):
```
ocelotl, line 281: I understand that this PR is part of a series of PRs, I just want to let you know that in other circumstances having tests that do nothing and always pass mess up with your testing results since Pytest considers them to be passed test cases when really nothing has been tested. I would recommend not having them at all, commenting them out or marking them with [`skip`](https://docs.pytest.org/en/latest/skipping.html#skipping-test-functions) (in that order of recommendation).

ocelotl (changes requested):
Please take a look at the comments and check what can be refactored from the test cases. :+1:

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+        self._exporter = PrometheusRemoteWriteMetricsExporter(
+            endpoint="/prom/test_endpoint"
+        )
+
+        def generate_record(aggregator_type):
+            return ExportRecord(
+                self._test_metric, None, aggregator_type(), Resource({}),
+            )
+
+        self._generate_record = generate_record
```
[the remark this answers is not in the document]
shovnik: It felt repetitive to use the MetricRecord contructor every time since most of the arguments were the same in each test, so I thought I could just create a method that creates a test_record with the required aggregator type.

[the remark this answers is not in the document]
ocelotl: Ok, but why define a private attribute for `generate_record`? Also, if you find yourself repeating arguments in several calls, then [`partial`](https://docs.python.org/3/library/functools.html#functools.partial) can help.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -0,0 +1,291 @@
...
+            SumAggregator: self.convert_from_sum,
+            HistogramAggregator: self.convert_from_histogram,
+            LastValueAggregator: self.convert_from_last_value,
+            ValueObserverAggregator: self.convert_from_last_value,
+        }
+        timeseries = []
+        for export_record in export_records:
+            aggregator_type = type(export_record.aggregator)
+            converter = converter_map.get(aggregator_type)
+            if not converter:
```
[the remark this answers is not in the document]
shovnik: Fixed

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -0,0 +1,291 @@
...
+                raise ValueError(
+                    str(aggregator_type) + " conversion is not supported"
+                )
+            timeseries.extend(converter(export_record))
+        return timeseries
+
+    def convert_from_sum(self, sum_record: ExportRecord) -> TimeSeries:
+        name = sum_record.instrument.name
+        value = sum_record.aggregator.checkpoint
+        return [self.create_timeseries(sum_record, name, value)]
```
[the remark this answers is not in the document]
shovnik: Refactored

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -0,0 +1,291 @@
...
+
+    def convert_from_min_max_sum_count(
+        self, min_max_sum_count_record: ExportRecord
+    ) -> TimeSeries:
+        timeseries = []
+        agg_types = ["min", "max", "sum", "count"]
```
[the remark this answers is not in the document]
shovnik: Refactored

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -0,0 +1,291 @@
...
+        name = histogram_record.instrument.name + "_count"
+        timeseries.append(
+            self.create_timeseries(histogram_record, name, float(count))
+        )
+        return timeseries
+
+    def convert_from_last_value(
+        self, last_value_record: ExportRecord
+    ) -> TimeSeries:
+        name = last_value_record.instrument.name
```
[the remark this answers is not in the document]
shovnik: Refactored

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -0,0 +1,291 @@
...
+    def create_sample(self, timestamp: int, value: float) -> Sample:
+        sample = Sample()
+        sample.timestamp = int(timestamp / 1000000)
+        sample.value = value
+        return sample
+
+    def create_label(self, name: str, value: str) -> Label:
+        label = Label()
+        # Label name must contain only alphanumeric characters and underscores
+        label.name = re.sub("[^0-9a-zA-Z_]+", "_", name)
```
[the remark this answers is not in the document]
shovnik: You bring up a good point. I rewrote the code do dedup labels by key prioritizing the required labels which identify what type of aggregation it is (min, max, sum, count, last). It is not possible for there to be duplicate labels within the metric names themselves, but it is possible that there are duplicate labels between the name labels, record labels and resource labels, so I used a set keeping track of seen label keys in the new implementation.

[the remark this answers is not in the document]
shovnik: Changed regex to "[^\\w_]" to ensure 1:1 underscore substitutions as suggested.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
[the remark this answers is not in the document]
shovnik: I decided to just use the constructor despite some repeating arguments in each method to make it clear what the properties of the test record are for someone just looking at one test.

Side Note: You are right, there is no reason for that test helper method to be private. My understanding of what should be public/private in the test class was incorrect.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+
+        def converter_method(record, name, value):
+            return (type(record.aggregator), name, value)
+
+        self._converter_mock = mock.MagicMock(return_value=converter_method)
+
+    # Ensures conversion to timeseries function works with valid aggregation types
+    def test_valid_convert_to_timeseries(self):
+        timeseries_mock_method = mock.Mock(return_value=["test_value"])
+        self._exporter.convert_from_sum = timeseries_mock_method
```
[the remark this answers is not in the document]
shovnik: I decided to bypass mocking these methods altogether as I realized that it is unnecessary to mock methods from the class I am testing since tests elsewhere will fail anyways if they were defective. I will keep in mind to use patch over assigning mocks in the future to avoid future tests though.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+            )
+
+
+class TestConversion(unittest.TestCase):
+    # Initializes test data that is reused across tests
+    def setUp(self):
+        self._test_metric = Counter(
+            "testname", "testdesc", "testunit", int, None
+        )
+        self._exporter = PrometheusRemoteWriteMetricsExporter(
```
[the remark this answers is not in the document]
shovnik: Remove leading underscore from all instances of self.exporter (in future subsequent PR as well where same issue was present).

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
```diff
@@ -0,0 +1,337 @@
...
+            self._generate_record(HistogramAggregator),
+            self._generate_record(LastValueAggregator),
+            self._generate_record(ValueObserverAggregator),
+        ]
+        data = self._exporter.convert_to_timeseries(test_records)
+        self.assertEqual(len(data), 5)
+        for timeseries in data:
+            self.assertEqual(timeseries, "test_value")
+
+        no_type_records = [self._generate_record(lambda: None)]
```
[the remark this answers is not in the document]
shovnik: Incidentally fixed by removing the generate_record method to make test_record contents clear.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
shovnik: Fixed by replacing generate_record method with ExportRecord constructor for clarity on contents of record.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
shovnik: I was using this to check if the mock method was called with the correct arguments, but after some research I now know there are `assert_called_with` methods for mocks which do exactly that. In this particular case, I got rid of the mock method altogether since it was mocking a method of the class being tested anyways.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
shovnik: Same as above comment.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
shovnik: I did not think of that. Got rid of the empty test leaving only a comment as a reminder of where to implement this test when the method is implemented.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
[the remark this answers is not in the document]
shovnik: Refactored

ocelotl (approved):

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -142,50 +153,144 @@ def export(
     def shutdown(self) -> None:
         raise NotImplementedError()
 
-    def convert_to_timeseries(
+    def _convert_to_timeseries(
```
lzchen, line 156: Do these functions need to be part of the exporter?

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -142,50 +153,144 @@ def export(
...
         self, export_records: Sequence[ExportRecord]
     ) -> Sequence[TimeSeries]:
-        raise NotImplementedError()
-
-    def convert_from_sum(self, sum_record: ExportRecord) -> TimeSeries:
-        raise NotImplementedError()
-
-    def convert_from_min_max_sum_count(
+        converter_map = {
+            MinMaxSumCountAggregator: self._convert_from_min_max_sum_count,
```
lzchen, line 160: We need to redefine this map on every call to convert? If you take the convert functions out of the exporter, you can just define it once at the top level.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -142,50 +153,144 @@ def export(
...
+            LastValueAggregator: self._convert_from_last_value,
+            ValueObserverAggregator: self._convert_from_value_observer,
+        }
+        timeseries = []
+        for export_record in export_records:
+            aggregator_type = type(export_record.aggregator)
+            converter = converter_map.get(aggregator_type)
+            if converter is None:
+                raise ValueError(
+                    str(aggregator_type) + " conversion is not supported"
```
lzchen, line 172:
```suggestion
                    str(aggregator_type) + " aggregator is not supported"
```

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -142,50 +153,144 @@ def export(
...
+                )
+            timeseries.extend(converter(export_record))
+        return timeseries
+
+    def _convert_from_sum(self, sum_record: ExportRecord) -> TimeSeries:
```
lzchen, line 177: Shouldn't this be `Sequence[TimeSeries]`?

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -142,50 +153,144 @@ def export(
...
-        raise NotImplementedError()
-
-    def convert_from_histogram(
+        timeseries = []
+        for agg_type in ["min", "max", "sum", "count"]:
+            name = min_max_sum_count_record.instrument.name + "_" + agg_type
+            value = getattr(
+                min_max_sum_count_record.aggregator.checkpoint, agg_type
+            )
+            timeseries.append(
```
lzchen, line 195: Prometheus accepts a separate timeseries for each of these values?

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -142,50 +153,144 @@ def export(
...
+            # Label name must contain only alphanumeric characters and underscores
+            label_name = re.sub("[^\\w_]", "_", label_name)
+            if label_name not in seen:
+                label = Label()
+                label.name = label_name
+                label.value = label_value
+                timeseries.labels.append(label)
+                seen.add(label_name)
+
+        add_label("__name__", name)
```
lzchen, line 273: What's this for?

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -142,50 +153,144 @@ def export(
...
-        raise NotImplementedError()
-
-    def create_label(self, name: str, value: str) -> Label:
-        raise NotImplementedError()
+        timeseries = TimeSeries()
+        seen = set()
+
+        def add_label(label_name, label_value):
+            # Label name must contain only alphanumeric characters and underscores
+            label_name = re.sub("[^\\w_]", "_", label_name)
```
lzchen, line 265: I'm wondering if we might get labels that are supposed to be unique but will be dropped due to removal of these characters. Is the alphanumeric + underscores limitation of the labels specific for Prometheus?

e.g. 
`testing` and `${testing}`

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -142,50 +153,144 @@ def export(
...
-
-    def create_sample(self, timestamp: int, value: float) -> Sample:
-        raise NotImplementedError()
-
-    def create_label(self, name: str, value: str) -> Label:
-        raise NotImplementedError()
+        timeseries = TimeSeries()
+        seen = set()
+
+        def add_label(label_name, label_value):
```
lzchen, line 263: Is `label` a Prometheus term?

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -142,50 +153,144 @@ def export(
...
+        add_label("__name__", name)
+        if labels:
+            for [label_name, label_value] in labels:
+                add_label(label_name, label_value)
+        if export_record.resource.attributes:
+            for (
+                label_name,
+                label_value,
+            ) in export_record.resource.attributes.items():
+                add_label(label_name, label_value)
```
lzchen, line 282: Do `Label` values have to be `str` type? Might have to convert the attribute values.

lzchen (changes requested):
Some comments.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Prometheus doesn't, but backends like Cortex do (exported from Prometheus). Cortex only accepts the TimeSeries format which leaves no other way to store aggregations of different types than creating multiple TimeSeries with different labels.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: An alternative to extracting all the converter methods out of the class could be defining it once in the constructor as a class attribute.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: The limitation is specific to the TimeSeries type https://prometheus.io/docs/concepts/data_model/ . You bring up a valid point. Since the non-alphanumeric characters are replaced with underscores: '$testing' and '_testing' will be considered the same label which would be an issue. Also, a label_name cannot start with two underscores. Would it be better to validate label names (raise Error) or add more complex sanitizing? I am not quite sure how to prevent duplicates when addressing label names that contain non-alphanumeric characters.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: No, they really don't. We were unsure were to put them. Will follow-up on this.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Good catch. Fixed.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Fixed.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Amended.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Each TimeSeries requires the "__name__" property to be set. It is what will appear as the metric name when querying the TimeSeries (on Grafana for example).

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: If I were to move the converter methods out of the class, would the map have to instantiated at file level in a separate converter.py file?

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: After further discussion, keeping the supported aggregator to converter method map as in the class namespace is reasonable since it is not used anywhere outside the class. Therefore, all the converter methods now call "self" and can remain as normal methods within the class. Let me know if you agree with this conclusion.

shovnik:
All comments have been addressed @lzchen.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -142,50 +161,132 @@ def export(
...
-
-    def convert_from_sum(self, sum_record: ExportRecord) -> TimeSeries:
-        raise NotImplementedError()
+        timeseries = []
+        for export_record in export_records:
+            aggregator_type = type(export_record.aggregator)
+            converter = self.converter_map.get(aggregator_type)
+            if converter is None:
+                raise ValueError(
+                    str(aggregator_type) + " aggregator is not supported"
```
lzchen, line 173: Nit: Should use string formatting.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -142,50 +161,132 @@ def export(
...
-        raise NotImplementedError()
-
-    def convert_from_sum(self, sum_record: ExportRecord) -> TimeSeries:
-        raise NotImplementedError()
+        timeseries = []
+        for export_record in export_records:
+            aggregator_type = type(export_record.aggregator)
+            converter = self.converter_map.get(aggregator_type)
+            if converter is None:
+                raise ValueError(
```
lzchen, line 172: Do we want to raise an exception? This could happen during runtime and we have a strict "no-crashing user's app" policy.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
lzchen: If the limitation is Prometheus specific, I am okay with the current implementation. The only problem was if it were a limitation we imposed on the OT side, then Prometheus users would have unexpected behaviour. But if it's on the Prometheus side, then user's should EXPECT this behaviour.

I wonder in this case if we should DROP labels that defy the `TimeSeries` definition (and maybe log a warning), instead of only dropping the duplicate.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
lzchen: Could you add a comment for that?

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
lzchen: What about other backends that support Prometheus data format? Seems like this would be a CortexPrometheusRemoteWriteExporter in this case then.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Would the standard then be to log a warning and then abort the export of that specific record?

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Sorry, what I meant was that although Prometheus does not receive data in TimeSeries format when using Prometheus instance and scraping metric_data, it's remote write api can only send data in TimeSeries format to an integrated backend. Therefore, any backend (influxDB, Thanos, Cortex) will need to receive the more complex aggregations split into separate TimeSeries with clear labelling. Cortex was just an example.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: In the rare case this occurs, I think whoever is misusing the exporter will value the label with the non-ascii characters as much as the valid one, so I suggest dropping the duplicate but logging a warning that after sanitizing, a duplicate label had to be dropped.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Added

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Tweaked

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Implemented the above mentioned logging ^

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -142,50 +164,135 @@ def export(
...
+        timeseries = []
+        for export_record in export_records:
+            aggregator_type = type(export_record.aggregator)
+            converter = self.converter_map.get(aggregator_type)
+            if converter is None:
+                logger.warning(
+                    "%s aggregator is not supported, record dropped",
+                    aggregator_type,
+                )
+                return None
```
lzchen, line 179: I don't think we want to return here? We just want to skip the record.

shovnik: I thought to skip the record, I would need to return both here and in the calling "export" method. I was planning on adding this to the export method:
if timeseries is None:
    return

Is there a better way to skip the record from an inner method?

shovnik: I misunderstood it as skip the batch. I fixed this to only skip the record and log a warning. I will add an additional check to not send empty requests in the subsequent PR.

lzchen (approved):
Nice!

PR #212, opened 2020-11-24
Add Prometheus Remote Write Exporter (4/6)

shovnik:
# Description
This is PR 3/6 of adding a Prometheus Remote Write Exporter in Python SDK and address Issue https://github.com/open-telemetry/opentelemetry-python/issues/1302

* Adds class skeleton

**[Part 3/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/207)**

👉 **[Part 4/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/212)**

**[Part 5/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/216)**
* Add integration tests using sample app and instance of Cortex

**Part 6/6**

## Type of change
# How Has This Been Tested?

- [ ] Added class `TestValidation` in `test_prometheus_remote_write_exporter.py`

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:
cc- @AzfaarQureshi , @alolita

### DIFF
--- a/exporter/opentelemetry-exporter-prometheus-remote-write/setup.cfg
+++ b/exporter/opentelemetry-exporter-prometheus-remote-write/setup.cfg
@@ -39,6 +39,8 @@ package_dir=
     =src
 packages=find_namespace:
 install_requires =
+    protobuf >= 3.13.0
+    requests == 2.25.0
     opentelemetry-api == 0.17.dev0
     opentelemetry-sdk == 0.17.dev0
 
--- a/exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
+++ b/exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
@@ -16,6 +16,9 @@ import logging
 import re
 from typing import Dict, Sequence
 
+import requests
+
+import snappy
 from opentelemetry.exporter.prometheus_remote_write.gen.remote_pb2 import (
     WriteRequest,
 )
@@ -48,7 +51,7 @@ class PrometheusRemoteWriteMetricsExporter(MetricsExporter):
         endpoint: url where data will be sent (Required)
         basic_auth: username and password for authentication (Optional)
         headers: additional headers for remote write request (Optional)
-        timeout: timeout for requests to the remote write endpoint in seconds (Optional)
+        timeout: timeout for remote write requests in seconds, defaults to 30 (Optional)
         proxies: dict mapping request proxy protocols to proxy urls (Optional)
         tls_config: configuration for remote write TLS settings (Optional)
     """
@@ -96,15 +99,15 @@ class PrometheusRemoteWriteMetricsExporter(MetricsExporter):
         if basic_auth:
             if "username" not in basic_auth:
                 raise ValueError("username required in basic_auth")
-            if (
-                "password" not in basic_auth
-                and "password_file" not in basic_auth
-            ):
+            if "password_file" in basic_auth:
+                if "password" in basic_auth:
+                    raise ValueError(
+                        "basic_auth cannot contain password and password_file"
+                    )
+                with open(basic_auth["password_file"]) as file:
+                    basic_auth["password"] = file.readline().strip()
+            elif "password" not in basic_auth:
                 raise ValueError("password required in basic_auth")
-            if "password" in basic_auth and "password_file" in basic_auth:
-                raise ValueError(
-                    "basic_auth cannot contain password and password_file"
-                )
         self._basic_auth = basic_auth
 
     @property
@@ -159,10 +162,20 @@ class PrometheusRemoteWriteMetricsExporter(MetricsExporter):
     def export(
         self, export_records: Sequence[ExportRecord]
     ) -> MetricsExportResult:
-        raise NotImplementedError()
+        if not export_records:
+            return MetricsExportResult.SUCCESS
+        timeseries = self._convert_to_timeseries(export_records)
+        if not timeseries:
+            logger.error(
+                "All records contain unsupported aggregators, export aborted"
+            )
+            return MetricsExportResult.FAILURE
+        message = self._build_message(timeseries)
+        headers = self._build_headers()
+        return self._send_message(message, headers)
 
     def shutdown(self) -> None:
-        raise NotImplementedError()
+        pass
 
     def _convert_to_timeseries(
         self, export_records: Sequence[ExportRecord]
@@ -304,13 +317,60 @@ class PrometheusRemoteWriteMetricsExporter(MetricsExporter):
         timeseries.samples.append(sample)
         return timeseries
 
-    def build_message(self, timeseries: Sequence[TimeSeries]) -> bytes:
-        raise NotImplementedError()
+    def _build_message(self, timeseries: Sequence[TimeSeries]) -> bytes:
+        write_request = WriteRequest()
+        write_request.timeseries.extend(timeseries)
+        serialized_message = write_request.SerializeToString()
+        return snappy.compress(serialized_message)
 
-    def get_headers(self) -> Dict:
-        raise NotImplementedError()
+    def _build_headers(self) -> Dict:
+        headers = {
+            "Content-Encoding": "snappy",
+            "Content-Type": "application/x-protobuf",
+            "X-Prometheus-Remote-Write-Version": "0.1.0",
+        }
+        if self.headers:
+            for header_name, header_value in self.headers.items():
+                headers[header_name] = header_value
+        return headers
 
-    def send_message(
+    def _send_message(
         self, message: bytes, headers: Dict
     ) -> MetricsExportResult:
-        raise NotImplementedError()
+        auth = None
+        if self.basic_auth:
+            auth = (self.basic_auth["username"], self.basic_auth["password"])
+
+        cert = None
+        verify = True
+        if self.tls_config:
+            if "ca_file" in self.tls_config:
+                verify = self.tls_config["ca_file"]
+            elif "insecure_skip_verify" in self.tls_config:
+                verify = self.tls_config["insecure_skip_verify"]
+
+            if (
+                "cert_file" in self.tls_config
+                and "key_file" in self.tls_config
+            ):
+                cert = (
+                    self.tls_config["cert_file"],
+                    self.tls_config["key_file"],
+                )
+        try:
+            response = requests.post(
+                self.endpoint,
+                data=message,
+                headers=headers,
+                auth=auth,
+                timeout=self.timeout,
+                proxies=self.proxies,
+                cert=cert,
+                verify=verify,
+            )
+            if not response.ok:
+                response.raise_for_status()
+        except requests.exceptions.RequestException as e:
+            logger.error("Export POST request failed with reason: %s", e)
+            return MetricsExportResult.FAILURE
+        return MetricsExportResult.SUCCESS
--- a/exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
+++ b/exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
@@ -360,22 +360,55 @@ class TestConversion(unittest.TestCase):
 class TestExport(unittest.TestCase):
     # Initializes test data that is reused across tests
     def setUp(self):
-        pass
+        self.exporter = PrometheusRemoteWriteMetricsExporter(
+            endpoint="/prom/test_endpoint"
+        )
 
     # Ensures export is successful with valid export_records and config
-    def test_export(self):
-        pass
+    @patch("requests.post")
+    def test_valid_export(self, mock_post):
+        mock_post.return_value.configure_mock(**{"status_code": 200})
+        test_metric = Counter("testname", "testdesc", "testunit", int, None)
+        labels = get_dict_as_key({"environment": "testing"})
+        record = ExportRecord(
+            test_metric, labels, SumAggregator(), Resource({})
+        )
+        result = self.exporter.export([record])
+        self.assertIs(result, MetricsExportResult.SUCCESS)
+        self.assertEqual(mock_post.call_count, 1)
 
-    def test_valid_send_message(self):
-        pass
+        result = self.exporter.export([])
+        self.assertIs(result, MetricsExportResult.SUCCESS)
+
+    def test_invalid_export(self):
+        record = ExportRecord(None, None, None, None)
+        result = self.exporter.export([record])
+        self.assertIs(result, MetricsExportResult.FAILURE)
+
+    @patch("requests.post")
+    def test_valid_send_message(self, mock_post):
+        mock_post.return_value.configure_mock(**{"ok": True})
+        result = self.exporter._send_message(bytes(), {})
+        self.assertEqual(mock_post.call_count, 1)
+        self.assertEqual(result, MetricsExportResult.SUCCESS)
 
     def test_invalid_send_message(self):
-        pass
+        result = self.exporter._send_message(bytes(), {})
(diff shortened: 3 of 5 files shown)


exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -0,0 +1,353 @@
...
+    ) -> MetricsExportResult:
+        timeseries = self.convert_to_timeseries(export_records)
+        message = self.build_message(timeseries)
+        headers = self.get_headers()
+        return self.send_message(message, headers)
+
+    def shutdown(self) -> None:
+        pass
+
+    def convert_to_timeseries(
```
ocelotl, line 165: Forgot to ask before, do these `convert_` methods need to be public? If not I would recommend to use a leading underscore to avoid making API commitments.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
```diff
@@ -0,0 +1,377 @@
...
+        expected_timeseries.samples.append(
+            self._exporter.create_sample(10, 5.0),
+        )
+        timeseries = self._exporter.create_timeseries(
+            export_record, "testname", 5.0,
+        )
+        self.assertEqual(timeseries, expected_timeseries)
+
+
+class ResponseStub:
```
ocelotl, line 316: I would recommend using `Mock`s as much as possible. This class can be replaced with a dictionary that is later used as the argument to `configure_mock`. Having classes like this one hides the fact that they are only for testing purposes, even when their name attempts to indicate so.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
```diff
@@ -0,0 +1,377 @@
...
+    @mock.patch("requests.post", return_value=ResponseStub(404))
+    def test_invalid_send_message(self, mock_post):
+        result = self._exporter.send_message(bytes(), {})
+        self.assertEqual(mock_post.call_count, 1)
+        self.assertEqual(result, MetricsExportResult.FAILURE)
+
+    # Verifies that build_message calls snappy.compress and returns SerializedString
+    @mock.patch("snappy.compress", return_value=bytes())
+    def test_build_message(self, mock_compress):
+        test_timeseries = [
```
ocelotl, line 357: Nit: single use variable here.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
```diff
@@ -0,0 +1,377 @@
...
+        message = self._exporter.build_message(test_timeseries)
+        self.assertEqual(mock_compress.call_count, 1)
+        self.assertIsInstance(message, bytes)
+
+    # Ensure correct headers are added when valid config is provided
+    def test_get_headers(self):
+        self._exporter.headers = {"Custom Header": "test_header"}
+
+        headers = self._exporter.get_headers()
+        self.assertEqual(headers.get("Content-Encoding", ""), "snappy")
```
ocelotl, line 370: This can be misleading if the test case fails. If `"Content-Encoding"` was supposed to be in the keys of `header`, then you should not artificially hide that fact under a different exception. If `"Content-Encoding"` is not in the keys of `header`, then you should be raising a `KeyError` and not an `AssertionError`. This is because if the test case fails, the current implementation of this assertion does not make it clear if `"Content-Encoding"` was not in the keys of `header` or if its original value was `""`.

ocelotl (changes requested):
Some test case result interpretations can be affected by the current implementation, please take a look.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Added a leading underscore to all methods except `export` and `shutdown` which the API describes.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
shovnik: Got rid of this Stub to use the Mock class combined with configure_mock as suggested. Thanks for the tip!

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
shovnik: Refactored

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
shovnik: Good point, fixed. I did not realize the standard function errors do not break the test.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
```diff
@@ -0,0 +1,416 @@
...
+
+class TestExport(unittest.TestCase):
+    # Initializes test data that is reused across tests
+    def setUp(self):
+        self.exporter = PrometheusRemoteWriteMetricsExporter(
+            endpoint="/prom/test_endpoint"
+        )
+
+    # Ensures export is successful with valid export_records and config
+    @patch("requests.post", return_value=Mock())
```
ocelotl, line 369: I think that this should work, please give it a try:

```suggestion
    @patch("requests.post")
```

The patched object should be replaced by a [`MagicMock`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) in this case.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
ocelotl: Any exception raised in a test will make it fail. What we don't want is to introduce ambiguity in a test case failure because the same exception (`AssertionError`) would be raised if the key was missing from `headers` or if its value was `""`. Both are wrong but should raise different exceptions to clearly indicate what went wront.

ocelotl (changes requested):
Well done! Just a request to remove unnecessary `Mock`s.

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
shovnik: It worked. Removed unnecessary specified return_value.

ocelotl (approved):
Great job!

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -159,10 +162,16 @@ def headers(self, headers: Dict):
     def export(
         self, export_records: Sequence[ExportRecord]
     ) -> MetricsExportResult:
-        raise NotImplementedError()
+        timeseries = self._convert_to_timeseries(export_records)
+        if not timeseries:
+            logger.warning("No valid records found, export aborted")
+            return MetricsExportResult.FAILURE
```
lzchen, line 168: Is it really a failure of there are no records to be exported?

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -304,13 +313,62 @@ def add_label(label_name: str, label_value: str):
...
+            data=message,
+            headers=headers,
+            auth=auth,
+            timeout=self.timeout,
+            proxies=self.proxies,
+            cert=cert,
+            verify=verify,
+        )
+        if response.status_code != 200:
+            logger.warning(
```
lzchen, line 367:
```suggestion
            logger.error(
```

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -304,13 +313,62 @@ def add_label(label_name: str, label_value: str):
...
+            self.endpoint,
+            data=message,
+            headers=headers,
+            auth=auth,
+            timeout=self.timeout,
+            proxies=self.proxies,
+            cert=cert,
+            verify=verify,
+        )
+        if response.status_code != 200:
```
lzchen, line 366: Anything that is not 200 is an error?

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -304,13 +313,62 @@ def add_label(label_name: str, label_value: str):
...
+            logger.warning(
+                "POST request failed with status %s with reason: %s and content: %s",
+                str(response.status_code),
+                response.reason,
+                str(response.content),
+            )
+            return MetricsExportResult.FAILURE
```
lzchen, line 373: Are we going to be handling retries? Timeouts?, etc?

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: This would only occur if all provided records contain only unsupported aggregators, so I figured it would better to signal it as a failure than continuously exporting nothing despite receiving records. I am adding a check at the start to return success if no records I provided to handle that valid case.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Fixed

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Also changed the log to an error and made issue more clear in message.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: We don't expect any positive response to not have a status_code of 200, but I suppose we got just check `if response` which evaluates 200-399 as True and greater than 400 as False. Are we overlooking a possible success that does not have a status_code of 200?

shovnik: Changed to check `if response.ok`

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: The timeout can be adjusted, but if a request fails (timeout or otherwise), we just report failure instead of retrying to avoid requests getting clogged due to records being exported continuously. If we set a timeout, but then retried multiple times, the timeout would no longer limit the time spent attempting to export one set of records. Our reasoning was that missing data is not as bad as delayed data especially in cases of cumulative data where missing data can be interpolated or cases where data is exported on a very short interval. Does not retrying failed requests sound reasonable for this use case?

lzchen: Depends on how robust we want the functionality of this exporter. We can handle it in a different PR in the future if is needed.

lzchen (approved):

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -304,13 +317,62 @@ def add_label(label_name: str, label_value: str):
...
+
+            if (
+                "cert_file" in self.tls_config
+                and "key_file" in self.tls_config
+            ):
+                cert = (
+                    self.tls_config["cert_file"],
+                    self.tls_config["key_file"],
+                )
+        response = requests.post(
```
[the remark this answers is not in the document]
shovnik: I did not consider that requests.post could throw exceptions in addition to returning HTTPErrors inside the response. Modified to use a try/except block which handles all request exceptions. (now raises HTTPErrors to be logged as well) This guarantees calling export() will never throw an exception.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -304,13 +317,62 @@ def add_label(label_name: str, label_value: str):
...
+            self.endpoint,
+            data=message,
+            headers=headers,
+            auth=auth,
+            timeout=self.timeout,
```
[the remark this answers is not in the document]
shovnik: Added, will also add to README

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -16,6 +16,9 @@
 import re
 from typing import Dict, Sequence
 
+import requests
```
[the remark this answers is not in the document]
shovnik: One issue that I am running into is that I cannot add python-snappy to the setup.cfg because it depends on the snappy library in C which I had to brew install to get. Is there a way to add a C dependency that a module requires? If not I might have to specify to first manually install snappy library and module before using exporter in README.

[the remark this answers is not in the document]
shovnik: Excluded python-snappy from install-requires for now.

exporter/opentelemetry-exporter-prometheus-remote-write/setup.cfg
```diff
@@ -39,6 +39,8 @@ package_dir=
     =src
 packages=find_namespace:
 install_requires =
+    protobuf >= 3.13.0
+    requests == 2.25.0
     opentelemetry-api == 0.16.dev0
```
lzchen, line 44: Could you update these dependencies to 0.17.dev0 as well as the current version? It's causing some build failures.

codeboten (approved):
Thanks for responding to all my comments.

exporter/opentelemetry-exporter-prometheus-remote-write/setup.cfg
shovnik: Updated

PR #216, opened 2020-11-25
Prometheus Remote Write Exporter (5/6)

AzfaarQureshi:
# Description
This is PR 5/6 of adding a Prometheus Remote Write Exporter in Python SDK and address Issue https://github.com/open-telemetry/opentelemetry-python/issues/1302

* Adds class skeleton

**[Part 3/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/207)**

**[Part 4/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/212)**

👉 **[Part 5/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/216)**
* Add docker integration tests

**[Part 6/6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/227)**
* Set up example app

## Type of change
# How Has This Been Tested?

- [ ] Added class `TestPrometheusRemoteWriteExporterCortex` in `opentelemetry-docker-tests`

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:
- [ ] ~~Documentation has been updated~~
cc- @shovnik, @alolita

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#237](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/237))
 - `opentelemetry-instrumentation-starlette` Added support for excluding some routes with env var `OTEL_PYTHON_STARLETTE_EXCLUDED_URLS`
   ([#237](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/237))
-
+- Add Prometheus Remote Write Exporter integration tests in opentelemetry-docker-tests
+  ((#216)[https://github.com/open-telemetry/opentelemetry-python-contrib/pull/216])
 
 ### Changed
 - `opentelemetry-instrumentation-asgi`, `opentelemetry-instrumentation-wsgi` Return `None` for `CarrierGetter` if key not found
@@ -92,7 +93,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#1239](https://github.com/open-telemetry/opentelemetry-python/pull/1239))
 - `opentelemetry-instrumentation-django` Bugfix use request.path replace request.get_full_path(). It will get correct span name
   ([#1309](https://github.com/open-telemetry/opentelemetry-python/pull/1309#))
-- `opentelemetry-instrumentation-django` Record span status and http.status_code attribute on exception 
+- `opentelemetry-instrumentation-django` Record span status and http.status_code attribute on exception
   ([#1257](https://github.com/open-telemetry/opentelemetry-python/pull/1257))
 - `opentelemetry-instrumentation-grpc` Rewrite gRPC server interceptor
   ([#1171](https://github.com/open-telemetry/opentelemetry-python/pull/1171))
@@ -120,7 +121,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-instrumentation-pymysql` Bumped version from 0.9.3 to 0.10.1
   ([#1228](https://github.com/open-telemetry/opentelemetry-python/pull/1228))
 - `opentelemetry-instrumentation-django` Changed span name extraction from request to comply semantic convention
-  ([#992](https://github.com/open-telemetry/opentelemetry-python/pull/992)) 
+  ([#992](https://github.com/open-telemetry/opentelemetry-python/pull/992))
 
 ## [0.13b0](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v0.13b0) - 2020-09-17
 
@@ -264,7 +265,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.8b0](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v0.8.0) - 2020-05-27
 
 ### Added
-- `opentelemetry-ext-datadog` Add exporter to Datadog 
+- `opentelemetry-ext-datadog` Add exporter to Datadog
   ([#572](https://github.com/open-telemetry/opentelemetry-python/pull/572))
 - `opentelemetry-ext-sqlite3` Initial release
 - `opentelemetry-ext-psycopg2` Implement instrumentor interface, enabling auto-instrumentation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,19 +12,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-sdk-extension-aws` Add method to return fields injected by propagator
   ([#226](https://github.com/open-telemetry/opentelemetry-python/pull/226))
 - `opentelemetry-exporter-prometheus-remote-write` Prometheus Remote Write Exporter Setup
-  ((#180)[https://github.com/open-telemetry/opentelemetry-python-contrib/pull/180])
-- `opentelemetry-exporter-prometheus-remote-write` Add Exporter constructor validation methods
-  ((#206)[https://github.com/open-telemetry/opentelemetry-python-contrib/pull/206])
-- `opentelemetry-exporter-prometheus-remote-write` Add conversion to TimeSeries methods
-  ((#207)[https://github.com/open-telemetry/opentelemetry-python-contrib/pull/207])
-- `opentelemetry-exporter-prometheus-remote-write` Add request methods
-  ((#212)[https://github.com/open-telemetry/opentelemetry-python-contrib/pull/212])
+  ([#180](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/180))
+- `opentelemetry-exporter-prometheus-remote-write` Add Exporter constructor validation methods in Prometheus Remote Write Exporter
+  ([#206](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/206))
+- `opentelemetry-exporter-prometheus-remote-write` Add conversion to TimeSeries methods in Prometheus Remote Write Exporter
+  ([#207](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/207))
+- `opentelemetry-exporter-prometheus-remote-write` Add request methods to Prometheus Remote Write Exporter
+  ([#212](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/212))
 - `opentelemetry-instrumentation-fastapi` Added support for excluding some routes with env var `OTEL_PYTHON_FASTAPI_EXCLUDED_URLS`
   ([#237](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/237))
 - `opentelemetry-instrumentation-starlette` Added support for excluding some routes with env var `OTEL_PYTHON_STARLETTE_EXCLUDED_URLS`
   ([#237](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/237))
 - Add Prometheus Remote Write Exporter integration tests in opentelemetry-docker-tests
-  ((#216)[https://github.com/open-telemetry/opentelemetry-python-contrib/pull/216])
+  ([#216](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/216))
 
 ### Changed
 - `opentelemetry-instrumentation-asgi`, `opentelemetry-instrumentation-wsgi` Return `None` for `CarrierGetter` if key not found
--- a/tests/opentelemetry-docker-tests/tests/prometheus-remote-write-cortex/prometheus_remote_write_cortex_test.py
+++ b/tests/opentelemetry-docker-tests/tests/prometheus-remote-write-cortex/prometheus_remote_write_cortex_test.py
@@ -1,0 +1,116 @@
+# Copyright The OpenTelemetry Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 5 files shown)


exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -0,0 +1,354 @@
...
+        serialized_message = write_request.SerializeToString()
+        return snappy.compress(serialized_message)
+
+    def get_headers(self) -> Dict:
+        headers = {
+            "Content-Encoding": "snappy",
+            "Content-Type": "application/x-protobuf",
+            "X-Prometheus-Remote-Write-Version": "0.1.0",
+        }
+        if self.headers:
```
ocelotl, line 303: This is pretty confusing, the method is named `get_headers` but it also sets `self.headers`. Maybe this method should be split?

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -0,0 +1,354 @@
...
+            for header_name, header_value in self.headers.items():
+                headers[header_name] = header_value
+        return headers
+
+    def send_message(
+        self, message: bytes, headers: Dict
+    ) -> MetricsExportResult:
+        auth = None
+        if self.basic_auth:
+            basic_auth = self.basic_auth
```
ocelotl, line 313: Why declare a new variable here?

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -0,0 +1,354 @@
...
+            if "password" in basic_auth:
+                auth = (basic_auth.username, basic_auth.password)
+            else:
+                with open(basic_auth.password_file) as file:
```
ocelotl, line 317: Maybe save the password from the file in an attribute to avoid having to open an close a file several times if this is called more than once?

tests/opentelemetry-docker-tests/tests/prometheus-remote-write-cortex/prometheus_remote_write_cortex_test.py
```diff
@@ -0,0 +1,71 @@
...
+            endpoint="http://localhost:9009/api/prom/push",
+            headers={"X-Scope-Org-ID": "5"},
+        )
+        self.labels = {"environment": "testing"}
+        self.meter = self.meter_provider.get_meter(__name__)
+        metrics.get_meter_provider().start_pipeline(
+            self.meter, self.exporter, 1,
+        )
+
+    def test_export_counter(self):
```
ocelotl, line 28: What is being tested here? There is no assertion. If what you want to test is that no exceptions were raised use [`fail`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.fail). More information [here](https://stackoverflow.com/a/4319870).

ocelotl (changes requested):
Some comments regarding testing.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Renamed the method to build_headers since that's what it is really doing. The old name did seem like a getter.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Just a forgotten variable after a refactor, fixed.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
shovnik: Good catch, I now just set the password attribute using the file in the setter to avoid ever reopening the file.

tests/opentelemetry-docker-tests/tests/prometheus-remote-write-cortex/prometheus_remote_write_cortex_test.py
shovnik: We originally had the pattern you linked but were recommended internally to let the test Error out naturally since that fails as well. The main issue we found with catch ExceptionType was it only handled one Exception type. Is this wrong?

ocelotl: Yes, it is wrong.

Every test case *must* have at least one assertion, since that indicates what is the intention of the test case. Even if your intention is _test that this code does not raise an exception_ then that must be made obvious by using an assertion. It is not the same thing when a test case fails because, let's say, an `IndexError` was raised because a queried sequence was too short than when a test case fails with an `AssertionError` because some counter was less than expected. The first situation tells you that the actual testing was not completed because something else went wrong in the process of testing. The second one tells you that you were able to reach the point where you could do the testing that you wanted.

For example, imagine you are testing the performance of a server. As every test case, it begins with a criteria that must be met for the test to pass. Let's say, the server must be able to reply back to 100 requests in a second. So, your test case basically consists in this:

1. Start the server
2. Send the server 100 requests in one second
3. *Assert* that you got 100 replies

If your test case fails with an `AssertionError` whose message is something like `We only got 94 replies`, then you know that the server is not performing as expected **and your testing process can be considered complete**. If your test case fails with `ImportError` then you can possibly figure out that some dependency of the server code is not installed in your testing environment. But the most important thing that you know from this second scenario is that **you did not test what you wanted to test**. That means you can not yet make any conclusions about the performance of the server because you got an exception different from an `AssertionError`. Only this kind of exception should be telling you that the test case failed because what you wanted to test did not comply with what was expected from it. A testing process is only complete if there are 0 failures or if every failure that there is is caused by an `AssertionError`. Every other exception raised is telling you, we were not able to perform the test, fix the issue and **run again**.

This makes sense because a tester can't report back to a developer any other result than a pass or a failure with an `AssertionError`. If a tester reports back an `ImportError` the developer will argue rightly that the code under test was not tested so it can't be considered as being at fault.

So, to summarize, it is perfectly fine that your testing criteria is _this code must run without raising any exception_. Even so, you must **assert** that, so that it is obvious to anyone that reads your test case what the intention of the test case is.

Something else, Python allows you to [catch multiple exceptions types](https://docs.python.org/3/tutorial/errors.html#handling-exceptions). Also keep in mind that what except does is pretty much the same as `isinstance` does, so you can catch multiple exceptions if they share a parent:

``` python
class Parent(Exception):
    pass


class Child0(Parent):
    pass


class Child1(Parent):
    pass


try:
    raise Child0()
except Parent:
    print("Caught a Child0")

try:
    raise Child1()
except Parent:
    print("Caught a Child1")
```
```
Caught a Child0
Caught a Child1
```

ocelotl (changes requested):

tests/opentelemetry-docker-tests/tests/prometheus-remote-write-cortex/prometheus_remote_write_cortex_test.py
ocelotl: By the way, pytest makes it confusing because every exception raised in a test case is reported as a failure. In my opinion, this situation should be reported as an `ERROR` instead of a `FAILURE`. An `ERROR` is something that went wrong while getting ready to test something. This is what happens when an exception is raised in a [fixture](https://docs.pytest.org/en/stable/fixture.html?highlight=fixture#pytest-fixtures-explicit-modular-scalable), which is exactly that, code that is run to get something ready for testing. Here is an example:

``` python
from pytest import fixture


@fixture
def preparation():
    1 / 0


def test_case(preparation):

    assert 1 > 0
```
```
pytest test_error.py 
===================================================================================== test session starts ======================================================================================
platform linux -- Python 3.8.3, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
rootdir: /home/ocelotl/lightstep/metric_language_equivalent_tests/python/varman
plugins: cov-2.10.0
collected 1 item                                                                                                                                                                               

test_error.py E                                                                                                                                                                          [100%]

============================================================================================ ERRORS ============================================================================================
_________________________________________________________________________________ ERROR at setup of test_case __________________________________________________________________________________

    @fixture
    def preparation():
>       1 / 0
E       ZeroDivisionError: division by zero

test_error.py:6: ZeroDivisionError
=================================================================================== short test summary info ====================================================================================
ERROR test_error.py::test_case - ZeroDivisionError: division by zero
======================================================================================= 1 error in 0.14s =======================================================================================
```

As you can see here, the result of the run is `ERROR` and not `FAILURE` because pytest is telling us that it was unable to do the actual testing because something went wrong when getting ready to run the actual test. `ERROR` means the _tester_ must fix something, `FAILURE` means the _developer_ must fix something.

Maybe someday I'll write a pytest plugin that will make every non `AssertionError` exception to be reported as an `ERROR` :stuck_out_tongue:

AzfaarQureshi: ahh okay, thanks for the detailed answer! I will fix it rn :D

shovnik: Although I am not the one making the face, thanks for the detailed response. It was very insightful.

ocelotl: My pleasure, fantastic job on getting all these PRs done @AzfaarQureshi @shovnik :muscle:

ocelotl (approved):

tests/opentelemetry-docker-tests/tests/prometheus-remote-write-cortex/prometheus_remote_write_cortex_test.py
```diff
@@ -0,0 +1,102 @@
+from opentelemetry import metrics
```
codeboten, line 1: file should have the copyright at the top

codeboten (approved):
Change looks good, just one comment about the missing copyright header.

One question: is there any way to test that the data on the cortex side matches what's expected?

tests/opentelemetry-docker-tests/tests/prometheus-remote-write-cortex/prometheus_remote_write_cortex_test.py
AzfaarQureshi: oops, will add!

AzfaarQureshi:
> One question: is there any way to test that the data on the cortex side matches what's expected?

@codeboten I could query Cortex to make sure I've gotten the data. However, that would require putting a sleep() in code to ensure that I query cortex after the export cycle has been completed. I asked in gitter and it seemed like this was against best practice so I removed it. 

Instead, I added an example app in PR 6/6 which is setup with Grafana. This way you could create metrics using the sample app, store them in Cortex, and hop over to Grafana to visualize the data. 

Would you prefer me adding the sleep so I can query cortex within the integration test or is the current approach fine? Or is there an alternative approach im missing?

alolita:
@AzfaarQureshi @shovnik please make sure the OTEL authors copyright is added to add source files. 

@codeboten thanks for flagging. I'd like to see a bot for copyright checks. Will file an issue.

exporter/opentelemetry-exporter-prometheus-remote-write/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Changelog
+
+## Unreleased
```
lzchen, line 3: I think after [this](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/239) was merged, we are using a consolidated CHANGELOG file now. Please rebase.

CHANGELOG.md
```diff
@@ -12,18 +12,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-  ((#207)[https://github.com/open-telemetry/opentelemetry-python-contrib/pull/207])
-- `opentelemetry-exporter-prometheus-remote-write` Add request methods
-  ((#212)[https://github.com/open-telemetry/opentelemetry-python-contrib/pull/212])
+  ([#180](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/180))
+- `opentelemetry-exporter-prometheus-remote-write` Add Exporter constructor validation methods in Prometheus Remote Write Exporter
+  ([#206](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/206))
+- `opentelemetry-exporter-prometheus-remote-write` Add conversion to TimeSeries methods in Prometheus Remote Write Exporter
+  ([#207](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/207))
+- `opentelemetry-exporter-prometheus-remote-write` Add request methods to Prometheus Remote Write Exporter
+  ([#212](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/212))
```
AzfaarQureshi, line 21: Noticed that the markdown was incorrect for some of the previous Remote Write Exporter links `(text)[url]` instead of `[text](url)`

exporter/opentelemetry-exporter-prometheus-remote-write/CHANGELOG.md
AzfaarQureshi: fixed!

PR #233, opened 2020-12-04
Return none for Getter if key does not exist

lzchen:
Part of https://github.com/open-telemetry/opentelemetry-python/issues/1274

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ on:
     - 'release/*'
   pull_request:
 env:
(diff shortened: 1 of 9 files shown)


codeboten (commented):
Are tests already catching this? If not it would be good to have tests that do.

lzchen:
@codeboten 
Yes each of these already has the empty use case test.

lzchen:
@codeboten 
For asgi, the test is already covering this but I added headers for a more explicit check. For celery and wsgi, those cases will be covered by the propagator tests.

lzchen:
@codeboten 
Nvm, I added tests.

codeboten (approved):

PR #235, opened 2020-12-05
Fix AttributeError: __aexit__ for aiopg.connect and aiopg.create_pool

lonewolf3739:
# Description

Function `async def wrap_connect_(...)` used  for patch `wrapt.wrap_function_wrapper(aiopg, "connect", wrap_connect_)` is an awaitable function. It works when aiopg is connection is made using await `conn = await aiopg.connect(...)` but raises error when used with context manager `async with aiopg.connect(...) as conn`.

It is very common to use with context manager to automatically close conn. This makes the change from awaitable function to context manager for both `aiopg.connect` and `aiopg.create_pool`. It works for both the cases.

Fixes https://github.com/open-telemetry/opentelemetry-python/issues/1448

## Type of change

# How Has This Been Tested?

- [x] `tox -e test-instrumentation-aiopg`


lonewolf3739:
More context: Here is the actual implementation of aiopg which returns the context manager for [.connect](https://github.com/aio-libs/aiopg/blob/master/aiopg/connection.py#L28) and [.create_pool](https://github.com/aio-libs/aiopg/blob/master/aiopg/pool.py#L19)

lonewolf3739:
roses are red, ci is green, review this sometime maybe?

lzchen (approved):

Issue #154, opened 2020-10-01
Add Datadog metrics exporter


ocervell:
Add an exporter to ship OT metrics to Datadog. Currently only the trace exporter is available.

alertedsnake:
I'm working on this one!

ocervell:
@alertedsnake Any updates for this ? Gentle ping :)

alertedsnake:
Not really - I spent a lot of time working on the [bugs](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/262) that prevented me from using this stuff for real, and then just ran out of time.  I do have a branch where I started working on it, but I'm not sure I'm going to continue - my company has been shifting towards Prometheus, so I think that exporter is going to handle our use-case once we get to using this in production.

Here's the branch if someone wants to take over:
https://github.com/alertedsnake/opentelemetry-python-contrib/tree/datadog-metrics

srikanthccv:
See https://github.com/open-telemetry/opentelemetry-python-contrib/pull/900

PR #236, opened 2020-12-05
Add more tests, fix compliance to semantics

### LINKED ISSUE #139: Make sure rpc instrumentations follow semantic conventions
Specs: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md

### LINKED ISSUE #173: Need Unit Tests for gRPC server instrumentation
In PR open-telemetry/opentelemetry-python#1308 the Spans created as part of instrumenting gRPC were corrected to use the error codes as recently updated in [the OTel Specifications](https://github.com/open-telemetry/opentelemetry-python/pull/1308#discussion_r514807784) instead of the gRPC error codes.

This didn't break any unit tests which is not ideal, we should add unit tests so that changes like these are flagged in the tests and require a test change.

alertedsnake:
# Description

## A whole bunch of reworking the tests

I've updated tests to use proper handler names - In the real world, the interceptor is always going to be passed a HandlerCallDetails object with the `method` field set to something valid... previously these tests set this to an empty string, but that's not what happens in a real case. So I've updated the tests to do that right, but also made sure that the span processing allows this to be unset, just in case someone was expecting this behavior to be present.

I've added more checks into the existing tests, now we're checking that span attributes are set as expected.

Fixed a few bugs in the tests where things which were intended to be checked, actually were not.

And finally, I've added a test to catch the gRPC `abort()` and verify the error status in the span.

## Updated status codes to be compliant with specifications

The previous PR I made had the gRPC status codes using the English names for the status, but the spec has been clarified to require the integer value.

Also, the specs say to not use `span.set_status()` for a non-error condition, so I've added a check for that in a few places.

## Added additional attributes

Also as the spec has been clarified, I've added the `rpc.method` and `rpc.service` attributes.

Fixes #173
Fixes #139 

# Type of change

# How Has This Been Tested?

As most of this is either test changes or changes covered by new or existing tests, I've just run a bunch of tests :)

- [X] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,10 +25,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#237](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/237))
 - Add Prometheus Remote Write Exporter integration tests in opentelemetry-docker-tests
   ([#216](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/216))
+- `opentelemetry-instrumentation-grpc` Add tests for grpc span attributes, grpc `abort()` conditions
+  ([#236](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/236))
 
 ### Changed
 - `opentelemetry-instrumentation-asgi`, `opentelemetry-instrumentation-wsgi` Return `None` for `CarrierGetter` if key not found
   ([#1374](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/233))
+- `opentelemetry-instrumentation-grpc` Comply with updated spec, rework tests
+  ([#236](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/236))
 
 ## [0.16b1](https://github.com/open-telemetry/opentelemetry-python-contrib/releases/tag/v0.16b1) - 2020-11-26
 
--- a/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py
+++ b/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py
@@ -113,7 +113,7 @@ class _OpenTelemetryServicerContext(grpc.ServicerContext):
     def abort(self, code, details):
         self.code = code
         self.details = details
-        self._active_span.set_attribute("rpc.grpc.status_code", code.name)
+        self._active_span.set_attribute("rpc.grpc.status_code", code.value[0])
         self._active_span.set_status(
             Status(status_code=StatusCode.ERROR, description=details)
         )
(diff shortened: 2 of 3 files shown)


alertedsnake:
(And now replying properly with my real github account)

Confirmed a slight issue - I'm setting the span.status.description field to the error description, but it seems that's expected to be type: detail in the case of an exception, which is a bit harder to do with gRPC, since any abort is raised as a bare Exception even when it isn't technically. I suppose I could make this end up something like StatusCode.NAME: description (with name and description set appropriately).

Anyone know where the canonical doc on that format is?

This shows up because of this line here: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/master/exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/exporter.py#L252

I did notice #154 and was pondering tackling that as well, so if this is considered an error, I can fix it there instead :)

alertedsnake:
Ah, I know what to do, hang on.

alertedsnake:
I should mention that the lint test runs fine locally...

toumorokoshi:
> This shows up because of this line here: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/master/exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/exporter.py#L252

I think this is just an incorrect assumption that was made because the default exception handling logic in the SDK is to format the exception that way.

Really exceptions should be [recorded as events](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#record-exception), so the datadog_exporter should actually use that to extract it instead.

The python code [already does this](https://github.com/open-telemetry/opentelemetry-python/blob/master/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L690-L715).

I'd recommend for now just modifying the datadog_exporter to not make that assumption, and filing a bug against it so it can be fixed in the future. 

What do you think?

toumorokoshi (approved):
LGTM! IT would be great if you could file that issue on the dd exporter, but all of your changes are correct and technically the DD issue could occur regardless of merging this change in.

toumorokoshi:
> I should mention that the lint test runs fine locally...

regarding this, have you updated your version of PIP? There's a new version that now has strict ordering and also results in detecting conflicting dependencies consistency.

I see this in the tox lint output:

```
ERROR: Cannot install opentelemetry-exporter-datadog[test]==0.17.dev0 and opentelemetry-exporter-prometheus-remote-write[test]==0.16.dev0 because these package versions have conflicting dependencies.

The conflict is caused by:
    opentelemetry-exporter-datadog[test] 0.17.dev0 depends on opentelemetry-api==0.17.dev0
    opentelemetry-exporter-prometheus-remote-write[test] 0.16.dev0 depends on opentelemetry-api==0.16.dev0
```

I think you just need to rebase your commit to fix that. [your branch has 16.dev0 as the prometheus writer version](https://github.com/alertedsnake/opentelemetry-python-contrib/blob/fix-semantics/exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/version.py#L15). head of the main branch looks to be 0.17.dev0.

alertedsnake:
> I think you just need to rebase your commit to fix that. [your branch has 16.dev0 as the prometheus writer version](https://github.com/alertedsnake/opentelemetry-python-contrib/blob/fix-semantics/exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/version.py#L15). head of the main branch looks to be 0.17.dev0.

Oh!  That's a good clue, thanks for catching that, I'll try it.

alertedsnake:
> I'd recommend for now just modifying the datadog_exporter to not make that assumption, and filing a bug against it so it can be fixed in the future.
> 
> What do you think?

Good call.  I've already started working on the Datadog exporter to add metrics (#154), so I'll file an issue and then fix it while I'm in there.

alertedsnake:
> Good call. I've already started working on the Datadog exporter to add metrics (#154), so I'll file an issue and then fix it while I'm in there.

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/master/exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/exporter.py#L182

Ooh boy, got my work cut out for me :)

PR #246, opened 2020-12-10
psycopg2 and dbapi instrumentaiton fixes

### LINKED ISSUE #143: Instrumentation breaks psycopg2 when registering types.
**Describe your environment** 
PostgreSQL supports JSON and JSONB types of DB columns. psycopg2 supports registering custom typecasters for these types. Most of the registration functions accept a connection or a cursor instance. The library crashes if anything else is passed in place of a connection or a cursor. 

Most higher level frameworks and ORMs support these features out of the box meaning they crash as well. For example, django projects using psycopg2 driver will crash. 

**Steps to reproduce**

Running the following program
```python
import psycopg2.extras
from opentelemetry.instrumentation.psycopg2 import Psycopg2Instrumentor

Psycopg2Instrumentor().instrument()

cnx = psycopg2.connect(host='localhost', port='5432', dbname='postgres', password='1234', user='postgres')
psycopg2.extras.register_default_jsonb(conn_or_curs=cnx, loads=lambda x: x)
```

results in the following exception:

```
❯ python main.py
  File "main.py", line 10, in <module>
    psycopg2.extras.register_default_jsonb(conn_or_curs=cnx, loads=lambda x: x)
  File "/Users/olone/playground/otel-psycopg2/venv/lib/python3.7/site-packages/psycopg2/_json.py", line 156, in register_default_jsonb
    loads=loads, oid=JSONB_OID, array_oid=JSONBARRAY_OID, name='jsonb')
  File "/Users/olone/playground/otel-psycopg2/venv/lib/python3.7/site-packages/psycopg2/_json.py", line 125, in register_json
    register_type(JSON, not globally and conn_or_curs or None)
TypeError: argument 2 must be a connection, cursor or None
```

For the instrumentation to not crash psycopg2 ever.

It crashes when psycopg2 enforces connection or cursor types.

**Additional context**
This is happening because instrumentation wraps connection factories and returns wrapt.ObjectProxy instances that wrap connection or cursor objects. These objects fail the type checks psycopg2 runs internally (probably in C code).

Here: https://github.com/open-telemetry/opentelemetry-python/blob/6019a91980ec84bbf969b0d82d44483c93f3ea4c/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py#L287-L304
And here: https://github.com/open-telemetry/opentelemetry-python/blob/6019a91980ec84bbf969b0d82d44483c93f3ea4c/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py#L361-L389

The solution might be to change how DBAPI instrumentation works and patch actual connection and cursor classes instead of wrap with object proxies.

owais:
# Description

* Updated psycopg2 instrumentation to work with registered types like JSONB. This makes the instrumentation compatible with psycopg2 when registering types manually. It also enables the instrumentation to work with frameworks like Django that register types implicitly.
* Updated DBAPI to use statement and DB name as the span name as recommended by the spec instead of using the entire query.

Fixes #143

## Type of change

# How Has This Been Tested?

- [x] Updated tests and added new ones.
- [x] Manual testing with django and custom registration of types with psycopg2.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-aiopg/src/opentelemetry/instrumentation/aiopg/aiopg_integration.py
+++ b/instrumentation/opentelemetry-instrumentation-aiopg/src/opentelemetry/instrumentation/aiopg/aiopg_integration.py
@@ -4,8 +4,8 @@ import wrapt
 from aiopg.utils import _ContextManager, _PoolAcquireContextManager
 
 from opentelemetry.instrumentation.dbapi import (
+    CursorTracer,
     DatabaseApiIntegration,
-    TracedCursor,
 )
 from opentelemetry.trace import SpanKind
 from opentelemetry.trace.status import Status, StatusCode
@@ -94,25 +94,29 @@ def get_traced_pool_proxy(pool, db_api_integration, *args, **kwargs):
     return TracedPoolProxy(pool, *args, **kwargs)
 
 
-class AsyncTracedCursor(TracedCursor):
+class AsyncCursorTracer(CursorTracer):
     async def traced_execution(
         self,
+        cursor,
         query_method: typing.Callable[..., typing.Any],
         *args: typing.Tuple[typing.Any, typing.Any],
         **kwargs: typing.Dict[typing.Any, typing.Any]
     ):
         name = ""
-        if len(args) > 0 and args[0]:
-            name = args[0]
-        elif self._db_api_integration.database:
-            name = self._db_api_integration.database
-        else:
-            name = self._db_api_integration.name
+        if args:
+            name = self.get_operation_name(cursor, args)
+
+        if not name:
+            name = (
+                self._db_api_integration.database
+                if self._db_api_integration.database
+                else self._db_api_integration.name
+            )
 
         with self._db_api_integration.get_tracer().start_as_current_span(
             name, kind=SpanKind.CLIENT
         ) as span:
-            self._populate_span(span, *args)
+            self._populate_span(span, cursor, *args)
             try:
                 result = await query_method(*args, **kwargs)
                 return result
@@ -123,10 +127,10 @@ class AsyncTracedCursor(TracedCursor):
 
 
 def get_traced_cursor_proxy(cursor, db_api_integration, *args, **kwargs):
-    _traced_cursor = AsyncTracedCursor(db_api_integration)
+    _traced_cursor = AsyncCursorTracer(db_api_integration)
 
     # pylint: disable=abstract-method
-    class AsyncTracedCursorProxy(AsyncProxyObject):
+    class AsyncCursorTracerProxy(AsyncProxyObject):
 
         # pylint: disable=unused-argument
         def __init__(self, cursor, *args, **kwargs):
@@ -134,20 +138,20 @@ def get_traced_cursor_proxy(cursor, db_api_integration, *args, **kwargs):
 
         async def execute(self, *args, **kwargs):
             result = await _traced_cursor.traced_execution(
-                self.__wrapped__.execute, *args, **kwargs
+                self, self.__wrapped__.execute, *args, **kwargs
             )
             return result
 
         async def executemany(self, *args, **kwargs):
             result = await _traced_cursor.traced_execution(
-                self.__wrapped__.executemany, *args, **kwargs
+                self, self.__wrapped__.executemany, *args, **kwargs
             )
             return result
 
         async def callproc(self, *args, **kwargs):
             result = await _traced_cursor.traced_execution(
-                self.__wrapped__.callproc, *args, **kwargs
+                self, self.__wrapped__.callproc, *args, **kwargs
             )
             return result
 
-    return AsyncTracedCursorProxy(cursor, *args, **kwargs)
+    return AsyncCursorTracerProxy(cursor, *args, **kwargs)
--- a/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
@@ -60,6 +60,7 @@ def trace_integration(
     connection_attributes: typing.Dict = None,
     tracer_provider: typing.Optional[TracerProvider] = None,
     capture_parameters: bool = False,
+    db_api_integration_factory=None,
 ):
     """Integrate with DB API library.
     https://www.python.org/dev/peps/pep-0249/
@@ -86,6 +87,7 @@ def trace_integration(
         version=__version__,
         tracer_provider=tracer_provider,
         capture_parameters=capture_parameters,
+        db_api_integration_factory=db_api_integration_factory,
     )
 
 
@@ -99,6 +101,7 @@ def wrap_connect(
     version: str = "",
     tracer_provider: typing.Optional[TracerProvider] = None,
     capture_parameters: bool = False,
+    db_api_integration_factory=None,
 ):
     """Integrate with DB API library.
     https://www.python.org/dev/peps/pep-0249/
@@ -115,6 +118,9 @@ def wrap_connect(
         capture_parameters: Configure if db.statement.parameters should be captured.
 
     """
+    db_api_integration_factory = (
+        db_api_integration_factory or DatabaseApiIntegration
+    )
 
     # pylint: disable=unused-argument
     def wrap_connect_(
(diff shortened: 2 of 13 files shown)


instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -342,24 +351,41 @@ def _populate_span(
         if self._db_api_integration.capture_parameters and len(args) > 1:
             span.set_attribute("db.statement.parameters", str(args[1]))
 
+    def get_operation_name(self, cursor, args):
+        if args and isinstance(args[0], str):
+            return args[0].split(" ")[0]
+        return ""
```
lonewolf3739, line 357: > It is not recommended to attempt any client-side parsing of db.statement just to get these properties, they should only be used if the library being instrumented already provides them.

Spec explicitly says it doesn't recommend manipulating statement to get the properties. But I also agree that it has very high cardinal value without variable arguments. If we are going to parse statement may be it's worth considering something like [sqlparse](https://github.com/andialbrecht/sqlparse) and have structured way of making span name from tokens.

owais: I agree that we shouldn't parse it either but I don't think this really counts as "SQL parsing". We are making a bare-bones, extremely cheap, best effort attempt at guessing the SQL operation name. I think as long as it does not incur a perf penalty, decrease reliabilit, affect the DB connection/cursor/statement in any way or end up talking to the database, we should be good. At least that's how I understand the intention behind the spec recommendation but may be there is more to it that I might have overlooked?

lonewolf3739: Yes, this isn't really a parsing. I thought the motive behind the spec recommendation is to not derive/guess anything from db statement. Otherwise this is good and it doesn't really cost anything in terms of perf/reliability etc... May be one other point worth bringing up here is different commands like `drop database ...`, `drop table ...`, `drop index ...` and `drop view ...` etc.. will have same span name and effectively making it a few set of keywords.

lonewolf3739: And given that the most usual operations are mainly select, update, create we are good this change.

instrumentation/opentelemetry-instrumentation-psycopg2/tests/test_psycopg2_integration.py
```diff
@@ -22,15 +23,69 @@
...
+
+    executemany = mock.MagicMock(spec=types.MethodType)
+    executemany.__name__ = "executemany"
+
+    callproc = mock.MagicMock(spec=types.MethodType)
+    callproc.__name__ = "callproc"
+
+    rowcount = "SomeRowCount"
+
+    def __init__(self, *args, **kwargs):
```
ocelotl, line 39: Is this necessary?

ocelotl (approved):

instrumentation/opentelemetry-instrumentation-psycopg2/tests/test_psycopg2_integration.py
owais: I think at some point the mock was a subclass of the real cursor so this was done to disable initializing the real cursor. Will remove.

owais: OK, this is actually required to satisfy the interface :) I forgot about that.

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
aabmass: Might be worth updating the spec to recommend this cheap "parsing"?

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -342,24 +351,38 @@ def _populate_span(
         if self._db_api_integration.capture_parameters and len(args) > 1:
             span.set_attribute("db.statement.parameters", str(args[1]))
 
+    def get_operation_name(self, cursor, args):  # pylint: disable=no-self-use
+        if args and isinstance(args[0], str):
+            return args[0].split(" ")[0]
```
aabmass, line 356:
```suggestion
            return args[0].split()[0]
```

I think this (no `sep` arg) might have better behavior for variety of SQL formattings e.g.
```py
# leading whitespace
cursor.execute("""
SELECT * FROM foo
""")

# newline sep
cursor.execute(
  """SELECT
         *
       FROM foo
  """
)
```

instrumentation/opentelemetry-instrumentation-psycopg2/src/opentelemetry/instrumentation/psycopg2/__init__.py
```diff
@@ -39,12 +39,19 @@
...
+    cursor as pg_cursor,  # pylint: disable=no-name-in-module
+)
+from psycopg2.sql import Composed  # pylint: disable=no-name-in-module
 
 from opentelemetry.instrumentation import dbapi
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
 from opentelemetry.instrumentation.psycopg2.version import __version__
-from opentelemetry.trace import get_tracer
+
+OTEL_CURSOR_FACTORY_KEY = "_otel_orig_cursor_factory"
```
aabmass, line 54: nit

```suggestion
_OTEL_CURSOR_FACTORY_KEY = "_otel_orig_cursor_factory"
```

instrumentation/opentelemetry-instrumentation-dbapi/tests/test_dbapi_integration.py
```diff
@@ -50,7 +50,7 @@ def test_span_succeeded(self):
         spans_list = self.memory_exporter.get_finished_spans()
```
aabmass, line 50: Would be nice to add a test for formats mentioned above like leading space or newline separators

instrumentation/opentelemetry-instrumentation-aiopg/src/opentelemetry/instrumentation/aiopg/aiopg_integration.py
```diff
@@ -123,7 +127,7 @@ async def traced_execution(
 
 
 def get_traced_cursor_proxy(cursor, db_api_integration, *args, **kwargs):
-    _traced_cursor = AsyncTracedCursor(db_api_integration)
+    _traced_cursor = AsyncCursorTracer(db_api_integration)
 
     # pylint: disable=abstract-method
     class AsyncTracedCursorProxy(AsyncProxyObject):
```
aabmass, line 133: I think this one should also be renamed to match above, `AsyncCursorTracerProxy`

aabmass (changes requested):
Jusit one blocking request

lonewolf3739 (approved):

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
owais: Update: Given this discussion, the discussion on another PR (https://github.com/open-telemetry/opentelemetry-python-contrib/pull/254#discussion_r542925671) and another discussion that happened in the SIG meeting, we agreed that this does not constitute to SQL parsing and agreed to move forward with the solution.

owais: @aabmass good idea to clarify it in the spec. I'll try to submit a PR soon.

instrumentation/opentelemetry-instrumentation-dbapi/tests/test_dbapi_integration.py
owais: Done

owais:
@aabmass please take another look when you can. Thanks.

aabmass (approved):
🚢

codeboten (approved):

PR #231, opened 2020-12-02
Add test case for DataDog fields and AWS SDK Extension

ocelotl:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #230
## Type of change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [X] test_fields

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/master/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [ ] Followed the style guidelines of this project
- [X] Changelogs have been updated
- [X] Unit tests have been added
- [ ] Documentation has been updated

PR #248, opened 2020-12-11
Upgrade isort and enable black compat mode

owais:
isort and black can be incompatible. Often isort re-writes files in a
way black doesn't like. It takes quite some time and manual effort to
make changes that satisfy both isort and black.

Fortunately, newer versions of isort support a black compatibility mode by setting
isort profile to "black". This makes isort order imports in a way that
is compatible with how black formats Python code. This PR configures
isort to use the black profile by default

Depends on #231


owais:
pylint is crashing after the upgrade. I'll look into that and ping here when this is ready for review.

lzchen:
Why does this PR depend on [#231]?

scripts/eachdist.py
```diff
@@ -509,7 +509,7 @@ def lint_args(args):
     )
     runsubprocess(
         args.dry_run,
-        ("isort", "--recursive", ".")
```
ocelotl, line 512: Just out of curiosity, does the new version of isort does not require this `--recursive` flag or running isort in black mode makes this flag unnecessary?

ocelotl (commented):
I am ok with these changes, @owais, tag me when the lint issues are fixed and I'll approve. :+1:

scripts/eachdist.py
owais: New version works recursively by default and the CLI flag was deprecated.

owais:
> I am ok with these changes, @owais, tag me when the lint issues are fixed and I'll approve. 👍

Might take some time given that the upgrade revealed tons of issues but will do, thanks.

owais:
> Why does this PR depend on [#231]?

Doesn't hard depend on 231, just that there is duplicate work involved. Newer pylint reports exact issues as errors that 231 fixes so to make this PR pass, we either need to copy over changes from 231 or wait for it to merge and rebase.

owais:
@ocelotl @lzchen This is ready now. Feel free to review whenever.

.gitignore
```diff
@@ -23,6 +23,7 @@ __pycache__
 venv*/
 .venv*/
 opentelemetry-python-core*/
+/opentelemetry-python-core
```
aabmass, line 26: What's the difference between this and above? Maybe they can be combined

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -98,7 +98,7 @@ async def _(param: str):
             return {"message": param}
 
         @app.get("/healthzz")
-        async def health():
+        async def health():  # pylint: disable=unused-variable
```
aabmass, line 101: How did this even trigger pylint. Seems like the new version is very noisy

.pylintrc
```diff
@@ -70,7 +70,14 @@ disable=missing-docstring,
...
       line-too-long, # Leave this up to black
-      exec-used
+      exec-used,
+      super-with-arguments, # temp-pylint-upgrade
+      isinstance-second-argument-not-valid-type, # temp-pylint-upgrade
+      raise-missing-from, # temp-pylint-upgrade 
+      unused-argument, # temp-pylint-upgrade  
+      protected-access, # temp-pylint-upgrade 
+      super-init-not-called, # temp-pylint-upgrade
+      invalid-overridden-method, # temp-pylint-upgrade
```
aabmass, line 80: what is is `temp-pylint-upgrade`?

.pylintrc
```diff
@@ -70,7 +70,14 @@ disable=missing-docstring,
       wrong-import-order, # Leave this up to isort
       bad-continuation, # Leave this up to black
       line-too-long, # Leave this up to black
```
aabmass, line 72: Can these all be removed after upgrade? Looking at the black docs, the config for compatibility should be pretty minimal https://black.readthedocs.io/en/stable/compatible_configs.html#pylint

.gitignore
owais: I used symlink instead of checking out another copy of core which did not work with the existing rule. Probably could be combined but I didn't understand fully how others checkout the repo so didn't want to break anyone's flow.

.pylintrc
owais: Just a comment to note down the rules we disabled when we upgraded pylint. Intention is to come back and fix these lint issues and remove the ignore rules.

owais: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/251

.pylintrc
owais: I haven't looked into it but I'm for removing rules if we can automate the fixes with black+isort. I really really miss gofmt :(

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
owais: This *is* an unused var actually but there is no reason why it can't be named `_` like the rest of the handlers defined inside this method. Updated.

ocelotl (approved):
:+1:

.gitignore
toumorokoshi: might be better to move this to a .gitignore_global, if it's user-flow related.

But either not way not really worried about slightly bloating the gitignore. It may be good to add a comment to clarify this is for local dev scenarios.

toumorokoshi (approved):
Thanks!

owais:
Fixed new lint issues. We need to merge this soon as changes to master keeps breaking this PR.

Issue #249, opened 2020-12-11
Add benchmarks to every type of package in `build` workflow


NathanielRN:
In #243 we add performance tests for the AWS SDK Extension.

The AWS SDK extension falls under the category `sdkextension` of tests matched in the `build` workflow. For this reason, when the `jq` command looks through the entire repo after the workflow for `*-benchmark.json` files we are guaranteed to find at least ` file. This file is then given to the subsequent Github action to report on benchmarking results.

However, the `build` workflow also has `instrumentation` and `exporter` packages. These don't have performance tests yet so the `jq` command will fail. Even if we suppress the failure, the subsequent GitHub action will fail if no valid `output.json` file is created.

A temporary solution was to only run those 2 steps of the workflow `if: matrix.package == 'sdkextension'`. A more long-term solution will be to add benchmarks to `instrumentation` and `exporter`.

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/91bfc9afe5ae78bd203a78d0e301074059d54aef/.github/workflows/test.yml#L61-L62

and

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/91bfc9afe5ae78bd203a78d0e301074059d54aef/.github/workflows/test.yml#L68-L69

We know only these packages are required because as long as subsequent tests fit under these categories we won't have to keep adding `if` statements.

NathanielRN:
We should probably keep this open until we have benchmarks for all relevant contrib packages (and a relevant way to force new packages to include them)!

@lzchen @codeboten

lzchen:
@NathanielRN 
If you are interested in adding these benchmarks, feel free to create PRs for them and we'll be happy to review them :)

PR #250, opened 2020-12-11
Commit benchmark results to gh-pages branch

NathanielRN:
# Description

A follow up to the Core PR that adds benchmarking tests: https://github.com/open-telemetry/opentelemetry-python/pull/1443

We added an issue in #249 and link the issue in the code here

We also found out that the GitHub action [will fetch the branch even if it on that branch which produces an error](https://github.com/rhysd/github-action-benchmark/issues/51).

This PR goes back to using `gh-pages` and adds a link in the ReadTheDocs directory to point to the benchmarks graphs which will live on the `gh-pages` branch.

This is also a good idea to reduce the noise on the `master` branch.

This will fix the current Contrib Repo Tests.

## Type of change

# How Has This Been Tested?

See [my example repo](https://github.com/NathanielRN/github-benchmarks-example/runs/1532810344?check_suite_focus=true#step:8:25) which uses `gh-pages` correctly to publish to https://nathanielrn.github.io/github-benchmarks-example/benchmarks/index.html

# Does This PR Require a Contrib Repo Change?

- [x] No.

# Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,7 @@ for more detail on available tox commands.
 
 ### Benchmarks
 
-Performance progression of benchmarks for packages distributed by OpenTelemetry Python can be viewed as a [graph of throughput vs commit history](https://open-telemetry.github.io/opentelemetry-python-contrib/benchmarks/index.html). From this page, you can download a JSON file with the performance results.
+Performance progression of benchmarks for packages distributed by OpenTelemetry Python can be viewed as a [graph of throughput vs commit history](https://opentelemetry-python-contrib.readthedocs.io/en/latest/performance/benchmarks.html). From the linked page, you can download a JSON file with the performance results.
 
 Running the `tox` tests also runs the performance tests if any are available. Benchmarking tests are done with `pytest-benchmark` and they output a table with results to the console.
 
--- a/docs/performance/benchmarks.rst
+++ b/docs/performance/benchmarks.rst
@@ -1,0 +1,4 @@
+Performance Tests - Benchmarks
+==============================
+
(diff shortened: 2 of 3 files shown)


NathanielRN:
Tests failing for now because we don't have a `gh-pages` branch

CONTRIBUTING.md
```diff
@@ -51,7 +51,7 @@ for more detail on available tox commands.
 
 ### Benchmarks
 
-Performance progression of benchmarks for packages distributed by OpenTelemetry Python can be viewed as a [graph of throughput vs commit history](https://open-telemetry.github.io/opentelemetry-python-contrib/benchmarks/index.html). From this page, you can download a JSON file with the performance results.
+Performance progression of benchmarks for packages distributed by OpenTelemetry Python can be viewed as a [graph of throughput vs commit history](https://opentelemetry-python-contrib.readthedocs.io/en/latest/performance/benchmarks.html). From the linked page, you can download a JSON file with the performance results.
```
aabmass, line 54: Same as other PR.

aabmass (approved):

docs/performance/benchmarks.rst
```diff
@@ -0,0 +1,4 @@
+Performance Tests - Benchmarks
+==============================
+
+Click `here <https://open-telemetry.github.io/opentelemetry-python-contrib/benchmarks/index.html>_` to view the latest performance benchmarks for packages in this repo.
```
NathanielRN, line 4: Need to wait for #252 to add this to the toc tree, otherwise we can just add the toc tree entry after #252 gets merged.

CONTRIBUTING.md
NathanielRN: Comment on the other PR for reference: https://github.com/open-telemetry/opentelemetry-python/pull/1469/files#r541332856

codeboten (approved):

PR #252, opened 2020-12-12
Add readTheDocs

### LINKED ISSUE #194: Add read the docs build to Contrib repo
We updated all the links to the point to what _would_ be the right links in this PR https://github.com/open-telemetry/opentelemetry-python-contrib/pull/176

A follow up is to add a `readTheDocs` build here just like we have on the Core repo so that these links work.

NathanielRN:
# Description

Adds the readTheDocs build to the Contrib repo. I went back to the Core PR https://github.com/open-telemetry/opentelemetry-python/pull/1258 where I removed the documentation for the individual instrumentation packages and added them back now in this PR.

However it ended up being much more annoying than I thought :)

I kept getting errors such as

```
py:class reference target not found: opentelemetry.instrumentation.instrumentor.BaseInstrumentor
```

Even though the test correctly installed all 3 of the `API`, `SDK`, and `Instrumentation` repos from GitHub correctly through VCS pip installs. (I verified that if I deleted the relevant `.tox/docs/src/opentelemetry-instrumentation/opentelemetry-instrumentation/` folder the Docs output a helpful message "error this module does not exist!")

But I finally found [this StackOverflow post](https://stackoverflow.com/a/30624034/7460739) which explained that this is a known issue with Sphinx.

I think the problem is that Sphinx will output warnings when imports are from OTHER REPOS, even though it can verify that they do exist in the Python install path.

This solution goes with the solution suggested in the StackOverflow post, but with an easy to comment/uncomment `nitpick-exceptions.ini` file.

I think it's worth it to keep the "error on warning" setting on because it will show us if our docs have invalid imports (although maybe that's that's what unit tests are supposed to be for???)

Fixes #194 

## Type of change

# How Has This Been Tested?

Both `tox -e docs` and `make html` produce html files in `_bulid/html` that make sense!

- [x] No.

# Checklist:

~- [] Changelogs have been updated~
~- [] Unit tests have been added~

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,0 +1,175 @@
+# Configuration file for the Sphinx documentation builder.
+#
+# This file only contains a selection of the most common options. For a full
+# list see the documentation:
+# http://www.sphinx-doc.org/en/master/config
+
+# -- Path setup --------------------------------------------------------------
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+
+import os
+import sys
+from configparser import ConfigParser
+from os import listdir
+from os.path import isdir, join
+
+# configure django to avoid the following exception:
+# django.core.exceptions.ImproperlyConfigured: Requested settings, but settings
+# are not configured. You must either define the environment variable
+# DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.
+from django.conf import settings
+
+settings.configure()
+
+exp = "../exporter"
+exp_dirs = [
+    os.path.abspath("/".join(["../exporter", f, "src"]))
+    for f in listdir(exp)
+    if isdir(join(exp, f))
+]
+
+instr = "../instrumentation"
+instr_dirs = [
+    os.path.abspath("/".join(["../instrumentation", f, "src"]))
+    for f in listdir(instr)
+    if isdir(join(instr, f))
+]
+
+sdk_ext = "../sdk-extension"
+sdk_ext_dirs = [
+    os.path.abspath("/".join(["../sdk-extension", f, "src"]))
+    for f in listdir(sdk_ext)
+    if isdir(join(sdk_ext, f))
+]
+
+sys.path[:0] = exp_dirs + instr_dirs + sdk_ext_dirs
+
+# -- Project information -----------------------------------------------------
+
+project = "OpenTelemetry Python Contrib"
+copyright = "OpenTelemetry Authors"  # pylint: disable=redefined-builtin
+author = "OpenTelemetry Authors"
+
+
+# -- General configuration ---------------------------------------------------
+
+# Easy automatic cross-references for `code in backticks`
+default_role = "any"
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = [
+    # API doc generation
+    "sphinx.ext.autodoc",
+    # Support for google-style docstrings
+    "sphinx.ext.napoleon",
+    # Infer types from hints instead of docstrings
+    "sphinx_autodoc_typehints",
+    # Add links to source from generated docs
+    "sphinx.ext.viewcode",
+    # Link to other sphinx docs
+    "sphinx.ext.intersphinx",
+    # Add a .nojekyll file to the generated HTML docs
+    # https://help.github.com/en/articles/files-that-start-with-an-underscore-are-missing
+    "sphinx.ext.githubpages",
+    # Support external links to different versions in the Github repo
+    "sphinx.ext.extlinks",
+]
+
+intersphinx_mapping = {
+    "python": ("https://docs.python.org/3/", None),
+    "opentracing": (
+        "https://opentracing-python.readthedocs.io/en/latest/",
+        None,
+    ),
+    "aiohttp": ("https://aiohttp.readthedocs.io/en/stable/", None),
+    "wrapt": ("https://wrapt.readthedocs.io/en/latest/", None),
+    "pymongo": ("https://pymongo.readthedocs.io/en/stable/", None),
+}
+
+# http://www.sphinx-doc.org/en/master/config.html#confval-nitpicky
+# Sphinx will warn about all references where the target cannot be found.
+nitpicky = True
+# Sphinx does not recognize generic type TypeVars
+# Container supposedly were fixed, but does not work
+# https://github.com/sphinx-doc/sphinx/pull/3744
+nitpick_ignore = []
+
+cfg = ConfigParser()
+cfg.read("./nitpick-exceptions.ini")
+mcfg = cfg["default"]
+
+
+def getlistcfg(strval):
+    return [
+        val.strip()
+        for line in strval.split("\n")
+        for val in line.split(",")
+        if val.strip()
+    ]
+
+
+if "class_references" in mcfg:
+    class_references = getlistcfg(mcfg["class_references"])
+    for class_reference in class_references:
+        nitpick_ignore.append(("py:class", class_reference,))
+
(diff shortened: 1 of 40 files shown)


docs/conf.py
```diff
@@ -0,0 +1,175 @@
...
+
+if "class_references" in mcfg:
+    class_references = getlistcfg(mcfg["class_references"])
+    for class_reference in class_references:
+        nitpick_ignore.append(("py:class", class_reference,))
+
+if "anys" in mcfg:
+    anys = getlistcfg(mcfg["anys"])
+    for any in anys:
+        nitpick_ignore.append(("any", any,))
```
NathanielRN, line 124: Just want to highlight this as my implementation of the solution described in the Stack Overflow post as mentioned in the PR description.

lonewolf3739: Will this break the ref links? For example in the current stable django documentation https://opentelemetry-python.readthedocs.io/en/stable/instrumentation/django/django.html If you click on `Bases: [opentelemetry.instrumentation.instrumentor.BaseInstrumentor]` or `See [BaseInstrumentor]`  They both take you BaseInstrumentor documentation. If I understand it correctly as these instrumentation libraries are moved out of main repo it can't find such references anymore and giving error. They are now cross-referencing targets from main repo docs.

will `intersphinx_mapping` along with updating \`BaseInstrumentor\`  occurrences with \`opentelemetry.instrumentation.instrumentor.BaseInstrumentor\` be helpful?

NathanielRN: Yes I believe you are correct that all the references in this `nitpick-exceptions.ini` file will produce links that are broken. It's because these components are Core code components and Contrib doesn't know where to link the reference to.

The `intersphinx_mapping` solution sounds interesting! However it's not something I've explored :sweat: although we can create a follow up to fix these links? If we want to keep this PR small we can merge it like this, and then we can create an issue to investigate that mapping solution to link to the Core docs?

The main benefit this PR does is that _at least_ we can have auto documentation on the packages even though the links to the Core components are broken.

Also, the docs will break if the SDK components break/change. But as I mentioned in the PR description, maybe that shouldn't be the job of the Docs, maybe that should be the job of the unit tests?

lonewolf3739: > we can create an issue to investigate that mapping solution to link to the Core docs?

This would be ideal. I just checked, all the references in this nitpick-exceptions.ini will not produce the broken links rather it is just a plain text instead of linked documentation. The `intersphinx_mapping` should really help us here for example pymongo instrumentation documentation has an link to official pymongo driver documentation [here](https://opentelemetry-python.readthedocs.io/en/stable/instrumentation/pymongo/pymongo.html#opentelemetry.instrumentation.pymongo.CommandTracer)

aabmass:
> But I finally found [this StackOverflow post](https://stackoverflow.com/a/30624034/7460739) which explained that this is a known issue with Sphinx.
> 
> I think the problem is that Sphinx will output warnings when imports are from OTHER REPOS, even though it can verify that they do exist in the Python install path.
> 
> This solution goes with the solution suggested in the StackOverflow post, but with an easy to comment/uncomment `nitpick-exceptions.ini` file.

The linked SO post seems to only be talking about builtins. I think the other `py:class reference target not found` are either related to intersphinx mapping being missing or the packages are not installed. I have a sphinx docs build working fine [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/tree/master/docs) that references OTel packages and I didn't need this ini stuff

docs/conf.py
```diff
@@ -0,0 +1,175 @@
...
+
+intersphinx_mapping = {
+    "python": ("https://docs.python.org/3/", None),
+    "opentracing": (
+        "https://opentracing-python.readthedocs.io/en/latest/",
+        None,
+    ),
+    "aiohttp": ("https://aiohttp.readthedocs.io/en/stable/", None),
+    "wrapt": ("https://wrapt.readthedocs.io/en/latest/", None),
+    "pymongo": ("https://pymongo.readthedocs.io/en/stable/", None),
```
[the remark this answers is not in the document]
NathanielRN: Thanks for this!

docs/conf.py
```diff
@@ -0,0 +1,175 @@
...
+# django.core.exceptions.ImproperlyConfigured: Requested settings, but settings
+# are not configured. You must either define the environment variable
+# DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.
+from django.conf import settings
+
+settings.configure()
+
+exp = "../exporter"
+exp_dirs = [
+    os.path.abspath("/".join(["../exporter", f, "src"]))
```
aabmass, line 29:
```suggestion
    os.path.abspath(join("..", "exporter", f, "src"))
```

docs/conf.py
aabmass: Think you should be able to remove all of this with the intersphinx mapping?

docs/conf.py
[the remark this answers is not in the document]
NathanielRN: So adding this `intersphinx-mapping` almost completely solved the problem! However there's still a few items that need to be ignore by the `nitpick`...

In [nitpick-exceptions.ini](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/252/files#diff-9d4ea345825ab76a8639109a7e745f5a02e5f3d19b3fabc57dd22e896e88d7f7) I still have to silence the errors from these:

```
[default]
class_references=
    ; TODO: Understand why sphinx is not able to find this local class
    opentelemetry.trace.propagation.textmap.TextMapPropagator
    ; - AwsXRayFormat
    opentelemetry.trace.propagation.textmap.DictGetter
    ; - instrumentation.asgi.CarrierGetter
    ; API
    opentelemetry.trace.propagation.textmap.Getter
    ; - DatadogFormat
    ; - AWSXRayFormat
    TextMapPropagatorT
    ; - AwsXRayFormat.extract

anys=
    ; API
    opentelemetry.trace.propagation.textmap.TextMapPropagator.fields
    ; - AWSXRayFormat
    TraceId
    ; - AwsXRayIdsGenerator
    TraceIdRatioBased
    ; - AwsXRayIdsGenerator
    ; SDK
    SpanProcessor
    ; - DatadogExportSpanProcessor
    TracerProvider
    ; - AwsXRayIdsGenerator
    ; Instrumentation
    BaseInstrumentor
    ; - instrumentation.*
```

I added comments showing where the errors come from. This makes me wonder if I did the [doc string for AWS components here](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/252/files#diff-4d01b1c228f66548cd4b78d6fdab79f2302c5af09befe5e2203cbffb6884b559R27) wrong? Is there a certain way that I need to follow for the backticks `` when writing doc strings?

Something that would make the following doc string break such that I needed the `nitpick-exceptions` above?

```
AWS X-Ray IDs Generator
-----------------------
The **AWS X-Ray IDs Generator** provides a custom IDs Generator to make
traces generated using the OpenTelemetry SDKs `TracerProvider` compatible
with the AWS X-Ray backend service `trace ID format`_.
```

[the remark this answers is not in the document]
aabmass: I dont think the docstring should matter. Automodule should generate entries regardless of docstring, as long as you have the `.. automodule:: opentelemetry....`

Do you see the AWS docs built locally?

docs/instrumentation/wsgi/wsgi.rst
```diff
@@ -0,0 +1,7 @@
+OpenTelemetry WSGI Instrumentation
+==================================
+
```
lonewolf3739, line 3: Some other instrumentation such as `tornado` and `urllib` etc.. are missing, any particular reason?

NathanielRN: I think I noticed that earlier too, I just pulled the files that were on Core so I assume they hadn’t been updated on Core either.

For that reason I thought as a follow up we can add those docs.

codeboten (approved):
Approving this to get the read the docs started for this repo, we can always improve them as we go.

PR #202, opened 2020-11-21
Update sqlalchemy instrumentation to follow semantic conventions

lonewolf3739:
# Description
Part of  #159 

# Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

PR #254, opened 2020-12-14
SQLAlchemy: Use SQL operation and DB name as the Span name

owais:
# Description

Current instrumentation uses the entire SQL query as the operation name
which makes traces very hard to read and understand in addition to
introducing high-cardinality issues. This commit fixes the problem by
using only the SQL operation name and the DB name instead of the entire
query.

## Type of change

# How Has This Been Tested?

- [x] Manual testing
- [x] Updated unit tests

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - `opentelemetry-instrumentation-asgi`, `opentelemetry-instrumentation-wsgi` Return `None` for `CarrierGetter` if key not found
-  ([#1374](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/233))
+  ([#233](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/233))
 - `opentelemetry-instrumentation-grpc` Comply with updated spec, rework tests
   ([#236](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/236))
 - `opentelemetry-instrumentation-asgi`, `opentelemetry-instrumentation-falcon`, `opentelemetry-instrumentation-flask`, `opentelemetry-instrumentation-pyramid`, `opentelemetry-instrumentation-wsgi` Renamed `host.port` attribute to `net.host.port`
@@ -57,6 +57,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#235](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/235))
 - `opentelemetry-exporter-datadog` `opentelemetry-sdk-extension-aws` Fix reference to ids_generator in sdk
   ([#235](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/235))
+- `opentelemetry-instrumentation-sqlalchemy` Use SQL operation and DB name as span name.
+  ([#254](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/254))
 
 ## [0.16b1](https://github.com/open-telemetry/opentelemetry-python-contrib/releases/tag/v0.16b1) - 2020-11-26
 
--- a/instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
+++ b/instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from sqlalchemy.event import listen
+from sqlalchemy.event import listen  # pylint: disable=no-name-in-module
 
 from opentelemetry import trace
 from opentelemetry.instrumentation.sqlalchemy.version import __version__
@@ -72,22 +72,38 @@ class EngineTracer:
         listen(engine, "after_cursor_execute", self._after_cur_exec)
         listen(engine, "handle_error", self._handle_error)
 
+    def _operation_name(self, db_name, statement):
+        parts = []
+        if isinstance(statement, str):
+            # otel spec recommends against parsing SQL queries. We are not trying to parse SQL
+            # but simply truncating the statement to the first word. This covers probably >95%
+            # use cases and uses the SQL statement in span name correctly as per the spec.
+            # For some very special cases it might not record the correct statement if the SQL
+            # dialect is too weird but in any case it shouldn't break anything.
+            parts.append(statement.split()[0])
+        if db_name:
+            parts.append(db_name)
+        if not parts:
+            return self.vendor
+        return " ".join(parts)
+
     # pylint: disable=unused-argument
     def _before_cur_exec(self, conn, cursor, statement, *args):
+        attrs, found = _get_attributes_from_url(conn.engine.url)
+        if not found:
+            attrs = _get_attributes_from_cursor(self.vendor, cursor, attrs)
+
+        db_name = attrs.get(_DB, "")
         self.current_span = self.tracer.start_span(
-            statement, kind=trace.SpanKind.CLIENT
+            self._operation_name(db_name, statement),
+            kind=trace.SpanKind.CLIENT,
         )
         with self.tracer.use_span(self.current_span, end_on_exit=False):
             if self.current_span.is_recording():
                 self.current_span.set_attribute(_STMT, statement)
                 self.current_span.set_attribute("db.system", self.vendor)
-
-                if not _set_attributes_from_url(
-                    self.current_span, conn.engine.url
-                ):
-                    _set_attributes_from_cursor(
-                        self.current_span, self.vendor, cursor
-                    )
+                for key, value in attrs.items():
+                    self.current_span.set_attribute(key, value)
 
     # pylint: disable=unused-argument
     def _after_cur_exec(self, conn, cursor, statement, *args):
@@ -108,25 +124,22 @@ class EngineTracer:
             self.current_span.end()
 
 
-def _set_attributes_from_url(span: trace.Span, url):
+def _get_attributes_from_url(url):
     """Set connection tags from the url. return true if successful."""
-    if span.is_recording():
-        if url.host:
-            span.set_attribute(_HOST, url.host)
-        if url.port:
-            span.set_attribute(_PORT, url.port)
-        if url.database:
-            span.set_attribute(_DB, url.database)
-        if url.username:
-            span.set_attribute(_USER, url.username)
+    attrs = {}
+    if url.host:
+        attrs[_HOST] = url.host
+    if url.port:
+        attrs[_PORT] = url.port
+    if url.database:
+        attrs[_DB] = url.database
+    if url.username:
+        attrs[_USER] = url.username
+    return attrs, bool(url.host)
 
-    return bool(url.host)
 
-
-def _set_attributes_from_cursor(span: trace.Span, vendor, cursor):
+def _get_attributes_from_cursor(vendor, cursor, attrs):
     """Attempt to set db connection attributes by introspecting the cursor."""
-    if not span.is_recording():
-        return
     if vendor == "postgresql":
         # pylint: disable=import-outside-toplevel
         from psycopg2.extensions import parse_dsn
@@ -135,6 +148,7 @@ def _set_attributes_from_cursor(span: trace.Span, vendor, cursor):
             dsn = getattr(cursor.connection, "dsn", None)
             if dsn:
                 data = parse_dsn(dsn)
-                span.set_attribute(_DB, data.get("dbname"))
-                span.set_attribute(_HOST, data.get("host"))
-                span.set_attribute(_PORT, int(data.get("port")))
+                attrs[_DB] = data.get("dbname")
+                attrs[_HOST] = data.get("host")
+                attrs[_PORT] = int(data.get("port"))
+    return attrs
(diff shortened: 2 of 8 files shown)


instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -72,20 +72,33 @@ def __init__(self, tracer, engine):
         listen(engine, "after_cursor_execute", self._after_cur_exec)
         listen(engine, "handle_error", self._handle_error)
 
+    def _operation_name(self, db_name="", statement=""):
```
aabmass, line 75: It doesn't look like the defaults are needed, they get passed in ln L94

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -106,33 +119,32 @@ def _handle_error(self, context):
...
-        if url.database:
-            span.set_attribute(_DB, url.database)
-        if url.username:
-            span.set_attribute(_USER, url.username)
-
-    return bool(url.host)
-
-
-def _set_attributes_from_cursor(span: trace.Span, vendor, cursor):
+    attrs = attrs or {}
```
aabmass, line 124: Same here from the code I can see; you always get a non-None dict

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -72,20 +72,33 @@ def __init__(self, tracer, engine):
         listen(engine, "after_cursor_execute", self._after_cur_exec)
         listen(engine, "handle_error", self._handle_error)
 
+    def _operation_name(self, db_name="", statement=""):
+        parts = []
+        if isinstance(statement, str):
+            parts.append(statement.split(" ")[0].split("\t")[0])
```
aabmass, line 78: This doesn't seem to follow the [semantic convention recommendations](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#semantic-conventions-for-database-client-calls):

> SQL spans SHOULD be named the
following way, unless the statement is known to be of low cardinality:
`<db.operation> <db.name>.<db.sql.table>`, provided that `db.operation` and `db.sql.table` are available.
If `db.sql.table` is not available due to its semantics, the span SHOULD be named `<db.operation> <db.name>`.
It is not recommended to attempt any client-side parsing of `db.statement` just to get these properties,
they should only be used if the library being instrumented already provides them.
When it's otherwise impossible to get any meaningful span name, `db.name` or the tech-specific database name MAY be used.

I would have thought SQLAlchemy could provide a way to get `db.operation`. I'm also not too confident this parsing would always work to get the operation, e.g. if the whole query is wrapped in parens

owais: We had the same discussion on another PR. https://github.com/open-telemetry/opentelemetry-python-contrib/pull/246#discussion_r542562273

Copying my reply here. 

> I agree that we shouldn't parse it either but I don't think this really counts as "SQL parsing". We are making a bare-bones, extremely cheap, best effort attempt at guessing the SQL operation name. I think as long as it does not incur a perf penalty, decrease reliabilit, affect the DB connection/cursor/statement in any way or end up talking to the database, we should be good. At least that's how I understand the intention behind the spec recommendation but may be there is more to it that I might have overlooked?

CHANGELOG.md
```diff
@@ -30,11 +30,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - `opentelemetry-instrumentation-asgi`, `opentelemetry-instrumentation-wsgi` Return `None` for `CarrierGetter` if key not found
-  ([#1374](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/233))
+  ([#233](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/233))
```
lzchen, line 33:
```suggestion
  ([#254](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/254))
```

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
lzchen: I agree that this is kind of going against the semantic conventions as well. I do not believe we should try to parse the statement to obtain <db.operation>. Ideally, this should have been addressed in [this pr](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/202) but might have been missed. We should be seeing if <db.operation> is provided by the sqlalchemy library. If not, we should fallback on <db.name>.

lzchen:
Part of https://github.com/open-telemetry/opentelemetry-python-contrib/issues/159

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
owais: I still don't think this goes against the spec. This definitely complied with the semantic conventions. Only point of contention is that the spec prescribes to not do any SQL parsing in order figure out the SQL command, table name etc. I don't think we are really doing that. If we ran an SQL parser that parsed the raw statement and tried to understand SQL grammer to figure out the SQL command and table name, I'd say that'd be against the spec. Here we are merely truncating the raw statement to the first word. I don't think that really counts as SQL parsing. If SQLAlchemy provided us with the query as a list with broken down components, it would be fine to just take the first item from the list and use it as the statement. I think what we are doing here is a lot closer to that than any kind of SQL parsing. 

Also, if we try to understand the sentiment behind the recommendation, I think the reasons to not parse are one or all of the following:

- SQL parsing can be slow, perhaps too slow for instrumentation to do.
- Depending on the library/ORM being instrumented, parsing a statement could be a stateful operation. Parsing a query once could result in a 2nd parse attempt by the library itself failing.
- Depending on the library/ORM, parsing could result in network op/re-connection or additional SQL query to retrieve metadata.
- We'd have to support a huge number of dialects and have fallbacks.
- In general, it can be error prone and make the instrumentation unreliable.

I don't think what we are doing anything here that touches on any of the issues. 

Moreover, most python SQL libraries/ORMs do not provide us with broken down information so not even truncating the raw statement to the first word gives us spans where operation name is always the same as the database name. This is very limiting for APM backends as they cannot any longer detect problematic spans/queries/patterns because every single query looks the same. 

All things considered, I think truncating the query to first word and using that as the SQL statement is a very good tradeoff IMO with very little downsides if any at all.

owais:
PTAL @aabmass @lzchen

lzchen (approved):

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
ocelotl: Since we need to get these attributes from somewhere, unless there is another solution that does even less parsing/truncation/etc. we should use the solution implemented here. Nevertheless, I understand how this can look as going against the specification (which by the way does not clearly specify if not doing any parsing is a RECOMMEND NOT or SHOULD NOT type of requirement), so, @owais can you add a comment that explains your point of view here?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -72,22 +72,34 @@ def __init__(self, tracer, engine):
...
+        if db_name:
+            parts.append(db_name)
+        if not parts:
+            return self.vendor
+        return " ".join(parts)
+
     # pylint: disable=unused-argument
     def _before_cur_exec(self, conn, cursor, statement, *args):
+        attrs = {}
+        attrs, found = _get_attributes_from_url(conn.engine.url, attrs)
```
ocelotl, line 88:
```suggestion
        attrs, found = _get_attributes_from_url(conn.engine.url)
```
Single use functiong here, BTW.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -108,33 +120,32 @@ def _handle_error(self, context):
             self.current_span.end()
 
 
-def _set_attributes_from_url(span: trace.Span, url):
+def _get_attributes_from_url(url, attrs):
```
ocelotl, line 123:
```suggestion
def _get_attributes_from_url(url):
```

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -108,33 +120,32 @@ def _handle_error(self, context):
...
-        if url.database:
-            span.set_attribute(_DB, url.database)
-        if url.username:
-            span.set_attribute(_USER, url.username)
-
-    return bool(url.host)
-
-
-def _set_attributes_from_cursor(span: trace.Span, vendor, cursor):
+    attrs = attrs or {}
```
ocelotl, line 125:
```suggestion
    attrs = {}
```

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -72,22 +72,34 @@ def __init__(self, tracer, engine):
...
+        if not parts:
+            return self.vendor
+        return " ".join(parts)
+
     # pylint: disable=unused-argument
     def _before_cur_exec(self, conn, cursor, statement, *args):
+        attrs = {}
+        attrs, found = _get_attributes_from_url(conn.engine.url, attrs)
+        if not found:
+            attrs = _get_attributes_from_cursor(self.vendor, cursor, attrs)
```
ocelotl, line 90: Single use function here also.

ocelotl (changes requested):

lzchen:
@owais has @aabmass 's comments been addressed yet?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from sqlalchemy.event import listen
+from sqlalchemy.event import listen  # pylint: disable=no-name-in-module
```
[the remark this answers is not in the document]
owais: It's an issue with latest pylint. https://github.com/PyCQA/pylint/issues/2648 and https://github.com/PyCQA/pylint/issues/2862

Let's discuss in the SIG meeting this week. We can either revert to older version of pylint for contrib repo or we can rename instrumentation packages to something different from the libraries they are instrumenting. For example, instrumentation for requests could be called `requests_otel` or `requests_instrumentation` instead of `requests`.

owais:
@lzchen Yes, they've been addressed. Not sure why github shows them as unresolved. May be a bad rebase. Will double check in the morning. Will also address @ocelotl comments in the morning.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
owais: Added

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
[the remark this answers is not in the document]
owais: Resolving. Let's discuss in SIG. Happy to come back to fix however we decide in SIG.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
owais: The function is big enough and self-contained enough to be independent IMO. Manually in-lining it would make the calling function do too much and harder to read IMO.

ocelotl: Not a tall, just a suggestion :+1:

ocelotl (approved):

codeboten (approved):
👍

Issue #149, opened 2020-11-03
Add hook to set custom span name in requests instrumentation


lzchen:
Similar to [wsgi](https://github.com/open-telemetry/opentelemetry-python/blob/master/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py#L210), accept a callback function in `RequestsInstrumentor().instrument` to be able to pass custom span name.

The [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#name) say "MAY" for this feature, but we should be consistent across our instrumentations.

Part of https://github.com/open-telemetry/opentelemetry-python/issues/1297

danielbraun89:
The requests name_callback does not accept neither method nor url as inputs. The user dont have any info to generate the name from. Shouldnt this issue remain open until the functionality is completed?

lzchen:
@danielbraun89 
Shouldn't the user always have access to `method` and `url`? They can simply use those within the callback itself, they don't need to be passed in.

lonewolf3739:
@lzchen I don't think user would know the `method` and `url` with `requests` instrumentation. We had discussion on whether to pass environ with [flask](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/152#discussion_r519871454) name_callback. With Flask `flask.request` is accessible to user always.

lzchen:
Oh that is true, I misunderstood the question. I will create an issue for it.

PR #259, opened 2020-12-15
Fix requests and urllib instrumentations span name callback parameters

lzchen:
Addresses this [comment](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/149#issuecomment-745074789)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#242](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/242))  
 - `opentelemetry-instrumentation-flask` Do not emit a warning message for request contexts created with `app.test_request_context`
   ([#253](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/253))
+- `opentelemetry-instrumentation-requests`, `opentelemetry-instrumentation-urllib` Fix span name callback parameters
+- ([#259](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/259))
 
 ## [0.16b1](https://github.com/open-telemetry/opentelemetry-python-contrib/releases/tag/v0.16b1) - 2020-11-26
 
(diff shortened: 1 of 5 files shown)


instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -122,7 +122,7 @@ def _instrumented_requests_call(
         method = method.upper()
         span_name = ""
         if name_callback is not None:
-            span_name = name_callback()
+            span_name = name_callback(method, url)
```
toumorokoshi, line 125: Just a thought that although this is sufficient for the vast majority of cases, I have a feeling people may want to customize the span name further later on.

For example, one cannot extracted values from the header and add them to the span name.

I don't have a great solution (besides passing in all args via **kwargs, etc), but this may come up as an issue.

instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
```diff
@@ -145,7 +143,7 @@ def _instrumented_open_call(
 
         span_name = ""
         if name_callback is not None:
-            span_name = name_callback()
+            span_name = name_callback(method, url)
```
toumorokoshi, line 146: same possible issue as with requests.

toumorokoshi (approved):
LGTM! Thanks.

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
lzchen: Yeah I thought of that too. For now, I think this is sufficient and we can change the parameters if the need arises.

CHANGELOG.md
```diff
@@ -39,6 +39,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#242](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/242))  
 - `opentelemetry-instrumentation-flask` Do not emit a warning message for request contexts created with `app.test_request_context`
   ([#253](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/253))
+- `opentelemetry-instrumentation-requests`, `opentelemetry-instrumentation-urllib` Fix span name callback parameters
+- ([#259](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/259))
```
alertedsnake, line 43: Note that there's an extra dash on line 43 here.

lzchen: Probably will be fixed in a new pr

PR #260, opened 2020-12-16
gRPC streaming bugfix

### LINKED ISSUE #257: gRPC server instrumentation creates multiple traces on streaming requests
**Environment**
Current `master` code, basically the sample code in the documentation, testing with a unary request vs. a streaming request.

**Steps to reproduce**
Create a simple gRPC servicer with two RPCs, one which returns a single message (the unary response), and one which yields items in a list for a streaming response.

The key here is to make an instrumented request within the primary request handler (I'm using a simple HTTP get with the Requests instrumentation), so you get an _additional_ span which should be attached to the same trace.

A single trace with the main span, and a second child span for the HTTP request.

Two separate traces, each containing a single span.

**Additional context**
The problem _only_ occurs on streaming requests - I'm sure the reworking I did as part of https://github.com/open-telemetry/opentelemetry-python/pull/1171 is where the problem started, I didn't take into account the streaming case specifically with multiple spans, and naturally, there are no tests for anything streaming, only unary responses.

So as part of this, we'll need some useful tests as well.  I'll see if I can write up my test case as an actual test case.

And again, I've got a vested interest in this working, so I'll have a PR up soon.

alertedsnake:
# Description

Fix for a bug I introduced when rewriting the gRPC instrumentation where child spans weren't properly being tracked as part of the same trace.  Also there weren't tests for streaming responses before, so I've added one, and several to verify child spans are handled properly in both unary and streaming responses.

Fixes #257 

## Type of change

# How Has This Been Tested?

Added new tests:
* `test_create_two_spans`
* `test_create_span_streaming`
* `test_create_two_spans_streaming`

Also verified manually via exporting spans to Jaeger as part of a test application based on the sample server code.

- [X] No.

# Checklist:


alertedsnake:
Is there anything you need me to do to help the process along here?  This gRPC part of this is not really usable as-is, and I'd like to help with other things, but I'm blocked by these PRs.

codeboten (approved):
Code changes LGTM!

alertedsnake:
So for the record, it's also very frustrating when the lint check passes when the PR is made but sometime before it's merged, the linter rules get updated, and now a significant amount of the code fails the checks.

Is there an announcement of these changes somewhere I could subscribe to?

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py
```diff
@@ -263,3 +273,23 @@ def telemetry_interceptor(request_or_iterator, context):
         return _wrap_rpc_behavior(
             continuation(handler_call_details), telemetry_wrapper
         )
+
+    # Handle streaming responses separately - we have to do this
+    # to return a *new* generator or various upstream things
+    # get confused, or we'll lose the consistent trace
+    def _intercept_server_stream(
```
ocelotl, line 280: Single-use function here.

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py
```diff
@@ -263,3 +273,23 @@ def telemetry_interceptor(request_or_iterator, context):
...
+        with self._set_remote_context(context):
+            with self._start_span(handler_call_details, context) as span:
+                context = _OpenTelemetryServicerContext(context, span)
+
+                try:
+                    yield from behavior(request_or_iterator, context)
+
+                except Exception as error:
+                    # pylint:disable=unidiomatic-typecheck
+                    if type(error) != Exception:
```
ocelotl, line 293: Can you please explain why is this necessary? `record_exception` expects an `Exception`. Even so, if this check was appropriate, wouldn't it be better to use `isinstance`?

instrumentation/opentelemetry-instrumentation-grpc/tests/test_server_interceptor.py
```diff
@@ -134,25 +157,146 @@ def test_create_span(self):
...
+
+        spans_list = self.memory_exporter.get_finished_spans()
+        self.assertEqual(len(spans_list), 2)
+        child_span = spans_list[0]
+        parent_span = spans_list[1]
+
+        self.assertEqual(parent_span.name, rpc_call)
+        self.assertIs(parent_span.kind, trace.SpanKind.SERVER)
+
+        # Check version and name in span's instrumentation info
```
ocelotl, line 252: Nit: redundant comment

ocelotl (changes requested):

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py
alertedsnake: Yes - in the gRPC world bare `Exception` instances are thrown by the system when you call `grpc.abort()`, and some of those shouldn't be treated as actual errors, for example the HTTP 404 series responses which are returned to the client but not recorded as errors in a trace.

This allows us to record anything *except those* as uncaught errors, since it's really unlikely anything else would throw a bare exception itself.  It's not ideal but there just isn't any other way to distinguish these at this point in the program flow.

Using `isinstance()` would catch application exceptions here, which we want to raise normally.

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py
alertedsnake: I'm not sure what the issue is?

ocelotl: Should have been Nit: single-use function here.

Having a function that is used only once has the disadvantage of making the reader make a mental copy of the arguments passed to it, having to look for the function code, mentally pasting the arguments, reading the function code and then going back to the starting point without any of the advantages of not having repeated code that a function provides.

Anyways, it is still a nitpick, you can leave it as is if you please.

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py
ocelotl: Ok, I see. This is unusual code, but necessary because of a particularity of gRPC that the reader may not know about. I'll approve, but please add this explanation as a comment here. :+1:

ocelotl (approved):

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py
alertedsnake: But this must be a function, for the reason explained in a the comment.

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py
alertedsnake: The same exact code is on line 263, with a comment describing why it was done this way.  Happy to duplicate the comment here if that would be preferred.

PR #264, opened 2020-12-18
botocore: switch setter for injecting headers into propagator

### LINKED ISSUE #262: gRPC client + botocore + datadog exporter instrumentation error
This one is a bit complex - when using the gRPC client instrumentation to make a call to an instrumented gRPC service that then makes a call via boto to the DynamoDB API, the combination of instrumentation leads to this error:

```
2020-12-17 14:19:39,666 DEBUG    [botocore.httpsession:298] Exception received when sending urllib3 HTTP request
  File "/home/michael/work/myapp/.venv/lib/python3.8/site-packages/botocore/httpsession.py", line 254, in send
  File "/home/michael/work/myapp/.venv/lib/python3.8/site-packages/urllib3/connectionpool.py", line 699, in urlopen
    httplib_response = self._make_request(
  File "/home/michael/work/myapp/.venv/lib/python3.8/site-packages/urllib3/connectionpool.py", line 394, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/home/michael/work/myapp/.venv/lib/python3.8/site-packages/urllib3/connection.py", line 234, in request
    super(HTTPConnection, self).request(method, url, body=body, headers=headers)
  File "/usr/lib64/python3.8/http/client.py", line 1255, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/home/michael/work/myapp/.venv/lib/python3.8/site-packages/botocore/awsrequest.py", line 91, in _send_request
    rval = super(AWSConnection, self)._send_request(
  File "/usr/lib64/python3.8/http/client.py", line 1296, in _send_request
    self.putheader(hdr, value)
  File "/home/michael/work/myapp/.venv/lib/python3.8/site-packages/urllib3/connection.py", line 219, in putheader
    _HTTPConnection.putheader(self, header, *values)
  File "/usr/lib64/python3.8/http/client.py", line 1232, in putheader
    if _is_illegal_header_value(values[i]):
TypeError: expected string or bytes-like object
```

Without either client instrumentation, this call works perfectly.

**Steps to reproduce**
As described above - I don't have a simplified sample, but I'll try to produce one.

No exception :)

The above exception.

**Additional context**

I'm definitely working on this, but any suggestions would certainly be welcome, I'm not quite sure what the invalid header is or where it might get set.

alertedsnake:
# Description

Occasionally not all of the headers available to botocore are strings - I've noticed this when a call is being made with an active span passed from a different application (specifically via gRPC).  Botocore does allow this, using the `add_header` method rather than `__setitem__`.  According to the docs for `botocore.compat.HTTPHeaders`:

```
If a parameter value contains non-ASCII characters it can be specified
as a three-tuple of (charset, language, value), in which case it will be
encoded according to RFC2231 rules.  Otherwise it will be encoded using
the utf-8 charset and a language of ''
```

Whereas using `__setitem__` does not do this conversion.

I'm not convinced this is the right place to fix this, however - I've only encountered it in the situation described in #262, and haven't been able to find what would be passing in a non-string header in the first place.  I'll be happy to dig more if anyone has a better suggestion, but as far as I can tell, this is also an innocuous change.

One question I have though is about how I could test this - I don't know how to simulate (in the botocore tests) having an active trace passed through from another application, so advice on this would be welcome.

Fixes #262
## Type of change

# How Has This Been Tested?

All existing tests still pass, and I've been working with the situation described in #262 

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
@@ -67,7 +67,7 @@ logger = logging.getLogger(__name__)
 def _patched_endpoint_prepare_request(wrapped, instance, args, kwargs):
     request = args[0]
     headers = request.headers
-    propagators.inject(type(headers).__setitem__, headers)
+    propagators.inject(type(headers).add_header, headers)
     return wrapped(*args, **kwargs)
 
 


instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
```diff
@@ -67,7 +67,7 @@
 def _patched_endpoint_prepare_request(wrapped, instance, args, kwargs):
     request = args[0]
     headers = request.headers
-    propagators.inject(type(headers).__setitem__, headers)
+    propagators.inject(type(headers).add_header, headers)
```
toumorokoshi, line 70: I don't have a lot of context on this, but we should probably add at least a test for the case you're running into.

From https://github.com/open-telemetry/opentelemetry-python-contrib/issues/262, it sounds like the most relevant part is the DyanamoDB API call. have you tried adding a mock DynamiDB api call (moto [looks to support dynamodb](https://github.com/spulec/moto/tree/master/moto/dynamodb))? It seems like a call that maps what you're executing in your service should trigger the error.

toumorokoshi: Taking a closer look, the part I'm a little unclear on is, as you mentioned, what is actually causing this. I have a feeling something was injected that should not have been by the propagator. I'll comment more on the issue.

toumorokoshi (changes requested):
The code seems innocuous, but we should probably root cause + add a test to make sure this is right, as you mentioned. I'll comment more on the other issue.

alertedsnake:
Closing this, it's not the root cause of #262 and so doesn't seem necessary.

PR #270, opened 2020-12-21
Set max number of commits in performance graph

NathanielRN:
# Description

As a follow up to #1443, this PR adds a limit to the number of commits saved and displayed in the performance graph.

By default [there is no limit](https://github.com/rhysd/github-action-benchmark#max-items-in-chart-optional) so I thought this would be a good preemptative measure to avoid a noisy (and maybe unreadable?) graph.

## Type of change

# How Has This Been Tested?

N/A

- [x] No.

# Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,6 +73,7 @@ jobs:
           tool: pytest
           output-file-path: output.json
           github-token: ${{ secrets.GITHUB_TOKEN }}
+          max-items-in-chart: 100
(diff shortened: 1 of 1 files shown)


toumorokoshi (approved):
SGTM. I'd be fine if we decided to just wait until later on too - this is one of those situations where you kind of have to see how this works in practice.

If we could do this by tag, (e.g. version numbers) that would be super amazing though.

codeboten (approved):

PR #272, opened 2020-12-22
datadog-exporter: don't set tags which are None

### LINKED ISSUE #262: gRPC client + botocore + datadog exporter instrumentation error

alertedsnake:
# Description

This is often `None`, but tags are always strings, and so things get broken when spans get passed along to other client calls.

Fixes #262

## Type of change

# How Has This Been Tested?

This passes existing tests, and I've added one to verify when the field is set to `None`, it's skipped.

- [X] No.

# Checklist:

### DIFF
--- a/exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/propagator.py
+++ b/exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/propagator.py
@@ -103,7 +103,7 @@ class DatadogFormat(TextMapPropagator):
             self.SAMPLING_PRIORITY_KEY,
             str(constants.AUTO_KEEP if sampled else constants.AUTO_REJECT),
         )
-        if constants.DD_ORIGIN in span.context.trace_state:
+        if span.context.trace_state.get(constants.DD_ORIGIN) is not None:
             setter.set(
                 carrier,
                 self.ORIGIN_KEY,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -161,10 +161,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#246](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/246))
 - Update TraceState to adhere to specs
   ([#276](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/276))
+- `opentelemetry-exporter-datadog` Don't allow `NoneType` values in `context.trace_state` fields
+  ([#272](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/272))
 
 ### Removed
 - Remove Configuration
   ([#285](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/285))
+
 
 ## [0.16b1](https://github.com/open-telemetry/opentelemetry-python-contrib/releases/tag/v0.16b1) - 2020-11-26
 
--- a/exporter/opentelemetry-exporter-datadog/tests/test_datadog_format.py
+++ b/exporter/opentelemetry-exporter-datadog/tests/test_datadog_format.py
@@ -198,7 +198,6 @@ class TestDatadogFormat(unittest.TestCase):
 
         self.assertEqual(FORMAT.fields, inject_fields)
 
-
     @patch("opentelemetry.exporter.datadog.propagator.get_current_span")
     def test_trace_state_not_none(self, mock_get_current_span):
         """Make sure the fields attribute returns the fields used in inject"""
@@ -231,6 +230,6 @@ class TestDatadogFormat(unittest.TestCase):
 
         # verify 'x-datadog-origin' is not present, it was None
         expected_fields = FORMAT.fields.copy()
-        expected_fields.discard('x-datadog-origin')
+        expected_fields.discard("x-datadog-origin")
 
         self.assertEqual(expected_fields, inject_fields)
--- a/exporter/opentelemetry-exporter-datadog/tests/test_datadog_format.py
+++ b/exporter/opentelemetry-exporter-datadog/tests/test_datadog_format.py
@@ -197,3 +197,40 @@ class TestDatadogFormat(unittest.TestCase):
             inject_fields.add(call[1][1])
 
         self.assertEqual(FORMAT.fields, inject_fields)
+
+
+    @patch("opentelemetry.exporter.datadog.propagator.get_current_span")
+    def test_trace_state_not_none(self, mock_get_current_span):
+        """Make sure the fields attribute returns the fields used in inject"""
+
+        tracer = trace.TracerProvider().get_tracer("sdk_tracer_provider")
+
+        mock_set_in_carrier = Mock()
+
+        mock_get_current_span.configure_mock(
+            **{
+                "return_value": Mock(
+                    **{
+                        "get_span_context.return_value": None,
+                        "context.trace_flags": 0,
+                        "context.trace_id": 1,
+                        "context.trace_state": {constants.DD_ORIGIN: None},
+                    }
+                )
+            }
+        )
+
+        with tracer.start_as_current_span("parent"):
+            with tracer.start_as_current_span("child"):
+                FORMAT.inject(mock_set_in_carrier, {})
+
+        inject_fields = set()
+
+        for call in mock_set_in_carrier.mock_calls:
+            inject_fields.add(call[1][1])
+
+        # verify 'x-datadog-origin' is not present, it was None
+        expected_fields = FORMAT.fields.copy()
+        expected_fields.discard('x-datadog-origin')
+
+        self.assertEqual(expected_fields, inject_fields)


alertedsnake:
One of the tests failed with a "performance alert" when I merged master in, not sure what that's about?

exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/propagator.py
```diff
@@ -97,7 +97,7 @@ def inject(
             self.SAMPLING_PRIORITY_KEY,
             str(constants.AUTO_KEEP if sampled else constants.AUTO_REJECT),
         )
-        if constants.DD_ORIGIN in span.context.trace_state:
+        if span.context.trace_state.get(constants.DD_ORIGIN) is not None:
```
lonewolf3739, line 100: I believe this will not happen from the latest version. Earlier `TraceState` was simple dict and there were no validation upon add/update key-value pair. We have fixed that in latest release and key-value pair much adhere to W3C tracestate format rules otherwise they will be discarded.

alertedsnake: Ah very nice.  I'll have to see if I can recreate the scenario I described in #262 with the current versions.  Though, I am a little concerned that by disallowing that to be `None` we'll have moved the error from this code I patched, to whatever code was actually causing it in the first place.

alertedsnake: Okay, rebased on the current main branch, and after some more testing, this PR still fixes the issue described in #262, so whatever change you're describing did not fix things.

lzchen: That's strange that you are still running into that error. From [here](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/util/tracestate.py#L66) every value in the tracestate must be a str type.

alertedsnake: Yeah, and I still haven't figured out where it's coming from.

lonewolf3739: Did you update the `opentelemetry-api` and propagator versions you are using? Can you provide a reproducible example?

alertedsnake: Yep, this is with code from the main branch of both main and contrib repositories.

I'll see if I can strip down something - so far I've only seen when making gRPC client calls to a gRPC service running the botocore instrumentation and datadog exporter, so it's a little complex to reproduce, but I might be able to simplify more.  Give me a day or two.

alertedsnake:
Aha, so I've managed to nearly re-recreate the scenario in a simplified way, and I've seen this message in the log:

```
WARNING:opentelemetry.trace.span:Invalid key/value pair (dd_origin, ) found.
```

I assume this is the thing you were talking about @lonewolf3739  - it does make my patch not necessary.  So it's quite possible this is fixed, let me recreate all of my virtualenvs and pull in the latest, I must have missed something.

lzchen:
@alertedsnake 
Any updates on this?

alertedsnake:
@lzchen yeah, there's some more detail in #262 - at this point I have found where the problem happens, but I don't really know the intent of the code that's setting this field to None, was asking if you guys knew more about what it's supposed to be doing.

This comment specifically: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/262#issuecomment-778923127

alertedsnake:
This fix no longer seems necessary as there isn't an exception thrown anymore, but the underlying bug does still seem present, see further discussion in #262 .

alertedsnake:
Hi, so I'm going to re-open this, since it does actually fix an issue I'm finding with `None` making it into HTTP headers being sent via `botocore` requests;

```
TypeError: expected string or bytes-like object
```
The header that's incorrect looks like this:
```
'x-datadog-origin': None,
```

The code in this PR does indeed fix the issue.  But again, this is only fixing an incorrect issue from propagating, it's not fixing the issue from happening - see all of the previous discussion about that.

lonewolf3739:
Can you share some minimal reproducible example for this problem? I don't see the possibility of tracestate value to be set to None.

alertedsnake:
As I mentioned, there's a lot of previous discussion wherein I went through this in detail, both here and in the linked issue.  In this case, I've got a gRPC service that makes a query to DynamoDB, and it's when that Dynamo call gets made (with the botocore instrumentation) that one of the HTTP headers on the request is `'x-datadog-origin': None`.  

In this message I mention the code that causes the problem and asked for some insight from others who have worked on this - getting an answer here would help quite a bit:
https://github.com/open-telemetry/opentelemetry-python-contrib/issues/262#issuecomment-778923127

There isn't really a minimal case, I could probably make one up if necessary, but I think it's pretty clear where the problem occurs from that question.

lonewolf3739:
I haven't followed that discussion closely. If this indeed fixes the problem then there is a possibility that same problem exists in other propagators as well. If you could provide some reproducible example then we could find the root of problem and fix that instead of adding checks to prevent None getting propagated.

I don't think this fix changes anything by looking at the[`class TraceState(typing.Mapping[str, str]):`](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/trace/span.py#L196) implementation unless I am missing something.

alertedsnake:
Here is my test case: https://github.com/alertedsnake/otel-test with some instructions on how to run it and how to discover the problem.  Let me know if this helps.

lonewolf3739:
> Here is my test case: https://github.com/alertedsnake/otel-test with some instructions on how to run it and how to discover the problem. Let me know if this helps.

Thanks, I will check it.

lonewolf3739:
> Here is my test case: https://github.com/alertedsnake/otel-test with some instructions on how to run it and how to discover the problem. Let me know if this helps.

Looks like I need api keys/creds of various service providers otherwise I won't be able to run this example. I am fine with this getting in although I think this doesn't really fix the underlying problem. @alertedsnake  Did you try pip editable install or something similar to see if this really fixed your issue?

alertedsnake:
Yeah, you'll need data dog keys, or to use one of the agent mocking Docker images, I can try to get that working.

As for the fix in this PR, yes, it does prevent the exception. But yes, it doesn't fix the actual problem - I was still hoping to get some insight from the original author or someone who knows the suspect code well.

alertedsnake:
@lonewolf3739 Ok I updated the test case to not require Datadog credentials, and added a line to the readme about the AWS credentails - apparently you don't need either to see the error happen.

lonewolf3739:
@alertedsnake Thanks for sharing the updated example. I have done the troubleshooting and found the underlying problem. This line `if constants.DD_ORIGIN in span.context.trace_state:` evaluates to true even when there is no `dd_origin` key. The [`TraceState`](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/trace/span.py#L196) doesn't implement the `__contains__` so the fallback as described in [python docs](https://docs.python.org/3/reference/expressions.html#membership-test-details) returns true. 

> For user-defined classes which do not define __contains__() but do define __iter__(), x in y is True if some value z, for which the expression x is z or x == z is true, is produced while iterating over y. If an exception is raised during the iteration, it is as if in raised that exception.
>
> Lastly, the old-style iteration protocol is tried: if a class defines __getitem__(), x in y is True if and only if there is a non-negative integer index i such that x is y[i] or x == y[i], and no lower integer index raises the IndexError exception. (If any other exception is raised, it is as if in raised that exception).

And this line `span.context.trace_state[constants.DD_ORIGIN],` returns None which is the reason for this problem.

alertedsnake:
> This line `if constants.DD_ORIGIN in span.context.trace_state:` evaluates to true even when there is no `dd_origin` key. The [`TraceState`](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/trace/span.py#L196) doesn't implement the `__contains__` so the fallback as described in [python docs](https://docs.python.org/3/reference/expressions.html#membership-test-details) returns true.

This is the kind of thing I was hoping someone with more knowledge of OTEL could figure out, nice work!

lonewolf3739:
No this is not the actual fix. We need to change tracestate to behave correctly with membership. Once we do that this is no longer needed.

alertedsnake:
Okay, that sounds even better.  I'm happy to close this in favor of a ticket in the main repo.

alertedsnake:
> Okay, that sounds even better. I'm happy to close this in favor of a ticket in the main repo.

In fact, I've got a fix, will make a PR shortly.

alertedsnake:
https://github.com/open-telemetry/opentelemetry-python/pull/1773 - I don't think it needs anything more than this, but let me know if  so.

Happy to open an issue in that repo if we think it needs one.

lzchen:
@lonewolf3739 
Is this PR still relevant now that [this](https://github.com/open-telemetry/opentelemetry-python/pull/1773) is merged?

lonewolf3739:
This can be closed.

PR #273, opened 2020-12-22
Use instanceof to check if responses are valid Response objects

ThePumpingLemma:
There are some exception types that set the response attribute to a
non-Response type, e.g. "AttributeError: 'dict' object has no attribute
'status_code'"

# Description

At $WORK, we have an integration with a vendor (i.e. POST calls via `requests`) that will fail once-in-a-while with an exception where the `response` attribute on the exception is a dictionary instead of a `Response` object. Because the instrumentation doesn't have a type check, it tries to get the `status_code` attribute from the dictionary and fails with an `AttributeError`. Since it swallows the exception thrown from `requests`, we can't easily debug it (vendor issue is not easily reproducible). This change will allow the instrumentation to work in this edge case.

## Type of change

# How Has This Been Tested?

- [X] Added unit tests

- [X] No.

# Checklist:


lzchen (approved):

owais (approved):

lzchen:
@ThePumpingLemma 
If you can fix the conflicts we can merge this.

ThePumpingLemma:
> @ThePumpingLemma
> If you can fix the conflicts we can merge this.

Fixed.

PR #283, opened 2021-01-08
Fix ids_generator references from moving from api to sdk

lzchen:
Part of https://github.com/open-telemetry/opentelemetry-python/pull/1514

### DIFF
--- a/exporter/opentelemetry-exporter-datadog/tests/test_datadog_format.py
+++ b/exporter/opentelemetry-exporter-datadog/tests/test_datadog_format.py
@@ -18,6 +18,7 @@ from unittest.mock import Mock, patch
 from opentelemetry import trace as trace_api
 from opentelemetry.exporter.datadog import constants, propagator
 from opentelemetry.sdk import trace
+from opentelemetry.sdk.ids_generator import RandomIdsGenerator
 from opentelemetry.trace import get_current_span, set_span_in_context
 from opentelemetry.trace.propagation.textmap import DictGetter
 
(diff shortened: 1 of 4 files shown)


codeboten (approved):

exporter/opentelemetry-exporter-datadog/tests/test_datadog_format.py
```diff
@@ -18,6 +18,7 @@
 from opentelemetry import trace as trace_api
 from opentelemetry.exporter.datadog import constants, propagator
 from opentelemetry.sdk import trace
+from opentelemetry.sdk.trace.ids_generator import RandomIdsGenerator
```
toumorokoshi, line 21: should we be promoting direct imports from these sub-modules, or just ensuring that all required classes can be imported from the top-level-ish `opentelemetry.sdk.trace` module?

I feel like it's a bit rigid to import things from `opentelemetry.sdk.trace.ids_generator` (can't refactor the codebase if you have to split classes / etc). 

If you make `opentelemetry.sdk.trace` a kitchen sink with all of these classes, you not only enable that ability to refactor, but you also reduce the friction on the user having to remember what module some class actually is.

toumorokoshi (approved):
non-blocking comment worth taking a look, but LGTM!

exporter/opentelemetry-exporter-datadog/tests/test_datadog_format.py
lzchen: I agree with you but possibly could be addressed in a later date.

PR #285, opened 2021-01-12
Remove Configuration from instrumentations

### LINKED ISSUE #284: Remove usage of `Configuration`
As explained [here](https://github.com/open-telemetry/opentelemetry-python/issues/1494) we are removing the `Configuration` class. This will make some changes necessary here as well.

ocelotl:
# Description

As discussed [here](https://github.com/open-telemetry/opentelemetry-python/issues/1494) we are removing `Configuration` from the core repo. This means that it needs to be removed here as well. This PR does that.

This also adds an `opentelemetry.util` namespace where util code can be added to. Should this go into a separate PR? I also think so. Nevertheless, I am adding this here because there was some non-configuration-related code in the old `Configuration` class that I am moving into this namespace now. I understand that this can be controversial, so if you prefer this to be split into 2 PRs comment below, please. :+1: 

Fixes #284
## Type of change

- [x] Yes. - Link to PR: https://github.com/open-telemetry/opentelemetry-python/pull/1523
- [ ] No.

# Checklist:


lzchen:
Please update the title and description of the PR to be more informative.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
```diff
@@ -51,6 +53,46 @@
 ]
 
 
+class _ExcludeList:
```
codeboten, line 56: This code would be useful to share across instrumentations. This was discussed at the Jan 14th SIG meeting, and everyone present agreed it could be moved into the `opentelemetry-instrumentation` package

codeboten (commented):
PR looks pretty good, just waiting on responses to the comment about moving the excludelist code into the instrumentation package before approving.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
ocelotl: Before moving this into `opentelemetry-instrumentation` is there anything else you want me to move there as well? I see how also the common code that is used to handle URLs can be placed there.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/open-telemetry/opentelemetry-python-contrib/compare/v0.16b1...HEAD)
 
 ### Added
+- Remove Configuration
```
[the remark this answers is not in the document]
ocelotl: Fixed

CHANGELOG.md
```diff
@@ -41,6 +43,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#273](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/273))
 
 ### Changed
+- Remove Configuration
```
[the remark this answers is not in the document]
ocelotl: Fixed

ocelotl:
> Please update the title and description of the PR to be more informative.

Done! :+1:

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/__init__.py
```diff
@@ -43,11 +43,7 @@ def _instrument(self, **kwargs):
 
         # FIXME this is probably a pattern that will show up in the rest of the
         # ext. Find a better way of implementing this.
-        # FIXME Probably the evaluation of strings into boolean values can be
-        # built inside the Configuration class itself with the magic method
-        # __bool__
-
-        if Configuration().DJANGO_INSTRUMENT is False:
+        if environ.get("OTEL_PYTHON_DJANGO_INSTRUMENT") == "False":
```
lonewolf3739, line 46: Does it have to be strict case sensitive comparison? Can this be something similar to this `environ.get("OTEL_PYTHON_DJANGO_INSTRUMENT", "").lower() == "false"`?

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
```diff
@@ -12,22 +12,22 @@
...
 from logging import getLogger
+from time import time
 
 from django.conf import settings
 
-from opentelemetry.configuration import Configuration
 from opentelemetry.context import attach, detach
 from opentelemetry.instrumentation.django.version import __version__
 from opentelemetry.instrumentation.utils import extract_attributes_from_object
-from opentelemetry.instrumentation.wsgi import (
```
lonewolf3739, line 24: I believe this wsgi instrumentation library could be used for any wsgi framework to collect telemetry. For instance let's say I have a WSGI based framework X and there is no standalone instrumentation for it. If we remove this what would be the way for collecting telemetry?

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -11,17 +11,15 @@
...
 
 import fastapi
 from starlette.routing import Match
 
-from opentelemetry.configuration import Configuration
-from opentelemetry.instrumentation.asgi import OpenTelemetryMiddleware
-from opentelemetry.instrumentation.fastapi.version import __version__  # noqa
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
+from opentelemetry.util.http import get_excluded_urls
+from opentelemetry.util.http.asgi import OpenTelemetryMiddleware
```
lonewolf3739, line 20: Do we want this ASGI `OpenTelemetryMiddleware` to be part of `util`? I think it's better to have its own package?

lonewolf3739 (commented):
Overall LGTM except the decision to move WSGI/ASGI middleware instrumentation libraries to util.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
ocelotl: The WSGI code as it is now is not an instrumentation because it does no implement `BaseInstrumentor`. There is no WSGI library that is instrumented by it as opposed to specific libraries like Flask, Django, etc. For that reason, the WSGI code should not be in `opentelemetry.instrumentation`.

I don't understand what you mean with _"For instance let's say I have a WSGI based framework X and there is no standalone instrumentation for it. If we remove this what would be the way for collecting telemetry?"_, can you explain, please?

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
ocelotl: Maybe, but I am putting any HTTP-related utility code in the same place.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/__init__.py
ocelotl: We could, but that would be outside the scope of this PR.

owais, line 46: Don't we want to use constants/symbols here instead of strings like we did in the core PR?

owais (approved):

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
lonewolf3739: The [WSGI](https://opentelemetry-python.readthedocs.io/en/stable/instrumentation/wsgi/wsgi.html) and [ASGI](https://opentelemetry-python.readthedocs.io/en/stable/instrumentation/asgi/asgi.html) are different instrumentation libraries on their own as of now. I may be wrong but my understanding is that I can use [opentelemetry-instrumentation-wsgi](https://opentelemetry-python.readthedocs.io/en/stable/instrumentation/wsgi/wsgi.html#module-opentelemetry.instrumentation.wsgi) with any wsgi based web frameworks to [traces](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/master/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py#L203). I will try to use an example to clarify the last part, We don't have the instrumentation library for [cherrpy](https://docs.cherrypy.org/en/latest/index.html) framework but I can still instrument my `cherrpy` based application using `opentelemetry-instrumentation-wsgi`. It would look something similar as followed
```python
import cherrypy
from opentelemetry.instrumentation.wsgi import OpenTelemetryMiddleware

class HelloWorld(object):
    @cherrypy.expose
    def index(self):
        return "Hello World!"

app  = cherrypy.Application(HelloWorld(), '/')

app.wsgiapp = OpenTelemetryMiddleware(app.wsgiapp)

cherrypy.quickstart(app)
```
I didn't think of WSGI/ASGI `OpenTelemetryMiddleware` as an util although there are some util functions in that package. It  was easier to find that library, docs and use because it is very similar to any other popular middlewares [flask-cache](https://pythonhosted.org/Flask-Cache/) or [bottle](https://bottlepy.org/docs/dev/recipes.html) middlewares etc.. to name few. I thought may be util is not an ideal place to put middleware kind of stuff, just wanted to understand the motivation behind moving from individual packages to util. On the other hand I think it is good to move the helper functions under `util.http`.

util/opentelemetry-util-http/src/opentelemetry/util/http/asgi/__init__.py
```diff
@@ -27,10 +26,10 @@
 from asgiref.compatibility import guarantee_single_callable
 
 from opentelemetry import context, propagators, trace
-from opentelemetry.instrumentation.asgi.version import __version__  # noqa
 from opentelemetry.instrumentation.utils import http_status_to_status_code
 from opentelemetry.trace.propagation.textmap import DictGetter
 from opentelemetry.trace.status import Status, StatusCode
+from opentelemetry.util.http.asgi.version import __version__  # noqa
```
mariojonke, line 32: Not sure about moving the WSGI/ASGI `OpenTelemetryMiddleware` classes to the `opentelemetry-util-http` package. With this there are now 3 `version.py` files in the  package since both `OpenTelemetryMiddleware` implementations are de facto instrumentations which require the version to get the tracer for producing spans.


Imo WSGI/ASGI `OpenTelemetryMiddleware` should stay in their own dedicated packages (common helper methods for extracting span attriubtes, ... can be moved to the util package though), also for dependency reasons.  E.g. if one wants instrument an arbitrary WSGI application with the WSGI `OpenTelemetryMiddleware` from the `opentelemetry-util-http` the `asgiref` library will be pulled additionally even though it is not required.

lzchen:
@ocelotl 
Why was this closed?

ocelotl:
@lzchen sorry, this was a mistake, reopening...

util/opentelemetry-util-http/src/opentelemetry/util/http/wsgi/version.py
```diff
@@ -11,16 +11,5 @@
...
-BASE_DIR = os.path.dirname(__file__)
-VERSION_FILENAME = os.path.join(
-    BASE_DIR, "src", "opentelemetry", "instrumentation", "wsgi", "version.py"
-)
-PACKAGE_INFO = {}
-with open(VERSION_FILENAME) as f:
-    exec(f.read(), PACKAGE_INFO)
-
-setuptools.setup(version=PACKAGE_INFO["__version__"])
+__version__ = "0.170"
```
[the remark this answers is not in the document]
ocelotl: Right, fixed.

util/opentelemetry-util-http/src/opentelemetry/util/http/asgi/__init__.py
```diff
@@ -13,12 +13,11 @@
 # limitations under the License.
 
 """
-The opentelemetry-instrumentation-asgi package provides an ASGI middleware that can be used
+The opentelemetry-util-web package provides an ASGI middleware that can be used
```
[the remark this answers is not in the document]
ocelotl: Fixed

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/__init__.py
ocelotl: Fixed

util/opentelemetry-util-http/src/opentelemetry/util/http/asgi/__init__.py
ocelotl: @mariojonke @lonewolf3739 

I understand that the current ASGI and WSGI packages are closely related to instrumentation. Nevertheless, they are not complete instrumentations as the rest of the instrumentations because they are not instrumenting anything by themselves when auto instrumentation happens nor they implement `BaseInstrumentor`. For that reason I believe they should not be part of any `instrumentation` package. The idea of `util` is pretty much to have a place where we can put anything that does not fit in the category of exporter, instrumentation, api, sdk, etc. I believe this to be the case, even when I see how this approach causes some concern because of the close relationship between the ASGI and WSGI and the instrumentation of other third party libraries.

Maybe having a `util.instrumentation.asgi`/`wsgi` package would be a better solution? :shrug:

codeboten (changes requested):
For the sake of getting the configuration piece moved forward, please keep the asgi/wsgi packages as separate for now. I feel like that discussion can be had separately from removing the config.

ocelotl:
> For the sake of getting the configuration piece moved forward, please keep the asgi/wsgi packages as separate for now. I feel like that discussion can be had separately from removing the config.

Sure, separating this

codeboten (approved):
Thanks for addressing my comments!

PR #286, opened 2021-01-19
Aiohttp instrumentation readme

theelous3:
# Description

Improved documentation with a minimal viable example for aiohttp instrumentation. There is currently not a clear example of this anywhere in the wild, nor is there much in the way of documentation explaining what glue is required between instrumentation and jaeger / opentelemetry layers. Now there is.

Fixes # (issue)

## Type of change

Small docs change.

# How Has This Been Tested?

You can take a look at the well rendered example here: https://github.com/theelous3/opentelemetry-python-contrib/tree/aiohttp-instrumentation-readme/instrumentation/opentelemetry-instrumentation-aiohttp-client

Run the code in the example yourself if you like. Should be self explanatory.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

N/A

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-aiohttp-client/README.rst
+++ b/instrumentation/opentelemetry-instrumentation-aiohttp-client/README.rst
@@ -33,26 +33,26 @@ Example
    from opentelemetry.sdk.trace.export import BatchExportSpanProcessor
 
 
-   jaeger_exporter = jaeger.JaegerSpanExporter(
-      service_name="xxx",
+   _JAEGER_EXPORTER = jaeger.JaegerSpanExporter(
+      service_name="example-xxx",
       agent_host_name="localhost",
       agent_port=6831,
    )
 
    _TRACE_PROVIDER = TracerProvider()
-   _TRACE_PROVIDER.add_span_processor(BatchExportSpanProcessor(jaeger_exporter))
+   _TRACE_PROVIDER.add_span_processor(BatchExportSpanProcessor(_JAEGER_EXPORTER))
    trace.set_tracer_provider(_TRACE_PROVIDER)
 
    AioHttpClientInstrumentor().instrument()
 
 
-   async def span_emmitter():
+   async def span_emitter():
       async with aiohttp.ClientSession() as session:
          async with session.get("https://example.com") as resp:
                print(resp.status)
 
 
-   asyncio.run(span_emmitter())
+   asyncio.run(span_emitter())
 
 
 References
--- a/instrumentation/opentelemetry-instrumentation-aiohttp-client/README.rst
+++ b/instrumentation/opentelemetry-instrumentation-aiohttp-client/README.rst
@@ -17,6 +17,44 @@ Installation
      pip install opentelemetry-instrumentation-aiohttp-client
 
 
+Example
+-------
+
+.. code:: python
+
+   import asyncio
+   
+   import aiohttp
+
+   from opentelemetry.instrumentation.aiohttp_client import AioHttpClientInstrumentor
+   from opentelemetry import trace
+   from opentelemetry.exporter import jaeger
+   from opentelemetry.sdk.trace import TracerProvider
+   from opentelemetry.sdk.trace.export import BatchExportSpanProcessor
+
+
+   jaeger_exporter = jaeger.JaegerSpanExporter(
+      service_name="xxx",
+      agent_host_name="localhost",
+      agent_port=6831,
+   )
+
+   _TRACE_PROVIDER = TracerProvider()
+   _TRACE_PROVIDER.add_span_processor(BatchExportSpanProcessor(jaeger_exporter))
+   trace.set_tracer_provider(_TRACE_PROVIDER)
+
+   AioHttpClientInstrumentor().instrument()
+
+
+   async def span_emmitter():
+      async with aiohttp.ClientSession() as session:
+         async with session.get("https://example.com") as resp:
+               print(resp.status)
+
+
+   asyncio.run(span_emmitter())
+
+
 References
 ----------
 


lonewolf3739:
> Improved documentation with a minimal viable example for aiohttp instrumentation. There is currently not a clear example of this anywhere in the wild, nor is there much in the way of documentation explaining what glue is required between instrumentation and jaeger / opentelemetry layers.

I am curious to understand this. Did you check out the otel-python [readthedocs](https://opentelemetry-python.readthedocs.io/en/stable/index.html) and [getting started](https://opentelemetry-python.readthedocs.io/en/stable/getting-started.html) guide to understand setting up pipeline to export traces to console/jaeger etc...? Did that not make it clear to you?

theelous3:
> I am curious to understand this. Did you check out the otel-python readthedocs and getting started guide to understand setting up pipeline to export traces to console/jaeger etc...? Did that not make it clear to you?

No, and this is mostly because it is not clear what "instrumentation" is, or is supposed to be. The relationship between instrumentation and other layers of the protocols and implementation is unclear.

When looking to use something, the first thing that most people (including myself) look at is example code. Example code should, in my opinion, always offer a minimum working example, not some abstract snippet with no clear relation to other required segments of code.

For example, look at the aiohttp docs: https://docs.aiohttp.org/en/stable/

The first example is a well formed and compete minimal example that will run. Same for aiopg: https://aiopg.readthedocs.io/en/stable/

As it should be for all documentation.

lonewolf3739:
> No, and this is mostly because it is not clear what "instrumentation" is, or is supposed to be. The relationship between instrumentation and other layers of the protocols and implementation is unclear.
When looking to use something, the first thing that most people (including myself) look at is example code. Example code should, in my opinion, always offer a minimum working example

The [getting-started](https://opentelemetry-python.readthedocs.io/en/stable/getting-started.html) guide has what it consideres to be the ["Hello World"](https://opentelemetry-python.readthedocs.io/en/stable/getting-started.html#hello-world-emitting-a-trace-to-your-console) of opentelemetry-python tracing. I believe it is minimal and working example of introducing this library. And as one progresses through the guide there is an [example](https://opentelemetry-python.readthedocs.io/en/stable/getting-started.html#configure-exporters-to-emit-spans-elsewhere) for exporting traces to one of the popular backends Jaeger. It clearly highlights how to run jaeger locally, the relevant opentelemetry-python dependency to be installed and where to find the generated traces, then it has the instrumentation [example](https://opentelemetry-python.readthedocs.io/en/stable/getting-started.html#integrations-example-with-flask) for simple flask application.  It would be great if you can share what lacked in the getting started guide.

The example in this PR depends and `opentelemetry-sdk` and `opentelemetry-exporter-jaeger` and assumes that there is Jaeger running locally. There is nothing written about dependencies to be installed and how to run jaeger locally. All of that is already covered in the getting started guide. Each instrumentation library should ideally has a example of how to instrument it and leave the tracer pipeline, choosing the exporter and tracing backend to end user.

The docs haven't been updated for long time and might have some issues. It will be addressed soon and there is an issue for it.

theelous3:
> It would be great if you can share what lacked in the getting started guide.

I believe I have done this already. The aiohttp instrumentation examples are incomplete and non-functional.

> There is nothing written about dependencies to be installed and how to run jaeger locally. All of that is already covered in the getting started guide

I don't mean to get too cheeky here, but by way of your reasoning above, this should be apparent to the user. However, if you like, I can add some disclaimer text before the example:

This example uses the `jaegertracing/all-in-one` image running locally. It requires the `opentelemetry-sdk` and `opentelemetry-exporter-jaeger` packages to be installed.

lonewolf3739:
Probably you misunderstood me. The library usage example contains how to configure the instrumentation for that particular library not things like setting up a trace pipeline and exporter etc... If you look at the first three sections of [getting-started](https://opentelemetry-python.readthedocs.io/en/stable/getting-started.html) guide you should have enough understanding about boilerplate and glue. For example [requests](https://opentelemetry-python.readthedocs.io/en/stable/instrumentation/requests/requests.html) usage example shows how to integrate with requests.

theelous3:
Yes. If it was up to me, all of the instrumentations would have complete functional examples. I believe this to be a problem generally with this project, so I don't really accept another lacking doc page as a good standard to follow. That page even shows `RequestInstrumentor.instrument()` being called twice in a row for no apparent reason, and is missing `.instrument` from the `RequestInstrumentor` object api.

The general standard across mature and well formed libraries as far as I have ever seen them, is to provide working examples.

The only thing different here is the (in my opinion awkward) structure, wherein what really should be totally separate libraries such as `AiopgInstrumentor` are packaged plug-in-ish-ly under `opentelemetry.instrumentation`, but their own pypi packages.

These sub-libraries are distinct enough that they deserve working examples.

lonewolf3739:
I agree that instrumentations should have complete functional examples(to be able to view the traces generated in UI). There are multiple exporters and end users could be interested in different exporter than Jaeger (could be Zipkin or something else) and there are dozens of instrumentations (more will be added). So the same trace pipeline is documented once and could be used for any library.

theelous3:
I see your point, but it is often the case that layers below some library code can differ, and in these cases it is perfectly sensible to select the simplest / most popular / author's preferred underlayer to use in documentation.

If someone is using prometheus and they see the jaeger exporter, and have read the pre-example text saying "this example uses `jaeger`" I'm sure they can figure out that jaeger != prometheus. Whatever remote possibility of confusion here is way offset by having the example actually work.

The top text in the docs reads:

> The opentelemetry-instrumentation-aiohttp-client package allows tracing HTTP requests made by the aiohttp client library.

And following that, the example immediately following that should do that. It's ridiculously ambiguous what "allows" means. At face value it reads such that by using just `opentelemetry-instrumentation-aiohttp-client` you can do `tracing`. You cannot.

codeboten:
@theelous3 please sign the CLA so we can get this merged, thanks!

codeboten (approved):

theelous3:
Signed.

PR #293, opened 2021-01-24
Update instrumentation examples

### LINKED ISSUE #292: Update instrumentations usage examples

lonewolf3739:
# Description

Fixes #292

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-aiopg/README.rst
+++ b/instrumentation/opentelemetry-instrumentation-aiopg/README.rst
@@ -6,6 +6,11 @@ OpenTelemetry aiopg instrumentation
 .. |pypi| image:: https://badge.fury.io/py/opentelemetry-instrumentation-aiopg.svg
    :target: https://pypi.org/project/opentelemetry-instrumentation-aiopg/
 
+Documentation
+-------------
+
+https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/aiopg/aiopg.html
+
 Installation
 ------------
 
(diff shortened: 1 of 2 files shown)


lzchen:
@lonewolf3739 
Hey any progress on this?

lonewolf3739:
@lzchen Sorry I totally forgot about this. I wanted to get feedback for this one example. Continuing the discussion here, Personally the decision to not take dependency on SDK and keeping it minimal makes sense to me(after knowing the design decisions etc..) but as an end user I would expect copy-pasting the [example](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/aiopg/aiopg.html#usage) from the documentation to show traces to me. If one reads the some instrumentation documentation and follows the example it doesn't really help. My one major concern is that it not obvious for a user who wants to instrument their app should set the tracing pipeline with exporters of their choice to have complete functional example.

lzchen:
@lonewolf3739 
> I would expect copy-pasting the example from the documentation to show traces to me. 

Yes your point makes complete sense, which is why I think we should still have an fully-working example but in some separate examples folder. I believe we had an issue before (can't seem to find it), where a user attempted to copy and paste the usage docstring without having the SDK installed and they filed an issue. The usage doc string would only be used for users who want to collect, not export as well (although having this exist may be confusing). Should we maybe remove the usage docs?

lonewolf3739:
> @lonewolf3739
> 
> > I would expect copy-pasting the example from the documentation to show traces to me.
> 
> Yes your point makes complete sense, which is why I think we should still have an fully-working example but in some separate examples folder. I believe we had an issue before (can't seem to find it), where a user attempted to copy and paste the usage docstring without having the SDK installed and they filed an issue. The usage doc string would only be used for users who want to collect, not export as well (although having this exist may be confusing). Should we maybe remove the usage docs?

then I would suggest we remove the example from usage docs and update with link to complete functional example (could be readme or core repo examples folder).

lzchen:
@lonewolf3739 
> then I would suggest we remove the example from usage docs and update with link to complete functional example (could be readme or core repo examples folder).

@codeboten thoughts?

lzchen:
Discussed in SIG. Will be leaving the usage docs and adding examples and linking to main repo folder.

PR #333, opened 2021-02-12
Update main after 1.0.0rc1 release from core

lzchen:


lonewolf3739 (approved):

lonewolf3739:
Not sure if it is related to this PR; We should do a release for instrumentations with `opentelemetry-api` rc1 as dependency  otherwise pip fails with errors as all the latest instrumentations have `opentelemetry-api==0.17b0` which is incompatible.

toumorokoshi (approved):

owais (approved):

PR #358, opened 2021-03-02
Add missing dependency

### LINKED ISSUE #357: gRPC instrumentation is missing wrapt
The gRPC instrumentation uses wrapt but it is missing from its dependencies. Fix this.

ocelotl:
# Description

Add wrapt for the gRPC instrumentation

Fixes #357 

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-grpc/setup.cfg
+++ b/instrumentation/opentelemetry-instrumentation-grpc/setup.cfg
@@ -42,6 +42,7 @@ install_requires =
     opentelemetry-api == 1.0.0.dev0
     opentelemetry-sdk == 1.0.0.dev0
     grpcio ~= 1.27
+    wrapt >= 1.0.0, < 2.0.0
 
 [options.extras_require]
 test =


codeboten (approved):

lzchen (approved):

instrumentation/opentelemetry-instrumentation-grpc/setup.cfg
```diff
@@ -42,6 +42,7 @@ install_requires =
     opentelemetry-api == 1.0.0.dev0
     opentelemetry-sdk == 1.0.0.dev0
     grpcio ~= 1.27
+    wrapt >= 1.0.0, < 2.0.0
```
alertedsnake, line 45: Why not `~= 1.0`?  
It should be noted that `2.0.0rc1` would fit the `< 2.0.0` requirement.

owais: >  It should be noted that 2.0.0rc1 would fit the < 2.0.0 requirement.

That's a bit surprising to me but in any case, wouldn't that only be the case if `--pre` is used with `pip install`? Not advocating for one way to specify deps. Just curious about how it works exactly.

alertedsnake: PEP-444 says:

> The exclusive ordered comparison <V MUST NOT allow a pre-release of the specified version unless the specified version is itself a pre-release. Allowing pre-releases that are earlier than, but not equal to a specific pre-release may be accomplished by using <V.rc1 or similar.

So you're right, but I'm now wondering why this has failed so much for me in the past?  I've definitely released broken versions into production because an alpha version got dragged by a "less than the next major" requirement spec, and I've never used `--pre`.  Strange.

owais (approved):

PR #379, opened 2021-03-25
Reimplement the keys method in WSGI CarrierGetter

### LINKED ISSUE #376: WSGI Carrier Getter always returns an empty list as keys
**Describe your environment**
Not relevant.

**Steps to reproduce**
1. Call the method: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/f8e51c488da1f5ab628ba4f5bf25e65638974c44/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py#L91

I would expect all the keys from the provided carrier returned.

No matter of the carrier contents an empty list is returned.

**Additional context**
Due to this behavior other extensions from this repository might not work as expected.
For example [OT Trace Propagator](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/f8e51c488da1f5ab628ba4f5bf25e65638974c44/propagator/opentelemetry-propagator-ot-trace/src/opentelemetry/propagators/ot_trace/__init__.py#L92) cannot extract baggage information from the carrier, because empty list is always returned.

marcinzaremba:
# Description

Implements keys method in WSGI CarrierGetter (previously a static, empty list was returned)

Fixes #376 

## Type of change

# How Has This Been Tested?

See the unit test added in this PR.

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
@@ -66,6 +66,8 @@ from opentelemetry.propagators.textmap import DictGetter
 from opentelemetry.trace.status import Status, StatusCode
 
 _HTTP_VERSION_PREFIX = "HTTP/"
+_CARRIER_KEY_PREFIX = "HTTP_"
+_CARRIER_KEY_PREFIX_LEN = len(_CARRIER_KEY_PREFIX)
 
 
 class CarrierGetter(DictGetter):
@@ -89,7 +91,11 @@ class CarrierGetter(DictGetter):
         return None
 
     def keys(self, carrier):
-        return []
+        return [
+            key[_CARRIER_KEY_PREFIX_LEN:].lower().replace("_", "-")
+            for key in carrier
+            if key.startswith(_CARRIER_KEY_PREFIX)
+        ]
 
 
 carrier_getter = CarrierGetter()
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -89,7 +91,11 @@ def get(
         return None
 
     def keys(self, carrier):
-        return []
+        return [
+            key[_CARRIER_KEY_PREFIX_LEN:].lower().replace("_", "-")
+            for key in carrier
+            if key.startswith(_CARRIER_KEY_PREFIX)
+        ]
```
owais, line 98: Out of curiosity, did this prevent or limit propagation in some way when it was implemented as an empty list?

marcinzaremba: It prevents any use case when information to be extracted is not known beforehand (for example HTTP headers starting with a particular prefix, but with unknown exact names). The great example is OT Trace Propagator, where the current implementation (empty list) renders injecting baggage unusable: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/f8e51c488da1f5ab628ba4f5bf25e65638974c44/propagator/opentelemetry-propagator-ot-trace/src/opentelemetry/propagators/ot_trace/__init__.py#L92

owais: >  The great example is OT Trace Propagator, where the current implementation (empty list) renders injecting baggage unusable: 

Thanks. This is what I was interested in.

owais (approved):

codeboten (approved):
👍

PR #381, opened 2021-03-25
fix instrumentation of connection when pool.acquire was called multip…

### LINKED ISSUE #336: aiopg instrumentation creating nested traces
**Describe your environment**
Python 3.6.12
Macos 11.1
opentelemetry-python-contrib: Tried released version, as well as `main`.
aiohttp==3.7.3
aiopg==1.1.0

 Describe any aspect of your environment relevant to the problem, including your Python version, [platform](https://docs.python.org/3/library/platform.html), version numbers of installed dependencies, information about your cloud hosting provider, etc. If you're reporting a problem with a specific version of a library in this repo, please check whether the problem has been fixed on main.

**Steps to reproduce**
I am trying the `aiopg` instrumentation. I am using it along aiohttp. This may be an usage issue, but I have created a minimal example in [here](https://github.com/cdvv7788/aiopg-instrumentor-bug) (Just install requirements, run, and visit `localhost:4000`...you will need a postgres database named test and jaeger running somewhere).
The trace will keep growing with every request, nesting new traces every time:
image

Every connection has a single trace (no nesting)

Every time a request is made, a new trace is nested and sends the whole stack to jaeger

**Notes**
This may be a bad usage issue. If that is the case, please let me know what may be causing it and how to use this properly.

sartx:
…le times

# Description

Fix appearing multiple nested spans when instrumented `aiopg.pool` was used. The reason of the bug was that `pool.acquire` instrumented connection every time. But pool reuses old connections which had already been instrumented. This fix addes checking connection. Instrumenting connection only if it was not instrumented.

Fixes #336

## Type of change

# How Has This Been Tested?

See added integration test in this PR.

- [ ] No.

# Checklist:


owais (approved):

instrumentation/opentelemetry-instrumentation-aiopg/src/opentelemetry/instrumentation/aiopg/aiopg_integration.py
```diff
@@ -86,9 +86,11 @@ def acquire(self):
         async def _acquire(self):
             # pylint: disable=protected-access
             connection = await self.__wrapped__._acquire()
-            return get_traced_connection_proxy(
-                connection, db_api_integration, *args, **kwargs
-            )
+            if not hasattr(connection, "__wrapped__"):
```
owais, line 89: we probably could check if the wrapper type is the same as the instrumented type but not sure if it's worth it. This is probably enough,.

toumorokoshi: I'd probably have argued for something a bit clearer that it was wrapped by opentelemetry, such as `__wrapped_by_opentelemetry__`. But agreed it's probably overkill: highly unlikely to overlap.

tests/opentelemetry-docker-tests/tests/postgres/test_aiopg_functional.py
```diff
@@ -185,3 +189,19 @@ def test_callproc(self):
         ):
             async_call(self._cursor.callproc("test", ()))
             self.validate_spans("test")
+
+    def test_instrumented_pool_with_multiple_acquires(self, *_, **__):
+        async def double_asquire():
```
toumorokoshi, line 194: is `asquire` a typo?

toumorokoshi (approved):
LGTM! possible typo worth looking at.

tests/opentelemetry-docker-tests/tests/postgres/test_aiopg_functional.py
sartx: Yes. You are right.

instrumentation/opentelemetry-instrumentation-aiopg/src/opentelemetry/instrumentation/aiopg/aiopg_integration.py
sartx: I agree with you. `if not isinstance(connection, AsyncProxyObject):` - I think that this option more accurately defines the case of verification.

owais: It does but not sure if it is worth the extra check :) Probably we should do what @toumorokoshi suggested in a separate PR so all instrumentations benefit from it,.

codeboten (approved):
Thanks for the fix! 👍

cdvv7788:
Thanks!

PR #395, opened 2021-04-02
Django trace response headers

owais:
# Description

Added opt-in support to return trace response headers from Django.

This allows users to configure their Django apps to inject trace context
as headers in HTTP responses. This is useful when client side apps need
to connect their spans with the server side spans e.g, in RUM products.

Today the most practical way to do this is to use the `Server-Timing`
header but in near future we might use the `traceresponse` header as
described here: https://w3c.github.io/trace-context/#trace-context-http-response-headers-format

As a result the implementation does not use a hard-coded header and
instead let's the users pick one.

This can be done by setting the `OTEL_PYTHON_TRACE_RESPONSE_HEADER` to
the header name that users want to inject in HTTP responses. The option
does not have a default value and the feature is disbaled when a env var
is not set.

## Type of change

# How Has This Been Tested?

- [x] Tested manually
- [x] Added unit tests

# Does This PR Require a Contrib Repo Change?

- [x] Yes. - Link to PR: https://github.com/open-telemetry/opentelemetry-python/pull/1762
- [ ] No.

# Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ on:
     - 'release/*'
   pull_request:
 env:
-  CORE_REPO_SHA: cad261e5dae1fe986c87e6965664b45cc9ab73c3
+  CORE_REPO_SHA: 7b11971c504387341df0c38f5a34d7d1293c7e4f
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


aabmass:
My comment from the SIG meeting:

Since W3C trace context is core to OTel and has drafted `traceresponse` back-propagation, this will probably make its way into the OTel Propagator spec. I'm trying to imagine what that API might look like, maybe:
1. We keep a single global textmap propagator but add a new method e.g. `injectResponse(carrier, setter)`. Or
2. We add a global textmap "back-propagator" with the same signature as the existing one `inject(carrier, setter)`

I just want to make sure this code is easy to migrate when this happens, or maybe python can have a prototype for this feature in the spec. @owais any idea of the timeline for the `traceresponse` header to be stable?

owais:
@aabmass Thanks. That makes sense and definitely makes everything a lot more cleaner and easier to implement in multiple instrumentations. I added an experimental back propagator in https://github.com/open-telemetry/opentelemetry-python/pull/1762 

I'll get back to you on the timeframe for the spec proposal.

owais:
Closed in favor of #436

PR #398, opened 2021-04-03
Fix Xray ID Generator Format Issue

schanjr:
# Description

Upon using the SDK class AwsXRayIdGenerator(), I found that the trace_id and span_id generated is in completely wrong format although a correct Xray Documentation was pasted in the python class. A sample implementation in Javascript is here: https://github.com/aws-observability/aws-otel-js/tree/main/packages/opentelemetry-id-generator-aws-xray 

Before: 
trace_id and span_id gave random numbers 

After: 
trace_id is a combination of a 8 hexadecimal digits created from timestamp and 24 random hexadecimal digit as per [aws documentation](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-traceids)

span_id is equivalent to [xray segment id](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html) which is 16 random hexadecimal digits 

Fixes # (issue)
Fixed format for generating trace id and span id created by AwsXRayIdGenerator()

## Type of change

# How Has This Been Tested?

- [x] For trace Id, modified test to reconvert the portion of trace_id timestamp from hexadecimal back to int. The rest of the trace_id and span_ids are completely randomized, as such no longer need any specific tests except length. 
- [x] Added test to assert trace id and span id length

- [x] No.

# Checklist:

### DIFF
--- a/sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/trace/aws_xray_id_generator.py
+++ b/sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/trace/aws_xray_id_generator.py
@@ -56,9 +56,8 @@ API
 .. _trace ID format: https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-traceids
 """
 
-import binascii
-import os
 import time
+import random
 
 
 class AwsXRayIdGenerator():
@@ -67,14 +66,15 @@ class AwsXRayIdGenerator():
     See Same implementation in Javascript: https://github.com/aws-observability/aws-otel-js/tree/main/packages/opentelemetry-id-generator-aws-xray
     """
     TIME = time
+    TRACE_ID_LENGTH = 32
+    SPAN_ID_LENGTH = 16
 
     # Ex: 0c6d1c759808e783
-    @staticmethod
-    def generate_span_id() -> str:
-        return binascii.b2a_hex(os.urandom(8)).decode()
+    def generate_span_id(self) -> str:
+        return (hex(random.getrandbits(88))[2:])[:self.SPAN_ID_LENGTH]
 
     # Ex: 6068d890e80ef8ae6323f667558381bd
     def generate_trace_id(self) -> str:
-        hexa = hex(int(self.TIME.time()))[2:]
-        bina = binascii.b2a_hex(os.urandom(12)).decode()
-        return hexa + bina
+        stamp8 = hex(int(self.TIME.time()))[2:]
+        random24 = hex(random.getrandbits(116))[2:]
+        return (stamp8 + random24)[:self.TRACE_ID_LENGTH]
--- a/sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/trace/aws_xray_id_generator.py
+++ b/sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/trace/aws_xray_id_generator.py
@@ -56,29 +56,25 @@ API
 .. _trace ID format: https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-traceids
 """
 
-import random
+import binascii
+import os
 import time
 
-from opentelemetry.sdk.trace.id_generator import IdGenerator, RandomIdGenerator
 
-
-class AwsXRayIdGenerator(IdGenerator):
-    """Generates tracing IDs compatible with the AWS X-Ray tracing service. In
-    the X-Ray system, the first 32 bits of the `TraceId` are the Unix epoch time
-    in seconds. Since spans (AWS calls them segments) with an embedded timestamp
-    more than 30 days ago are rejected, a purely random `TraceId` risks being
-    rejected by the service.
-
+class AwsXRayIdGenerator():
+    """Generates tracing IDs compatible with the AWS X-Ray tracing service.
     See: https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-traceids
+    See Same implementation in Javascript: https://github.com/aws-observability/aws-otel-js/tree/main/packages/opentelemetry-id-generator-aws-xray
     """
+    TIME = time
 
-    random_id_generator = RandomIdGenerator()
-
-    def generate_span_id(self) -> int:
-        return self.random_id_generator.generate_span_id()
-
+    # Ex: 0c6d1c759808e783
     @staticmethod
-    def generate_trace_id() -> int:
-        trace_time = int(time.time())
-        trace_identifier = random.getrandbits(96)
-        return (trace_time << 96) + trace_identifier
+    def generate_span_id() -> str:
+        return binascii.b2a_hex(os.urandom(8)).decode()
+
+    # Ex: 6068d890e80ef8ae6323f667558381bd
+    def generate_trace_id(self) -> str:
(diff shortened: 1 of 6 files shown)


schanjr:
I already signed the CLA by clicking "MISSING ID ON COMMIT". Git log also has my email and user name... No idea what else is needed to get rid of that error. Please advise.

schanjr:
Reviewed some failure builds: 
1. https://github.com/open-telemetry/opentelemetry-python-contrib/pull/398/checks?check_run_id=2261556289  - Looked like data dog flaky tests? It is not related to xray and should not import the modified class. 
2. https://github.com/open-telemetry/opentelemetry-python-contrib/pull/398/checks?check_run_id=2261556344 - Same comments, related to data dog 
3. https://github.com/open-telemetry/opentelemetry-python-contrib/pull/398/checks?check_run_id=2261556351 - Seems to be caused by the new way of generating random hex. I'll run some benchmarks again later to see what are other ways to improve the performance. 
4. https://github.com/open-telemetry/opentelemetry-python-contrib/pull/398/checks?check_run_id=2261556211 - I had a lot of issue when running the tests locally so I created some setup to add the related file to the PYTHON_PATH as doing the `from opentelemetry.sdk.extension.aws.trace import AwsXRayIdGenerator` kept on trying to use the pip installed library which had the old code. Let me know if this is not needed in the actual build, I can remove it... 
5. Not sure how to fix the EasyCLA

schanjr:
Attempted to address the performance. Here are my benchmarks without any kind of shared buffer implementation. Let me know if performance is still considered a major issue, I didn't want to over engineer at this point.

```python 
import binascii
import os
import time
import random
import secrets

ITERATIONS = 10000000


def baseline():
    stamp = hex(int(time.time()))[2:]
    random24 = hex(random.getrandbits(116))[2:]
    return (stamp + random24)[:32]

def format_baseline():
    stamp = format(int(time.time()), "x")
    random24 = format(random.getrandbits(116),"x")
    return (stamp + random24)[:32]

def original_algo():
    hexa = hex(int(time.time()))[2:]
    bina = binascii.b2a_hex(os.urandom(12)).decode()
    return hexa + bina

def secret_token():
    hexa = hex(int(time.time()))[2:]
    token = secrets.token_hex(16)
    return hexa + token


if __name__ == '__main__':
    import timeit
    print('baseline', timeit.timeit("baseline()", setup="from __main__ import baseline", number=ITERATIONS))
    print('format_baseline', timeit.timeit("format_baseline()", setup="from __main__ import format_baseline", number=ITERATIONS))
    print('original algo', timeit.timeit("original_algo()", setup="from __main__ import original_algo", number=ITERATIONS))
    print('secret_token', timeit.timeit("secret_token()", setup="from __main__ import secret_token", number=ITERATIONS))

```

results 

```bash
/Users/chanst/Github/opentelemetry-python-contrib/venv/bin/python "/Users/chanst/Library/Application Support/JetBrains/IntelliJIdea2020.3/scratches/python/benchmark_gen.py"
baseline 9.068622905
format_baseline 9.382549085
original algo 14.069556111
secret_token 17.044063333000004

Process finished with exit code 0
 
```

alolita:
@anuraaga @awssandra Can you please review? Thx.

sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/trace/aws_xray_id_generator.py
```diff
@@ -56,29 +56,25 @@
...
-    @staticmethod
-    def generate_trace_id() -> int:
-        trace_time = int(time.time())
-        trace_identifier = random.getrandbits(96)
-        return (trace_time << 96) + trace_identifier
+    TIME = time
+    TRACE_ID_LENGTH = 32
+    SPAN_ID_LENGTH = 16
+
+    # Ex: 0c6d1c759808e783
```
anuraaga, line 72: Hi @schanjr - the interface for ID generation in python and JS are quite different, in python the return values must be integers

https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/trace/id_generator.py#L21

So I think this new implementation won't actually work when the SDK tries to use it and gets a string back.

I don't see anything wrong with the previous implementation, can you describe where you're seeing wrong IDs?

sdk-extension/opentelemetry-sdk-extension-aws/tests/trace/test_aws_xray_ids_generator.py
```diff
@@ -15,28 +15,46 @@
...
             trace_id = id_generator.generate_trace_id()
             self.assertTrue(trace_id != INVALID_TRACE_ID)
             span_id = id_generator.generate_span_id()
             self.assertTrue(span_id != INVALID_TRACE_ID)
 
-    def test_id_timestamps_are_acceptable_for_xray(self):
+    def test_trace_id_timestamps_are_acceptable_for_xray(self):
         id_generator = AwsXRayIdGenerator()
         for _ in range(1000):
             trace_id = id_generator.generate_trace_id()
```
anuraaga, line 43: I think this test could be improved to something like

```python
before_time = int(time.time())
trace_id = id_generate_trace_id()
trace_id_time = trace_id >> 96
after_time = int(time.time())
self.assertGreater(trace_id_time, before_time)
self.assertLessEqual(trace_id_time, after_time)
```

but given this runs 1000 times we'd expect the test to already be checking the timestamp reasonably well.

sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/trace/aws_xray_id_generator.py
schanjr: @anuraaga Thank you for taking a look at this. Upon inspecting the logs again, I realized the final output did indeed convert the int values to hex at some point. The OTEL collector in turn converts the hex to xray format before pushing to aws. I was a bit surprised as the JS version of SDK uses the same OTEL collector image and pushes the actual xray format to the collector. The collector did not need to do the conversion. 


 Sorry for the trouble :) 


span console output
```json
{
    "name": "/health asgi",
    "context": {
        "trace_id": "0x606a8c10b37a1e9cad7c550289dae5ae",
        "span_id": "0xe6f70a22665060ae",
        "trace_state": "[]"
    },

```

otel collector image 
```
    "{\"name\":\"fastapi-app\",\"id\":\"e6f70a22665060ae\",\"start_time\":1617595408.623005,\"trace_id\":\"1-606a8c10-b37a1e9cad7c550289dae5ae\",\"end_time\":1617595408.623856,\"http\":{\"request\":{\"method\":\"GET\",\"url\":\"http://127.0.0.1:31000/health\",\"user_agent\":\"PostmanRuntime/7.26.10\",\"client_ip\":\"127.0.0.1\"},\"response\":{\"status\":200,\"content_length\":0}},\"fault\":false,\"error\":false,\"aws\":{\"xray\":{\"sdk\":\"opentelemetry for python\",\"sdk_version\":\"1.0.0\",\"auto_instrumentation\":false}},\"metadata\":{\"default\":{\"http.flavor\":\"1.1\",\"http.route\":\"/health\",\"net.host.port\":\"\",\"otel.resource.service.name\":\"fastapi-app\",\"otel.resource.telemetry.sdk.language\":\"python\",\"otel.resource.telemetry.sdk.name\":\"opentelemetry\",\"otel.resource.telemetry.sdk.version\":\"1.0.0\"}}}\n"
  ]

```

PR #400, opened 2021-04-03
Fix DataDog trace exporter

owais:
# Description

Fixed DataDog trace exporter

`ddtrace` package introduced some breaking changes that break the
DataDog trace exporter. This commit downgrades the dependency to last
know working version to fix the exporter.

## Type of change

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/open-telemetry/opentelemetry-python-contrib/compare/v0.19b0...HEAD)
 
 ### Changed
+- Downgrade DataDog exporter's `ddtrace` dependency so the exporter continues to work.
+  ([#400](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/400))
 - GRPC instrumentation now correctly injects trace context into outgoing requests.
   ([#392](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/39))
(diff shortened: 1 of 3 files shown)


ericmustin:
@owais thanks for the ping here. Afaik, nothing wrong with keeping things pinned for now like you’ve done, but I’ll sync with @majorgreys this week on how we can address long term in dd-trace-py. Will followup when I have more.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/open-telemetry/opentelemetry-python-contrib/compare/v0.19b0...HEAD)
 
 ### Changed
+- Downgrade DataDog exporter's `ddtrace` dependency so the exporter continues to work.
```
toumorokoshi, line 10: Would it be more accurate to say that the version range has been restricted? We've never prescribed a specific version.

toumorokoshi (changes requested):
minor nit about the changelog, otherwise LGTM!

I think regardless of how ddtrace chooses to handle it, we should get this merged in so we can keep CI working and users unbroken.

CHANGELOG.md
owais: Yes. Updated.

owais:
@ericmustin IMO direct dependencies like this should always specify a fixed version instead of an open range. My recommendation would be to keep the dependency pinned and have the maintainer of opentelemetry-exporter-datadog update it to newer versions as they see fit.

toumorokoshi (approved):
LGTM, thanks!

codeboten (approved):

Issue #401, opened 2021-04-04
Instrumentations do not use `tracer_provider` argument to create a tracer instances


lonewolf3739:
Users can optionally pass an instance of `TracerProvider` to `SomeInstrumentator.instrument(tracer_provider=...)`. Instrumentations should use this tracer_provider and fall back to global tracer provider if not given. As of now some instrumentations directly use global tracer provider which is not expected because providers can have different configuration such as `Resource`.

lzchen:
Remaining instrumentations: sklearn, pyramid and (perhaps) the logging instrumentation.

PR #402, opened 2021-04-04
Update instrumentations to use tracer_provider for creating tracer if given, otherwise use global tracer provider

lonewolf3739:
# Description
An optional tracer provider can be passed to instrumentations to use for creating tracer instances. There is no consistency in instrumentations.

Contributes to #401


instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -179,7 +180,8 @@ def __init__(self, *args, **kwargs):
         self.wsgi_app = _rewrapped_app(self.wsgi_app)
 
         _before_request = _wrapped_before_request(
-            _InstrumentedFlask.name_callback
+            _InstrumentedFlask.name_callback,
+            tracer_provider=_InstrumentedFlask._tracer_provider,
```
owais, line 184: How about passing in an instance of a tracer to this function instead of a provider. That way it won't have to call get_tracer for every incoming request.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -210,7 +214,9 @@ def instrument_app(
             app._original_wsgi_app = app.wsgi_app
             app.wsgi_app = _rewrapped_app(app.wsgi_app)
 
-            _before_request = _wrapped_before_request(name_callback)
+            _before_request = _wrapped_before_request(
+                name_callback, tracer_provider=tracer_provider
```
owais, line 218: same comment as above

owais (commented):
Thanks. Minor comment. Looks good otherwise.

owais (changes requested):
Didn't realize this was a draft. Will review again once final version is available.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
lonewolf3739: Yeah, I was looking into this. Few instrumentation have this problem and others have followed different ways to avoid creating tracer instance on each call. It should be achievable here as well. I am fine with just passing tracer as long as we keep it consistent across instrumentations. WDYT?

lonewolf3739: Updated this to create one tracer instance when instrumentation is enabled and use it for all.

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
```diff
@@ -87,16 +85,11 @@ class AsyncPGInstrumentor(BaseInstrumentor):
     def __init__(self, capture_parameters=False):
         super().__init__()
         self.capture_parameters = capture_parameters
+        self._tracer = None
 
     def _instrument(self, **kwargs):
-        tracer_provider = kwargs.get(
-            "tracer_provider", trace.get_tracer_provider()
-        )
-        setattr(
```
lzchen, line 95: Any reason why we designed this like this? Storing the `tracer` as an attribute I mean.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
```diff
@@ -107,9 +108,7 @@ def process_request(self, request):
 
         token = attach(extract(request_meta, getter=wsgi_getter))
 
-        tracer = get_tracer(__name__, __version__)
-
-        span = tracer.start_span(
+        span = self._tracer.start_span(
```
lzchen, line 111: `None` check?

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -83,19 +83,26 @@ class FalconInstrumentor(BaseInstrumentor):
 
     def _instrument(self, **kwargs):
         self._original_falcon_api = falcon.API
+        _InstrumentedFalconAPI._tracer_provider = kwargs.get("tracer_provider")
```
lzchen, line 86: Any reason not to store the `tracer` instead of `tracer_provider` here?

instrumentation/opentelemetry-instrumentation-mysql/src/opentelemetry/instrumentation/mysql/__init__.py
```diff
@@ -77,22 +77,24 @@ def _uninstrument(self, **kwargs):
...
+            tracer_provider: The optional TracerProvider to use. If omitted
+            the current globally configured one is used.
 
         Returns:
             An instrumented connection.
         """
-        tracer = get_tracer(__name__, __version__)
-
         return dbapi.instrument_connection(
-            tracer,
```
lzchen, line 92: Was this a bug?

lzchen:
I wonder if we should still allow `tracer_provider` to be passed into the [logging](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/__init__.py#L78) instrumentation. Technically if they pass in a `tracer_provider`, it could be different than the one that belongs to the [span](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/__init__.py#L93) that is being recorded.

lzchen:
I believe [sklearn](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-sklearn/src/opentelemetry/instrumentation/sklearn/__init__.py#L122) is still using the global tracer but maybe that can be for a separate PR :)

tox.ini
```diff
@@ -319,7 +319,7 @@ commands =
 
 [testenv:lint]
 basepython: python3.8
-recreate = False 
+recreate = True
```
lzchen, line 322: What is this for?

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
lonewolf3739: I am not aware if there was any particular reason.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
lonewolf3739: Ideally this wouldn't None because it is set [here](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/402/files#diff-c93e4cdcf380e4064033aacbb347553945bb8aa80866cdf24355f867d0917a90R114) when `.instrument()` is called but let me check if there is chance of user misconfiguring and subsequently running into error.

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
lonewolf3739: I don't see any problem with storing tracer directly. Let me check and comeback.

instrumentation/opentelemetry-instrumentation-mysql/src/opentelemetry/instrumentation/mysql/__init__.py
lonewolf3739: Yes, I don't know when `dbapi.instrument_connection` signature changed but this is not updated here and few other places.

tox.ini
lonewolf3739: I have no idea why docs is failing and this was a try at fixing that issue assuming cache was causing the problem. I will check again.

lonewolf3739:
> I believe [sklearn](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-sklearn/src/opentelemetry/instrumentation/sklearn/__init__.py#L122) is still using the global tracer but maybe that can be for a separate PR :)

Once this gets in I will update the original issue with what are covered and pending. I have zero expertise with sklearn and I hope someone with domain knowledge would show up to improve it.

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
lonewolf3739: There is no special reason except the little lazy initialization of tracer. Either way is fine to me. What do you think?

lzchen: To be consistent, I think storing `tracer` would be nice.

lzchen:
@owais @lonewolf3739 
Thoughts on [this](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/402#issuecomment-817972691)?

lonewolf3739:
> @owais @lonewolf3739
> Thoughts on [this](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/402#issuecomment-817972691)?

I believe the tracer_provider is used for resource attributes. Adding to your point here is the small discussion we had about using the resource from span instead https://github.com/open-telemetry/opentelemetry-python-contrib/pull/345#discussion_r593931930. If there is no other reason to pass tracer provider I think we can get rid of it.

owais:
@lonewolf3739 @lzchen regarding logging instrumentation, it would be ideal if we could update it to extract service name from the span instead of provider but I don't know if that is possible without adding a dependency on the SDK. 

One possible solution could be to implement log injection as a feature of the SDK instead of an instrumentation. Technically it's not really an instrumentation as it doesn't instrument the logging module (doesn't trace it) but provides utility to augment generated logs with additional data. The behavior is actually part of upcoming Otel log spec as well.

tox.ini
lonewolf3739: @codeboten I remember you did something similar to this in core repo to fix the unusual pipeline failures. I can't get the docs to pass and I couldn't understand error. I would really appreciate some help here.

lzchen:
@owais 
Isn't [Resource](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/__init__.py#L80) an [SDK](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/resources/__init__.py#L136) only concept anyways? I feel the logging enriching feature is not an instrumentation, and needs to depend on the SDK.

owais:
> Isn't Resource an SDK only concept anyways? I feel the logging enriching feature is not an instrumentation, and needs to depend on the SDK.

I agree. We can add a dependency on the SDK to begin with and later move the package out of instrumentations into either it's own package or into the SDK.

lzchen (approved):

lzchen:
@owais 
Can you review once more?

owais (approved):

codeboten (approved):

PR #406, opened 2021-04-05
Remove http.status_text from span attributes

lonewolf3739:
# Description

`http.status_text` was removed in this spec pr https://github.com/open-telemetry/opentelemetry-specification/pull/972

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
@@ -171,7 +171,6 @@ def _instrument(tracer_provider=None, span_callback=None, name_callback=None):
 
                 if span.is_recording():
                     span.set_attribute("http.status_code", code_)
-                    span.set_attribute("http.status_text", result.reason)
                     span.set_status(Status(http_status_to_status_code(code_)))
 
                 ver_ = str(getattr(result, "version", ""))
--- a/instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
@@ -203,7 +203,6 @@ def _apply_response(span: Span, response: urllib3.response.HTTPResponse):
         return
 
     span.set_attribute("http.status_code", response.status)
-    span.set_attribute("http.status_text", response.reason)
     span.set_status(Status(http_status_to_status_code(response.status)))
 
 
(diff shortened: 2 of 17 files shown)


instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
```diff
@@ -171,7 +171,6 @@ def _instrumented_open_call(
 
                 if span.is_recording():
                     span.set_attribute("http.status_code", code_)
-                    span.set_attribute("http.status_text", result.reason)
                     span.set_status(Status(http_status_to_status_code(code_)))
```
codeboten, line 174: should the reason be set on the status?

instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
```diff
@@ -203,7 +203,6 @@ def _apply_response(span: Span, response: urllib3.response.HTTPResponse):
         return
 
     span.set_attribute("http.status_code", response.status)
-    span.set_attribute("http.status_text", response.reason)
     span.set_status(Status(http_status_to_status_code(response.status)))
```
codeboten, line 206: same question about setting the reason on the status here

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -157,8 +157,7 @@ def add_response_attributes(
     passed to a PEP3333-conforming start_response callable."""
     if not span.is_recording():
         return
-    status_code, status_text = start_response_status.split(" ", 1)
```
codeboten, line 160: same question about the status_text being set on the status

codeboten (commented):
Thanks for taking on this change, I noticed some of the instrumentation uses the text from status_text in the status description, we should be consistent one way or the other about it.

instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
lonewolf3739: Personally I don't think setting a reason phrase adds any value when there is already status code. Copying the same from [RFC](https://tools.ietf.org/html/rfc7230#section-3.1.2). 
> The reason-phrase element exists for the sole purpose of providing a
   textual description associated with the numeric status code, mostly
   out of deference to earlier Internet application protocols that were
   more frequently used with interactive text clients.  A client SHOULD
   ignore the reason-phrase content.

Some places reason is not really http reason phrase it is just formatted exception. I would suggest removing reason phrase on status, WDYT?

codeboten: Ah i see, for some reason I was thinking that `reason` would be more than the status message for the http response

codeboten (approved):

PR #420, opened 2021-04-07
Update gRPCContext wrapper class

alertedsnake:
# Description

There are a few cases where one needs to dig into `grpc.ServicerContext` objects, and these fields were missing from our wrapper, which can cause issues with implmementation.

Yes, they're private attributes, but there's no way to get the data they hold otherwise, so we should properly map them so our wrapper Context behaves like a real one does.

## Type of change

# How Has This Been Tested?

There don't seem to be any test cases which cover this - I only noticed because I have an interceptor that reports exceptions  to Sentry, and so it has to dig into the `grpc.ServicerContext` object for details, rather than wrapping it like we do here.

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py
+++ b/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py
@@ -71,29 +71,8 @@ class _OpenTelemetryServicerContext(grpc.ServicerContext):
         self.details = None
         super().__init__()
 
-    @property
-    def _state(self):
-        return self._servicer_context._state
-
-    @_state.setter
-    def _state(self, value):
-        self._servicer_context._state = value
-
-    @property
-    def _rpc_event(self):
-        return self._servicer_context._rpc_event
-
-    @_rpc_event.setter
-    def _rpc_event(self, value):
-        self._servicer_context._rpc_event = value
-
-    @property
-    def _request_deserializer(self):
-        return self._servicer_context._request_deserializer
-
-    @_request_deserializer.setter
-    def _request_deserializer(self, value):
-        self._servicer_context._request_deserializer = value
+    def __getattr__(self, attr):
+        return getattr(self._servicer_context, attr)
 
     def is_active(self, *args, **kwargs):
         return self._servicer_context.is_active(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


codeboten (commented):
Code looks good, is it worth adding a test around this? @alertedsnake

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py
```diff
@@ -71,6 +71,30 @@ def __init__(self, servicer_context, active_span):
         self.details = None
         super().__init__()
 
+    @property
```
ocelotl, line 74: Still, we should not use someone else's private attributes. Also, next time we need to access a private attribute we would need to add yet another property.

I feel like this is a job for [`__getattr__`](https://docs.python.org/3/reference/datamodel.html#object.__getattr__).

ocelotl (changes requested):
I agree with the main objective of this PR, @alertedsnake, but please consider reimplementing this with `__getattr__` instead of relying on private attributes.

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py
alertedsnake: I could agree with that, though since this class is a wrapper instead of a subclass, this is always going to be a problem when the wrapped class gets changed and people expect those changes to propagate.  In fact, now that I look again, we're missing a private function as well, `_finalize_state()`.

Not sure who the original author was, but I wonder why the choice was made to wrap this rather than subclass it?  Possibly because the `grpc` implementation is also marked private.

What would your suggestion be, to redefine `__getattr__` so it calls `getattr()` on the wrapped class if the attribute is not defined in our wrapper?

ocelotl: Correct, use the wrapped attribute if the attribute is not found in the wrapper. :+1:

alertedsnake: Oh I know the answer to my question - because the base class is a metadata class, the real class is that private `_Context` in the private `grpc._server` module.

Expect an update later today, and thanks for the tip!

ocelotl: :v:

ocelotl (approved):

codeboten (approved):

PR #436, opened 2021-04-13
Added support for traceresponse headers

owais:
Added opt-in support to return traceresponse headers for server instrumentations.

This allows users to configure their web apps to inject trace context
as headers in HTTP responses. This is useful when client side apps need
to connect their spans with the server side spans e.g, in RUM products.

Today the most practical way to do this is to use the `Server-Timing`
header but in near future we might use the `traceresponse` header as
described here: https://w3c.github.io/trace-context/#trace-context-http-response-headers-format

Added trace response propagation support for:

- Django
- Falcon
- Flask
- Pyramid
- Tornado

# Description

## Type of change

# How Has This Been Tested?

- [x] Added unit tests
- [x] Manual testing

- [x] Yes. - Link to PR: open-telemetry/opentelemetry-python#1762
- [ ] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
@@ -99,6 +99,10 @@ import opentelemetry.instrumentation.wsgi as otel_wsgi
 from opentelemetry import context, trace
 from opentelemetry.instrumentation.falcon.version import __version__
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
+from opentelemetry.instrumentation.propagators import (
+    FuncSetter,
+    get_global_response_propagator,
+)
 from opentelemetry.instrumentation.utils import (
     extract_attributes_from_object,
     http_status_to_status_code,
@@ -119,6 +123,7 @@ _ENVIRON_EXC = "opentelemetry-falcon.exc"
 
 _excluded_urls = get_excluded_urls("FALCON")
 _traced_request_attrs = get_traced_request_attrs("FALCON")
+_response_propagation_setter = FuncSetter(falcon.api.Response.append_header)
 
 
 class FalconInstrumentor(BaseInstrumentor):
@@ -272,6 +277,10 @@ class _TraceMiddleware:
                     description=reason,
                 )
             )
+
+        propagator = get_global_response_propagator()
+        if propagator:
+            propagator.inject(resp, setter=_response_propagation_setter)
 
         if self._response_hook:
             self._response_hook(span, req, resp)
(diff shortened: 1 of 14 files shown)


owais:
I created individual PRs first but they were very repetitive and quite small. I thought it'd be better to have them bundled as a single PR. This way they can be reviewed at once by the same reviewers. It should be more efficient I think.

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -273,5 +279,9 @@ def process_response(
                 )
             )
 
+        propagator = get_global_response_propagator()
+        if propagator:
+            propagator.inject(resp, setter=self.back_propagation_setter)
```
lzchen, line 284: You are able to access a class attribute like this via `self`?

aabmass (approved):
LGTM

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -264,3 +264,11 @@ def _end_span_after_iterating(iterable, span, tracer, token):
             close()
         span.end()
         context.detach(token)
+
+
+class BackPropagationSetter:
```
lzchen, line 269: Inherit from `Setter`

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
owais: Yes but it can lead to bugs later. Will change.

owais:
After sub-classing, docs job is failing with:


```
Warning, treated as error:
/home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py:docstring of opentelemetry.instrumentation.wsgi.ResponsePropagationSetter:1:py:class reference target not found: opentelemetry.instrumentation.propagators.Setter
ERROR: InvocationError for command /home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/.tox/docs/bin/sphinx-build -E -a -W -b html -T . _build/html (exited with code 2)
```

The core repo GIT SHA was updated to tip of core main which contains the change. The file for which docs are being generated is obviously able to import the said class as it passes all tests and lint checks. 

I tried invalidating tox cache but it isn't helping either. Looks like docs job is not installing the latest core main or is not using the locally checked out copy for some reason.

owais:
@lzchen I'm at a loss. I'm reverting inheriting from ABC setter for now and added a TODO to unblock release for now. Will need more time to figure out why docs doesn't like it. I remember Srikanth had a similar issue last week so probably something is wrong with docs tox job that we need to look into anyway.

lzchen (approved):

PR #442, opened 2021-04-14
Keep client interceptors in sync with grpc client interceptors

mihirg:
# Description
Keep the opentelemetry interceptors in sync with grpc interceptors

Fixes # (issue)
https://github.com/open-telemetry/opentelemetry-python-contrib/issues/320
The discussion on this issue states a more serious refactoring is needed. This fix is to get people unblocked.

## Type of change

# How Has This Been Tested?
Yes. Updated test cases

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py
+++ b/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py
@@ -13,9 +13,7 @@
 # limitations under the License.
 
 # pylint:disable=relative-beyond-top-level
-# pylint:disable=arguments-differ
 # pylint:disable=no-member
-# pylint:disable=signature-differs
 
 """Implementation of gRPC Python interceptors."""
 
@@ -48,21 +46,23 @@ class _InterceptorUnaryUnaryMultiCallable(grpc.UnaryUnaryMultiCallable):
         self._base_callable = base_callable
         self._interceptor = interceptor
 
-    def __call__(self, request, timeout=None, metadata=None, credentials=None):
-        def invoker(request, metadata):
-            return self._base_callable(request, timeout, metadata, credentials)
-
-        client_info = _UnaryClientInfo(self._method, timeout)
-        return self._interceptor.intercept_unary(
-            request, metadata, client_info, invoker
-        )
-
-    def with_call(
-        self, request, timeout=None, metadata=None, credentials=None
+    def __call__(
+        self,
+        request,
+        timeout=None,
+        metadata=None,
+        credentials=None,
+        wait_for_ready=None,
+        compression=None,
     ):
         def invoker(request, metadata):
-            return self._base_callable.with_call(
-                request, timeout, metadata, credentials
+            return self._base_callable(
+                request,
+                timeout,
+                metadata,
+                credentials,
+                wait_for_ready,
+                compression,
             )
 
         client_info = _UnaryClientInfo(self._method, timeout)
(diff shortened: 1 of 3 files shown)


mihirg:
Can someone review this PR?

owais (approved):
I wonder if we should use **kwargs instead?

lzchen:
@mihirg

What does the more "serious refactoring" entail? Would it make it so that we would not need these parameters anymore, which in that case, we should go with `kwargs` like @owais  suggested?

ocelotl (changes requested):
Thanks for the fix @mihirg! Please remove these comments from `instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py b/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py`

`# pylint:disable=arguments-differ`
`# pylint:disable=signature-differs`

Removing them should alert us of any further discrepancies between our implementation and the base classes :+1:

mihirg:
> @mihirg
> 
> What does the more "serious refactoring" entail? Would it make it so that we would not need these parameters anymore, which in that case, we should go with `kwargs` like @owais suggested?

Done.

mihirg:
> Thanks for the fix @mihirg! Please remove these comments from `instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py b/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py`
> 
> `# pylint:disable=arguments-differ`
> `# pylint:disable=signature-differs`
> 
> Removing them should alert us of any further discrepancies between our implementation and the base classes 👍

Done.

lzchen (approved):

instrumentation/opentelemetry-instrumentation-grpc/tests/test_client_interceptor.py
```diff
@@ -41,13 +41,55 @@
 from .protobuf.test_server_pb2 import Request
 
 
+class Interceptor(
```
aabmass, line 44: Is this to test that the instrumentation doesn't break existing interceptors? Can we just add a comment explaining what this is for

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py
```diff
@@ -41,41 +39,67 @@ class _StreamClientInfo(
 ):
     pass
 
+def _get_metadata_timeout(**kwargs):
```
aabmass, line 42: This is being called as `_get_metadata_timeout(kwargs)` which would raise `_get_metadata_timeout() takes 0 positional arguments but 1 was given`.
```suggestion
def _get_metadata_timeout(kwargs):
```

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py
```diff
@@ -41,41 +39,67 @@ class _StreamClientInfo(
...
     def __init__(self, method, base_callable, interceptor):
         self._method = method
         self._base_callable = base_callable
         self._interceptor = interceptor
 
-    def __call__(self, request, timeout=None, metadata=None, credentials=None):
+    def __call__(
+        self,
+        request,
+        **kwargs,
```
aabmass, line 57: What do you think of doing something like this instead of `_get_metadata_timeout()` throughout?

```py
def __call__(
    self,
    request,
    *,
    metadata=None,
    timeout=None,
    **kwargs
):
    ...
```

aabmass (changes requested):

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py
ocelotl: Unless it is fine to return `None` as the value for `metadata` and/or `timeout`, then this should not use `**kwargs`. This looks like we should be using positional arguments as before instead of using `**kwargs`.

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py
ocelotl: I don't see any advantage on using `kwargs` here if every call will necessarily require `timeout` and `metadata`.

ocelotl (changes requested):

mihirg:
@ocelotl / @aabmass Incorporated your feedback.

ocelotl (approved):
Thanks @mihirg! Great first contribution :muscle: :sunglasses:

ocelotl:
@mihirg apparently there are a few lint issues. No worries, the best way of getting them fixed is by doing this:
1. Run `tox -e lint` (this should fail)
2. From the root repo directory run `source .tox/bin/lint/activate`
3. Run `black .` (this should fix any `black`-related issues)
4. Run `isort .` (this should fix any `isort`-related issues)
5. Run `deactivate`
6. Run `tox -e lint` again

If it passes, then you are good to commit the fixes and push your branch again. If there are failures, they should be `pylint` failures, please take a look at the errors and either make a change for them or ignore them with `# pylint:disable=<the-error-name-here>`.

aabmass (approved):
Thanks for the contribution!

mihirg:
@ocelotl I have fixed the the linter issues.

PR #458, opened 2021-04-16
implements context propagation for lambda invoke + tests

kubawach:
# Description

Implements context propagation for lambda invoke (botocore) which needs to have the headers as a part of the payload.
Adds tests.
Bumps moto to ~2.0 - previous version had issues preventing lambda tests.

## Type of change

# How Has This Been Tested?

- Automated test with moto

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
@@ -46,6 +46,7 @@ API
 ---
 """
 
+import json
 import logging
 
 from botocore.client import BaseClient
@@ -99,6 +100,27 @@ class BotocoreInstrumentor(BaseInstrumentor):
     def _uninstrument(self, **kwargs):
         unwrap(BaseClient, "_make_api_call")
 
+    @staticmethod
+    def _is_lambda_invoke(service_name, operation_name, api_params):
+        return (
+            service_name == "lambda"
+            and operation_name == "Invoke"
+            and isinstance(api_params, dict)
+            and "Payload" in api_params
+        )
+
+    @staticmethod
+    def _patch_lambda_invoke(api_params):
+        try:
+            payload_str = api_params["Payload"]
+            payload = json.loads(payload_str)
+            headers = payload.get("headers", {})
+            inject(headers)
+            payload["headers"] = headers
+            api_params["Payload"] = json.dumps(payload)
+        except ValueError:
+            pass
+
     # pylint: disable=too-many-branches
     def _patched_api_call(self, original_func, instance, args, kwargs):
         if context_api.get_value("suppress_instrumentation"):
@@ -110,6 +132,12 @@ class BotocoreInstrumentor(BaseInstrumentor):
 
         error = None
         result = None
+
+        # inject trace context into payload headers for lambda Invoke
+        if BotocoreInstrumentor._is_lambda_invoke(
+            service_name, operation_name, api_params
+        ):
+            BotocoreInstrumentor._patch_lambda_invoke(api_params)
 
         with self._tracer.start_as_current_span(
             "{}".format(service_name), kind=SpanKind.CLIENT,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/open-telemetry/opentelemetry-python/compare/v1.2.0-0.21b0...HEAD)
 
+### Added
+- `opentelemetry-instrumentation-botocore` now supports
+  context propagation for lambda invoke via Payload embedded headers. 
+  ([#458](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/458))
+  
 ## [0.21b0](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.2.0-0.21b0) - 2021-05-11
 ### Changed
 
--- a/instrumentation/opentelemetry-instrumentation-boto/setup.cfg
+++ b/instrumentation/opentelemetry-instrumentation-boto/setup.cfg
@@ -47,7 +47,7 @@ install_requires =
 [options.extras_require]
 test =
     boto~=2.0
-    moto~=1.0
+    moto~=2.0
     opentelemetry-test == 0.22.dev0
 
 [options.packages.find]
--- a/instrumentation/opentelemetry-instrumentation-botocore/setup.cfg
+++ b/instrumentation/opentelemetry-instrumentation-botocore/setup.cfg
@@ -45,7 +45,7 @@ install_requires =
 
 [options.extras_require]
 test =
-    moto ~= 1.0
+    moto[all] ~= 2.0
     opentelemetry-test == 0.22.dev0
 
 [options.packages.find]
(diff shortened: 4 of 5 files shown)


kubawach:
For Py38, the instrumentation tests fail with:
```
ImportError while importing test module '/home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_instrumentation.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test_botocore_instrumentation.py:61: in <module>
    class TestBotocoreInstrumentor(TestBase):
test_botocore_instrumentation.py:332: in TestBotocoreInstrumentor
    def test_lambda_client(self):
../../../.tox/py38-test-instrumentation-botocore/lib/python3.8/site-packages/moto/__init__.py:8: in f
    module = importlib.import_module(module_name, "moto")
/opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
../../../.tox/py38-test-instrumentation-botocore/lib/python3.8/site-packages/moto/awslambda/__init__.py:2: in <module>
    from .models import lambda_backends
../../../.tox/py38-test-instrumentation-botocore/lib/python3.8/site-packages/moto/awslambda/models.py:10: in <module>
    import docker
E   ModuleNotFoundError: No module named 'docker'
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!

```

Any hints regarding how to fix that?

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
```diff
@@ -110,6 +111,19 @@ def _patched_api_call(self, original_func, instance, args, kwargs):
         error = None
         result = None
 
+        # inject trace context into payload headers for lambda Invoke
+        if (
+            service_name == "lambda"
+            and operation_name == "Invoke"
+            and "Payload" in api_params
+        ):
```
owais, line 119: *If* there is a way, I think we should be patching the lamba invoke function or another place before this func directly to inject context. deserializing and then re-serializing JSON doesn't feel right.

kubawach: I don't think it's possible - according to boto doc,`Payload (bytes or seekable file-like object) -- The JSON that you want to provide to your Lambda function as input.` So either way we aren't going to get a dict, even at the entry point.
We could avoid (de)serializing JSON by using pure string manipulation (find headers by string, deserialize, modifu, serialize again, inject). This however is more tricky and error-prone (but possibly much faster). @owais I defer the decision to you, I'm too newbie here :)

owais: I was thinking about raw string manipulation as well. In simple cases we probably only need to strip last character '}', concatinate our string and add back the last char but yea it can be error prone. 

I think parsing and dumping again is fine if that's the only way. 

BTW what if we inject context into HTTP request headers instead of the payload. Will lamba be able to read them?

kubawach: As far as I can tell (again, no expert here!) in order to get HTTP header values as part of the event in lambda handler, each customer would need to configure a mapping template for their lambda in API gateway (header val -> payload val). So I think that this would be too much of an overhead. Will go with raw string and see how it goes :)

owais: Makes sense.

kubawach: @owais pushed raw string version. In order to handle corner cases (existing headers prop in payload dict) I had to add quite a few lines. Looks solid but it's definitely more complex than straight json (de)serialization. Please have a look :)

kubawach:
@codeboten @ocelotl @owais please have a look at the PR :)

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
owais: Oh you went all the way with raw string manipulation :) I can't say I really understand it at a quick glance haha :)
I'm not entirely sure if we should use this or the json module. Let's see what others have to say about it.

Curious if you benchmarked it and how faster it was compared to using json module?

kubawach: Frankly speaking I have not. I'll run some microbenchmarks with timeit to compare both implementations.

kubawach:
@owais I finally run some microbenchmarking of the code and results are... interesting. They are heavily dependant on the input provided, as in case of "headers" property included in the payload the algo checks if it's actually a first-level children by counting curly braces.

Test setup:
- 100000 iterations
- payload containing 100 entries, flat structure
- run with headers property near beginning, near end and without the property at all

Results (in ms):
```
JSON-based:  4.218646998
String-based no headers:  0.5552989840000002
String-based with headers:  19.781614840000003
String-based with headers at the beginning:  1.2955868240000008
```

In general without headers property set or with property near the beginning improved (string-based) algorithm is much faster (3,5x - 8x). However for bigger payloads with headers near the end the situation changes - JSON based implementation becomes 5x faster.

I believe that for most cases we should not expect "headers" property to appear in the lambda invoke payload as supplied by the customer. So in the end the decision which way to include should be based on our readiness to support slightly more complex code for a really small performance gain (100000 iterations - 4 ms vs 0,5 ms).

kubawach:
@codeboten @ocelotl please have a look, it's been in review for last 16 days :)

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
kubawach: @owais added some perf microbenchmarking results :)

owais:
> In general without headers property set or with property near the beginning improved (string-based) algorithm is much faster (3,5x - 8x). However for bigger payloads with headers near the end the situation changes - JSON based implementation becomes 5x faster.
> I believe that for most cases we should not expect "headers" property to appear in the lambda invoke payload as supplied by the customer. So in the end the decision which way to include should be based on our readiness to support slightly more complex code for a really small performance gain (100000 iterations - 4 ms vs 0,5 ms).

Given the small gains in only a subset of cases, I think we probably don't want the additional complexity of custom parsing. I'm inclined towards going with stock JSON module but happy to hear from others.

owais (approved):

codeboten (approved):

codeboten:
@kubawach if you could take a look at the lint failure, we can get this PR merged as soon as it passes CI

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
kubawach: Switched back to JSON, as discussed elsewhere.

kubawach:
@codeboten fixed linting and rebased to current main, resolving some conflicts by the way

kubawach:
Seems that all of the conflicts have been resolved, please merge as convenient :)

codeboten:
@kubawach looks like there's still some issues that need resolving w/ lint and tests failing.

kubawach:
@codeboten sorry for premature alarm. Now it really looks as ready to merge (I even took screenshot to be 101% sure ;) )
image

codeboten:
Thanks @kubawach!

codeboten (approved):

PR #465, opened 2021-04-23
Move opentelemetry-instrumentation from core

lzchen:
Part of https://github.com/open-telemetry/opentelemetry-python/pull/1797

### DIFF
--- a/opentelemetry-instrumentation/setup.cfg
+++ b/opentelemetry-instrumentation/setup.cfg
@@ -1,0 +1,56 @@
+# Copyright The OpenTelemetry Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+[metadata]
+name = opentelemetry-instrumentation
+description = Instrumentation Tools & Auto Instrumentation for OpenTelemetry Python
+long_description = file: README.rst
+long_description_content_type = text/x-rst
+author = OpenTelemetry Authors
+author_email = cncf-opentelemetry-contributors@lists.cncf.io
+url = https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/opentelemetry-instrumentation
+platforms = any
+license = Apache-2.0
+classifiers =
+    Development Status :: 4 - Beta
+    Intended Audience :: Developers
+    License :: OSI Approved :: Apache Software License
+    Programming Language :: Python
+    Programming Language :: Python :: 3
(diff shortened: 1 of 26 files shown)


owais (approved):

opentelemetry-instrumentation/setup.cfg
```diff
@@ -0,0 +1,56 @@
...
+classifiers =
+    Development Status :: 4 - Beta
+    Intended Audience :: Developers
+    License :: OSI Approved :: Apache Software License
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
```
lonewolf3739, line 34: py39 is not part of ci. would you mind adding it to env list https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/.github/workflows/test.yml#L13?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+
+    parser.add_argument(
+        "-s",
+        "--service-name",
+        required=False,
+        help="""
+        The service name that should be passed to a trace exporter.
+        """,
+    )
```
lonewolf3739, line 71: Looks like this `--service-name` doesn't do anything right now, Is this deprecated or not fully developed? Either way this can be addressed in another PR but we should decide on removing or implementing the missing part.

lonewolf3739 (approved):

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
lonewolf3739: @owais you might know better about this.

opentelemetry-instrumentation/setup.cfg
codeboten: Added this in a separate PR https://github.com/open-telemetry/opentelemetry-python-contrib/pull/466

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
codeboten: Looks like this was removed here: https://github.com/open-telemetry/opentelemetry-python/commit/c61a7126445f57748af8f7ffd1b7fc0cdadd5476#diff-8e926c8a585e70ab1e5df5d63ed7709fb982976f01bbdee56e238ad74153ef57L86

codeboten (approved):
Thanks for moving this

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
owais: Oh, it was useful. We should probably add it back.

lzchen: @owais 
Probably won't be adding it back due to that PR? ^

PR #468, opened 2021-04-26
Add SQLAlchemy multithreading test

### LINKED ISSUE #383: sqlalchemy: add test to ensure thread safety
A follow up to https://github.com/open-telemetry/opentelemetry-python-contrib/pull/315/#issuecomment-801845366

jomasti:
# Description

I was able to get the test to fail reliably for PostgreSQL by reverting
back the behavior before #315. When going back to the current behavior,
the test passes fine.

Should more be done to try to trigger the others
to also log the warning?

Resolves #383 

## Type of change

# How Has This Been Tested?

- [x] Docker SQLAlchemy test added

- [x] No.

# Checklist:

### DIFF
--- a/tests/opentelemetry-docker-tests/tests/sqlalchemy_tests/mixins.py
+++ b/tests/opentelemetry-docker-tests/tests/sqlalchemy_tests/mixins.py
@@ -208,6 +208,9 @@ class SQLAlchemyTestMixin(TestBase):
         We also expect no logged warnings about calling end() on an ended span.
         """
 
+        if self.VENDOR == "sqlite":
+            return
+
         def insert_player(session):
             _session = session()
             player = Player(name="Player")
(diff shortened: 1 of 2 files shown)


tests/opentelemetry-docker-tests/tests/sqlalchemy_tests/mixins.py
```diff
@@ -199,3 +201,45 @@ def test_parent(self):
...
+        session_factory = sessionmaker(bind=self.engine)
+        # pylint: disable=invalid-name
+        Session = scoped_session(session_factory)
+        thread_one = threading.Thread(target=insert_player, args=(Session,))
+        thread_two = threading.Thread(target=insert_players, args=(Session,))
+
+        logger = logging.getLogger("opentelemetry.sdk.trace")
+        with self.assertLogs(logger, level="WARNING") as cm:
+            # Dummy warning so test doesn't fail on no logs
+            logger.warning("Dummy warning")
```
codeboten, line 235: Can you explain why this assertLogs is needed if you're intentionally creating a warning as part of the test?

codeboten (commented):
Thanks for adding this test! Just one question about the use of logging in the test.

tests/opentelemetry-docker-tests/tests/sqlalchemy_tests/mixins.py
jomasti: Yeah, I wanted to be able to assert no warning logs are generated. There would be a warning about using a closed span before the multithreading fix that was merged a while back. If there are no logs when using `assertLogs`, it fails. Is there a preferred alternative when using `unittest`? Mocking the logger instead?

ocelotl: I would do this:

``` python
from unittest import TestCase


class Test(TestCase):

    def test_case(self):

        with self.assertRaises(AssertionError):
            with self.assertLogs():
                pass
```

tests/opentelemetry-docker-tests/tests/sqlalchemy_tests/test_sqlite.py
```diff
@@ -59,3 +59,7 @@ def test_engine_execute_errors(self):
         self.assertEqual(
             span.status.description, "no such table: a_wrong_table"
         )
+
+    def test_multithreading(self):
+        # Overriding instead of skipping based on DB name in mixin
+        pass
```
ocelotl, line 65: If this will run a test case that always passes it wold be better to remove it if possible. Having test cases that always pass cause unnecessary reporting from pytest.

ocelotl (changes requested):

ocelotl (approved):

codeboten (approved):

PR #474, opened 2021-04-27
Generate instrumentation packages setup.py files

owais:
# Description

All instrumentations packages have almost exactly same setup.py files. This commit adds a python script that generates it from a source template. This dramatically reduces the time and effort required to update all instrumentation setup.py files, and also reduces chances of making manual mistakes.

I'm working on another feature where I needed to update setup.py files for all insrumentations. I'll use this once it is merged.

Another alternative is to take out common code as yet another package and add it as a `setup_requires` dependency so it is installed in target envs before setup.py runs but this adds another runtime dependency for users and doesn't give them any benefits. It's also not heavily used in the eco-system so non standard tools (other than pip) may not work well with it.

## Type of change

- [x] Tooling enhancement

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-wsgi/setup.py
```diff
@@ -11,6 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+
+# DO NOT EDIT. THIS FILE WAS AUTOGENERATED FROM templaes/wsgi.
```
codeboten, line 16: looks like a typo sneaked into the generated code, though it doesnt appear in the generator 🤔

instrumentation/opentelemetry-instrumentation-urllib3/setup.py
```diff
@@ -11,6 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+
+# DO NOT EDIT. THIS FILE WAS AUTOGENERATED FROM templaes/urllib3.
```
codeboten, line 16: should this say it was generated from instrumentation_setup.py?

instrumentation/opentelemetry-instrumentation-wsgi/setup.py
```diff
@@ -11,6 +11,12 @@
...
+# Run `python scripts/generate_setuppy.py` to regenerate.
```
codeboten, line 17: Nit: I see why you named this the way you did, but I would rename the script `generate_setup.py`, it's easier to read.

codeboten (commented):
Would love to see this extended to include the version file at some point as well, though I don't know if the instrumentation versions will always been the same or not.

instrumentation/opentelemetry-instrumentation-wsgi/setup.py
owais: Yea, looks like I updated the generator but didn't run it before checking in.

instrumentation/opentelemetry-instrumentation-urllib3/setup.py
owais: Yes. This is incorrect. Fixing.

codeboten (approved):
Nice, this really does cleanup things and ensures consistency in the instrumentations.

PR #478, opened 2021-04-28
Set status code on ASGI server span

### LINKED ISSUE #427: FastAPI Instrumentor not populating attributes required for Azure exporter
**Describe your environment** 

**Package Version - azure-monitor-opentelemetry-exporter = {version = "^1.0.0-beta.3", allow-prereleases = true}
opentelemetry-instrumentation-fastapi = "^0.19b0"
Operating System - macOS Mojave v10.14.6**
**Python Version - v3.9.0**

**Steps to reproduce**
When using the azure-monitor-opentelemetry-exporter in conjunction with opentelemetry-instrumentation-fastapi 
the response code is not populated by the exporter, raised this issue with Azure monitor team and was redirected here;

> Looking at the output, it seems like the spans are missing some attributes. This is probably an issue in the opentelemetry-instrumentation-asgi, could you file an issue here for this problem specifically? Basically, the exporter sees if http.status_code is populated in the span attributes to populate response_Code in app insights. Since the attribute is not there, this is why you are seeing 0s. Normally, it is up to the instrumentation to populate this.

[Azure monitor issue#17839](https://github.com/Azure/azure-sdk-for-python/issues/17839)

Spans obtained from console exporter: 

- one for successful - 201 request
- one for failed - 422 request
- one on exit

[json.zip](https://github.com/open-telemetry/opentelemetry-python-contrib/files/6281056/json.zip)

Screen grab from Azure App Insights indication reponse code as 0:

image

Compatibility with Azure monitor exporter

**Additional context**
Detailed context for this issue

[Azure monitor issue#17839](https://github.com/Azure/azure-sdk-for-python/issues/17839)

Please let me know in case of additional info required.

Thanks.

jomasti:
# Description

Set the status code on the ASGI server span so that the proper response status code is displayed when exported to something like Azure Monitor

Fixes #427

## Type of change

# How Has This Been Tested?

- [x] Updated the tests to verify the status code is in the server span

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -225,8 +225,10 @@ async def wrapped_send(message):
                         if send_span.is_recording():
                             if message["type"] == "http.response.start":
                                 status_code = message["status"]
+                                set_status_code(span, status_code)
                                 set_status_code(send_span, status_code)
```
owais, line 229: Does `send` here mean sending a response back to the client?
Can the op represented by the outer span have multiple sends or receives?

jomasti: As far as I know, the callable will be called once per connection, which is one request for HTTP. That should mean only one `http.response.start` event. Happy to make a change if that is wrong.

lzchen: I believe multiple calls to `receive` and `send` can occur per HTTP call like from [this issue](https://github.com/Azure/azure-sdk-for-python/issues/18171#issuecomment-827759899). Should the parent span then reflect ANY error that occurs during the flow? If this is the case, we probably need the response code logic in the `receive` as well.

jomasti: Yes, there are multiple send events, but the [second one](https://asgi.readthedocs.io/en/latest/specs/www.html#response-body-send-event) is the response body, and that event does not have another status code.

lzchen: @owais 
As discussed in the SIG meeting on 05/13/2021, we were wondering why there is a parent span for the channel in the first place? Would it be possible to remove this?

owais: I'll do a deep dive into ASGI and get back to this. As far as I understand right now, I think the receive span should be the SERVER span and "send" span should be it's child or grandchild. Both of these spans should use links to establish a relation with the channel span. I need some time to dig into this and confirm if my understanding is correct.

owais (approved):

owais:
LGTM. It makes sense for the status code to be on the server span for HTTP requests. We do however need to re-work this instrumentation when it comes to websockets. Ticket to track/discuss it: #505

lzchen (approved):

codeboten (approved):

PR #480, opened 2021-04-28
Load instrumentors via Distro

owais:
# Description

This feature allows distros to customize instrumentation initialization when using automatically instrumenting with the `opentelemetry-instrument` command. 

## Type of change

# How Has This Been Tested?

- [x] Tested manually
- [x] Added test for newly introduced method 

- [ ] Yes.
- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#472](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/472))
 - Set the `traced_request_attrs` of FalconInstrumentor by an argument correctly.
   ([#473](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/473))
+- Distros can now implement `load_instrumentor(EntryPoint)` method to customize instrumentor
+  loading behaviour.
+  ([#480](https://github.com/open-telemetry/opentelemetry-python/pull/480))
 
 ### Added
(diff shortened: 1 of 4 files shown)


owais:
Ported PR https://github.com/open-telemetry/opentelemetry-python/pull/1751 from core.

owais:
https://github.com/open-telemetry/opentelemetry-python/pull/1751#issuecomment-822636085

> Actually, we have been using entry points to allow for environment variables to point to functions or classes (they can point to any Python object). So, any callback can be specified with an environment variable in this way.

@ocelotl answering your Q ^ from the core PR.

My specific use case is that I want to add additional functionality to some instrumentations with the help of hooks in the splunk-otel-python distribution. For example, I might want to override the logic to compute span names or want to add additional Splunk specific tags to spans. How do I do that today so that users running their apps with `opentelemetry-instrument` command but with splunk's distro installed automatically get Splunk specific behavior in instrumentations? 

This feature will allow the Splunk distro to pass splunk specific kwargs to instrumentations when using the opentelemetry-instrument command. Sure, the distro can inject env vars for every single kwarg it wants to pass to every single instrumentation and we could update all instrumentations to support passing customization options via both kwargs and env vars including for python objects such as callbacks. It is technically doable but it is far more complicated and touches so many more packages, not to mention how ugly it'd make the distro code. This feature gives distros a clean and obvious way to override instrumentation initialization in a single place.

codeboten (approved):
👍

PR #485, opened 2021-04-30
Avoid case sensitivity when handling baggage

### LINKED ISSUE #484: Case sensitivity affects OT propagator baggage handling
[Here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/propagator/opentelemetry-propagator-ot-trace/src/opentelemetry/propagators/ot_trace/__init__.py#L98) is the problem. `startswith` is not case insensitive.

ocelotl:
Fixes #484

# Description

Fixes #484

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#480](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/480))
 - Fix entrypoint for ottrace propagator
   ([#492](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/492))
+- `opentelemetry-propagator-ot-trace` Fix case sensitivity when handling baggage
+  ([#484](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/484))
 
 ### Added
 - Move `opentelemetry-instrumentation` from core repository
--- a/propagator/opentelemetry-propagator-ot-trace/src/opentelemetry/propagators/ot_trace/__init__.py
+++ b/propagator/opentelemetry-propagator-ot-trace/src/opentelemetry/propagators/ot_trace/__init__.py
@@ -95,7 +95,7 @@ class OTTracePropagator(TextMapPropagator):
 
             for key in getter.keys(carrier):
 
-                if not key.startswith(OT_BAGGAGE_PREFIX):
+                if not key.lower().startswith(OT_BAGGAGE_PREFIX.lower()):
                     continue
 
                 baggage[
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#472](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/472))
 - Set the `traced_request_attrs` of FalconInstrumentor by an argument correctly.
   ([#473](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/473))
+- `opentelemetry-propagator-ot-trace` Fix case sensitivity when handling baggage
+  ([#484](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/484))
```
lzchen, line 16:
```suggestion
  ([#485](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/485))
```

lzchen:
Does this mean "oT-BaGgAgE-" is a valid prefix for a header? Can we not be explicit about it?

ocelotl:
> Does this mean "oT-BaGgAgE-" is a valid prefix for a header? Can we not be explicit about it?

Hm, to say it is valid would be a bit too much of a general statement. With this PR I just mean to say that we will not care about casing when checking the prefixes of a baggage dictionary. This is done because some of the frameworks that handle the baggage headers have their own casing conventions and will transform `ot-baggage` into `Ot-Baggage`, for example.

What do you mean with "can we not be explicit about it"?, @lzchen

ocelotl:
> Does this mean "oT-BaGgAgE-" is a valid prefix for a header? Can we not be explicit about it?

Hm, to say it is valid would be a bit too much of a general statement. With this PR I just mean to say that we will not care about casing when checking the prefixes of a baggage dictionary. This is done because some of the frameworks that handle the baggage headers have their own casing conventions and will transform `ot-baggage` into `Ot-Baggage`, for example.

What do you mean with "can we not be explicit about it"?, @lzchen

lzchen:
> This is done because some of the frameworks that handle the baggage headers have their own casing conventions and will transform `ot-baggage` into `Ot-Baggage`,

So this use case makes sense, but with the current implementation `"oT-BaGgAgE-"` would be accepted as well correct? I'm not saying that this is wrong. If we accept that we consider that as a valid header then I don't have a problem with it.

ocelotl:
> > This is done because some of the frameworks that handle the baggage headers have their own casing conventions and will transform `ot-baggage` into `Ot-Baggage`,
> 
> So this use case makes sense, but with the current implementation `"oT-BaGgAgE-"` would be accepted as well correct? I'm not saying that this is wrong. If we accept that we consider that as a valid header then I don't have a problem with it.

Hm, now that I think about this, this fix is not right, marking this as a draft for the time being.

ocelotl:
After some more consideration, I think this is not the right approach, the code as it is right now is fine. Instead of making the handling of character cases in `extract` this is a job for the `getter.keys` method. Closing this PR.

PR #488, opened 2021-05-04
Make propagators conform to spec

mariojonke:
# Description

Propagators use the root context as default value for the extract operation. Additionally avoid modifying the context in case there is nothing to extract or the carrier does not hold valid data.

Fixes https://github.com/open-telemetry/opentelemetry-python/issues/1765

Fixes also an issue in the `ot-trace` propagator where trace/span id validity was comparing str with int values.

## Type of change

# How Has This Been Tested?
unit tests have been added/adapted

- [x] No.

# Checklist:

### DIFF
--- a/exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/propagator.py
+++ b/exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/propagator.py
@@ -42,6 +42,9 @@ class DatadogFormat(TextMapPropagator):
         context: typing.Optional[Context] = None,
         getter: Getter = default_getter,
     ) -> Context:
+        if context is None:
+            context = Context()
+
         trace_id = extract_first_element(
             getter.get(carrier, self.TRACE_ID_KEY)
         )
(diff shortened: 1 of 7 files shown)


lonewolf3739 (commented):
LGTM. Probably it's better we log warning when there is some error extracting span id or trace id or anything.

exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/propagator.py
```diff
@@ -42,6 +42,9 @@ def extract(
         context: typing.Optional[Context] = None,
         getter: Getter = default_getter,
     ) -> Context:
+        if context is None:
+            context = Context()
```
ocelotl, line 46: In this PR description you mention this, @mariojonke:

_Propagators should return the passed in context (or the current one if none was given) if nothing could be extracted._

Should we use the current context instead of instantiating a new one?

mariojonke: We decided in the SIG meeting to use the root context as default because the current context would continue a possible active trace instead of starting a new one if extracting from the carrier fails  (see also https://github.com/open-telemetry/opentelemetry-python/issues/1765#issuecomment-827464045)

See also PR in the core repo: https://github.com/open-telemetry/opentelemetry-python/pull/1811

ocelotl: Ok, good :+1:

ocelotl (approved):

lonewolf3739 (approved):

codeboten (approved):
👍

PR #502, opened 2021-05-21
Add create context key to contrib

eunice98k:
# Description

This PR makes the corresponding changes from otel-python [open-telemetry/opentelemetry-python#1372](https://github.com/open-telemetry/opentelemetry-python/pull/1853) in the contrib to implement `create_key` functionality as noted [here](https://github.com/open-telemetry/opentelemetry-python/issues/1737).

# How Has This Been Tested?

`tox` was ran to ensure the new and existing unit tests all pass.


CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased](https://github.com/open-telemetry/opentelemetry-python/compare/v1.2.0-0.21b0...HEAD)
-
+- Added support for CreateKey functionality.
```
lzchen, line 8: Nit: Please include this in the 'Added' section.

exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/spanprocessor.py
```diff
@@ -17,13 +17,20 @@
...
+    detach,
+    set_value,
+)
 from opentelemetry.sdk.trace import Span, SpanProcessor
 from opentelemetry.sdk.trace.export import SpanExporter
 from opentelemetry.trace import INVALID_TRACE_ID
 from opentelemetry.util._time import _time_ns
 
 logger = logging.getLogger(__name__)
+EXPORT_KEY = create_key("suppress_instrumentation")
```
lzchen, line 33:
```suggestion
_EXPORT_KEY = create_key("suppress_instrumentation")
```

lzchen (approved):

codeboten (changes requested):
Just wanted to confirm my understanding here, the `suppress_instrumentation` context key was being utilized previously by the instrumentations to check if somewhere the SDK had set it. But from the looks here, this change updates the code to create separate keys for each instrumentation.

I believe this mean that this change breaks the functionality we currently have where the span process can set the `suppress_instrumentation` key before export to prevent the exporter from generating spans. https://github.com/open-telemetry/opentelemetry-python/blob/77c9867dcb0f3179c1b395729aae6aec2c3133a0/opentelemetry-sdk/src/opentelemetry/sdk/trace/export/__init__.py#L89 

Can you confirm? If so, we should figure out the right way to move forward to not break this functionality.

eunice98k:
> Just wanted to confirm my understanding here, the `suppress_instrumentation` context key was being utilized previously by the instrumentations to check if somewhere the SDK had set it. But from the looks here, this change updates the code to create separate keys for each instrumentation.
> 
> I believe this mean that this change breaks the functionality we currently have where the span process can set the `suppress_instrumentation` key before export to prevent the exporter from generating spans. https://github.com/open-telemetry/opentelemetry-python/blob/77c9867dcb0f3179c1b395729aae6aec2c3133a0/opentelemetry-sdk/src/opentelemetry/sdk/trace/export/__init__.py#L89
> 
> Can you confirm? If so, we should figure out the right way to move forward to not break this functionality.

Sorry I didn't know the full context behind the `suppress_instrumentation` context key before making the changes. Yes your understanding is correct in that a new key is created for each instrumentation. This was because now with the `create_key` functionality creates a unique key name, which is now not simply `suppress_instrumentation`. 

What I can do is since in the otel-python PR, I am creating a unique `suppress_instrumentation` key, I could simply export this key to each instrumentation that uses it instead of calling `create_key` each time. (I will also rename this from EXPORT_KEY to _SUPPRESS_INSTRUMENTATION_KEY.)
https://github.com/open-telemetry/opentelemetry-python/pull/1853/files#diff-ccdfba02d42525f2fee85e7369c94b7c4976ec22079248bbddf610cd11435739R41

ocelotl:
> > Just wanted to confirm my understanding here, the `suppress_instrumentation` context key was being utilized previously by the instrumentations to check if somewhere the SDK had set it. But from the looks here, this change updates the code to create separate keys for each instrumentation.
> > I believe this mean that this change breaks the functionality we currently have where the span process can set the `suppress_instrumentation` key before export to prevent the exporter from generating spans. https://github.com/open-telemetry/opentelemetry-python/blob/77c9867dcb0f3179c1b395729aae6aec2c3133a0/opentelemetry-sdk/src/opentelemetry/sdk/trace/export/__init__.py#L89
> > Can you confirm? If so, we should figure out the right way to move forward to not break this functionality.
> 
> Sorry I didn't know the full context behind the `suppress_instrumentation` context key before making the changes. Yes your understanding is correct in that a new key is created for each instrumentation. This was because now with the `create_key` functionality creates a unique key name, which is now not simply `suppress_instrumentation`.
> 
> What I can do is since in the otel-python PR, I am creating a unique `suppress_instrumentation` key, I could export this key to each instrumentation that uses it instead of calling `create_key` each time. (I will also rename this from EXPORT_KEY to _SUPPRESS_INSTRUMENTATION_KEY.)
> https://github.com/open-telemetry/opentelemetry-python/pull/1853/files#diff-ccdfba02d42525f2fee85e7369c94b7c4976ec22079248bbddf610cd11435739R41
> 
> Let me know if you think this is alright!

Yes, this makes sense :+1:

codeboten:
> > Just wanted to confirm my understanding here, the `suppress_instrumentation` context key was being utilized previously by the instrumentations to check if somewhere the SDK had set it. But from the looks here, this change updates the code to create separate keys for each instrumentation.
> > I believe this mean that this change breaks the functionality we currently have where the span process can set the `suppress_instrumentation` key before export to prevent the exporter from generating spans. https://github.com/open-telemetry/opentelemetry-python/blob/77c9867dcb0f3179c1b395729aae6aec2c3133a0/opentelemetry-sdk/src/opentelemetry/sdk/trace/export/__init__.py#L89
> > Can you confirm? If so, we should figure out the right way to move forward to not break this functionality.
> 
> Sorry I didn't know the full context behind the `suppress_instrumentation` before making the changes. Yes your understanding is correct in that a new key is created for each instrumentation. This was because now with the `create_key` functionality creates a unique key name, which is now not simply `suppress_instrumentation`.
> 
> What I can do is since in the otel-python PR, I am creating a unique `suppress_instrumentation` key, I could export this key to each instrumentation that uses it instead of calling `create_key` each time. (I will also rename this from EXPORT_KEY to _SUPPRESS_INSTRUMENTATION_KEY.)
> https://github.com/open-telemetry/opentelemetry-python/pull/1853/files#diff-ccdfba02d42525f2fee85e7369c94b7c4976ec22079248bbddf610cd11435739R41
> 
> Let me know if you think this is alright!

The problem with exporting the key from the SDK is that the instrumentations will then depend on the SDK, which I don't believe they do today. This should likely be part of the API instead, this was discussed in the specification here: https://github.com/open-telemetry/opentelemetry-specification/issues/530

I'll add this to the agenda to discuss it at Thursday's SIG meeting.

ocelotl:
> > > Just wanted to confirm my understanding here, the `suppress_instrumentation` context key was being utilized previously by the instrumentations to check if somewhere the SDK had set it. But from the looks here, this change updates the code to create separate keys for each instrumentation.
> > > I believe this mean that this change breaks the functionality we currently have where the span process can set the `suppress_instrumentation` key before export to prevent the exporter from generating spans. https://github.com/open-telemetry/opentelemetry-python/blob/77c9867dcb0f3179c1b395729aae6aec2c3133a0/opentelemetry-sdk/src/opentelemetry/sdk/trace/export/__init__.py#L89
> > > Can you confirm? If so, we should figure out the right way to move forward to not break this functionality.
> > 
> > 
> > Sorry I didn't know the full context behind the `suppress_instrumentation` before making the changes. Yes your understanding is correct in that a new key is created for each instrumentation. This was because now with the `create_key` functionality creates a unique key name, which is now not simply `suppress_instrumentation`.
> > What I can do is since in the otel-python PR, I am creating a unique `suppress_instrumentation` key, I could export this key to each instrumentation that uses it instead of calling `create_key` each time. (I will also rename this from EXPORT_KEY to _SUPPRESS_INSTRUMENTATION_KEY.)
> > https://github.com/open-telemetry/opentelemetry-python/pull/1853/files#diff-ccdfba02d42525f2fee85e7369c94b7c4976ec22079248bbddf610cd11435739R41
> > Let me know if you think this is alright!
> 
> The problem with exporting the key from the SDK is that the instrumentations will then depend on the SDK, which I don't believe they do today. This should likely be part of the API instead, this was discussed in the specification here: [open-telemetry/opentelemetry-specification#530](https://github.com/open-telemetry/opentelemetry-specification/issues/530)
> 
> I'll add this to the agenda to discuss it at Thursday's SIG meeting.

Hm, yeah, good point, @codeboten. I think [this](https://github.com/open-telemetry/opentelemetry-specification/pull/1653) is the PR that added the suppress instrumentation key to the API.

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -45,15 +45,20 @@
...
+)
 from opentelemetry.propagate import inject
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.trace import SpanKind, get_tracer
 from opentelemetry.trace.status import Status
 
 # A key to a context variable to avoid creating duplicate spans when instrumenting
 # both, Session.request and Session.send, since Session.request calls into Session.send
-_SUPPRESS_HTTP_INSTRUMENTATION_KEY = "suppress_http_instrumentation"
+_SUPPRESS_HTTP_INSTRUMENTATION_KEY = context.create_key(
```
lzchen, line 59: Remove this?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -16,8 +16,11 @@
 
 from wrapt import ObjectProxy
 
+from opentelemetry.context import create_key
 from opentelemetry.trace import StatusCode
 
+_SUPPRESS_INSTRUMENTATION_KEY = create_key("suppress_instrumentation")
```
lzchen, line 22: Can we leave a comment saying that this is a temporary solution, and we are waiting on a decision to how we are handling supressing instrumentations?

lzchen (approved):

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -16,8 +16,14 @@
 
 from wrapt import ObjectProxy
 
+from opentelemetry.context import create_key
 from opentelemetry.trace import StatusCode
 
+# This is a temporary location for the suppress instrumentation key.
```
ocelotl, line 22:
```suggestion
# FIXME This is a temporary location for the suppress instrumentation key.
```

ocelotl (approved):
Just please add the `FIXME`, it makes it easier to find stuff that is to be fixed later.

PR #510, opened 2021-05-25
[Chore] Change main after opentelemetry-instrumentation-tornado hotfix

lzchen:


.github/workflows/contributing-message.yml
```diff
@@ -0,0 +1,14 @@
...
+name: contributing-message
+
+jobs:
+  contributing-message:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/first-interaction@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        pr-message: 'Message that will be displayed on users' first pr. Look, a `code block` for markdown.'
```
codeboten, line 14: this change seems like it shouldn't be here

ocelotl (changes requested):
Same comment as @codeboten

.github/workflows/contributing-message.yml
lzchen: Oh my mistake. Removed.

ocelotl (approved):

codeboten (approved):

PR #514, opened 2021-05-26
Generate instrumentation list

owais:
# Description

https://user-images.githubusercontent.com/46186/120125067-c5c23a00-c1d4-11eb-8476-f7d33da114ed.mp4

## Details

- We now automatically generate bootstrap_gen.py file from the list of instrumentations present in the source tree.
- Bootstrap command now uses consumes this auto-generated list instead of keeping it's own local copy.
- We no longer uninstall packages before installing them as instrumentation package no longer specify libraries as dependencies so the edge cases are no longer there.
- We no longer try to install an incompatible version or force upgrade/downgrade an installed version. This used to leave systems in broken states which should happen no more.
- Updated `scripts/prepare_release.py` to generate instrumentation setup.py files and instrumentation list for the bootstrap command.

## Type of change

# How Has This Been Tested?

- [ ] Updated tests

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


owais:
This will shrink after #475 is merged and this is rebased on main.

lonewolf3739 (approved):

instrumentation/opentelemetry-instrumentation-aiohttp-client/setup.py
```diff
@@ -66,6 +68,32 @@
 
 extras_require["test"] = test_deps
 
+
+class JSONMetadataCommand(distutils.cmd.Command):
```
codeboten, line 72: Curious if you'd considered putting this class in the instrumentation package for convenience? Wondering if this is something others who want to create instrumentations that are not necessarily in this repo would be able to use.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py
```diff
@@ -15,118 +15,19 @@
...
-# A mapping of "target library" to "desired instrumentor path/versioned package
-# name". Used as part of the `opentelemetry-bootstrap` command which looks at
-# libraries used by the application that is to be instrumented, and handles
-# automatically installing the appropriate instrumentations for that app.
-# This helps for those who prefer to turn on as much instrumentation as
-# possible, and don't want to go through the manual process of combing through
-# the libraries their application uses to figure which one can be
-# instrumented.
-# NOTE: system-metrics is not to be included.
-def all_instrumentations():
```
codeboten, line 37: Seeing this code being removed makes me **very** happy 🥳

CONTRIBUTING.md
```diff
@@ -170,7 +170,7 @@ Below is a checklist of things to be mindful of when implementing a new instrume
 - Extends from [BaseInstrumentor](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/instrumentor.py#L26)
 - Supports auto-instrumentation
   - Add an entry point (ex. https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-requests/setup.cfg#L56)
-  - Add instrumentation package to `bootstrap.py` (https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py#L37)
+  - Run `python scripts/setup.py` followed by `python scripts/generate_instrumentation_bootstrap.py` after adding a new instrumentation package.
```
codeboten, line 173: It would be great to add a github action that checks that these scripts have been run.

codeboten (commented):
Overall this looks great, a much needed improvement. Please take a look at the comments.

codeboten:
Also, kudos on using the video to explain the PR, it made the review much easier!

instrumentation/opentelemetry-instrumentation-aiohttp-client/setup.py
owais: That would add the instrumentation package as a build/setup dependency to all instrumentations. It is a regular dependency anyway so adding a dep is not a concern but I've never used setup dependencies so not sure how well it works or if it has any downsides. 

Instrumentations outside this repo don't really need this as bootstrap command will probably never know about them.

owais: Either way, having it in setup.py keeps the doors open if we want to move it to the opentelemetry-instrumentation package later.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py
owais: I know!!

CONTRIBUTING.md
owais: Added. Example failure https://github.com/open-telemetry/opentelemetry-python-contrib/runs/2712912080?check_suite_focus=true

tox.ini
```diff
@@ -156,11 +156,13 @@ envlist =
...
 
+    generate
+
 [testenv]
 deps =
   -c dev-requirements.txt
-  test: pytest
-  test: pytest-benchmark
+  pytest
+  pytest-benchmark
```
owais, line 165: With fresh tox envs, running `tox -e propagator` or `tox -e sdkextension` results in tox running the following command:

```
sdkextension installdeps: -cdev-requirements.txt
```
which fails as pip doesn't receive any real requirements to install. 

Running `tox -e test-propagator` fixes the issue as tox matches the `test` dep with the env. This means we'd always have to run such envs with the test- prefix which can be annoying and confuse contributors.

The above change installs pytest and pytest-benchmark by default  in all envs. All envs need pytest anyway so there is no harm in doing this. Plus this always adds at least one dependency to install which works with `pip -c dev-requirements.txt`.

Happy to hear other ideas to address this issue.

instrumentation/opentelemetry-instrumentation-aiohttp-client/setup.py
codeboten: Fair enough.

gen-requirements.txt
```diff
@@ -0,0 +1,6 @@
+-c dev-requirements.txt
```
owais, line 1: Added a separate file for two reasons:

1. tox -e generate only needs these dependencies. Installing everything else from dev-requirements.txt is very unnecessary.
2. astor pulls in deps that fails on pypy so if we add astor to dev-requirements, pypy test jobs start failing.

codeboten (approved):
Thanks for addressing my comments!

ocelotl (commented):
Nice video @owais :sunglasses:

PR #515, opened 2021-05-26
update base version of python to 3.9

codeboten:
Noticed the tox target was still on python3.8, updating to 3.9


lzchen:
[Here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/.readthedocs.yml#L12) too?

lzchen (approved):

owais (approved):

ocelotl:
I have updated this PR to use the commit added by #475. It should pass lint now, but it is not. I think this may be because of Github using a cached virtual environment.

PR #533, opened 2021-06-05
Migrate Django middleware to new-style

adamantike:
# Description

New-style middlewares were [introduced][django_1_10_changelog] in Django `1.10`, and also `settings.MIDDLEWARE_CLASSES` was [removed][django_2_0_changelog] in Django 2.0.

This change migrates the Django middleware to conform with the new style, from Django `2.0` onwards. This is useful because it will help us solve the pending issue in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/391.

By having a single entrypoint to the middleware, `__call__`, which is [wrapped with `sync_to_async` just once][call_wrapped] for async requests, we avoid the [issue][asgiref_issue] where a `ContextVar` cannot be reset from a different context.

With the current deprecated `MiddlewareMixin` way, both `process_request` and `process_response` were being [wrapped separately with `sync_to_async`][mixin_wrapped], which was the source of the mentioned issue.

It's worth mentioning that this change **maintains compatibility with legacy Django versions**.

[django_1_10_changelog]: https://docs.djangoproject.com/en/3.2/releases/1.10/#new-style-middleware
[django_2_0_changelog]: https://docs.djangoproject.com/en/3.2/releases/2.0/#features-removed-in-2-0
[call_wrapped]: https://github.com/django/django/blob/213850b4b9641bdcb714172999725ec9aa9c9e84/django/core/handlers/base.py#L54-L57
[mixin_wrapped]: https://github.com/django/django/blob/213850b4b9641bdcb714172999725ec9aa9c9e84/django/utils/deprecation.py#L137-L147
[asgiref_issue]: https://github.com/django/asgiref/issues/267

## Type of change

# How Has This Been Tested?

- [x] `tox -e test-instrumentation-django`

- [x] No.

# Checklist:


owais:
I don't think this is a good enough reason to drop support for older versions even if they aren't supported by their authors anymore. This is 9 versions we'll be dropping support for here. The upside doesn't seem to be worth it. 

I think we have two choices here:

1. Implement both old and new style middlewares. Use new style for 1.10+ and old one for <1.10
2. Wait for this [proposal](https://github.com/open-telemetry/opentelemetry-python/discussions/1754) to be implemented and then remove support for <1.10 in a future version. It'll allow users to use older instrumentation version with newer API/SDK packages. This means that opentelemetry-instrumentation will have to 1.0 as well though.

If we really want to use new style middlewares, I'd recommend to go with option 1. today.

adamantike:
> This is 9 versions we'll be dropping support for here.

I agree with trying to maintain support for legacy library versions as much as possible. However, there are a few more considerations to bring to the table:

* Django `1.5` was the first one to [support Python 3](https://docs.djangoproject.com/en/3.2/releases/1.5/#overview), so we can agree that Django versions up to `1.4` included won't be supported, as https://github.com/open-telemetry/opentelemetry-python requires Python 3.6+.
* Not a hard blocker, but Django officially supports Python 3.6 only [since Django `1.11`](https://docs.djangoproject.com/en/1.11/faq/install/#what-python-version-can-i-use-with-django). It could work with previous Django versions, but I doubt it has been extensively tested.

That being said, I've updated the PR to fallback to inheriting from deprecated `MiddlewareMixin` for older Django versions. I've found other unrelated issues that won't let those Django versions to correctly instrument OpenTelemetry, but I plan to tackle those in separate PRs.

adamantike:
@owais, please let me know what do you think about this approach, when you get the time to review it.

adamantike:
Gentle ping, to be able to move forward with #391

adamantike:
@owais, can you please review this PR again?

owais (approved):

adamantike:
Updated to fix changelog's conflict.

codeboten (commented):
@owais is this another case for having different packages for different supported versions as mentioned here https://github.com/open-telemetry/opentelemetry-python-contrib/issues/550?

owais:
@codeboten I think a single package makes more sense. I'll share my thoughts and preferences in the issue you linked. For this PR, I think the use case is too simple to warrant a new package. I think we should not block this PR until #550  is resolved.

adamantike:
I have resolved merge conflicts again. Is there anything else I can do to move forward with this PR?

codeboten (approved):
Thanks for getting this PR in 👍!

PR #536, opened 2021-06-07
fix asynchonous unary call traces

sengjea:
# Description

When calling the GRPC client with `future()`, the client fails as the context object is re-entered during the asynchonous callback. See `test_unary_unary_future`. This PR fixes that by using lower level function calls to control the re-entry of context during the synchronous and asynchonous bits.

## Type of change
Bug fix

# How Has This Been Tested?

-  `tox -e test-instrumentation-grpc test_client_interceptor.py::TestClientProto::test_unary_unary_future`

No.

# Checklist:

Followed the style guidelines of this project
Unit tests have been added


instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py
```diff
@@ -115,41 +94,46 @@ def _trace_result(self, guarded_span, rpc_info, result):
...
+                        exc.code().value[0],
+                    )
+                span.set_status(
+                    Status(
+                        status_code=StatusCode.ERROR,
+                        description="{}: {}".format(type(exc).__name__, exc),
+                    )
+                )
+                span.record_exception(exc)
+                span.end()
```
aabmass, line 134: Is there any way we could move these `span.end()` into a finally block or set the `end_on_exit=True` flag to use_span? I'm just worried about accidentally leaking non ended spans if this exception handler raises.

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py
```diff
@@ -162,7 +146,8 @@ def _intercept_server_stream(
         else:
             mutable_metadata = OrderedDict(metadata)
 
-        with self._start_span(client_info.full_method) as span:
+        span = self._start_span(client_info.full_method)
+        with trace.use_span(span, end_on_exit=True):
```
aabmass, line 150: This is equivalent to the previous code I think

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py
sengjea: I've moved this into a finally block. I don't think we can set `end_on_exit=True` here as the span should only end after the call is completed and the asynchonous callbacks invoked.

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py
sengjea: I've reverted `_start_span` to call `start_as_current_span` rather than `start_span`

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py
```diff
@@ -115,41 +97,51 @@ def _trace_result(self, guarded_span, rpc_info, result):
...
+                span.set_status(
+                    Status(
+                        status_code=StatusCode.ERROR,
+                        description="{}: {}".format(type(exc).__name__, exc),
+                    )
+                )
+                span.record_exception(exc)
+                raise exc
+            finally:
+                if not result:
```
aabmass, line 142: What is this check for, not sure if I understand?

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py
```diff
@@ -115,41 +97,51 @@ def _trace_result(self, guarded_span, rpc_info, result):
...
+                    span.end()
+        return self._trace_result(span, rpc_info, result)
```
aabmass, line 144: Not 100% sure just from looking at this, but L143 and L144 should never both run is that right?

aabmass (approved):
Nice LGTM overall, but would be great if someone else can also take a look.

`intercept_unary()` and `intercept_stream()` seem to duplicate a lot of code, would it be reasonable to refactor them to share it?

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py
sengjea: Yeah the thing we're kind of missing here is some kind of`end_on_exception=True` for `use_span`... I had to do this for another similar code and made a wrapper for this:

```
@contextmanager
def use_span(span, end_on_exception=False, **kwargs):
    try:
        with trace.use_span(span, **kwargs):
            yield span
    except: # noqa
        if end_on_exception:
            span.end()
        raise
```

sengjea:
> Nice LGTM overall, but would be great if someone else can also take a look.
> 
> `intercept_unary()` and `intercept_stream()` seem to duplicate a lot of code, would it be reasonable to refactor them to share it?

ok i'll do that

aabmass:
> ok i've refactored some of the code but I think it's getting into can-o-worms territory. There's stuff like `RpcInfo` which seems like an overhang from the opentracing era where span decorator functions could be passed into the instrumentor for additional span processing. Will do a bigger refactor in another PR.

Nice thank you for the contribution!

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py
codeboten: Has this comment been addressed?

sengjea: I think so, it's the same as the one below I thought?
The check is to ensure that L143 and L144 never both run.

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py
```diff
@@ -94,17 +73,17 @@ def _start_span(self, method):
...
-            name=method, kind=trace.SpanKind.CLIENT, attributes=attributes
+        return self._tracer.start_span(
+            name=method, kind=trace.SpanKind.CLIENT, attributes=attributes,
         )
 
     # pylint:disable=no-self-use
-    def _trace_result(self, guarded_span, rpc_info, result):
+    def _trace_result(self, span, rpc_info, result):
         # If the RPC is called asynchronously, release the guard and add a
         # callback so that the span can be finished once the future is done.
```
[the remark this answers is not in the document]
sengjea: It's now been updated to reflect the changes

sengjea:
the `docs` check is failing because it was unable to reach some endpoint.

ocelotl (approved):

lonewolf3739 (approved):

codeboten (approved):

Issue #367, opened 2021-03-09
HTTP Span Attributes: `http.url` must not contain username / password


pellared:
As is stated in the recent [specification change](https://github.com/open-telemetry/opentelemetry-specification/pull/1502) :

> `http.url` MUST NOT contain credentials passed via URL in form of `https://username:password@www.example.com/`. In such case the attribute's value should be `https://www.example.com/`

ryokather:
Hi @codeboten is this issue still open? I'd love to take on my first issue and be assigned to it. I noticed that its been fixed in other language implementations but not here yet. 

cc: @alolita

ryokather:
Just wanted to follow up @lzchen or @owais that it's ok to take this on!

codeboten:
Yup, still open, all yours.

marcorichetta:
This should be closed, right? @codeboten

lzchen:
@marcorichetta 
I believe `httpx` is the only instrumentation that still needs implementing.

lzchen:
@tammy-baylis-swi 
Interesting. If that's the case then simply some tests should be added to confirm that it does indeed solve this problem.

alexmojaki:
Should this be closed based on https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2020?

ocelotl:
I asked Claude to review this issue and Claude suggests it be closed with this explanation:

> httpx was the last holdout; credential redaction added in merged PR #2020.
>
> Proof in `main`: [`util-http: redact_url#L329-L333`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/94e851dfe333d47fdb52edb82635330668af284e/util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py#L329-L333)

To guard against a mistaken suggestion, Claude then independently re-verified this conclusion:

> Verified: `util-http/__init__.py` provides `remove_url_credentials`/`redact_url`, so credentials are stripped from `http.url`.

Claude evaluated this issue with high confidence that it can be closed.

Claude has classified this issue as already resolved (the requested fix or feature has already shipped).

The intention of this comment is to make it easier for a human being to review this issue before making a decision.

PR #538, opened 2021-06-07
Ensure clean http url

ryokather:
# Description

This PR fixes issue [#367](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/367) in which a new specification was established such that http.url, a span attribute, must not contain username or password information for security purposes. This would ensure that URLs passed in the form of `https://username:password@www.example.com/` would result in a http span attribute value of `https://www.example.com/`.

This issue was fixed by implementing an additional util function in `util/open-telemetry-http` called `remove_url_credentials(url: str) -> str`. This function utilizes the `urllib.parse` module to cleanly strip a username and password from a given url if present. This fix was applied to the following instrumentations: aiohttp-client, asgi, requests, tornado, urllib, and wsgi. Note that the current implementation of the urllib3 instrumentation would not process the username and password of a url so this logic was not needed.

## Type of change

# How Has This Been Tested?

An additional test called `test_credential_removal` was provided in every instrumentation that sets the `HTTP_URL` attribute. This includes the instrumentations aiohttp-client, asgi, requests, tornado, urllib, urllib3, and wsgi. Tests were run using tox.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


opentelemetry-instrumentation/setup.cfg
```diff
@@ -43,6 +43,7 @@ include_package_data = True
 install_requires =
     opentelemetry-api == 1.4.0.dev0
     wrapt >= 1.0.0, < 2.0.0
+    yarl ~= 1.6
```
ocelotl, line 46: I think this dependency can be avoided by using `urllib.parse`:

``` python
In [57]: from urllib.parse import urlparse, urlunparse

In [58]: parsed_url = urlparse("http://username:password@httpbin.com")

In [59]: urlunparse((parsed_url.scheme, ":".join((parsed_url.hostname, str(parsed_url.port or ""))), parsed_url.path, parsed_url.params, parsed_url.query, parsed_url.fragment))
Out[59]: 'http://httpbin.com:'
```

ocelotl (changes requested):

opentelemetry-instrumentation/setup.cfg
ryokather: Figured that the yarl implementation is more concise and is utilized by aiohttp but it makes sense to limit dependencies. Just updated with `urllib.parse`.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -60,3 +61,33 @@ def unwrap(obj, attr: str):
     func = getattr(obj, attr, None)
     if func and isinstance(func, ObjectProxy) and hasattr(func, "__wrapped__"):
         setattr(obj, attr, func.__wrapped__)
+
+
+def remove_url_credentials(url: str) -> str:
+    """Given a string url, remove the username and password only if it is a valid url"""
+
+    def validate_url(url):
```
ocelotl, line 69: Nit: Single-use function here.

ocelotl (approved):

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -60,3 +61,33 @@ def unwrap(obj, attr: str):
     func = getattr(obj, attr, None)
     if func and isinstance(func, ObjectProxy) and hasattr(func, "__wrapped__"):
         setattr(obj, attr, func.__wrapped__)
+
```
lzchen, line 64: Would this logic be better off [here](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/util/opentelemetry-util-http)?

ryokather: Only thought I had was that it is most similar to the other utility function `http_status_to_status_code` in the same file and the library you linked is only used by web server instrumentations. However, it might make more logical sense to separate it from the `opentelemetry-instrumentation` itself. What would you suggest? @lzchen

lzchen: I believe the library would be used for all http related instrumentations, it just so happens the logic contained is only used in server instrumentations.

ryokather: Sounds good 👍 Just updated with the last remaining suggestions.

lzchen (approved):

lzchen:
@ryokather 
Please rebase and then we can merge this!

ryokather:
Should be done!

Issue #439, opened 2021-04-13
Consistent mechanism to avoid double instrumenting


codeboten:
Followup to https://github.com/open-telemetry/opentelemetry-python/issues/895, the instrumentations currently use many different ways to prevent double instrumentations. It would be good to have a consistent mechanism instead, which could even remove the need to check if a library is instrumented in each of the individual instrumentation libraries.

This could be done instead in the base instrumentor

PR #549, opened 2021-06-23
Consistent way of not instrumenting multiple times

lzchen:
Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/439
We use `_is_instrumented_by_opentelemetry` for all functionality involving indicating whether a component has been instrumented. This is the text that should be used for any context variable/property that is used revolving around determining whether an instrumentation is already instrumented or not.

Another thing to note is the relationship between `instrument()` and `instrument_app`/`instrument_connection()` etc.
`instrument_X()` should take priority over `instrument()`, i.e. if `uninstrument_app(app)` is called after `instrument()`, `app` should produce no telemetry.

As well, calling `instrument_app(app)` after `instrument()` should not produce double telemetry.

Also a bunch of other consistency-based changes:

1. Made all `instrument/uninstrument_X` static methods, where `X` is a component
2. fastapi - added `uninstrument_app`
3. Changed `warning` to `error` on dependency conflict for `instrument()`


codeboten (commented):
> We use the string `_is_instrumented_by_opentelemetry` for all functionality involving indicating whether a component has been instrumented.

@lzchen this statement in the PR description was confusing to me, that variable is a boolean, not a string correct?

lzchen:
@codeboten 
Yes it is a boolean. In the description I was outlining that we are using this specific text`_is_instrumented_by_opentelemetry` (instead of some other text like `is_instrumented`) for all things related to instrumenting. I've updated my comment.

instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/__init__.py
```diff
@@ -79,6 +79,7 @@
 """
 
 import typing
+from os import stat
```
lonewolf3739, line 82: nit: unused import

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -250,28 +254,22 @@ def instrument_app(
...
+        if hasattr(app, "_original_wsgi_app"):
             app.wsgi_app = app._original_wsgi_app
 
             # FIXME add support for other Flask blueprints that are not None
             app.before_request_funcs[None].remove(app._before_request)
             app.teardown_request_funcs[None].remove(_teardown_request)
             del app._original_wsgi_app
-
-            app._is_instrumented = False
+            app._is_instrumented_by_opentelemetry = False
```
lonewolf3739, line 272: Should we rather delete this attribute when uninstrumented?

codeboten:
Should this be specified in the spec or is this implementation specific details? https://github.com/open-telemetry/oteps/pull/165/files?short_path=fac8c59#diff-fac8c5914f3523b6888b428ec9eea8229238297768d93e2171418f7c45f805ce

instrumentation/opentelemetry-instrumentation-aiopg/src/opentelemetry/instrumentation/aiopg/wrappers.py
```diff
@@ -153,6 +153,10 @@ def instrument_connection(
     Returns:
         An instrumented connection.
     """
+    if isinstance(connection, wrapt.ObjectProxy):
```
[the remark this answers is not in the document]
lzchen: Yeah I was thinking about that too (differentiating OpenTelemetry proxy objects). We can actually just check for [TracedConnectionProxy](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-aiopg/src/opentelemetry/instrumentation/aiopg/aiopg_integration.py#L60) instead. Thoughts?

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
lzchen: Not a strong opinion on this. The field is really only used by OpenTelemetry.

lonewolf3739 (approved):

codeboten (approved):
👍 Thanks for addressing my comments

PR #555, opened 2021-06-28
Add tests for credentials from HTTPX in http.url

xoob:
# Description

According to the [specs for HTTP spans][1], `http.url` MUST NOT contain credentials. This change adds a test case to ensure that they are not in the URL attribute.

For code clarity we rename the type `RawURL` (a low-level tuple) to avoid confusion with `httpx.URL` (a high-level object). This is the same naming used in httpx internals.

[1]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#common-attributes

## Type of change

# How Has This Been Tested?

Tested locally.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
@@ -31,7 +31,7 @@ from opentelemetry.trace import SpanKind, Tracer, TracerProvider, get_tracer
 from opentelemetry.trace.span import Span
 from opentelemetry.trace.status import Status
 
-URL = typing.Tuple[bytes, bytes, typing.Optional[int], bytes]
+RawURL = typing.Tuple[bytes, bytes, typing.Optional[int], bytes]
 Headers = typing.List[typing.Tuple[bytes, bytes]]
 RequestHook = typing.Callable[[Span, "RequestInfo"], None]
 ResponseHook = typing.Callable[[Span, "RequestInfo", "ResponseInfo"], None]
@@ -45,7 +45,7 @@ AsyncResponseHook = typing.Callable[
 
 class RequestInfo(typing.NamedTuple):
     method: bytes
-    url: URL
+    url: RawURL
     headers: typing.Optional[Headers]
     stream: typing.Optional[
         typing.Union[httpx.SyncByteStream, httpx.AsyncByteStream]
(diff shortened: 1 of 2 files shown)


ocelotl:
> I suggest following the pattern established in #538 and adding a test.

Yes, please :+1:

owais (commented):
Thanks. Looks good. Can we please add a test so we don't unintentionally remove this in future?

xoob:
Alright, this was interesting. While implementing the unit test case I discovered that the URL set on the span was already stripped of credentials. Some further digging revealed that due the instrumentor is implemented as a Transport, which receives the stripped `request.url.raw` instead of the full `request.url`.

To make this more obvious, I have renamed the type `URL` to `RawURL` (a low-level tuple) to avoid confusion with `httpx.URL` (a high-level object). This is the same naming used by httpx internally.

This PR now only contains the unit test and variable renaming.

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -31,7 +31,7 @@
 from opentelemetry.trace.span import Span
 from opentelemetry.trace.status import Status
 
-URL = typing.Tuple[bytes, bytes, typing.Optional[int], bytes]
+RawURL = typing.Tuple[bytes, bytes, typing.Optional[int], bytes]
```
codeboten, line 34: Does this need to be exposed outside of the module? Could it be `_RawURL` instead?

codeboten (commented):
Thanks for the contribution! Just one question but otherwise the code looks good.

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
xoob: The library defines a few internal type hints like this; they could all be private. This would be a rather large patch of unrelated changes though, so may be more appropriate as a separate PR?

owais, line 34: Worth adding a comment documenting this does not contain credentials or other additional info?

ocelotl: Hm, I agree with @owais. "Raw" does not significantly tells the reader that the URL has no credentials. I'd say the we just keep the `URL` name and add the tests below.

ocelotl (changes requested):

owais:
Hi @xoob, thanks for the PR. Could you please take care of the remaining minor comments and resolve the conflicts so we can merge this? Thanks

lzchen:
@xoob 
We will be closing this PR for now. Feel free to open it back up if you will be resuming work for it.

PR #560, opened 2021-07-01
Require aiopg to be less than 1.3.0

RyanSiu1995:
# Description

Aiopg latest version has breaking changes in its internal object.
It breaks the build in default branch.
I have fixed the version here to make the build pass first.
It has a break changes in its API so we may need a further discussion on how to handle this case.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-aiopg/src/opentelemetry/instrumentation/aiopg/package.py
```diff
@@ -13,4 +13,4 @@
 # limitations under the License.
 
 
-_instruments = ("aiopg >= 0.13.0",)
+_instruments = ("aiopg <= 1.2.1",)
```
owais, line 16: shouldn't this be `>=0.13.0, <1.3.0` consider 1.3 introduced the breaking change and for some reason we did not want to support versions older than 0.13?

RyanSiu1995: I wasn't aware of the support of multiple conditions in python.
I have updated accordingly.

owais: Thanks. Can you please open an issue that tracks adding support for 1.3.0 and newer versions? Please also include details about the issue this instrumentation has today.

RyanSiu1995: Created

codeboten (approved):

owais (approved):

PR #562, opened 2021-07-02
Fix RequestsInstrumentor for custom transport adapters

mariojonke:
The `RequestsInstrumentor` contained some dead code from an early metrics implementation that in certain situations failed with an `AttributeError`. The `AttributeError` happened when trying to determine the HTTP flavor from the `raw.version` attribute on the returned `Response` object. 
By default, the `raw` attribute is an `urllib3` response that has the `version` attribute. However, `requests` allows mounting custom [transport adapters](https://docs.python-requests.org/en/master/user/advanced/#transport-adapters) for certain URLs. 
A custom adapter can return a `Response` where the `raw` object isn't an `urllib3` response and the version attribute might not be present. Thus, the `RequestsInstrumentor` could run into an `AttributeError`.

This change removes the dead metrics code.

## Type of change

# How Has This Been Tested?

Additional unit test to verify that the `RequestsInstrumentor` works with custom transport adapters.

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
@@ -131,10 +131,6 @@ def _instrument(tracer, span_callback=None, name_callback=None):
 
         url = remove_url_credentials(url)
 
-        labels = {}
-        labels[SpanAttributes.HTTP_METHOD] = method
-        labels[SpanAttributes.HTTP_URL] = url
-
         with tracer.start_as_current_span(
             span_name, kind=SpanKind.CLIENT
         ) as span:
@@ -164,15 +160,6 @@ def _instrument(tracer, span_callback=None, name_callback=None):
                     )
                     span.set_status(
                         Status(http_status_to_status_code(result.status_code))
-                    )
-                labels[SpanAttributes.HTTP_STATUS_CODE] = str(
-                    result.status_code
-                )
-                if result.raw and result.raw.version:
-                    labels[SpanAttributes.HTTP_FLAVOR] = (
-                        str(result.raw.version)[:1]
-                        + "."
-                        + str(result.raw.version)[:-1]
                     )
             if span_callback is not None:
                 span_callback(span, result)
--- a/instrumentation/opentelemetry-instrumentation-requests/tests/test_requests_integration.py
+++ b/instrumentation/opentelemetry-instrumentation-requests/tests/test_requests_integration.py
@@ -356,7 +356,6 @@ class RequestsIntegrationTestBase(abc.ABC):
         self.assertEqual(span.status.status_code, StatusCode.ERROR)
 
     def test_adapter_with_custom_response(self):
-        # See ONE-56722
         response = Response()
         response.status_code = 210
         response.reason = "hello adapter"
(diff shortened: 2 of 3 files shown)


instrumentation/opentelemetry-instrumentation-requests/tests/test_requests_integration.py
```diff
@@ -335,6 +355,27 @@ def test_requests_timeout_exception(self, *_, **__):
         span = self.assert_span()
         self.assertEqual(span.status.status_code, StatusCode.ERROR)
 
+    def test_adapter_with_custom_response(self):
+        # See ONE-56722
```
ocelotl, line 359: Hm, what is this? :thinking:

ocelotl (approved):

instrumentation/opentelemetry-instrumentation-requests/tests/test_requests_integration.py
mariojonke: thx, i removed it (just some ticket number)

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -131,10 +131,6 @@ def _instrumented_requests_call(
 
         url = remove_url_credentials(url)
 
-        labels = {}
-        labels[SpanAttributes.HTTP_METHOD] = method
-        labels[SpanAttributes.HTTP_URL] = url
```
codeboten, line 136: These changes don't seem directly related to the description of this change, were these attributes superfluous?

codeboten (commented):
Change looks good, I just have one question before approving.

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
lonewolf3739: I think this is the dead code @mariojonke is talking about in the PR description.

lonewolf3739:
Do client semantic conventions allow http.falvor? I think we could do something like `hasattr` and add the span attribute instead of entirely removing it? WDYT?

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
mariojonke: Yes, the `labels` are a leftover from the old metrics implementation. Since all the metrics related code got removed, the labels are unused.

mariojonke:
Semantic conventiosn for `http.flavor` exist in the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#common-attributes) for server and client spans.
However, parsing the flavor from the response might be unreliable since
1) a custom response might also have a version attribute
2) even if it is a urllib3 `HTTPResposnse` the version attribute is provided only from yet [another](https://github.com/urllib3/urllib3/blob/main/src/urllib3/response.py#L208) response object. Usually it is a `http.client.HTTPResponse` but i'm not sure if this is always the case.
So the format of the `version` attribute might actually be different from what is expected.

Also, I think the removed code for parsing the version appears to be incorrect. In the [HTTPResponse docs](https://docs.python.org/3/library/http.client.html#http.client.HTTPResponse.version) the version is defined as `10` for `HTTP/1.0`  and `11` for `HTTP/1.1`. However, for `10` the `http.flavor` attribute would be parsed to `1.1` due to how the version string is sliced.

codeboten (approved):

codeboten:
@lonewolf3739 please approve if your comments have been addressed, thanks!

lonewolf3739 (approved):

PR #563, opened 2021-07-02
Suppress botocore downstream instrumentation like urllib3

NathanielRN:
# Description

I noticed that if I instrument _both_ `urllib3` and `botocore` that I will get 2 traces (e.g. S3 and s3.amazon.com) when I should only get 1.

Before this fix:

Screen Shot 2021-07-02 at 2 40 58 PM

After this fix:

Screen Shot 2021-07-02 at 2 40 34 PM

## Type of change

# How Has This Been Tested?

The first way I thought to write a test for this would be to create an instance of the `urllib3` `Instrumentor` and prove that only 1 trace gets created, but since this is a common pattern among all the instrumentation libraries I didn't add it.

Let me know if this would be useful though!

- [x] No.

# Checklist:

~- [] Unit tests have been added~


lzchen:
Does `botocore` and `urllib3` share the same code path when executing a client call?

NathanielRN:
@lzchen Yes as far as I am aware!

A quick dive into the `botocore` package helped me find that `botocore > httpsession.py` has the following code:

```
class URLLib3Session(object):
...
def send(self, request):
        try:
            ...
            # conn is a `urllib3.connectionpool.HTTPConnectionPool` derived class type
                method=request.method,
                url=request_target,
                body=request.body,
                headers=request.headers,
                retries=Retry(False),
                assert_same_host=False,
                preload_content=False,
                decode_content=False,
                chunked=self._chunked(request.headers),
            )
```

And in our instrumentation we do the following for the `opentelemtry-instrumentation-urllib` package:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/bf97e172f0506c94590db4335dd96219b9ea7c79/instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py#L141-L145

So the trace is recording twice.

lzchen:
I believe `requests` also has this problem. We reserved [this](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py#L59) key word for http instrumentation suppression. Maybe we should use this instead?

NathanielRN:
@lzchen That's a fair ask, I see that the `urllib3` instrumentation will check both keys:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/bf97e172f0506c94590db4335dd96219b9ea7c79/instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py#L220-L224

Further `botocore` only requires `jmespath, urllib3, python-dateutil` so the only one we will instrument is probably the HTTP library. Will make the change!

lzchen:
Not a blocker for this pr, but we should probably move the context variable to a common location like [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py#L25).

lzchen (approved):

codeboten (approved):

PR #570, opened 2021-07-08
[exporter/datadog]: fix service name resolution

ericmustin:
# Description

This PR fixes the `service_name` resolution for the datadog exporter. The order of operations here is:

1. Resource attribute `service.name`
2. User supplied datadog `service`

However, in practice due to the work done in this PR: https://github.com/open-telemetry/opentelemetry-python/pull/1480

The resource attribute resolution will provide a default resource service.name of "unknown_service" in the event no resource attribute has been set for `service.name`. This means that even when a user does supply a fallback datadog `service` it will never be discovered beacuse the default `"unknown_service"` will take precedence. This PR updates the datadog exporter to explicitly check if a default `unknown_service` is being used for resource attribute `service.name` and if it does, allows the datadog supplied `service` to override it. 

Fixes # (issue)

Internall bug reports impacting end users

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/exporter.py
```diff
@@ -312,19 +313,23 @@ def _parse_tags_str(tags_str):
     return parsed_tags
 
 
-def _extract_tags_from_resource(resource):
+def _extract_tags_from_resource(resource, fallback_service_name):
     """Parse tags from resource.attributes, except service.name which
     has special significance within datadog"""
     tags = {}
     service_name = None
```
NathanielRN, line 320: Since it's not used below on line 321 anymore, I think it would make sense to push the initialization of `service_name = None` lower to line 322 perhaps?

exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/exporter.py
```diff
@@ -312,19 +313,23 @@ def _parse_tags_str(tags_str):
...
-        return [tags, service_name]
+        return [tags, fallback_service_name]
 
     for attribute_key, attribute_value in resource.attributes.items():
         if attribute_key == SERVICE_NAME_TAG:
             service_name = attribute_value
         else:
             tags[attribute_key] = attribute_value
+
+    if service_name is None or service_name == UNKNOWN_SERVICE_NAME:
```
NathanielRN, line 330: It look like `UNKNOWN_SERVICE_NAME` is only used 1 time in this exporter? Because that's the case you could consider just doing `"unknown_service"` like we do in `opentelemetry-sdk`:

https://github.com/open-telemetry/opentelemetry-python/blob/f11ed2f3bacb11d53a7a2b4837cf6308fa34cc71/opentelemetry-sdk/src/opentelemetry/sdk/resources/__init__.py#L172

But not a blocker!

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#560](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/560))
 - `opentelemetry-instrumentation-django` Migrated Django middleware to new-style.
   ([#533](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/533))
+- `opentelemetry-exporter-datadog` Fix service name resolution.
```
NathanielRN, line 31: Maybe this describes the change more explicitly?

```suggestion
- `opentelemetry-exporter-datadog` Datadog exporter should not use `unknown_service` as fallback resource service name
```

NathanielRN (approved):
Minor suggestions, otherwise looks good to me 😄

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  ([#540](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/570))
```
lzchen, line 32:
```suggestion
  ([#570](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/570))
```

exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/constants.py
```diff
@@ -14,3 +14,4 @@
 DD_ERROR_TYPE_TAG_KEY = "error.type"
 DD_ERROR_MSG_TAG_KEY = "error.msg"
 DD_ERROR_STACK_TAG_KEY = "error.stack"
+UNKNOWN_SERVICE_NAME = "unknown_service"
```
lzchen, line 17:
```suggestion
_UNKNOWN_SERVICE_NAME = "unknown_service"
```

exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/exporter.py
lzchen: Either that or change both to use `UNKNOWN_SERVICE_NAME`

lzchen:
Both [jaeger](https://github.com/open-telemetry/opentelemetry-python/blob/main/exporter/opentelemetry-exporter-jaeger-proto-grpc/src/opentelemetry/exporter/jaeger/proto/grpc/__init__.py#L112) and [zipkin](https://github.com/open-telemetry/opentelemetry-python/blob/main/exporter/opentelemetry-exporter-zipkin-proto-http/src/opentelemetry/exporter/zipkin/proto/http/__init__.py#L99) do not allow users to set the `service_name` through the constructor. Should we be consistent and recommend users to set the `service_name` through a `Resource`?

ericmustin:
@lzchen yes generally speaking I think that's a good direction to take, as 'd like to nudge folks to use the opentelemetry semantic conventions when using opentelemetry libraries.  If that means removing the `service` argument for this exporter that's fine with me. 

My only blocker is that I would have to coordinate with an update to the docs/blog posts/etc on my employers doc website so that any onboarding users were using the correct code snippets. So, if possible, i would prefer if we could just ship this non-breaking change (all the feedback seems reasonable, will update), and then I can try to get everything in order to update the exporter to be in line with other non-otlp exporters. If you'd like to make an issue for that work and assign it to me so that we can make sure this is tracked/completed, that's good with me.

Is that approach ok for y'all?

lzchen:
@ericmustin 
Sure that sounds good. Please add a TODO to that section to indicate that we plan on changing to using the `Resource`.
I've created an issue [here](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/574). Please comment on it so I can assign you.

exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/exporter.py
ericmustin: makes sense, updated!

exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/exporter.py
ericmustin: thanks, updated, i just removed the use of `UNKNOWN_SERVICE_NAME` entirely and went with "unknown_service".

CHANGELOG.md
ericmustin: lgtm, updated

CHANGELOG.md
ericmustin: good catch updated

ericmustin: good catch updated

CHANGELOG.md
ericmustin: lgtm, updated

exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/constants.py
ericmustin: removed this constant entirely but otherwise good catch

exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/exporter.py
```diff
@@ -312,19 +312,23 @@ def _parse_tags_str(tags_str):
...
 
-def _extract_tags_from_resource(resource):
+def _extract_tags_from_resource(resource, fallback_service_name):
     """Parse tags from resource.attributes, except service.name which
     has special significance within datadog"""
     tags = {}
-    service_name = None
     if not (resource and getattr(resource, "attributes", None)):
-        return [tags, service_name]
+        return [tags, fallback_service_name]
```
NathanielRN, line 320: Technically `tags` could just be `{}` here since it's not really use here at this point too! But that's not in scope for your PR I would say :-)

NathanielRN (approved):

exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/exporter.py
ericmustin: good catch. As you can tell my code is not particularly pythonic (i'm mostly ruby these days 😅 ). But there's a lot in this little exporter that could get cleaned up. I'll make a note to also try to give this code another pass for https://github.com/open-telemetry/opentelemetry-python-contrib/issues/574

lzchen (approved):
Could you add a test for this?

ericmustin:
@lzchen sorry for delay was on pto, added a test lmk if anything else needed here

Issue #551, opened 2021-06-24
Making it easier to create custom distros


NathanielRN:
# Description

In the June 24th, 2021 SIG meeting we mentioned that the way `Distro` and `Configurator` are set up right now are not friendly for creating **new** distros because the implementation takes the **1st distro** and the **1st configurator** it finds installed instead of taking the one the user might want.

See:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/63e7561931cd2d5e8edca9281c5156b9852eb348/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py#L47

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/63e7561931cd2d5e8edca9281c5156b9852eb348/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py#L92-L95

Auto-instrumentation _requires_ a distro to be installed. The `opentelemetry-distro` package provides a `Configurator` and an `OpenTelemetryDistro` class. Auto-instrumentation then hooks into the Configurator AND the Distro to configure instrumentation.

Currently the `Configurator` [initializes components](https://github.com/open-telemetry/opentelemetry-python/blob/1e600f385a81e340a2a4bbb7d10075f113149a21/opentelemetry-distro/src/opentelemetry/distro/__init__.py#L141-153)
```python
def _initialize_components():
    exporter_names = _get_exporter_names()
    trace_exporters = _import_exporters(exporter_names)
    id_generator_name = _get_id_generator()
    id_generator = _import_id_generator(id_generator_name)
    _init_tracing(trace_exporters, id_generator)


class Configurator(BaseConfigurator):

    # pylint: disable=no-self-use
    def _configure(self, **kwargs):
        _initialize_components()
```

And the `OpenTelemetryDistro` [sets environment variables](https://github.com/open-telemetry/opentelemetry-python/blob/1e600f385a81e340a2a4bbb7d10075f113149a21/opentelemetry-distro/src/opentelemetry/distro/__init__.py#L156-L164):

```python
class OpenTelemetryDistro(BaseDistro):
    """
    The OpenTelemetry provided Distro configures a default set of
    configuration out of the box.
    """

    # pylint: disable=no-self-use
    def _configure(self, **kwargs):
        os.environ.setdefault(OTEL_TRACES_EXPORTER, "otlp_proto_grpc_span")
```

## Problem

[...]

PR #571, opened 2021-07-09
Enforce one distro is installed

NathanielRN:
# Description

Small PR to enforce that only one distro be installed when trying to run auto instrumentation.

Additionally, checks that the `Configurator` that is loaded automatically is a derived class of `BaseConfigurator`. The idea is that `Configurator`s should come from distros and every distro should provide one. (Even if they just subclass the `opentelemetry-sdk` package Configurator which does a pretty good configuration).

Follow up to: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/551

## Type of change

# How Has This Been Tested?

Unfortunately I can't add unit tests because if you import functions from `sitecustomize.py` it will try to run the whole file and auto instrument which we can't do in a testing environment for now...

But I did confirm locally that an error is thrown _only_ if multiple distros are installed:

```
$ OTEL_RESOURCE_ATTRIBUTES=service.name=my_auto_app \
OTEL_EXPORTER_OTLP_INSECURE=True \
OTEL_EXPORTER_OTLP_ENDPOINT=localhost:4317 \
LISTEN_ADDRESS=127.0.0.1:8080 \
opentelemetry-instrument python3 auto-instrumentation/flask/application.py

Multiple distros were found: (opentelemetry.distro,opentelemetry.distro_aws,opentelemetry.distro_test). Only one should be installed.
Failed to auto initialize opentelemetry
  File "/Users/enowell/git/opentelemetry-python-contrib/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py", line 132, in initialize
    distro = _load_distros()
  File "/Users/enowell/git/opentelemetry-python-contrib/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py", line 45, in _load_distros
    raise RuntimeError("Cannot Auto Instrument with multiple distros installed.")
RuntimeError: Cannot Auto Instrument with multiple distros installed.
 * Serving Flask app "create_flask_app" (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: on
 * Running on http://127.0.0.1:8080/ (Press CTRL+C to quit)
```

- [x] No.

# Checklist:

~- [ ] Unit tests have been added~ See comment about testing above.

### DIFF
--- a/opentelemetry-instrumentation/README.rst
+++ b/opentelemetry-instrumentation/README.rst
@@ -38,6 +38,11 @@ opentelemetry-instrument
 
     opentelemetry-instrument python program.py
 
+**NOTE:** `opentelemetry-instrument` will check for compatible "distros" (like `opentelemetry-distro`)
+when automatically instrumenting. If multiple distros are found, instrumentation will fail. Check out
+`Github Issue #551 <https://github.com/open-telemetry/opentelemetry-python-contrib/issues/551>`_ for
+more.
+
 The instrument command will try to automatically detect packages used by your python program
 and when possible, apply automatic tracing instrumentation on them. This means your program
 will get automatic distributed tracing for free without having to make any code changes
--- a/opentelemetry-instrumentation/README.rst
+++ b/opentelemetry-instrumentation/README.rst
@@ -39,9 +39,11 @@ opentelemetry-instrument
     opentelemetry-instrument python program.py
 
 **NOTE:** `opentelemetry-instrument` will check for compatible "distros" (like `opentelemetry-distro`)
-when automatically instrumenting. If multiple distros are found, instrumentation will fail. Check out
+when automatically instrumenting. A "distro" is defined as a class that successfully inherits from the
+[BaseDistro](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/distro.py#L30).
+If multiple distros are found, auto-instrumentation will fail. Check out
 `Github Issue #551 <https://github.com/open-telemetry/opentelemetry-python-contrib/issues/551>`_ for
-more.
+more details.
 
 The instrument command will try to automatically detect packages used by your python program
 and when possible, apply automatic tracing instrumentation on them. This means your program
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
@@ -40,21 +40,24 @@ def _load_distros() -> BaseDistro:
 
         if more_distros:
             logger.error(
-                    "Multiple distros were found: (%s). Only one should be installed.",
-                    ",".join([first_distro.module_name] + more_distros),
-                )
-            raise RuntimeError("Cannot Auto Instrument with multiple distros installed.")
+                "Multiple distros were found: (%s). Only one should be installed.",
+                ",".join([first_distro.module_name] + more_distros),
+            )
+            raise RuntimeError(
+                "Cannot Auto Instrument with multiple distros installed."
+            )
 
         return first_distro.load()()
     except StopIteration:
-        logger.warning("Initializing Auto Instrumentation without using a distro.")
+        logger.warning(
+            "Initializing Auto Instrumentation without using a distro."
+        )
+        return DefaultDistro()
     except Exception as exc:  # pylint: disable=broad-except
         logger.exception(
             "Distribution %s configuration failed", first_distro.module_name
         )
         raise exc
-    logger.warning("Initializing Auto Instrumentation without using a distro.")
-    return DefaultDistro()
 
 
 def _load_instrumentors(distro):
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
@@ -35,7 +35,7 @@ logger = getLogger(__file__)
 def _load_distros() -> BaseDistro:
     entry_points = iter_entry_points("opentelemetry_distro")
     try:
-        first_distro = entry_points[0]
+        first_distro = next(entry_points)
 
         more_distros = [ep.module_name for ep in entry_points]
 
@@ -49,14 +49,14 @@ def _load_distros() -> BaseDistro:
         loaded_distro = first_distro.load()()
         if isinstance(loaded_distro, BaseDistro):
             logger.debug(
-                "Distribution %s will be configured", loaded_distro.module_name
+                "Distribution %s will be configured", first_distro.module_name
             )
             return loaded_distro
     except IndexError:
         logger.warning("Initializing Auto Instrumentation without using a distro.")
     except Exception as exc:  # pylint: disable=broad-except
         logger.exception(
-            "Distribution %s configuration failed", loaded_distro.module_name
+            "Distribution %s configuration failed", first_distro.module_name
         )
         raise exc
     logger.warning("Initializing Auto Instrumentation without using a distro.")
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
@@ -45,12 +45,7 @@ def _load_distros() -> BaseDistro:
                 )
             raise RuntimeError("Cannot Auto Instrument with multiple distros installed.")
 
-        loaded_distro = first_distro.load()()
-        if isinstance(loaded_distro, BaseDistro):
-            logger.debug(
-                "Distribution %s will be configured", first_distro.module_name
-            )
-            return loaded_distro
+        return first_distro.load()()
     except IndexError:
         logger.warning("Initializing Auto Instrumentation without using a distro.")
     except Exception as exc:  # pylint: disable=broad-except
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
@@ -34,38 +34,31 @@ logger = getLogger(__file__)
 
 def _load_distros() -> BaseDistro:
     entry_points = iter_entry_points("opentelemetry_distro")
-    loaded_distro = None
-    loaded_distro_name = None
-    for entry_point in entry_points:
-        if loaded_distro_name is not None:
+    try:
+        first_distro = entry_points[0]
+
+        more_distros = [ep.module_name for ep in entry_points]
+
+        if more_distros:
             logger.error(
-                "Multiple distros were found: (%s). Only one should be installed.",
-                ",".join([loaded_distro_name, entry_point.module_name] + [ep.module_name for ep in entry_points]),
-            )
+                    "Multiple distros were found: (%s). Only one should be installed.",
+                    ",".join([first_distro.module_name] + more_distros),
+                )
             raise RuntimeError("Cannot Auto Instrument with multiple distros installed.")
 
-        loaded_distro_name = entry_point.module_name
-
-        try:
-            loaded_distro = entry_point.load()()
-            if not isinstance(loaded_distro, BaseDistro):
-                logger.debug(
-                    "%s is not an OpenTelemetry Distro. Skipping",
-                    entry_point.module_name,
-                )
-                continue
+        loaded_distro = first_distro.load()()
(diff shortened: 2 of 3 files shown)


NathanielRN:
@owais Here is a PR to enforce only 1 distro be installed on the system! Would appreciate your feedback 😄

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
```diff
@@ -32,24 +33,40 @@
 
 
 def _load_distros() -> BaseDistro:
-    for entry_point in iter_entry_points("opentelemetry_distro"):
+    entry_points = iter_entry_points("opentelemetry_distro")
+    loaded_distro = None
+    loaded_distro_name = None
+    for entry_point in entry_points:
```
owais, line 39: Using a loop for this seems weird. Can we run a check for length instead? Check if `len(entry_points)` is > `1` and log the error. Also use `entry_points[0]` to get reference to the first found distro and guard it for `IndexError` exception.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
```diff
@@ -96,7 +113,14 @@ def _load_configurators():
             )
             continue
         try:
-            entry_point.load()().configure()  # type: ignore
+            configurator: BaseConfigurator = entry_point.load()()
+            if not isinstance(configurator, BaseConfigurator):
```
owais, line 117: Do we want to enforce nominal typing at runtime here? I don't see this adding a ton of benefit but these things keep adding up to the "startup" cost of the SDK which can be critical in environments like AWS Lambda with really low resources.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
NathanielRN: I agree it's a little weird 😓 I took the tradeoff though because the `entry_points` item is a generator so you can't run `len(entry_points)` on it 😕 

I guess I didn't want to run the cost of iterating through the iterator until we know for certain that there was > 1 distro installed, but I'll think of a re-write to make this cleaner 🙂

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
NathanielRN: If we consider startup costs should we remove the type check for `BaseDistro`?

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/753e22896a11cadf28c4f64acdfcfd73350d7d90/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py#L38-L42

I guess it has the benefit of checking a Distro and we know that Distro's have Configurators. I'll remove this but let me know if you think we should remove the `BaseDistro` check too.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
owais: how about this? :D
```
try
  distro = entry_points.__next__()
except StopIteration:
  # no distro found
```

It doesn't look clean either. I'm fine with loop or using __next__. Just that loop does not clarify the intention of this code block until you read it thoroughly. 

`list(entry_points)[0]` would probably be the best thing in terms of readability but not sure about performance as it might hit the disk on every iteration. Then again, ideally no one should have more than one distros installed and if they have, they probably are going to get undefined and inconsistent behavior anyway.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
owais: I'm not sure. What benefit do we get from this message? it's also logging on debug level but we are paying the cost of isinstance always. i personally feel we should remove this. This isn't something that is going to be very helpful to end users IMO. If anyone publishes a package with a distro entrypoint that doesn't satisfy the interface, an attribute/type error would probably be good enough I think

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
NathanielRN: I went with a similar solution using `next()` on the latest commit so hopefully it looks cleaner! I'll resolve this comment and we can continue there 🙂

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
NathanielRN: Okay I agree with you! I will remove it since I'm changing those lines anyways.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
owais: Sorry, just reviewed it in more detail. We are enforcing this by raising a RuntimeError if more than one distros are found. I think in that case startup performance shouldn't be a concern since the service will fail anyway. So I think converting to a list and running checks on length should be fine.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
```diff
@@ -32,24 +32,28 @@
...
-            return distro
-        except Exception as exc:  # pylint: disable=broad-except
-            logger.exception(
-                "Distribution %s configuration failed", entry_point.name
-            )
-            raise exc
+            raise RuntimeError("Cannot Auto Instrument with multiple distros installed.")
+
+        return first_distro.load()()
+    except IndexError:
```
owais, line 49: this would be `StopIteration` I think.

NathanielRN: Right, thank you! I changed it and configured that's what we wanted in my tests 😄 

```
opentelemetry-instrument python3 auto-instrumentation/flask/application.py
Initializing Auto Instrumentation without using a distro.
 * Serving Flask app "create_flask_app" (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: on
 * Running on http://127.0.0.1:8080/ (Press CTRL+C to quit)
```

ocelotl:
Hm, not sure if this is the approach we previously discussed. I don't think we should enforce only one distro being installed, but we should have a default one and let the user select one with an environment variable.

NathanielRN:
@ocelotl I think that is Option 1 in issue #551. The votes were in favor of Option 2 when I brought it up again so I added an implementation of it in this PR. I think it's the right way to go, but maybe we can discuss it more in the issue?

ocelotl:
> @ocelotl I think that is Option 1 in issue #551. The votes were in favor of Option 2 when I brought it up again so I added an implementation of it in this PR. I think it's the right way to go, but maybe we can discuss it more in the issue?

Ah, yes. Sorry, I forgot to vote. Also, detecting the error when running `opentelemetry-instrument` is too late. The error should be raised when another additional distro is installed.

NathanielRN:
> The error should be raised when another additional distro is installed.

Hm I'm not familiar with how to make packages exclusive from each other at install time... I don't know if you can do that at `pip install` time. Even if you can, there's always a possibility that a user just unzips the 2nd distro package's contents in the `python3.*/` folder and I'm pretty sure you can't throw an error/detect it at that time 😕

ocelotl:
Also, this approach is very inconvenient for someone who may want to switch between more than one distro repeatedly (for testing purposes, for example).

lonewolf3739:
I have a dumb question, why do we limit to only one distro?

opentelemetry-instrumentation/README.rst
```diff
@@ -38,6 +38,11 @@ opentelemetry-instrument
 
     opentelemetry-instrument python program.py
 
+**NOTE:** `opentelemetry-instrument` will check for compatible "distros" (like `opentelemetry-distro`)
+when automatically instrumenting. If multiple distros are found, instrumentation will fail. Check out
```
lzchen, line 42:
```suggestion
when automatically instrumenting. If multiple distros are found, auto-instrumentation will fail. Check out
```

opentelemetry-instrumentation/README.rst
```diff
@@ -38,6 +38,11 @@ opentelemetry-instrument
...
+`Github Issue #551 <https://github.com/open-telemetry/opentelemetry-python-contrib/issues/551>`_ for
+more.
```
lzchen, line 44:
```suggestion
more details.
```

opentelemetry-instrumentation/README.rst
lzchen, line 42:
```suggestion
when automatically instrumenting. A "distro" is defined as a class that successfully inherits from the [BaseDistro](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/distro.py#L30). If multiple distros are found, instrumentation will fail. Check out
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
```diff
@@ -32,25 +32,32 @@
...
+            raise RuntimeError(
+                "Cannot Auto Instrument with multiple distros installed."
             )
-            raise exc
-    return DefaultDistro()
+
+        return first_distro.load()()
+    except StopIteration:
+        logger.warning(
+            "Initializing Auto Instrumentation without using a distro."
```
lzchen, line 53: A description of the error would be good.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
```diff
@@ -32,25 +32,32 @@
 
 
 def _load_distros() -> BaseDistro:
```
lzchen, line 34:
```suggestion
def _load_distro() -> BaseDistro:
```

lzchen:
@NathanielRN 
Is the PR description/title still up to date? I don't see the changes pertaining to `Configurator`

NathanielRN:
@lonewolf3739 

> I have a dumb question, why do we limit to only one distro?

It's not a dumb questions 😄 The goal is just to make it easier for the user, because if multiple distros are installed they probably will set conflicting values and produce confusing bugs.

i.e.

`opentelemetry-distro-aws` will set `OTEL_PROPAGATORS=xray`

`opentelemetry-distro-microsoft` will set `OTEL_PROPAGATORS=azure`

We mentioned this comment in [#551](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/551#issuecomment-872390715) that this would be confusing.

But like @ocelotl said there are cases like testing multiple distros where this setup would be inconvenient. It probably warrants a SIG discussion at this point.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
NathanielRN: Added the following description:

```
logger.warning(
    "Failed to find even one installed distro. Initializing Auto Instrumentation without using a distro."
)
```

NathanielRN:
@lzchen 

> Is the PR description/title still up to date? I don't see the changes pertaining to Configurator

Enforcing 1 `Configurator` is already true as of right now:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/2ee2cf3cb54c7e9704567894636bcc1cfff702e6/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py#L91-L97

What I meant in this PR is that we enforce one Distro which means we enforce one Configurator. However I can see why that's confusing, I'll update the title.

lonewolf3739:
> because if multiple distros are installed they probably will set conflicting values and produce confusing bugs.

The way I imagined this is each distro would add to existing values or creates one if it doesn't exist, for example, the propagator is initially set to `jaeger`, and the vendor distro such as azure would update it to `jaeger,azure`. And when there are conflicting cases where there can be only one value then the latest update value is considered (following how we do it resource merge etc...).

NathanielRN:
We decided to go in a different direction for distro, follow the results here: https://github.com/open-telemetry/opentelemetry-python/discussions/2005

PR #583, opened 2021-07-14
Use basename to get package name

### LINKED ISSUE #582: Fix eachdist.py
`scripts/eachdist.py` is ignoring certain packages because it tries to get the package name by splitting with `split("/", maxsplit=1)`. This fails for a regular package path:

```python
In [1]: "/a/b/c/d".split("/", maxsplit=1)
Out[1]: ['', 'a/b/c/d']
```
This script also tries to split at a different character if the operating system is Windows.

Both problems can be fixed by using [`basename`](https://docs.python.org/3/library/os.path.html#os.path.basename) which is OS-portable:

```python
In [1]: from os.path import basename

In [2]: basename("/a/b/c/d")
Out[2]: 'd'
```

ocelotl:
Fixes #582

### DIFF
--- a/scripts/eachdist.py
+++ b/scripts/eachdist.py
@@ -11,6 +11,7 @@ from configparser import ConfigParser
 from datetime import datetime
 from inspect import cleandoc
 from itertools import chain
+from os.path import basename
 from pathlib import Path, PurePath
 
 DEFAULT_ALLSEP = " "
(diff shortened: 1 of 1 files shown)


horsfallnathan (approved):

lonewolf3739 (approved):

aabmass (approved):

codeboten (approved):
Will the same change be needed in the core repo?

ocelotl:
> Will the same change be needed in the core repo?

Thanks for pointing that out, it is not strictly necessary since the script in the core repo does not use `maxsplit=` but it would be an improvement, will open a PR. :v:

PR #585, opened 2021-07-14
updating changelogs and version to 1.4.0-0.23b0

ocelotl:


codeboten (approved):
👍

docs-requirements.txt
```diff
@@ -6,9 +6,9 @@ sphinx-autodoc-typehints
 # doesn't work for pkg_resources.
 -e "git+https://github.com/open-telemetry/opentelemetry-python.git#egg=opentelemetry-api&subdirectory=opentelemetry-api"
 -e "git+https://github.com/open-telemetry/opentelemetry-python.git#egg=opentelemetry-semantic-conventions&subdirectory=opentelemetry-semantic-conventions"
--e "git+https://github.com/open-telemetry/opentelemetry-python-contrib.git#egg=opentelemetry-instrumentation&subdirectory=opentelemetry-instrumentation"
+-e "git+https://github.com/open-telemetry/opentelemetry-python-contrib.git@7e9a1915b6863681f01e264296282089a299d3d5#egg=opentelemetry-instrumentation&subdirectory=opentelemetry-instrumentation"
 -e "git+https://github.com/open-telemetry/opentelemetry-python.git#egg=opentelemetry-sdk&subdirectory=opentelemetry-sdk"
--e "git+https://github.com/open-telemetry/opentelemetry-python-contrib.git#egg=opentelemetry-util-http&subdirectory=util/opentelemetry-util-http"
+-e "git+https://github.com/open-telemetry/opentelemetry-python-contrib.git@7e9a1915b6863681f01e264296282089a299d3d5#egg=opentelemetry-util-http&subdirectory=util/opentelemetry-util-http"
```
NathanielRN, line 11: Why do these values have to change? Shouldn't it already get the correct packages from the CORE-PR that recently got merge (and which corresponds to this PR)?: https://github.com/open-telemetry/opentelemetry-python/commit/bb41b81093d4010dadb233f146e8ba04973ec5a3

lonewolf3739 (approved):

docs-requirements.txt
ocelotl: Well... This is a workaround to make `tox -e docs` pass on the contrib repo 🤷

What happens is that after the latest release `opentelemetry -sdk` now depends on `opentelemetry-instrumentation`. The former is in the core repo and the latter was moved to the contrib one.

If we don't make these changes, then the `docs` virtual environment will end up having `opentelemetry-instrumentation` 23dev0 installed (because right now the release branch of the contrib repo has not yet been merged into `main`) and then will try to install `opentelemetry-sdk` 1.4.0 which depends on `opentelemetry-instrumentation` **23b0** (because right now the release branch of the core repo **has** been merged into `main`).

NathanielRN: Thanks for explaining the issue! I took the time to look into and I think we can have a different solution that doesn't involved setting the SHA. I [made a PR](https://github.com/ocelotl/opentelemetry-python-contrib/pull/1) on your fork to solve this issue.

I actually don't even know why this works... the commit we're using here _still_ has `opentelemetry-instrumentation==0.23dev0` doesn't it? Would look to hear what you think but that has me confused 🙂

NathanielRN (changes requested):
Following up on my [comment here](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/585#discussion_r670064019), I made [this PR on your fork](https://github.com/ocelotl/opentelemetry-python-contrib/pull/1) to hopefully solve the docs issue in a different way.

I can't make the PR on `opentelemetry-python-contrib` separately because it needs `opentelemtry-instrument==0.23b0` which you are adding in this PR.

If we agree with my solution, we should update the PR description with my PR description for history purposes.

owais (approved):

docs-requirements.txt
aabmass: > What happens is that after the latest release `opentelemetry -sdk` now depends on `opentelemetry-instrumentation`. The former is in the core repo and the latter was moved to the contrib one.

So core has a dependency on contrib now?

ocelotl:
> Following up on my [comment here](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/585#discussion_r670064019), I made [this PR on your fork](https://github.com/ocelotl/opentelemetry-python-contrib/pull/1) to hopefully solve the docs issue in a different way.
> 
> I can't make the PR on `opentelemetry-python-contrib` separately because it needs `opentelemtry-instrument==0.23b0` which you are adding in this PR.
> 
> If we agree with my solution, we should update the PR description with my PR description for history purposes.

All right, marking this PR as draft while we look into your solution :+1:

PR #588, opened 2021-07-15
Make it explicit to install distro to get auto-instrumentation working

lonewolf3739:
# Description

Fixes https://github.com/open-telemetry/opentelemetry-python/issues/1931

### DIFF
--- a/opentelemetry-instrumentation/README.rst
+++ b/opentelemetry-instrumentation/README.rst
@@ -11,14 +11,19 @@ Installation
 
 ::
 
-    pip install opentelemetry-distro[otlp] opentelemetry-instrumentation
+    pip install opentelemetry-instrumentation
 
 
 This package provides a couple of commands that help automatically instruments a program:
 
 .. note::
-    You need to install distro package (may be provided by your vendor) to get auto instrumentation working. The ``opentelemetry-distro``
-    package automatically configures some of the common options for users.
+    You need to install a distro package to get auto instrumentation working. The ``opentelemetry-distro``
+    package contains the default distro and automatically configures some of the common options for users.
+    For more info about ``opentelemetry-distro`` check `here <https://opentelemetry-python.readthedocs.io/en/latest/examples/distro/README.html>`_
+    ::
+
+        pip install opentelemetry-distro[otlp]
+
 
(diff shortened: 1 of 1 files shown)


opentelemetry-instrumentation/README.rst
```diff
@@ -11,11 +11,14 @@ Installation
 
 ::
 
-    pip install opentelemetry-instrumentation
+    pip install opentelemetry-distro[otlp] opentelemetry-instrumentation
```
lzchen, line 14: I think we should separate out these sections. Right underneath this, an explanation like what you have in the note would be good (users need to install "A" distro), and then have the `pip install` follow that excerpt. An explanation that `opentelemetry-distro` contains the default distro would be good too. This way users won't be confused as to why or what the `opentelemetry-distro` packge is.

owais (approved):
This looks fine but I still think our default recommendation should be to just install the distro and run `opentelemetry-instrument` as part of getting started docs. It should be followed up with an advanced section that documents other packages and how one may use them without the distro. IMO `opentelemetry-distro` should be the only or the first package users install when getting started.

lonewolf3739:
Completely agree. That will make a huge difference in terms of user experience for the first the time users. Meanwhile, I wanted to add this note since it is not obvious to someone that without a distro auto instrumentation doesn't produce any telemetry.

lzchen (approved):

lzchen:
@ocelotl 
Quick ping to make sure this pr doesn't get overlooked during the migration of `opentelemetry-instrumentation` to core.

https://github.com/open-telemetry/opentelemetry-python/pull/1959

ocelotl:
> @ocelotl
> Quick ping to make sure this pr doesn't get overlooked during the migration of `opentelemetry-instrumentation` to core if merged.
> 
> @lonewolf3739
> Can you fix the builds so we can merge this asap?
> 
> [open-telemetry/opentelemetry-python#1959](https://github.com/open-telemetry/opentelemetry-python/pull/1959)

Roger that, @lzchen.

ocelotl (approved):

lonewolf3739:
^ @lzchen it's green now.

PR #598, opened 2021-07-21
Worklow to publish a package from a CODEOWNER tag

NathanielRN:
# Description

Adds a new workflow (`.github/publish-a-package.yml`) and a new build script (`scripts/build_a_package.sh`) to allow releases of contrib repo packages from a push of a tag.

The hope is that once we approve CODEOWNERS who own contrib packages, they **can release packages independent of the maintainers by simply pushing a tag** with a package name and package version that matches a package in this repo.

The format of the tag should be:

`opentelemetry-<PKG_NAME>==<PKG_VERSION>`

NOTE: You can only release a package that is `>=1.0`! This is to allow beta packages to be released by the regular release workflow when the maintainers have a chance to do it.

I also updated the original script to skip `.tar.gz` files which are `>=1.0` for this very reason.

This is very similar to the setup over [on opentelemtry-dotnet-contrib where they have a separate workflow for every package](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/.github/workflows). That workflow gets triggered on a push to a relevant tag. However, my goal here is to one have **one workflow for each package**.

Also see the .NET PR that added this functionality: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/64

The PRO is less code duplication.

The CON is less control by CODEOWNERS as to what the workflow does for their specific package (although we can address those later on a case-by-case need)

## Type of change

~- [ ] This change requires a documentation update~ No?

# How Has This Been Tested?

- You can run `./scripts/build_a_package.sh` locally, you just need to:
  - Set `GITHUB_REF=` to something like `refs/tags/opentelemetry-sdk-extension-aws==0.23b0`
  - Update the `BASE_DIR` variable to the location of your contrib repo

- To test the actual workflow, I'll try to create my own repo sometime this week to validate everything works.

- [x] No.

# Checklist:

~- [ ] Unit tests have been added~ Didn't see any for the previous release flow?


.github/workflows/publish-a-package.yml
```diff
@@ -0,0 +1,29 @@
+name: Publish a Package from Tag
+
+on:
+  push:
+    tags:
+      - 'opentelemetry-+==1.+'
```
NathanielRN, line 6: `+` should be a valid glob value for tags according to [GitHub's docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags)

.github/workflows/publish-a-package.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  publish:
+    name: Publish package from tag ${{ env.GITHUB_REF }}
```
NathanielRN, line 10: This variable is _crucial_ to make this whole setup work. The variable is defined [on GitHub's docs](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables) and I took inspiration from [this SO post on how to get the currently pushed tag in GitHub actions](https://stackoverflow.com/a/58178121/7460739) for the rest of the PR.

.github/workflows/publish-a-package.yml
aabmass: > [`+`: Matches one or more of the preceding character.](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#:~:text=character-,%2B%3A%20Matches%20one,of%20the%20preceding%20character,-Matches)

Wouldn't that only match the `-` and `.` respectively?

.github/workflows/publish-a-package.yml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.7'
```
aabmass, line 16: Can this be a more current version?

scripts/build_a_package.sh
```diff
@@ -0,0 +1,79 @@
...
+set -ev
+
+if [[ -z $GITHUB_REF ]]; then
+  echo 'Failed to run script, missing workflow env variable GITHUB_REF'
+  exit -1
+fi
+
+# Follows PEP 440 versioning conventions: https://www.python.org/dev/peps/pep-0440/#pre-releases
+# Only allow publishing packages that are 1.0+ using this method
+if ! [[ $GITHUB_REF =~ ^refs/tags/(opentelemetry-.*)==([1-9]([0-9]+)?\.[0-9]+((a|b|rc)[0-9]+)?(\.(post)?[0-9]+)?)$ ]]; then
```
aabmass, line 30: I think something like this is a lot simpler if you don't care about validating against PEP440:

```suggestion
if ! [[ $GITHUB_REF =~ ^refs/tags/(opentelemetry-.*)==([1-9][0-9]*\..+)$ ]]; then
```

If you want to validate it, the regex might have bugs. There is a nice parser for this available in python [`packaging` package](https://packaging.pypa.io/en/latest/version.html). You might consider doing a `python -c` oneliner if you want to validate

scripts/build_a_package.sh
```diff
@@ -0,0 +1,79 @@
...
+
+# From the capture groups in the regex above
+PKG_NAME=${BASH_REMATCH[1]}
+PKG_VERSION=${BASH_REMATCH[2]}
+
+# Get the latest versions of packaging tools
+python3 -m pip install --upgrade pip setuptools wheel
+
+# If testing locally, change `BASEDIR` to your own repo directory
+BASEDIR=$(dirname $(readlink -f $(dirname $0)))
```
aabmass, line 43: Why would you need to change it when testing locally, this should work fine either way?

nit I've found this oneliner to be a bit simpler in other scripts: https://github.com/open-telemetry/opentelemetry-python/blob/c49e6fd6417282934b2ec4320b62b0e3adbe082d/scripts/proto_codegen.sh#L20-L21

scripts/build_a_package.sh
```diff
@@ -0,0 +1,79 @@
...
+cd $BASEDIR
+
+DISTDIR=${BASEDIR}/dist
+mkdir -p $DISTDIR
+rm -rf $DISTDIR/*
+
+for DIRECTORY_WITH_PACKAGE in exporter/*/ instrumentation/*/ propagator/*/ sdk-extension/*/ util/*/; do
```
aabmass, line 50: Would something like this work?

```bash
SETUP_PY_FILE=*/*/$PKG_NAME/setup.py
DIRECTORY_WITH_PACKAGE=$(dirname $setup_py_file)
```

aabmass:
> Adds a new workflow (`.github/publish-a-package.yml`) and a new build script (`scripts/build_a_package.sh`) to allow releases of contrib repo packages from a push of a tag.

How would a CODEOWNER push to a tag? Wouldn't this still require the CODEOWNER to have push access?

scripts/build_a_package.sh
NathanielRN: Sorry I probably should have been more specific, here's the error I get when running with `/bin/sh`:

```sh
$ sh
sh-3.2$ BASEDIR=$(dirname $(readlink -f $(dirname $0)))
readlink: illegal option -- f
usage: readlink [-n] [file ...]
usage: dirname path
```

That line you provided works great though! I'll be using it, thanks 🙂

NathanielRN:
@aabmass Thanks for the review! It prompted me to test the results out on my own repo and validate that it works 🙂 

Here is a workflow where the publish fails because [the tag package name does not correspond to any package](https://github.com/NathanielRN/opentelemetry-python-contrib-1/runs/3127188240?check_suite_focus=true).

Here is another workflow where the publish fails because [the tag matches a package but does not match the current version of the package in the repo](https://github.com/NathanielRN/opentelemetry-python-contrib-1/runs/3127197925?check_suite_focus=true).

Let me know if you have feedback on these?

.github/workflows/publish-a-package.yml
NathanielRN: Thank you! I wrongly assumed that it would work like the other globs but after reading the [Filter Pattern Cheat Sheet](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) I found out it does exactly what you say which isn't what I want.

I went with something simple and just used `*` instead, the script should be more informative as to why it fails if it does. That's probably preferred over the workflow just not triggering at all because it doesn't match.

.github/workflows/publish-a-package.yml
NathanielRN: I did this to copy `./scripts/build.sh`, but I used `3.9` on my own repo and everything worked fine so I updated this value to that.

scripts/build_a_package.sh
NathanielRN: Nice I really like the `packaging` package solution! It worked on my repo's workflow too so I'll keep it like that 👍 The only thing is we don't have a custom error message now, we just throw the python traceback instead.

See this workflow that fails on [invalid version in the tag](https://github.com/NathanielRN/opentelemetry-python-contrib-1/runs/3127327199?check_suite_focus=true)

scripts/build_a_package.sh
NathanielRN: Thanks for the tip! I don't know if I solved it the way you had envisioned but I got this to work:

```bash
SETUP_PY_FILE_PATTERN=$(ls -U **/$PKG_NAME/setup.py)
DIRECTORY_WITH_PACKAGE=$(dirname $SETUP_PY_FILE_PATTERN)
```

I found out that `ls -U` is faster because the results come unsorted, and for some reason doing `ls -U $SETUP_PY_FILE_PATTERN` didn't work but interpolating with other components _does_ work.

I guess there is a worry that `ls` could match multiple files? But hopefully not... in which case with only 1 file this works fine.

aabmass: The `ls -U` isn't doing anything, the glob gets evaluated in the shell before it even calls `ls`, so for example:

```sh
PKG_NAME=opentelemetry-instrumentation-flask
SETUP_PY_FILE_PATTERN=**/$PKG_NAME/setup.py
DIRECTORY_WITH_PACKAGE=$(dirname $SETUP_PY_FILE_PATTERN)
echo $DIRECTORY_WITH_PACKAGE
```

prints `instrumentation/opentelemetry-instrumentation-flask`

scripts/build_a_package.sh
```diff
@@ -0,0 +1,62 @@
...
+python3 -c "from packaging.version import Version; Version('${PKG_VERSION}')"
+
+BASEDIR=$(git rev-parse --show-toplevel)
+cd $BASEDIR
+
+DISTDIR=${BASEDIR}/dist
+mkdir -p $DISTDIR
+rm -rf $DISTDIR/*
+
+SETUP_PY_FILE_PATTERN=$(ls -U **/$PKG_NAME/setup.py)
```
aabmass, line 45: nit name is confusing, this variable will end up being the actual path to the setup.py file

scripts/build_a_package.sh
NathanielRN: > The ls -U isn't doing anything, the glob gets evaluated in the shell before it even calls ls

Ah thanks for that, I'll remove it right now

> prints `instrumentation/opentelemetry-instrumentation-flask`

Hm am I doing something wrong? I did the same but I get `**/opentelemetry-instrumentation-flask` printed out instead:

In my repo directory: `~/git/opentelemetry-python-contrib`

```
$ PKG_NAME=opentelemetry-instrumentation-flask
$ SETUP_PY_FILE_PATTERN=**/$PKG_NAME/setup.py
$ DIRECTORY_WITH_PACKAGE=$(dirname $SETUP_PY_FILE_PATTERN)
$ echo $DIRECTORY_WITH_PACKAGE
```

Prints `**/opentelemetry-instrumentation-flask`

scripts/build_a_package.sh
NathanielRN: Ah agreed, I had it be something different initially so I'll update it. Thanks!

scripts/build_a_package.sh
aabmass: That's weird, what shell are you using? How about `*/$PKG_NAME/setup.py`?

aabmass: Ah i see what you mean. I'm not sure the best way, this seems to work OK:

```bash
$ find -wholename "*/$PKGNAME/setup.py"
./instrumentation/opentelemetry-instrumentation-flask/setup.py
$ find -wholename "*/$PKGNAME/foobar.py"
# prints nothing
```

I think `ls` would fail with the nonexistant file case because the glob doesn't expand:
```bash
$ ls */$PKGNAME/foobar.py
ls: cannot access '*/opentelemetry-instrumentation-flask/foobar.py': No such file or directory
```

I'm fine with whatever, just wanted to call out you could probably avoid the for loop 😃

NathanielRN: Yes thank you for the call out it reads and works much better without the loop! 😄 

Sounds good, I think the `ls` error could be useful for people debugging and it's not too noisy. It also runs faster than `find .` at least on my computer 🙂 

Either way on failure `SETUP_FILE_PATH` is empty so I've added this check for a better log message and early exit!

```
if [ -z $SETUP_PY_FILE_PATH ]; then
  echo "Error! setup.py not found for $PKG_NAME, can't build."
  exit -1
fi
```

scripts/build_a_package.sh
```diff
@@ -0,0 +1,68 @@
...
+# wheel is then published to PyPI.
+
+set -ev
+
+if [ -z $GITHUB_REF ]; then
+  echo 'Failed to run script, missing workflow env variable GITHUB_REF'
+  exit -1
+fi
+
+PKG_NAME_AND_VERSION=${GITHUB_REF#refs/tags/*}
```
aabmass, line 28: One other bash nit, i'd use lowercase variable names for non-environment variables.

NathanielRN: That's a good idea! I was following `./scripts/build.sh` but we should probably update that one instead. Updated!

ocelotl (approved):

scripts/build_a_package.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
NathanielRN: I think we definitely could! I just wrote it like this for now so people could compare it against the `./build.sh` in the same repo and see the difference 🙂 I think it's a small enough script to change to Python in the near future if we want!

aabmass:
Code LGTM but I'm not sure if we've agreed on the approach enough to merge yet. My understanding is this isn't needed for a while until that checklist is complete, right?

lzchen:
LGTM.

@codeboten 
Thoughts on this?

aabmass (approved):

.github/workflows/publish-a-package.yml
```diff
@@ -0,0 +1,31 @@
...
+        python-version: '3.9'
+    - name: Log tag that triggered publish workflow
+      run:  echo "Attempting to publish package from tag $GITHUB_REF"
+    - name: Build wheel for tag
+      run: ./scripts/build_a_package.sh
+    - name: Install twine
+      run: |
+        pip install twine
+    # We don't need to publish to TestPyPI because we only publish 1 package.
+    # If it fails no other work needs to be reversed.
```
owais, line 25: This PR is fine but I think we should add this and then verify the package by installing it with pip. Once everything works as expected, only then we publish to real pypi service.

owais: Not asking we do it in this PR. Can we create a ticket for it?

owais (approved):
Would suggest to run a test with a test/local pypi server and a fake package.

.github/workflows/publish-a-package.yml
NathanielRN: Sounds good! I made #622 🙂

PR #602, opened 2021-07-22
Allow instrumentation of newer FastAPI versions

adamantike:
# Description

Since the introduction of the `_instruments` runtime checks in #475, the FastAPI instrumentation has stopped working for versions >= `0.59.0`. However the current test suite passes even for the latest released version at the moment (`0.67.0`).

It seems this isn't related to a limitation in the instrumentation code, but actually because of it being created when `0.58` was the latest version: https://github.com/open-telemetry/opentelemetry-python/commit/7bec76a22032e498911d06b9e3dab4e33528d995.

## Type of change

# How Has This Been Tested?

- [x] `tox -e test-instrumentation-fastapi`, which now installs latest FastAPI version.

- [x] No.

# Checklist:


adamantike:
Gentle ping to some of the reviewers for the initial PR with FastAPI support, just to increase awareness and visibility, as it would be great to allow newer FastAPI versions in the next minor/patch version of the library!

cc @codeboten @lzchen

lzchen:
@adamantike 
Which versions are actually not working? Just `0.58.1`?

owais:
- Does FastAPI follow semver? 
- Can we may be declare compatibility with all FastAPI versions that are <1.0?

potatochip:
> @adamantike
> Which versions are actually not working? Just `0.58.1`?

Anything greater than 0.58 is no longer working. Without testing, pretty sure that anything less than 0.58 is not working either. The fastapi instrumentor only works for fastapi version 0.58.x now.

potatochip:
> * Does FastAPI follow semver?
> * Can we may be declare compatibility with all FastAPI versions that are <1.0?

They follow semver. Should pin to less than v1 like you propose.

owais:
> They follow semver. Should pin to less than v1 like you propose.

I assumed all newer version were working. Sorry about the confusion. We should pin to the specific known working versions only and open an issue to support newer unsupported versions.

adamantike:
> Does FastAPI follow semver?
> Can we may be declare compatibility with all FastAPI versions that are <1.0?

They do follow SemVer, according to https://fastapi.tiangolo.com/deployment/versions/. However, as FastAPI is still at pre-1.0, any new minor version could include breaking changes.

I haven't found any issues when testing version `0.66.1`, and I think allowing `~=0.58` (to support >=0.58,<1.0) allows us to support newer minor versions that don't break the existing contracts for middlewares. As soon as a new minor version adds breaking changes that affect this instrumentation, we can add a stricter pin on FastAPI.

adamantike:
@owais, is there any context or changes I can provide here, to move forward with this PR?

owais (approved):

lonewolf3739 (approved):

owais:
@adamantike the CI was having multiple issues one after the other. Looks like we've sorted out most of the issues now. Thanks for your patience.

PR #612, opened 2021-08-04
Exit AWS ECS and EKS Detector container file reads early if container id found

NathanielRN:
# Description

The container ID in the `/proc/self/cgroup` file available on every container should be the same on every line that actually has it. For that reason, we can exit the file read early if we find it instead of loop and replacing it over and over again.

If we merge this before the next release I think we can just use the "SKIP CHANGELOG" tag 🙂 

## Type of change

# How Has This Been Tested?

Covered by previous unit tests.

- [x] No.

# Checklist:


sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/ecs.py
```diff
@@ -51,6 +51,7 @@ def detect(self) -> "Resource":
                         line = raw_line.strip()
                         if len(line) > _CONTAINER_ID_LENGTH:
                             container_id = line[-_CONTAINER_ID_LENGTH:]
+                            break
```
codeboten, line 54: is it possible to add a unit test for this?

NathanielRN: I modified the existing unit tests which had the container ID multiple times to use `bogusContainerIdThatShouldNotBeOneSetBecauseTheFirstOneWasPicked` on subsequent lines. That way if the code is missing this break statement it will pick up the wrong container ID because it didn't grab the first one 😄

lzchen (approved):

codeboten (approved):

PR #613, opened 2021-08-05
EKS Detector should return empty string, not resource if env var missing

NathanielRN:
# Description

We noticed a logic mistake in the "bad" path of the EKS detector. Instead of returning `Resource.get_empty()`, we should return the empty string `""` if the process can't be verified to be running on EKS.

This check was probably moved by me when I was setting up the tests. I've added a unit test for it now.

## Type of change

# How Has This Been Tested?

Created a unit test to make sure this code path raises an exception.

- [x] No.

# Checklist:

~- [ ] Changelogs have been updated~ Should not need a CHANGELOG if we merge before the next release

### DIFF
--- a/sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/eks.py
+++ b/sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/eks.py
@@ -74,7 +74,7 @@ def _get_cluster_info(cred_value):
     )
 
 
-def _get_cluster_name(cred_value):
+def _get_cluster_name(cred_value) -> str:
     cluster_info = json.loads(_get_cluster_info(cred_value))
     cluster_name = ""
     try:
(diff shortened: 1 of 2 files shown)


lzchen (approved):

sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/eks.py
```diff
@@ -74,11 +74,7 @@ def _get_cluster_info(cred_value):
     )
 
 
-def _get_cluster_name():
-    cred_value = _get_k8s_cred_value()
-    if not _is_eks(cred_value):
-        return Resource.get_empty()
-
+def _get_cluster_name(cred_value):
```
aabmass, line 77: maybe for clarity (too bad the type checker doesn't actually run on here)

```suggestion
def _get_cluster_name(cred_value) -> str:
```

aabmass (approved):

sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/eks.py
NathanielRN: Great idea! Thanks! 🙂

ocelotl (approved):

PR #631, opened 2021-08-26
Remove unnecessary dependencies

### LINKED ISSUE #547: opentelemetry-util-http requiring asgiref
**Steps to reproduce**
Install `opentelemetry-util-http`. It will install `asgiref`.

This dependency is not installed.

`asgiref` is installed.

**Additional context**
In the code for `opentelemetry-util-http` I don't see `asgiref` being used at all. Unless there is a good reason to keep it around, I am using `ExcludeList` outside of ASGI/WSGI environments, and I would prefer getting rid of the dependency in this place. If it is ok to remove it, it should be trivial. I can create the PR, just need confirmation.

ocelotl:
Fixes #547


instrumentation/opentelemetry-instrumentation-boto/setup.cfg
```diff
@@ -41,7 +41,6 @@ install_requires =
     opentelemetry-api ~= 1.3
     opentelemetry-semantic-conventions == 0.24b0
     opentelemetry-instrumentation == 0.24b0
-    opentelemetry-instrumentation-botocore == 0.24b0
```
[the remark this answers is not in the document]
ocelotl: Hm, I'm not sure. Even if that is the case, do we want that to happen? It seems like a very problematic way of adding spans, just because some dependency is installed by an instrumentation package, imagine debugging that.

ocelotl:
> I think [hhtpx](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-httpx/setup.cfg#L45) can be removed as well.

Sorry, where is `httpx` a removable dependency?

instrumentation/opentelemetry-instrumentation-boto/setup.cfg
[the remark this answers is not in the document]
owais: I think it is reasonable to do. Otel Node does it quite a bit IIRC. Debugging would be the same whether it is installed automatically or manually but with automatic install, we ship a much better default UX IMO.

lzchen:
@ocelotl 
Click the link that I linked. `wrapt` doesn't need to be in `httpx` I believe.

ocelotl:
> @ocelotl
> Click the link that I linked. `wrapt` doesn't need to be in `httpx` I believe.

:thinking: `wrapt` does [need](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py#L19) to be in `httpx`.

instrumentation/opentelemetry-instrumentation-aiopg/setup.cfg
```diff
@@ -39,14 +39,14 @@ package_dir=
 packages=find_namespace:
 install_requires =
     opentelemetry-api ~= 1.3
-    opentelemetry-semantic-conventions == 0.24b0
```
codeboten, line 42: why is this not required at install but is required for the test?

lzchen:
@ocelotl 

> wrapt does need to be in httpx.

If you take a look at that file, `wrapt` is not being used anywhere. Not sure how this was not caught in the linter.

ocelotl:
> @ocelotl
> 
> > wrapt does need to be in httpx.
> 
> If you take a look at that file, `wrapt` is not being used anywhere. Not sure how this was not caught in the linter.

Ok, I see. I have removed `wrapt` from `httpx` :+1: 

I also submitted #646 to deal with unused imports in contrib.

instrumentation/opentelemetry-instrumentation-aiopg/setup.cfg
ocelotl: Because it is not used in any code in `opentelemetry-instrumentation-aiopg/src` but it is used in `opentelemetry-instrumentation-aiopg/tests/test_aiopg_integration.py`.

codeboten: Ah i see... it relies on dbapi's use of the semantic conventions. 👍

codeboten (approved):

lzchen:
@owais 
Has your question been addressed?

owais:
Not a hard blocker for me but I'd like a better understanding of why boto has botocore as a dep. Does removing it mean users will have degraded experience if they just install boto instrumentation?

ocelotl:
> Not a hard blocker for me but I'd like a better understanding of why boto has botocore as a dep. Does removing it mean users will have degraded experience if they just install boto instrumentation?

I think this is something to address in a different issue/PR. Even if that is the case, it definitely needs more than just a dependency, that kind of behavionr needs to be documented somewhere, having spans just showing up when another package is installed is not user friendly either.

lzchen:
Discussed in SIG, we are okay to merge this.

Issue #444, opened 2021-04-14
Can't run python scripts/eachdist.py develop


dlopes7:
When I try running `python scripts/eachdist.py develop`

It fails with:

```
ERROR: Could not find a version that satisfies the requirement opentelemetry-api==1.0.1.dev0 (from opentelemetry-exporter-datadog[test])
ERROR: No matching distribution found for opentelemetry-api==1.0.1.dev0
```

**Steps to reproduce**

* `clone git@github.com:open-telemetry/opentelemetry-python-contrib.git`
* `cd opentelemetry-python-contrib.git`
* `git clone git@github.com:open-telemetry/opentelemetry-python.git opentelemetry-python-core`
* `python scripts/eachdist.py develop`

**What is the expected behavior?**

The script runs without errors

**What is the actual behavior?**

The script exists with error code 1, because of:

```
ERROR: Could not find a version that satisfies the requirement opentelemetry-api==1.0.1.dev0 (from opentelemetry-exporter-datadog[test])
ERROR: No matching distribution found for opentelemetry-api==1.0.1.dev0
[...]

PR #644, opened 2021-09-01
Falcon 3 support

adriangb:


adriangb:
@ocelotl take a look when you have a chance. I'm seeing the following failure:

```
_ ERROR collecting instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py _
1956
ImportError while importing test module '/home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py'.
1957
Hint: make sure your test modules/packages have valid Python names.
1958
Traceback:
1959
/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/importlib/__init__.py:127: in import_module
1960
    return _bootstrap._gcd_import(name[level:], package, level)
1961
test_falcon.py:19: in <module>
1962
    from opentelemetry.instrumentation.falcon import FalconInstrumentor
1963
E   ModuleNotFoundError: No module named 'opentelemetry.instrumentation.falcon'
```

I'm also seeing lint and generate failures. I ran `tox generate`, it did a bunch of stuff and it's still complaining. It seems like running it locally doesn't agree with CI 🤷 .

Also, `CONTRIBUTING.md` is still broken 😢

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -140,11 +140,22 @@ def instrumentation_dependencies(self) -> Collection[str]:
         return _instruments
 
     def _instrument(self, **kwargs):
+        self._original_falcon_api = None
+        self._original_falcon_app = None
```
owais, line 144: can we use only one attribute for both 2 and 3, and conditionally read/write falcon.App or falcon.API?

owais (commented):
Nice. Looks good. Left one comment.

owais:
@adriangb what command is that traceback from? I just ran `tox -e py39-test-instrumentation-falcon` on main and it worked perfectly for me locally.

owais:
I think you probably need to update these lines as well:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/tox.ini#L248
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/tox.ini#L239

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
adriangb: Sure, if you prefer that

adriangb:
> @adriangb what command is that traceback from? I just ran `tox -e py39-test-instrumentation-falcon` on main and it worked perfectly for me locally.

https://github.com/open-telemetry/opentelemetry-python-contrib/pull/644/checks?check_run_id=3485738088#step:7:1963

adriangb:
> I think you probably need to update these lines as well:

Makes sense, thank you, updated

tox.ini
```diff
@@ -174,6 +174,8 @@ deps =
   ; FIXME: Elasticsearch >=7 causes CI workflow tests to hang, see open-telemetry/opentelemetry-python-contrib#620
   ; elasticsearch7: elasticsearch-dsl>=7.0,<8.0
   ; elasticsearch7: elasticsearch>=7.0,<8.0
+  falcon2: falcon >=2.0.0,<=3.0.0
+  falcon2: falcon >=3.0.0,<=4.0.0
```
owais, line 178:
```suggestion
  falcon3: falcon >=3.0.0,<=4.0.0
```

adriangb: thanks, good catch

tox.ini
```diff
@@ -175,7 +175,7 @@ deps =
   ; elasticsearch7: elasticsearch-dsl>=7.0,<8.0
   ; elasticsearch7: elasticsearch>=7.0,<8.0
   falcon2: falcon >=2.0.0,<=3.0.0
```
owais, line 177:
```suggestion
  falcon2: falcon >=2.0.0,<3.0.0
```

tox.ini
```diff
@@ -175,7 +175,7 @@ deps =
...
-  falcon2: falcon >=3.0.0,<=4.0.0
+  falcon3: falcon >=3.0.0,<=4.0.0
```
owais, line 178:
```suggestion
  falcon3: falcon >=3.0.0,<4.0.0
```

adriangb:
Thanks for the help @owais 

Looks like the tests are actually broken now. The tests access a private attribute of `falcon.App` then double `[0]` index on it and then accesses `__self__`. There are no comments as to why all of this is going on. To be honest, I'm pretty shocked that tests are monkey patching private attributes of stuff that was already monkey patched by the instrumentation. **monkey patching is bad!**

owais:
Looks like Falcon changed how it handles middlewares internally so we need to update test code so it can correctly fetch the injected Otel middleware for both 2 and 3.

adriangb:
Yes, I agree. However I think I think it's unfair to blame this on Falcon: the test breaks because it's fragile, and it's fragile because it accesses private attributes of a 3rd party library. They could have made the change in a patch bump, we should be thankful that they seem to have done it during a major version bump...

While it may be possible to fix this here, I think the overall approach of monkey patching stuff (in the library, in tests) and accessing private stuff in 3rd party libraries is fundamentally flawed and brittle.

It's also going to be hard to fix this test: I still can't run this locally (because as we've discussed before the local setup instructions are broken; and no one has been able to give me a series of valid steps of cloning repos into repos and installing stuff that works).

It looks like this portion of the code was authored by you (@owais ) and @lonewolf3739. I'd appreciate a walkthrough of _why_ this is being done, what `[0][0]` is accessing, etc if I'm going to try to fix this test by blindly pushing to CI.

owais:
I'm not blaming anything. I was merely pointing out what needs to be fixed in order to make the test pass again. 

> I'd appreciate a walkthrough of why this is being done, what [0][0] is accessing, etc if I'm going to try to fix this test by blindly pushing to CI.

Looking at the test code, it need a reference to the injected middleware which can be accessed from a falcon app instance with the help of `[0][0]` because of how it internally stored references to middlewares. If there is a public API to fetch references to installed middleware, we should use that but I'm not aware of any.

owais:
> It's also going to be hard to fix this test: I still can't run this locally (because as we've discussed before the local setup instructions are broken; and no one has been able to give me a series of valid steps of cloning repos into repos and installing stuff that works).

Is there an issue for this I can look at and answer?

adriangb:
> Is there an issue for this I can look at and answer?

#444 is one, also discussions in https://github.com/open-telemetry/opentelemetry-python/pull/1952#issuecomment-892796468

owais:
> #444 is one,

I've never used `eachdist.py develop` so can't tell how to fix it. I always install every dep manually with `python setup.py develop` to create temp environments.

For tests, I just run specific tox target. For Falcon, I'd run `tox -e py39-test-instrumentation-falcon`. Then to speed up iterative development, I activate the env created by tox and run pytest directly as below:

```
# run one time to create env
tox -e py39-test-instrumentation-falcon
source .tox/py39-test-instrumentation-falcon/bin/activate


# then iterate on tests and run the following to test each change
pytest -s instrumentation/opentelemetry-instrumentation-falcon/tests/
```

`tox -e lint` is extremely slow for me locally especially when I just wan to lint a single package so I never run it locally and use CI for it.

adriangb:
> I've never used eachdist.py develop so can't tell how to fix it.

Well that's what the `CONTRIBUTING.md` docs say, so if it's not supported, the docs should be edited

> I always install every dep manually with python setup.py develop to create temp environments.

That doesn't sound like a sustainable solution for contributors who aren't familiar w/ the packaging being done here

> For tests, I just run specific tox target. For Falcon, I'd run tox -e py39-test-instrumentation-falcon

I get: `ERROR: Could not find a version that satisfies the requirement opentelemetry-test==0.24b0 (from opentelemetry-util-http[test]) (from versions: none)`

Evidently, there's some more to this. I recall there being some stuff about cloning other repos into subdirectories of this repo and whatnot.

Regardless, I again don't think that's a sustainable solution for contributors, especially if it's not documented anywhere.

> tox -e lint is extremely slow for me locally

Maybe this package should consider using [pre-commit](http://pre-commit.com) instead so that black & friends only need to run on changed files?

owais:
> Why does the test need to monkey patch its own middleware? If it needs a callback to record that the middleware worked, the middleware should have a constructor parameter for that. And the .instrument call should have a parameter to accept a pre-build middleware bound to the callback.

I think it's because the test is verifying extraction of those values from env instead of arguments passed to constructor. The env is detected only once per module and [stored globally ](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py#L127-L129)

If we move env extraction to intrumentor constructor, we can solve this by patching env in test's setUp before initializing the instrumentor. Essentially [this](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py#L43-L50) needs to happen before [this](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py#L37-L40)

_I think_ if we make these changes, we wouldn't need to patch the middleware to "fake" injection from env.

owais:
>   I get: ERROR: Could not find a version that satisfies the requirement opentelemetry-test==0.24b0 (from opentelemetry-util-http[test]) (from versions: none)

This is helpful. Can you run the following command in contrib repo root dir and share the output:

```
cat opentelemetry-python-core/tests/util/src/opentelemetry/test/version.py
```

owais:
@adriangb if the output from the above command is not 

```
__version__ = "0.24b0"
```

Then your `opentelemetry-python-core` dir has outdated version of the core repo. You'll need to make sure it is on latest `main` branch. Once you do that, tox commands should start working.

adriangb:
It is `__version__ = "0.24.dev0"`.

To me an outside contributor it is super arbitrary that `your opentelemetry-python-core dir has outdated version of the core repo`. I may have cloned the core repo, but only because I was at some point arbitrarily told to do so. If cloning the core repo is into a specific folder is required to run things, it should be hidden behind some (working) setup script or at least well documented (as far as I know, it is neither).

In any case, re-cloning the repo now gets me `__version__ = "0.24b0"` and I can now run `tox -e py39-test-instrumentation-falcon`.

Now when running `tox -e py39-test-instrumentation-falcon` I get a key error for a missing env var `AWS_ACCESS_KEY_ID` from [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/b6e49ba66d70998813766d763bdb08f453829518/instrumentation/opentelemetry-instrumentation-boto/tests/conftest.py#L28). Looks like someone forgot to provide a default to `dict.pop` (`{}.pop("")` raises a value error)

owais:
I'm just trying to unblock you NOW so you get the PR through. Obviously we need to update docs and improve tooling but I'd suggest to send different PRs or open issues with ideas to fix or even just general feedback. Let's focus on removing the immediate blockers from preventing this PR from progressing here. 

> Now when running tox -e py39-test-instrumentation-falcon I get a key error for a missing env var AWS_ACCESS_KEY_ID from here. Looks like someone forgot to provide a default to dict.pop ({}.pop("") raises a value error)

I've seen this but only with boto/botocore instrumentation. That line should not even run if we are only running falcon tests. Can you please share the entire output?

adriangb:
Here's a dump of the output from `tox -e py39-test-instrumentation-falcon > out.txt`

I think that line will run since purest always runs `conftest.py` during collection.

owais:
They shouldn't run for sure as tests for the package are run from the package's root so pytest wouldn't find conftests from other packages. I also verified it by trying again locally and they did not run. _I think_ your env probably has boto instrumentation package installed with `python setup.py develop`. Just a wild guess. Can't think of other reason why conftest from boto would be activated when running falcon tests. 

May be try remove the repos and clone fresh?

You can also just comment out those lines to make your local tests pass and then uncomment before pushing. 

Also, I think since tox env is created, you should be able to run the following to just run falcon tests:

`.tox/py39-test-instrumentation-falcon/bin/pytest instrumentation/opentelemetry-instrumentation-falcon/tests/`

adriangb:
Well I was able to activate the virtual env (`. .tox/py39-test-instrumentation-falcon/bin/activate`), but it seems that the package itself was never installed (?) (`E   ModuleNotFoundError: No module named 'opentelemetry.instrumentation.falcon'`). And `pip install -e .` doesn't work...

adriangb:
@owais w/ the upstream fixes and trashing the repos and re-cloning, I was able to use tox to generate the env and then run pytest. In order to run the tests in the VSCode debugger (which I find extremely helpful) I had to edit `pytest.ini` to add `testpaths = instrumentation/opentelemetry-instrumentation-falcon/tests/` so that it won't execute that boto conftest.py and fail on missing env vars.

With that, I was able to track down at least some of the failures to a real problem: the tests are using `falcon.API`, and monkey patching `falcon.App` does not monkey patch `falcon.API`. We could also monkey patch `falcon.API`, but it gets messy. I think it should be okay to only support `falcon.App` on Falcon 3. Side note: this is yet another example of why **monkey patching like this is bad design**.

So the sole remaining issue now is exception info: currently `exc_info()` is being used from within the middleware's `handle_response` method, but it seems that Falcon 3 does not execute `handle_response` within the context of the raised exception, so `exc_info() == None` -> tests fail.

owais:
> With that, I was able to track down at least some of the failures to a real problem: the tests are using falcon.API, and monkey patching falcon.App does not monkey patch falcon.API. We could also monkey patch falcon.API, but it gets messy. I think it should be okay to only support falcon.App on Falcon 3. 

Isn't API an alias to App in Falcon 3? My initial reaction is that we should only instrument App and not API but we need to understand the implications of that.

- If we instrument only App, does it mean the instrumentation will NOT work for users who are still using falcon.API?
- If we instrument both App and API, will it lead to duplicate spans or other issues? If not, I think we should instrument both to support most users.


> So the sole remaining issue now is exception info: currently exc_info() is being used from within the middleware's handle_response method, but it seems that Falcon 3 does not execute handle_response within the context of the raised exception, so exc_info() == None -> tests fail.

This is pretty important. We need to capture exceptions/errors to provide full visibility into a Falcon app. We'll need to figure out how Falcon 3 handles exceptions and perhaps patch some other place to capture that information.

adriangb:
> Isn't API an alias to App in Falcon 3?

No, it's a subclass

> * If we instrument only App, does it mean the instrumentation will NOT work for users who are still using falcon.API?

Correct

> * If we instrument both App and API, will it lead to duplicate spans or other issues? If not, I think we should instrument both to support most users.

I think we could, but API subclasses App and adds a deprecation warning. So we don't wan't to duplicate instrument. I think we'd have to copy the definition of API from their codebase to this one and re-bind it to the monkey patched App. Doesn't sound pretty, I bet there's rough edges. I think it's better to just keep things simple and well documented, API is deprecated anyway.
 
> This is pretty important. We need to capture exceptions/errors to provide full visibility into a Falcon app. We'll need to figure out how Falcon 3 handles exceptions and perhaps patch some other place to capture that information.

I looked, I don't think it can be done without doing one of:
1. Monkey patching private methods in their class 👎 
2. Adding an exception handler to catch `Exception` and then wrapping that in a `falcon.HTTPInternalServerErr`, taking care to add the same `{"detail" : ...}` that Falcon itself would have added.
3. Add an exception handler, record the info we need and then call their default exception handler (which I think we can get, but only via private datastores, and it's probably different for Falcon 2 & 3).

None of these options sounds great...

Maybe we can get Falcon to change its behavior, I think it should be backwards compatible to just execute the middleware response processing within the context of the exception.

owais:
> I think we could, but API subclasses App and adds a deprecation warning. So we don't wan't to duplicate instrument. I think we'd have to copy the definition of API from their codebase to this one and re-bind it to the monkey patched App. Doesn't sound pretty, I bet there's rough edges. I think it's better to just keep things simple and well documented, API is deprecated anyway.

Ideally I'd like to ship an experience that "just works" but if it is this complicated, we can document that Falcon 3 instrumentation will work only with Falcon.App even though Falcon itself works with both App and API.



> I looked, I don't think it can be done without doing one of:
>    Monkey patching private methods in their class 👎
>    Adding an exception handler to catch Exception and then wrapping that in a falcon.HTTPInternalServerErr, taking care to add the same {"detail" : ...} that Falcon itself would have added.
>    Add an exception handler, record the info we need and then call their default exception handler (which I think we can get, but only via private datastores, and it's probably different for Falcon 2 & 3).
> Maybe we can get Falcon to change its behavior, I think it should be backwards compatible to just execute the middleware response processing within the context of the exception.

I want to clear the thing about monkey patching. I know it's not ideal and can be distasteful but this is what the OpenTelemetry (and OpenCensus+OpenTracing before that) community has adopted as the solution for 100% automatic instrumentation and it has worked well in practice despite occasional breakage here and there. So as a project, we have embraced dynamic runtime patching to implement automatic instrumentation. This repo alone has 30+ instrumentations and most of them do some for of patching at runtime.  

Note that we only patch to extract information and our patches are more like wrappers. We never patch to modify behaviour which is actually the most dangerous thing about monkey-patching as it can result in incompatibilities with the rest of the ecosystem. Only downside to our patches is that in a future version, one of our patches may stop working due to internal interface changes and we've accepted this risk in order to ship the experience we want. 

That said, I'd be happy to look at any Python specific proposals to ditch runtime patching that still allows us to ship the same user experience.

With that out of the way, if wrapping some internal method of Falcon is what it takes to extract error/exception information then let's do it. I don't see any issues with it given we already do it in multiple places for 30+ packages.

> Maybe we can get Falcon to change its behavior, I think it should be backwards compatible to just execute the middleware response processing within the context of the exception.

Not sure if I'm following this but if I understood the comment, this goes against the principles of instrumentation. We only want to patch/wrap to extract information. Other than a possible 1-5% performance hit, enabling an instrumentation should have no effect on an application. It should not change behavior or code paths the app takes.

adriangb:
> Ideally I'd like to ship an experience that "just works" but if it is this complicated, we can document that Falcon 3 instrumentation will work only with Falcon.App even though Falcon itself works with both App and API.

To me "no surprises that require me to dig into the source code to see what is going on" >> "just works".

> I want to clear the thing about monkey patching. I know it's not ideal and can be distasteful but this is what the OpenTelemetry (and OpenCensus+OpenTracing before that) community has adopted as the solution for 100% automatic instrumentation and it has worked well in practice despite occasional breakage here and there. So as a project, we have embraced dynamic runtime patching to implement automatic instrumentation. This repo alone has 30+ instrumentations and most of them do some for of patching at runtime.

I don't see an argument for the benefits of monkey patching as the default approach, all I see is "this ship sailed". IMO an API like `App = instrument(falcon.App)` where App is a closure dynamically created from the argument is much more robust than monkey patching, it can be nested (e.g. if something other than OpenTelemetry wants to "instrument", doesn't depend on import order or dereferencing from the module, would work with falcon.App and falcon.API, etc. It's not "automatic" as it seems you've defined it here, and the API can't be used for every package, but it's explicit and robust.

> We never patch to modify behaviour which is actually the most dangerous thing about monkey-patching as it can result in incompatibilities with the rest of the ecosystem.

Agreed, but it's a fine line. It's easy to introduce behavior changes when injecting hundreds of lines of code, especially in a language like Python where getters can hide side effects and such. That doesn't just apply to monkey patching though.

> Only downside to our patches is that in a future version, one of our patches may stop working due to internal interface changes and we've accepted this risk in order to ship the experience we want.

I think this is a pretty big risk. A patch version in the instrumented package could break instrumentation, or a users app. They may _need_ to upgrade due to security reasons. So now users have to pick between security and OpenTelemetry. I wouldn't want to put users in that position.

> That said, I'd be happy to look at any Python specific proposals to ditch runtime patching that still allows us to ship the same user experience.

I think one thing that could help (other than more explicit testable APIs like the one I propose above) is to use a layered approach: a robust middleware, subclass or whatever that errs way on the robustness and composability side of things and another higher level interface that applies this lower level automatically, perhaps with sensible defaults and less configurability. This sort of design lets users start with the high level approach and peel back the onion if they have to, for more advanced use cases or because the higher level approach breaks (e.g. because it had no choice but to monkey patch a private attribute). But none of that is Python specific.

> I don't see any issues with it given we already do it in multiple places for 30+ packages.

The issue is that it is brittle and can easily break. Not to mention making long term maintenance harder because of version incompatibility like we're experiencing here.

> > Maybe we can get Falcon to change its behavior, I think it should be backwards compatible to just execute the middleware response processing within the context of the exception.
> 
> Not sure if I'm following this but if I understood the comment, this goes against the principles of instrumentation. We only want to patch/wrap to extract information. 

Sorry, I should have clarified. I meant submitted a PR to Falcon, not monkey patch it here. It shouldn't break anything for their users unless someone is doing `assert exc_info() == None` in their middleware.

owais:
This PR has already has already deviated too much from the actual issue it is trying to solve. I'm happy to discuss any topic but lets please create separate issues or discussions for those and here just focus on getting this PR through. 

> I don't see an argument for the benefits of monkey patching as the default approach, all I see is "this ship sailed".

It has not sailed. instrumentation is still pre-1.0. Even if it was 1.0, we could release 2.0 to improve things. So if there is a solution other than patching that allows automatic instrumentation, I'd be happy to review. In any case it should be a separate proposal and not debated here.

adriangb:
Agreed, sorry for derailing. I'll loop back once I here back in https://github.com/falconry/falcon/issues/1942 or if I come up with a solution. At least now I'm able to test locally, so thank you for your help 😃

owais:
Thanks but I don't think falconry/falcon#1942 might not be good enough because at best it'll be release in the next minor version. That means our instrumentation will have a gap in supported versions (3.0 and 3.0.1).  That said, if it works out and is released relatively soon, may be we can support `~2.0` and `~3.1` and skip 3.0

adriangb:
Yeah, agreed. I think that as long as the latest version of each major version of supported it should be OK

owais:
😬 we try to support all versions between the minimum and maximum but I think we can make an exception about starting 3.x support from 3.1 or 3.0.2 provided they are simple upgrades on 3.0.

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -193,7 +211,6 @@ def __call__(self, env, start_response):
         env[_ENVIRON_ACTIVATION_KEY] = activation
 
         def _start_response(status, response_headers, *args, **kwargs):
-            otel_wsgi.add_response_attributes(span, status, response_headers)
```
adriangb, line 196: This was a pretty insidious bug that was hard to figure out.

If you look at the middleware, it is already setting `span.status`. This _re sets it_. `otel_wsgi.add_response_attributes` sets the same status code as the middleware, but does _not_ set the description. So after this line executes `span.status.description` is always None. But none of the tests were checking this, so it was not caught.

But what about `test_500` then? Well, in Falcon 2 (but not 3) `super().__call__(...)` would re-raise the error, triggering the `except` clause right below this that skips calling `_start_response` and thus `otel_wsgi.add_response_attributes(span, status, response_headers)` is never called and `span.status.description` is carried over from the middleware.

I'll add an assertion for `span.status.description` to other tests so that this can be avoided in the future.

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -169,6 +178,15 @@ def __init__(self, *args, **kwargs):
...
         super().__init__(*args, **kwargs)
 
+    def _handle_exception(
+        self, req, resp, ex, params
+    ):  # pylint: disable=C0103
+        # Falcon 3 does not execute middleware within the context of the exception
+        # so we capture the exception here and save it into the env dict
+        _, exc, _ = exc_info()
+        req.env[_ENVIRON_EXC] = exc
+        return super()._handle_exception(req, resp, ex, params)
```
adriangb, line 188: I couldn't figure out another way to do this other than accessing this private method, FYI

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -169,6 +178,15 @@ def __init__(self, *args, **kwargs):
         self._excluded_urls = get_excluded_urls("FALCON")
         super().__init__(*args, **kwargs)
 
+    def _handle_exception(
+        self, req, resp, ex, params
+    ):  # pylint: disable=C0103
+        # Falcon 3 does not execute middleware within the context of the exception
+        # so we capture the exception here and save it into the env dict
+        _, exc, _ = exc_info()
+        req.env[_ENVIRON_EXC] = exc
```
adriangb, line 187: I'm using the already defined but previously unused key `_ENVIRON_EXC` to store the status info.

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -126,6 +126,13 @@ def response_hook(span, req, resp):
 
 _response_propagation_setter = FuncSetter(falcon.Response.append_header)
 
+if hasattr(falcon, "App"):
+    # Falcon 3
+    _instrument_app = "App"
+else:
+    # Falcon 2
+    _instrument_app = "API"
```
adriangb, line 134: I centralized this logic up here so we don't duplicate it below

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -138,14 +145,16 @@ def instrumentation_dependencies(self) -> Collection[str]:
...
+            falcon, _instrument_app, partial(_InstrumentedFalconAPI, **kwargs)
+        )
 
     def _uninstrument(self, **kwargs):
-        falcon.API = self._original_falcon_api
+        setattr(falcon, _instrument_app, self._original_falcon_api)
 
 
-class _InstrumentedFalconAPI(falcon.API):
+class _InstrumentedFalconAPI(getattr(falcon, _instrument_app)):
```
adriangb, line 157: I really dislike using a `getattr` in a class def, but it works and it avoids a deprecation warning for Falcon 3 users.

adriangb:
@owais this is now working 😄

adriangb:
> I'm happy to discuss any topic but lets please create separate issues or discussions for those and here just focus on getting this PR through

I opened #683 to discuss further

adriangb:
@owais I merged in main. Could we move forward with this? Thanks.

adriangb:
thanks for updating the branch 😄

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
adriangb: Checks added in 8eda7b6916989b74d03c9d3f506532e8b833e614

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
owais: and this works for both 2 and 3?

adriangb: the `_handle_exception` also exists in Falcon 2, so it seems to work 😃

owais (approved):

owais:
LGTM but giving others time to review as well.

lzchen (approved):

owais:
@adriangb please resolve conflcits and we'll merge this.

lzchen:
@adriangb 

Are you interested in becoming a CODEOWNER for the falcon instrumentation package? The responsibilities are outlined [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md#expectations-from-contributors)

adriangb:
> @adriangb please resolve conflcits and we'll merge this.

Done 😄 , let's let the tests run and make sure everything is still ✅ 

> Are you interested in becoming a CODEOWNER for the falcon instrumentation package?

At this point in time I have enough OSS work on my hands, but I will certainly consider it. I appreciate the offer.

PR #655, opened 2021-09-02
Add component owner action

lzchen:
Assigns and requires review from owners of components on PRs that are against those components.

Adds @NathanielRN  as the first code owner for aws sdk extension component.


NathanielRN (approved):
🚀

PR #656, opened 2021-09-02
Update Doc for aws propagator default sampling behavior

NathanielRN:
# Description

By default the `TraceProvider` uses a Sampler which respects the Parent Sampling decision. If the user does not provide `Sampled=1;` into their `X-Amzn-Trace-Id` header when using the `AwsXrayFormat` propagator, the parent context (and all subsequent child spans) are assumed to be `sampled=false` by default.

This PR updates the docs to warn users about that.

Fixes #649

## Type of change

Please delete options that are not relevant.

- [x] This change requires a documentation update

# How Has This Been Tested?

N/A

# Does This PR Require a Core Repo Change?

- [x] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [x] Followed the style guidelines of this project
~- [ ] Changelogs have been updated~
~- [ ] Unit tests have been added~
- [x] Documentation has been updated

PR #658, opened 2021-09-02
Update Doc for aws propagator default sampling behavior

willarmiros:
# Description

Dupe of #656, for also testing CODEOWNERS

### DIFF
--- a/sdk-extension/opentelemetry-sdk-extension-aws/README.rst
+++ b/sdk-extension/opentelemetry-sdk-extension-aws/README.rst
@@ -64,6 +64,11 @@ Or by setting this propagator in your instrumented application:
     from opentelemetry.sdk.extension.aws.trace.propagation.aws_xray_format import AwsXRayFormat
 
     set_global_textmap(AwsXRayFormat())
+    
+**NOTE**: Because the parent context parsed from the `X-Amzn-Trace-Id` header
+assumes the context is _not_ sampled by default, users should make sure to add
+`Sampled=1` to their `X-Amzn-Trace-Id` headers so that the child spans are
+sampled.
 
 Usage (AWS Resource Detectors)
 ------------------------------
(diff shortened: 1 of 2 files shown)


owais (approved):

NathanielRN (approved):

NathanielRN:
Hm interesting, maybe because I am part of the commit history I didn't get set a reviewer even though I am a CODEOWNER? But I can still give an approving review?

owais:
May be. I think we should update the check so that it is only required when the author is not a code-owner.

NathanielRN:
Also it looks like [the "Component Owners" test](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/658/checks?check_run_id=3500599381) is failing 😕 @lzchen 

```
Run dyladan/component-owners@main
  with:
    assign-owners: true
    request-owner-reviews: true
    repo-token: ***
    config-file: .github/component_owners.yml
1 owners found NathanielRN
Adding assignees
Error: Resource not accessible by integration
```

willarmiros:
Yeah sorry, totally didn't think of that when giving you the co-author @NathanielRN

NathanielRN:
No worries at all @willarmiros ! I think it's useful to see what this action does. We know the CODEOWNER works and the test passes from #657, but we should figure out the change in this PR.

@lzchen had a good idea that this is because it's from your fork wheras #657 was from an upstream branch. I remember seeing the `Error: Resource not accessible by integration` a lot such as [in this GitHub Issue](https://github.com/actions/first-interaction/issues/10#issuecomment-545576314) when doing operations on the repo.

codeboten (approved):

NathanielRN:
Closing in favor of #656 now that we checked the CODEOWNERs tool is working as expected.

codeboten:
> Closing in favor of #656 now that we checked the CODEOWNERs tool is working as expected.

I... this was really confusing. The file diff made it look like this was being added in a different place lol

NathanielRN:
> I... this was really confusing. The file diff made it look like this was being added in a different place lol

Ah really sorry about that! Testing CODEOWNERs got in the middle of me deciding the change would fit better at a different point in the docs 😓 I really appreciate your green check mark! Especially since it helped me realize I should've closed this one immediately once #656 got merged but better late than never I guess 😅

PR #659, opened 2021-09-02
Change action to address forks

lzchen:


NathanielRN (approved):
Nice! I guess we do have to be careful running code from works in this workflow from now on, [based on GitHub's warning for this event](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target):

> Warning: The pull_request_target event is granted a read/write repository token and can access secrets, even when it is triggered from a fork. Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event. Additionally, any caches share the same scope as the base branch, and to help prevent cache poisoning, you should not save the cache if there is a possibility that the cache contents were altered. For more information, see "Keeping your GitHub Actions and workflows secure: Preventing pwn requests" on the GitHub Security Lab website.

lzchen:
@NathanielRN 

Right, we should be providing our utmost attention to security flaws when reviewing PRs anyways =p

PR #661, opened 2021-09-06
Add net.peer.ip in requests & urllib3 instrumentations.

Oberon00:
# Description

Capture net.peer.ip in requests & urllib3 instrumentations, using a shared new ~~package `opentelemetry-instrumentation-http-base`~~ instrumentor for httplib.client that could also be used to easily add net.peer.ip to other instrumentations based on the stdlib's httplib.client (but this PR starts with only the aforementioned two).

## Type of change

# How Has This Been Tested?

Unit tests included. Due to how sharing of test code is currently set up, this requires a change to the main repo to add the test base class.

- [X] Yes. - Link to PR: https://github.com/open-telemetry/opentelemetry-python/pull/2101

# Checklist:

-  ~~Documentation has been updated~~


Oberon00:
What's up with these pylint errors? ` instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_instrumentation.py:133:16: E1101: Instance of 'TestBotocoreInstrumentor' has no 'get_finished_spans' member (no-member)`

lzchen:
Any chance we can add this functionality to [http utils](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/util/opentelemetry-util-http)? We already have that package to contain common http functionality.

Oberon00:
I think that should not be a technical problem. It will add a dependency on asgiref though and so far util-http only instruments server-side libs. If you prefer I can merge the new http-base package into util-http.

Oberon00:
@lzchen I'll go ahead and change the package.

Oberon00:
@lzchen Done. Instead of adding a new package, I now extend the util-http package, like you suggested.

.github/workflows/test.yml
```diff
@@ -6,7 +6,7 @@ on:
     - 'release/*'
   pull_request:
 env:
-  CORE_REPO_SHA: c49ad57bfe35cfc69bfa863d74058ca9bec55fc3
+  CORE_REPO_SHA: f4b9ac20023395e077f47831a04a1a6b473c3406 # TODO: Update to main
```
lzchen, line 9: Nit: TODO not needed I believe.

CHANGELOG.md
```diff
@@ -20,6 +20,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-instrumentation-urllib3` Updated `_RequestHookT` with two additional fields - the request body and the request headers
 ([#660](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/660))
 
+### Added
+
+- `opentelemetry-instrumentation-urllib3`, `opentelemetry-instrumentation-requests`
+  The `net.peer.ip` attribute is set to the IP of the connected HTTP server or proxy
+  using the new `opententelemetry-instrumentation-http-base` package
```
lzchen, line 27: No new package anymore.

util/opentelemetry-util-http/src/opentelemetry/util/http/httplib.py
```diff
@@ -0,0 +1,175 @@
...
+from opentelemetry.instrumentation.utils import unwrap
+from opentelemetry.semconv.trace import SpanAttributes
+from opentelemetry.trace.span import Span
+
+_STATE_KEY = "httpbase_instrumentation_state"
+
+logger = logging.getLogger(__name__)
+
+
+class HttpClientInstrumentor(BaseInstrumentor):
```
lzchen, line 40: I'm not too sure how this is supposed to be used by the user. Is this `Instrumentor` instantiated as well along side `RequestsInstrumentor`? Or is `RequestsInstrumentor` supposed to extend this class?

lzchen:
@Oberon00 

>  It will add a dependency on asgiref though

Sorry I might be missing something. Where is the dependency for `asgiref` in this library?

Oberon00:
@lzchen 

> Sorry I might be missing something. Where is the dependency for asgiref in this library?

The code in this PR does not depend on asgiref, but the util-http package does.

util/opentelemetry-util-http/src/opentelemetry/util/http/httplib.py
Oberon00: Oh, actually now that you say it, this was the reason why it was in a separate package...

This is not supposed to be used by the user at all. Instead it should just be loaded by the instrumentation infrastructure and then it does its thing behind the scentes. The requests and urllib3 instrumentations are the actual users of this code, since they register their wish to have the IP set onto a span which this instrumentation fulfills once the standard library httplib opens a socket.

Oberon00: I think I have to add an entrypoint to the setup.cfg of util-http as well as the other instrumentation boilerplate. Do you think it would be better to keep it as separate package?

.github/workflows/test.yml
Oberon00: Aren't PRs squashed when merging? In that case, a new commit ID is generated.

lzchen:
@Oberon00 

> The code in this PR does not depend on asgiref, but the util-http package does.

util/opentelemetry-util-http/src/opentelemetry/util/http/httplib.py
lzchen: I don't mind having an entrypoint for client libraries, unless you feel strongly having another package. I personally don't like having many packages :)

.github/workflows/test.yml
lzchen: Right, but this SHA is used only for testing. So as long as a commit in core does not cause a build failure in this tox build, this SHA doesn't need to be updated. Actually, I don't believe it needs to be updated in this PR either, because your change isn't dependent on any new commits in core.

Oberon00:
@lzchen You are right, seems I somehow looked in the wrong place 😄

util/opentelemetry-util-http/src/opentelemetry/util/http/httplib.py
Oberon00: > How come we need to create a new instrumentor for this feature?

Because this instruments the standard library httplib.client. I could put it in the urllib3 instrumentation, but if we ever want to instrument another library based on httplib.client (e.g. there is a urllib instrumentation here that could also be enhanched with this), then where should we put it if not in a separate, independent instrumentor?

> Do all client libraries have a "send" method?

No, just those based on httplib.client, which appies to requests and urllib3 at least.

Oberon00: So maybe the point to emphasize here: This is *not* a completely generic HTTP IP detection (at least yet). It is a instrumentation for httplib.client except that it does not create spans on its own but cooperates with other instrumentations that instrument libs which use httplib.client internally.

.github/workflows/test.yml
Oberon00: As mentioned in the PR description, the test code depends on https://github.com/open-telemetry/opentelemetry-python/pull/2101 (httpretty does not provide an IP address)

util/opentelemetry-util-http/src/opentelemetry/util/http/httplib.py
Oberon00: > If not, maybe we can just have trysetip and set_ip_on_next_http_connection as utility functions, and decorate the request functions directly?

Sorry, I don't follow this. What do you mean by request functions I should decorate? The IP cannot be obtained from the API functions we usually instrument since there is no access to the underlying socket, or accessing it would be very complicated.

util/opentelemetry-util-http/setup.cfg
```diff
@@ -40,3 +40,7 @@ packages=find_namespace:
 
 [options.packages.find]
 where = src
+
+[options.entry_points]
+opentelemetry_instrumentor =
```
Oberon00, line 45: Compared to the original PR version with a separate package for this:
On the minus side, this now means that everyone depending on util-http will get the httplib.client instrumentation, even if they don't use it. The overhead should be small though. On the plus side, there is no new package.

Oberon00, line 45: BTW, is there some integration test that loads the urllib3 or requests instrumentation via the entrypoint, that I could extend to assert on the net.peer.ip attribute being present, so I could verify the wiring is not messed up here?

.github/workflows/test.yml
lzchen: Right, I just meant that the `TODO: Update to main` is not needed. I don't believe we need to update the SHA to main even after this is merged.

util/opentelemetry-util-http/src/opentelemetry/util/http/httplib.py
lzchen: > The requests and urllib3 instrumentations are the actual users of this code,

Sorry I might be missing something. I'm not understanding how the two instrumentations actually use `HttpClientInstrumentor`? Even the tests have:

```
HttpClientInstrumentor().instrument()
URLLib3Instrumentor().instrument()
```
so it's a bit unclear in how these are supposed to be used.

util/opentelemetry-util-http/setup.cfg
lzchen: The benefit of not having a new package outweighs the small overhead I feel :)

lzchen: Just to clarify, you are wanting to test if attribute is present if auto-instrumentation is used? If so, I don't believe we have integration tests for those. We currently only have [unit tests](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-instrumentation/tests/test_run.py) for auto-instrumentation.

util/opentelemetry-util-http/src/opentelemetry/util/http/httplib.py
Oberon00: > Currently, it looks like only set_ip_on_next_http_connection is being used in the instrumentations.

Exactly, this is the *only* feature this new instrumentation provides. It sets net.peer.ip to the actually connected-to IP (vs some separate DNS lookup which will usually not yield the same IP again)

Oberon00: > Even the tests have:
> ```
> HttpClientInstrumentor().instrument()
> URLLib3Instrumentor().instrument()
> ```

Yes. The tests need to call `HttpClientInstrumentor().instrument()` manually since no distribution (is it called that way) is loaded that auto-enables the instrumentation. `URLLib3Instrumentor().instrument()` cannot call `HttpClientInstrumentor().instrument()` IIUC, as that would lead to double instrumentation. Also, I wanted to have the feature optional. If users disable the HttpClientInstrumentor (not sure if this is possible today), they won't get net.peer.ip and potential associated overhead and bugs.

.github/workflows/test.yml
Oberon00: Now that the core PR is merged, this point is moot, but it would have seemed odd to point to a commit that is not on any branch anywhere (after PR merge). Technically, it wouldn't have been a real problem as AFAIK GH keeps commits that were visible on PRs indefinitely, but still.

```suggestion
  CORE_REPO_SHA: d9c22a87b6bfc5ec332588c764f82c32f068b2c3
```

Oberon00:
Is there anything I can do to move this forward?

util/opentelemetry-util-http/src/opentelemetry/util/http/httplib.py
codeboten: Just confirming here. As a user, if I manually instrument code, and load the `RequestsInstrumentor`, I will also need to manually load the `HttpClientInstrumentor` myself correct?

Would it be possible to call `trysetip` from the requests/urlib3 code and skip the additional instrumentor? Or is the problem that those libraries don't have access to the underlying socket to pull IP information from it?

codeboten (commented):
The code looks good, just one clarifying question regarding the flow of how the instrumentation is used.

util/opentelemetry-util-http/src/opentelemetry/util/http/httplib.py
Oberon00: Yes, the main reason was that its not easy to get the underlying socket. I did not check for every library separately, in some it may be possible to use some undocumented properties or instrument some constructors to get to the socket. Instrumenting httplib.client, although also awkward, seemed like the most clean way.

codeboten: Got it, that makes sense. Thanks!

codeboten (approved):
👍

lzchen (approved):

util/opentelemetry-util-http/src/opentelemetry/util/http/httplib.py
Oberon00: @codeboten To follow up here: Is manually using the instrumentors common enough that it would make sense to add a hint to use HttpClientInstrumentor to the READMEs of the requests and urllib3 instrumentations?

PR #671, opened 2021-09-09
Remove upper bound for dependency requirements

### LINKED ISSUE #668: Allow usage of httpx 0.19.*
httpx 0.19.0
opentelemetry-instrumentation-httpx 0.24b0

**Steps to reproduce**
Start the application.

No error messages.

`DependencyConflict: requested: "httpx >= 0.18.0, < 0.19.0" but found: "httpx 0.19.0"`

I'm wondering why opentelemetry-instrumentation-httpx is configured to support only 0.18.* and not anything above 0.19.*
If there are no concerns, could it allow 0.19.* as well?

https://github.com/open-telemetry/opentelemetry-python-contrib/blame/97e9f2f9ef6c775108ab30946a8668f64f37a4d9/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/package.py#L16

(Sorry. I opened it as a bug, but this is not a real bug).

ocelotl:
Fixes #668

### DIFF
--- a/instrumentation/README.md
+++ b/instrumentation/README.md
@@ -15,7 +15,7 @@
 | [opentelemetry-instrumentation-fastapi](./opentelemetry-instrumentation-fastapi) | fastapi ~= 0.58 |
 | [opentelemetry-instrumentation-flask](./opentelemetry-instrumentation-flask) | flask >= 1.0, < 3.0 |
 | [opentelemetry-instrumentation-grpc](./opentelemetry-instrumentation-grpc) | grpcio ~= 1.27 |
-| [opentelemetry-instrumentation-httpx](./opentelemetry-instrumentation-httpx) | httpx >= 0.18.0, < 0.19.0 |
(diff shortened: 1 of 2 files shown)


owais (approved):

jouve:
the limitation is enabled because of the handle_request API change in https://github.com/encode/httpx/blob/master/CHANGELOG.md#0200-13th-october-2021

lzchen:
@ocelotl 
Is this PR still needed?

jouve:
I think it would be better to update instrumentation-httpx to the new httpx API and make it depend on httpx >= 1.19 instead of 1.18

owais:
When was 1.20 released? Is it too complex to support <1.20?

ocelotl:
@lzchen it is an old draft, I honestly don't remember :sweat_smile: 

Closing to avoid confusion, if others have a solution for the corresponding issue it would be better to open a new PR :v:

PR #634, opened 2021-08-27
Make botocore instrumentation (mostly) compliant to semconv

mariojonke:
# Description

This PR adapts the span attributes of the botocore (AWS SDK) instrumentation to make them (mostly) comply with the [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-sdk.md) defined in the spec.

All spans created with the botocore instrumentation will generally set the [common attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-sdk.md#common-attributes) for the invoked AWS service and operation.

To handle service specifics, `AwsSdkExtension`s are introduced. An extension is a set of callbacks that get invoked for a traced AWS SDK call. A concrete extension might extract additional attributes before/after the service call, inject the span into a certain carrier, do some additional error handling, ...

Following extensions are added:
* DynamoDB: extracts the request and response attributes for the operations defined in the [AWS SDK spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-sdk.md#dynamodb) and follows the database semantic conventions for [connection](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#connection-level-attributes) and [call-level](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes) attributes.
* Lambda:  follows the semantic conventions for [outgoing invocations](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/faas.md#outgoing-invocations).
* SQS: only extracts the `aws.queue_url` which does not comply with the spec and was kept for backward compatibility.

General span attributes `aws.region`, `aws.request_id`, `retry_attempts`, for which no semantic convetions exist, were kept for backward compability.

Also fixes some issues:
* Lambda invoke: injects now the actual botocore span instead of its parent
* uninstruments now also `Endpoint.prepare_request` which was missing before

## Type of change

- [x] Breaking change: breaks existing span attribute names to make them spec compliant

# How Has This Been Tested?

New unit tests + adapting existing tests

# Does This PR Require a Core Repo Change?

- [x] No.

# Checklist:

[...]

PR #674, opened 2021-09-10
botocore: Make common span attributes compliant with semconv in spec

mariojonke:
# Description

This PR changes common span attributes for an AWS SDK call to make them comply with the semantic conventions in the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-sdk.md#common-attributes):
*  adds `rpc.system`
* changes `aws.operation` -> `rpc.method`
* changes `aws.service` -> `rpc.service`

Attributes for which no semantic conventions exist are kept unchanged. These are:
* `aws.region`
* `aws.request_id`
* `retry_attempts`

Simplifies unit tests as most of the assertions follow a similar pattern.

## Type of change

- [x] Breaking change (changes semantic conventions of existing attributes)

# How Has This Been Tested?

Existing tests have been adapted

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/types.py
```diff
@@ -0,0 +1,45 @@
+from typing import Any, Dict
+
+from opentelemetry.trace import SpanKind
+
+BotoClientT = "botocore.client.BaseClient"
+
+OperationParamsT = Dict[str, Any]
+
+
+class AwsSdkCallContext:
```
owais, line 10: Are users supposed to use these objects directly in code or is this strictly for internal instrumentation use? For the later, I'd suggest o make the private with leading underscore.

mariojonke: currently they are only (intended to be) used internally by the instrumentation. maybe in the future, some could be made available for public use, e.g. by request/response hooks.
i'll make them private for now as suggested

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
```diff
@@ -118,12 +122,12 @@ def _uninstrument(self, **kwargs):
...
+    def _is_lambda_invoke(call_context: _AwsSdkCallContext):
         return (
-            service_name == "lambda"
-            and operation_name == "Invoke"
-            and isinstance(api_params, dict)
-            and "Payload" in api_params
+            call_context.service == "lambda"
+            and call_context.operation == "Invoke"
+            and isinstance(call_context.params, dict)
+            and "Payload" in call_context.params
```
owais, line 130: curious if there is a case where service is "lambda", operator is "Invoke" but params is not a dict or "Payload" is not present in params. Can that happen? If it can, why don't we want to consider it a "lambda invoke"? If it can't, why check for params at all?

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
```diff
@@ -143,31 +147,35 @@ def _patched_api_call(self, original_func, instance, args, kwargs):
         if context_api.get_value(_SUPPRESS_INSTRUMENTATION_KEY):
             return original_func(*args, **kwargs)
 
-        # pylint: disable=protected-access
-        service_name = instance._service_model.service_name
-        operation_name, api_params = args
+        call_context = _determine_call_context(instance, args)
```
owais, line 150: nit: a comment here explaining why/when this can be None would be nice

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
```diff
@@ -176,46 +184,63 @@ def _patched_api_call(self, original_func, instance, args, kwargs):
...
+        call_context = _AwsSdkCallContext(client, operation, params)
+
+        logger.debug(
+            "AWS SDK invocation: %s %s",
+            call_context.service,
+            call_context.operation,
+        )
+
+        return call_context
+    except Exception as ex:  # pylint:disable=broad-except
```
owais, line 244: this is absolutely required? would be better if we caught specific exceptions here instead of everything.

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
```diff
@@ -176,46 +184,63 @@ def _patched_api_call(self, original_func, instance, args, kwargs):
...
+        logger.warning("Error when initializing call context", exc_info=ex)
```
owais, line 245: wondering if this should be an error instead of warning as I assume this should never happen and is there only to shout very loudly so users report the bug

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/types.py
```diff
@@ -0,0 +1,45 @@
+from typing import Any, Dict
+
+from opentelemetry.trace import SpanKind
+
+_BotoClientT = "botocore.client.BaseClient"
+
+_OperationParamsT = Dict[str, Any]
+
+
+class _AwsSdkCallContext:
```
owais, line 10: could this be a named tuple instead?

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
mariojonke: I think the name of the method is misleading as it actually checks if it is a lambda invoke operation and if the trace context can be injected into the `Payload` (the code already existed before so not sure).

From how botocore is implemented the `params` object should always be a dictionary. It contains all the parameters that are passed to the boto3/botocore API method. So if no `Payload` argument is passed to `lambda.invoke(...)` it will not show up in the `params` object.

The idea is to move the lambda-specific tracing part into a separate extension like proposed in the [initial PR](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/634/files#diff-abc82a132764a6244c38e8e72e4b15c966c275c02aa13fdae8327c87ee5fe1ddR65) to extract the additional attributes that are defined in the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/faas.md#outgoing-invocations).

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/types.py
mariojonke: there might be some extensions later on (e.g. SQS) that want to return a different `span_kind` or `span_name`, so for that they would be required to update the fields on this object.

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
mariojonke: since this is trying to extract parameters from the internals of the botocore lib there could be unforeseen errors, e.g. when internals of botocore change, that wouldn't be handled here. This in turn would affect existing user code which is against one of the design principles of opentelemetry. So that is actually the reason for the broad except here.

mariojonke: well, i guess is certainly possible to use `getattr` and `hasattr` to safely access attributes but the question is then what to do if one of the attributes is actually missing. From what i can see there are 2 possibilities. 1) assign a default value or use None, or 2) throw a (typed) exception.
Option 1) is probably not a good idea since it will either produce incorrect data or fail at a later point in time when a `None` field is accessed by an extension.
Option 2) will clutter the code for extracting the attributes since each attribute needs to be checked if it exists and throw an exception if not. Additionally, i don't think that there is much benefit in this approach over broadly catching a possible exception. Since the try/catch block only covers the creation of `_AWSCallContext` there is imo no harm in catching the general exception as for creating the span an instance of `_AWSCallContext` is required in any case.

owais: IMO

- if something is not available expectedly (may or may not be set by design), it should be ignored and not recorded as an attribute.
- if something is not available unexpectedly (internal structure changed), it should again be ignored and not recorded as an attribute but in this case an additional warning should be logged so the issue surfaces up and we become aware of it.

owais: >  i don't think that there is much benefit in this approach over broadly catching a possible exception. Since the try/catch block only covers the creation of _AWSCallContext there is imo no harm in catching the general exception as for creating the span an instance of _AWSCallContext is required in any case.

One big downside is that if one attribute fails to be extracted, we'll not record anything. If we do it defensively per attribute, we'll still record what we can extract and ignore what we can't. IMO in the end, some telemetry is better than no telemetry for the end-user.

mariojonke: Updated the PR and most of the attributes in `_AWSCallContext` are now extracted defensively. Attributes `service` and `operation` are extracted the normal way though since instrumentation without them does not really make much sense.
I kept the broad exception as safety net to avoid breaking user code (e.g. if an extracted attribute is actually a property that might raise some unforeseen exception).

owais (approved):

lzchen (approved):

PR #516, opened 2021-05-26
Instrumentation meta package

owais:
# Description

Add a new `opentelemetry-instrumentation-all` metapackage that does not ship any code but pulls in all instrumentations maintained by us.

Depends on #514 

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [ ] Followed the style guidelines of this project
- [ ] Changelogs have been updated
- [ ] Unit tests have been added
- [ ] Documentation has been updated

PR #678, opened 2021-09-13
remove need to clone core repo

codeboten:
# Description

Companion PR to https://github.com/open-telemetry/opentelemetry-python/pull/2108, this removes the need to clone the core repo in order to run tox. It uses pip instead.

NOTE: This change also removes lint ignore rules for the `opentelemetry-python-core` subdirectory. This means after this PR, contributors should remove that folder from their local environment.


lonewolf3739 (approved):

owais (approved):

owais:
Generate for bootstrap needs to be updated. I guess we should move it to core and use something [like this](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/516) as source of truth for generate

I was planning to come back and finish the meta package PR next week but if it helps, feel free to take it over. https://github.com/open-telemetry/opentelemetry-python-contrib/pull/516

codeboten:
@owais yah that's the last thing i need to fix for this PR to be done. will take a look sometimes this week.

lzchen:
@owais 
Which files for bootstrap needs to be updated exactly? (or not, if we are changing it)

lzchen (approved):

owais:
@lzchen this file: https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap_gen.py

@codeboten I think I'll something that might help you with this today.

lzchen:
@owais 
The `gen` file has to be updated every release. Is that related to this PR though?

lzchen:
@owais 
Oh I guess `opentelemetry-python-core` needs to be checked out to run `eachdist.py` for generating?

owais:
@codeboten  see if this helps: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/681/files#diff-02fe7adf3ca73fd04ef2811691de61fbc8ae90a52016caa844922c4b811d9e8dR29

You can install the egg from git like other packages as a dev dependency and just copy over the `__init__.py` file from the metapackage.

owais:
> Oh I guess opentelemetry-python-core needs to be checked out to run eachdist.py for generating?

Right. That's how it works today but with the changes @codeboten is making, it shouldn't be necessary going forward.

codeboten:
@owais please take another look. I've updated `generate_instrumentation_bootstrap.py` to:

- pull down `bootstrap_gen.py` from the core repo via `requests` and store it in a temp folder
- generate the new `bootstrap_gen.py` and compare the two files
- if they differ, the script returns an error, which will cause CI to fail, rather than running `git diff` as was done before

Let me know what you think.

owais:
IIUR, this will tell the author that bootstrap_gen in core needs an update and then the author will have to go and create a PR in core by manually copying/generating the file. Is that right?

owais:
It would be great if a bot could create PR in core repo every time the file changed in contrib main but I'm working on a proposal related to distros that might help avoid this completely.

codeboten:
> IIUR, this will tell the author that bootstrap_gen in core needs an update and then the author will have to go and create a PR in core by manually copying/generating the file. Is that right?

@owais yup that's correct. it's not great, but it seems the benefits of removing the core repo from regular dev workflow outweighs the burden to manually copy this file when necessary. I like the idea of removing the need for this altogether though, that sounds promising! Thoughts?

owais:
Definitely! I wasn't against it. Just trying to verify if I understood the new flow properly. We should just document the workflow here so outside contributors know what they are supposed to do to fix the CI failure.

scripts/generate_instrumentation_bootstrap.py
```diff
@@ -50,16 +52,17 @@
...
-    "opentelemetry-python-core",
-    "opentelemetry-instrumentation",
-    "src",
-    "opentelemetry",
-    "instrumentation",
-    "bootstrap_gen.py",
+current_path = os.path.join(tmpdir.name, "current.py",)
+
+core_repo = os.getenv("CORE_REPO_SHA", "main")
+url = "https://raw.githubusercontent.com/open-telemetry/opentelemetry-python/{}/opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap_gen.py".format(
```
ocelotl, line 61: Better to make this a f-string

ocelotl (approved):

ocelotl:
@codeboten, please check `lint`.

codeboten:
@ocelotl fixed the lint errors

PR #679, opened 2021-09-14
botocore hooks

ItayGibel-heliosphere:
# Description

`opentelemetry-instrumentation-botocore`: added request_hooks response_hooks  dictionaries passed as arguments to the instrument method.

These dictionaries contain a mapping between AWS service names and callable objects that will be invoked before/after an api call to the respective service

## Type of change

# How Has This Been Tested?

added a new unit test, testing the hooks activation

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
```diff
@@ -44,6 +44,40 @@
...
+
+    def ec2_request_hook(span, operation_name, api_params):
+        # request hook logic
+
+    def ec2_response_hook(span, operation_name, result):
+        # response hook logic
+
+    # Instrument Botocore with hooks
+    BotocoreInstrumentor().instrument(
+        request_hooks={"ec2": ec2_request_hook}, response_hooks={"ec2": ec2_response_hook}
```
owais, line 71: This is deviating a bit too much from other instrumentation packages. How important is to map hooks to sub-components? Is there a very compelling use case for this?

What if we just accepting a single request/response hook and passed the "service name" as an argument to it? Users would then have some logic branching based on the service name. I'd imagine it wouldn't make it too terrible to implement hooks that handle different services in different ways but it'd simplify it for the common case where people just want a simple, single hook for a given instrumentation package. What do you think?

ItayGibel-heliosphere: I think you are right

My idea here was to let the users pass hooks, without the need to write the logic of branching based on the service name, but since they anyway have to create a dictionary with a mapping between service names to the respective code, it won't require much more work to write this logic branching.

I'll change it to a single request hook and single response hook, to make it consistent with the other libraries

ItayGibel-heliosphere:
@owais - I've applied the change you requested, can you please have a look?
Thanks

owais (approved):

PR #684, opened 2021-09-17
update open calls to pass encoding

codeboten:
# Description

Contrib repo counterpart to https://github.com/open-telemetry/opentelemetry-python/issues/2126


lonewolf3739 (approved):

sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/beanstalk.py
```diff
@@ -42,7 +42,7 @@ def detect(self) -> "Resource":
             conf_file_path = "/var/elasticbeanstalk/xray/environment.conf"
 
         try:
-            with open(conf_file_path) as conf_file:
+            with open(conf_file_path, encoding="utf-8") as conf_file:
```
aabmass, line 45: This seems to be the only one in actual code. @NathanielRN is this fine with you since its in AWS code?

NathanielRN: Yes this should be fine! Looks like this is [the encoding we use on the JavaScript Resource Detectors](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/1b7ae9c3c249cd1166b610e2a4c95c01fec93d1c/detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsBeanstalkDetector.ts#L65-L68).

```js
const rawData = await AwsBeanstalkDetector.readFileAsync(
    this.BEANSTALK_CONF_PATH,
    'utf8'
);
```

aabmass (approved):

lzchen (approved):

PR #693, opened 2021-09-21
use f-strings instead of format

### LINKED ISSUE #692: use f-strings wherever possible

codeboten:
# Description

Continuing work to support pylint 2.11.0, addressing string formatting in code.

Fixes #692

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
+++ b/instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
@@ -349,10 +349,7 @@ class TestMiddleware(TestBase, WsgiTestBase):
         )
         self.assertEqual(
             response.headers["traceresponse"],
-            "00-{0}-{1}-01".format(
-                format_trace_id(span.get_span_context().trace_id),
-                format_span_id(span.get_span_context().span_id),
-            ),
+            f"00-{format_trace_id(span.get_span_context().trace_id)}-{format_span_id(span.get_span_context().span_id)}-01",
         )
         self.memory_exporter.clear()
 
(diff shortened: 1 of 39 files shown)


lzchen (approved):

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
```diff
@@ -349,10 +349,7 @@ def test_trace_response_headers(self):
         )
         self.assertEqual(
             response.headers["traceresponse"],
-            "00-{0}-{1}-01".format(
```
aabmass, line 352: Cases like these where you don't have short variable, f-strings can be pretty ugly :(

codeboten: I hear that... I don't feel strongly either ways. I found the formatting previously there confusing as well w/ the multiple 0's and 1's 😄 

Could create a temporary variable or just disable pylint for this line.

ocelotl (approved):

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
alertedsnake: It's probably more readable when it's not complex code buried in a string - I prefer the format one.

owais (approved):

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
aabmass: > I found the formatting previously there confusing as well w/ the multiple 0's and 1's

I agree, maybe short variable names with f-strings would be better here, or could do `"00-{}-{}-01".format(...)` or named format params. I'll leave it up to you, don't want to block this @codeboten

codeboten: i'm happy to fix this in a following PR. I will refactor all the tests that are duplicated into a single test in the base class.

codeboten: will use the following method to simplify the code https://github.com/open-telemetry/opentelemetry-python/pull/2159

PR #706, opened 2021-10-01
Tornado attributes

remram44:
# Description

* Set `http.client_ip` instead of `net.peer.ip`: the value is changed by Tornado to `X-Forwarded-For` or `X-Real-IP`, see [Tornado code](https://github.com/tornadoweb/tornado/blob/790715ae0f0a30b9ee830bfee75bb7fa4c4ec2f6/tornado/httpserver.py#L332-L352), see [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/b46bcab5fb709381f1fd52096a19541370c7d1b3/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions)
    * The actual `net.peer.ip` is in `_orig_remote_ip`, however I thought it best not to read a private field. Let me know if it's ok.
* Set `handler` to the fully-qualified name of the request handler class
    * The Tornado instrumentation does not provide any route information, such as `http.route`, only the base handler class's name as the span name. I didn't find an easy way to read the route pattern, the name of the handler seems like the next best thing

## Type of change

# How Has This Been Tested?

- [x] `tox -e py38-test-instrumentation-tornado` (`TestTornadoInstrumentation` updated)

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
@@ -249,6 +249,11 @@ def _get_operation_name(handler, request):
     return f"{class_name}.{request.method.lower()}"
 
 
+def _get_full_handler_name(handler):
+    klass = type(handler)
+    return f"{klass.__module__}.{klass.__qualname__}"
+
+
 def _start_span(tracer, handler, start_time) -> _TraceContext:
     token = context.attach(extract(handler.request.headers))
 
@@ -261,6 +266,7 @@ def _start_span(tracer, handler, start_time) -> _TraceContext:
         attributes = _get_attributes_from_request(handler.request)
         for key, value in attributes.items():
             span.set_attribute(key, value)
+        span.set_attribute("handler", _get_full_handler_name(handler))
 
     activation = trace.use_span(span, end_on_exit=True)
     activation.__enter__()  # pylint: disable=E1101
--- a/instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
@@ -235,9 +235,6 @@ def _get_attributes_from_request(request):
         SpanAttributes.HTTP_TARGET: request.path,
     }
 
-    if request.host:
-        attrs[SpanAttributes.HTTP_HOST] = request.host
-
     if request.remote_ip:
         attrs[SpanAttributes.NET_PEER_IP] = request.remote_ip
 
--- a/instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
@@ -236,7 +236,7 @@ def _get_attributes_from_request(request):
     }
 
     if request.remote_ip:
-        attrs[SpanAttributes.NET_PEER_IP] = request.remote_ip
+        attrs[SpanAttributes.HTTP_CLIENT_IP] = request.remote_ip
 
     return extract_attributes_from_object(
         request, _traced_request_attrs, attrs
--- a/instrumentation/opentelemetry-instrumentation-tornado/tests/test_instrumentation.py
+++ b/instrumentation/opentelemetry-instrumentation-tornado/tests/test_instrumentation.py
@@ -143,7 +143,7 @@ class TestTornadoInstrumentation(TornadoTest):
                 SpanAttributes.HTTP_HOST: "127.0.0.1:"
                 + str(self.get_http_port()),
                 SpanAttributes.HTTP_TARGET: "/",
-                SpanAttributes.NET_PEER_IP: "127.0.0.1",
+                SpanAttributes.HTTP_CLIENT_IP: "127.0.0.1",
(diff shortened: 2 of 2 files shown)


instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
```diff
@@ -252,6 +249,11 @@ def _get_operation_name(handler, request):
     return f"{class_name}.{request.method.lower()}"
 
 
+def _get_full_handler_name(handler):
+    klass = type(handler)
+    return f"{klass.__module__}.{klass.__qualname__}"
```
owais, line 254: would this work fine for dynamically generated handlers within a function? I know that is not very common but sounds like an edge case. Any concerns regarding that?

owais:
>     The actual net.peer.ip is in _orig_remote_ip, however I thought it best not to read a private field. Let me know if it's ok.

that is OK as long as we do it defensively

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
remram44: I believe `__qualname__` would do the right thing with closures, see the [examples in PEP 3155](https://www.python.org/dev/peps/pep-3155/#example-with-nested-functions)

remram44:
I added reading `net.peer.ip` from `_orig_remote_ip`, if it's present

Off-topic: I see the hacktoberfest topic is set on some open-telemetry repos but not here, I'm not sure if it's on purpose :wink:

owais:
Last time there was a "fest" we had a few dozen people open PRs to add new lines in markdown files to earn some free merch 😂 is this similar?

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
```diff
@@ -235,11 +235,12 @@ def _get_attributes_from_request(request):
...
-    if request.host:
-        attrs[SpanAttributes.HTTP_HOST] = request.host
-
     if request.remote_ip:
-        attrs[SpanAttributes.NET_PEER_IP] = request.remote_ip
+        attrs[SpanAttributes.HTTP_CLIENT_IP] = request.remote_ip
+        if getattr(request, "_orig_remote_ip"):
+            attrs[SpanAttributes.NET_PEER_IP] = request._orig_remote_ip
+        else:
+            attrs[SpanAttributes.NET_PEER_IP] = request.remote_ip
```
owais, line 243: nit: Isn't this already captured above as HTTP client IP? May be we can skip it here and only capture when `_orig_remote_ip` is present.

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
```diff
@@ -235,11 +235,12 @@ def _get_attributes_from_request(request):
         SpanAttributes.HTTP_TARGET: request.path,
     }
 
-    if request.host:
-        attrs[SpanAttributes.HTTP_HOST] = request.host
-
     if request.remote_ip:
-        attrs[SpanAttributes.NET_PEER_IP] = request.remote_ip
```
owais, line 242: Please add a comment here for future contributors describing why we assign these to different attributes.

owais (approved):
LGTM. Please take care of the comments.

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
remram44: Sure I don't mind. `_orig_remote_ip` is always set in practice, in current versions.

remram44:
Done!

hacktoberfest is usually ok in my experience, last year really did not go well. I believe they have made some changes to the procedure this year but I can't vouch for the masses.

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
```diff
@@ -235,11 +235,13 @@ def _get_attributes_from_request(request):
...
+        # NET_PEER_IP is the address of the network peer
+        # HTTP_CLIENT_IP is the address of the client, which might be different
+        # if Tornado is set to trust X-Forwarded-For headers (xheaders=True)
+        attrs[SpanAttributes.HTTP_CLIENT_IP] = request.remote_ip
+        if getattr(request, "_orig_remote_ip"):
```
lonewolf3739, line 243: Missing a default here will raise `AttributeError`.
```suggestion
        if getattr(request, "_orig_remote_ip", None):
```

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
```diff
@@ -264,6 +271,7 @@ def _start_span(tracer, handler, start_time) -> _TraceContext:
         attributes = _get_attributes_from_request(handler.request)
         for key, value in attributes.items():
             span.set_attribute(key, value)
+        span.set_attribute("handler", _get_full_handler_name(handler))
```
lonewolf3739, line 274: nit: prefix the attribute key with `tornado`?

remram44:
(this should take care of test failures)

lonewolf3739 (approved):

PR #712, opened 2021-10-04
Expand allowed versions for jinja2 instrumentation

andrew-matteson:
# Description

The Jinja2 API hasn't changed the functions that have been wrapped in at least 8 years.
The version supported should be much more broad.

Specifically, 
`
jinja2.environment.Template.render
jinja2.Template.generate
jinja2.Environment.compile
jinja2.Environment._load_template
`
which are the only wrapped functions seem to be appropriate for the instrumentation wrappers without further changes.

I'm not a jinja2 expert, so it's possible I'm missing something. Happy to help with changes if it seems that I am.

## Type of change

# How Has This Been Tested?

This change produces useful traces in our system.

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-jinja2/src/opentelemetry/instrumentation/jinja2/package.py
+++ b/instrumentation/opentelemetry-instrumentation-jinja2/src/opentelemetry/instrumentation/jinja2/package.py
@@ -13,4 +13,4 @@
 # limitations under the License.
 
 
-_instruments = ("jinja2~=2.7",)
+_instruments = ("jinja2>=2.7",)


instrumentation/opentelemetry-instrumentation-jinja2/src/opentelemetry/instrumentation/jinja2/package.py
```diff
@@ -13,4 +13,4 @@
 # limitations under the License.
 
 
-_instruments = ("jinja2~=2.7",)
+_instruments = ("jinja2>=2.7",)
```
owais, line 16: Can we set it to `>=2.7,<4.0` so we don't automatically support a future version that might not work (4.0)?

andrew-matteson: Done. Thank you.

owais (approved):

andrew-matteson:
@owais If you can point me towards instructions about how to clear the remaining checks, I can take care of them tomorrow.

andrew-matteson:
@owais Can you let me know if there's anything I need to do to get the checks to run?

bootstrap_gen.py
```diff
@@ -0,0 +1,138 @@
+# Copyright The OpenTelemetry Authors
```
codeboten, line 1: @andrew-matteson this file should be committed [here](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap_gen.py). Apologies for the confusion, improving this is being actively worked on.

codeboten (changes requested):
I triggered the build just now by updating the branch, it appears to be running.

andrew-matteson:
@owais @codeboten Could you give me advice on how to proceed?

When I run `tox -e generate` locally, it generates a bootstrap file that doesn't have the `pika` section. The jinja2 changes are committed in the core repo. I'm not sure how the comparison between the files works, but if it's just checking that the generated file exactly matches what's in the core repo, I could see a scenario in which both the `pika` PR and this one are blocked, waiting for each other. Please let me know if there's anything I can do.

The performance regression in the most recent run seems spurious (adding a space to a comment is probably unrelated...)

owais:
> I'm having some challenges with tests + linting. The tests seem to be running with jinja2==2.11.3. The linting is running with jinja2==3.0.1. The linting fails because the test references, in a version specific way, a variable that stopped existing in 3.0.0. I get the error:

> E1101: Module 'jinja2.environment' has no '_spontaneous_environments' member (no-member)

Wait, doesn't this mean instrumentation will fail in Jinja2==3.0.1?

owais:
> @owais Nope. _spontaneous_environments is only referenced in the tests. This logic is used to prevent test drool due to jinja2's caching logic.

Yeah, I realized it right after posting the comment :)

owais (approved):
LGTM. Will merge if everything else other than docker-tests passes which is known to be broken right now.

PR #719, opened 2021-10-10
Bugfix/set default context in pika

oxeye-nikolay:
# Description

This PR fixes a bug in the pika instrumentation library. When the propagator could not detect a context from the message headers (for example in the sending action `properties.headers`) the context would be set to `{}` instead of the current context - thus breaking the trace chain.

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [x] Maunal Instrumentation test

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/utils.py
```diff
@@ -41,15 +41,20 @@ def decorated_callback(
     ) -> Any:
```
owais, line 41: So `_decorate_callback` wraps the "receive" functionality only?

PR #734, opened 2021-10-12
Fix RuntimeError: Working outside of request context.

benjaminsky:
# Description

Check for `flask.request` to avoid `RuntimeError: Working outside of request context.` error.

## Type of change

# How Has This Been Tested?

Validated locally.

- [x] No.

# Checklist:


owais:
Same question as @lonewolf3739. Is this happening because you are using the instrumentation outside a real request-response cycle (some mock/test env) or are we missing something here as instrumentation authors?

benjaminsky:
I ran into this issue instrumenting a flask app that’s also instrumented with New Relic. I think that’s the cause. I can dig into it further but it is a real environment. It’s an ancient version of flask too though - 0.12.1.

Ben Campbell
________________________________

PR #735, opened 2021-10-13
botocore: Add DynamoDB extension

mariojonke:
# Description

Adds an extension to extract additional request/response attributes for DynamoDB as defined in the [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-sdk.md#dynamodb) for the AWS SDK. Depending on which DynamoDB operation is called via the AWS SDK, different attributes are extracted.

To follow the database semantic conventions, relevant [connection](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#connection-level-attributes) and [call-level](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes) attributes are also extracted. Those attributes are:
* `db.system`: always 'dynamodb'
* `db.operation`: corresponds to the called DynamoDB method/operation of the AWS SDK
* `net.peer.name`: basically corresponds to 'dynamodb.<region>.amazonaws.com'

Removes the attribute `aws.table_name` as according to the AWS SDK spec,  `aws.dynamodb.table_names` should be used.  (According to the spec the attribute is now also a list of table names instead of a single table name.) 

## Type of change

- [x] Breaking change: breaks existing span attribute names to make them spec compliant

# How Has This Been Tested?

Adds additional unit tests to verify that DynamoDB related attributes are extracted accordingly

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/dynamodb.py
```diff
@@ -0,0 +1,410 @@
+import abc
```
ocelotl, line 1: I think this is missing a license

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/dynamodb.py
```diff
@@ -0,0 +1,410 @@
...
+        return "UpdateTable"
+
+
+################################################################################
+# DynamoDB extension
+################################################################################
+
+_OPERATION_MAPPING = {
+    op.operation_name(): op
+    for op in globals().values()
```
ocelotl, line 328: Just curious, why use `globals` and a series of `if`s if the list of classes can be known beforehand? Is it expected that this module will receive more classes at runtime?

PR #740, opened 2021-10-14
Bugfix/set default headers for properties in pika

oxeye-nikolay:
# Description

The pika instrumentation injects the context into the `properties.headers` of the connection. If the properties are not provided, it generates a new `BasicProperties` object, which by default sets headers to `None` instead of an empty dict - which results in a failure to inject the context into the headers.
The fix was to set the `headers={}` by default when creating the `BasicProperties` object.

## Type of change

# How Has This Been Tested?

A pika client that creates a pika connection and executes `basic_publish` without properties.

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/utils.py
```diff
@@ -40,16 +40,15 @@ def decorated_callback(
...
-            properties.headers = {}
+            properties = BasicProperties(headers={})
         ctx = propagate.extract(properties.headers, getter=_pika_getter)
         if not ctx:
             ctx = context.get_current()
         span = _get_span(
             tracer,
             channel,
             properties,
+            span_kind=SpanKind.CLIENT,
```
owais, line 51: shoudl this be producer/consumer instead of client/server?

oxeye-nikolay: Yeah, You create a single connection and declare one or many channels, over which you declare one or many callbacks. So it is more of a producer/consumer than client/server.

owais (approved):

PR #754, opened 2021-10-18
Consolidate instrumentation documentation in docstrings

lzchen:
Updates instrumentation docstrings to include section on configuration. Most are related to `exclude_url` functionality and `request/response` hooks. We want all of our documentation to exist in one place (readthedocs).

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
@@ -95,6 +95,7 @@ created span and some other contextual information. Example:
         client_request_hook=client_request_hook,
         client_response_hook=client_resposne_hook
     )
+
 API
 ---
 """
--- a/instrumentation/opentelemetry-instrumentation-urllib/README.rst
+++ b/instrumentation/opentelemetry-instrumentation-urllib/README.rst
@@ -16,32 +16,6 @@ Installation
 
      pip install opentelemetry-instrumentation-urllib
 
-Configuration
--------------
-
-Request/Response hooks
-**********************
-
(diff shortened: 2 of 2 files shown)


lzchen:
@codeboten 
I believe configuration options for instrumentations either exist in the `README` or in the docstrings of the instrumentation. We should choose one place and be consistent. Only choosing `README` in this PR because it seemed easier to find (from some feedback) but I don't have a strong opinion. Thoughts?

ocelotl (approved):

instrumentation/opentelemetry-instrumentation-django/README.rst
```diff
@@ -15,52 +15,6 @@ Installation
 
     pip install opentelemetry-instrumentation-django
```
codeboten, line 17: Where did this one go?

codeboten (commented):
Just confirming this is moving all the documentation that would have ended up on pypi into the read the docs repo correct? Users should be able to access it via the links in each package.

instrumentation/opentelemetry-instrumentation-django/README.rst
lzchen: It's already in [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/__init__.py#L14)

lzchen:
@owais 
> We could also have our 4th generate script that allows us to keep docstrings as the single source of truth and generate everything else from that but this is fine as well as long as readthedocs contains the docs.

Haha so many generate scripts xD

codeboten:
> @owais
> 
> > We could also have our 4th generate script that allows us to keep docstrings as the single source of truth and generate everything else from that but this is fine as well as long as readthedocs contains the docs.
> 
> Haha so many generate scripts xD

Soon we'll generate scripts via a generate script
5r7ki1
.

PR #761, opened 2021-10-19
Fix sqlalchemy for postgres Unix sockets

mattoberle:
# Description

This PR addresses an issue that breaks `opentelemetry-instrumentation-sqlalchemy` when connecting to Postgres via Unix socket.

The following bit of replaced code contained a type inconsistency:

```py
attrs[SpanAttributes.NET_PEER_PORT] = int(data.get("port"))
# data.get -> Optional[str]
# int(None)
```

When using postgresql via unix socket `dsn` looks something like this:

```py
'user=postgres host=/tmp/socket dbname=postgres'
```

The `parse_dsn` function returns this:

```py
{'user': 'postgres', 'dbname': 'postgres', 'host': '/tmp/socket'}
```

## Type of change

# How Has This Been Tested?

I applied the patch locally and then did this:

1. Set a SQLAlchemy URL like this: `postgresql+psycopg2://{user}:@/{dbname}?host={socket_path}`.
2. Apply the instrumentation to the engine.
3. Execute a query through the ORM and commit.

> Note: I'm happy to write a test for this as well, but there aren't currently tests against specific engines or private functions. I wasn't sure if that was by design.

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -163,5 +163,7 @@ def _get_attributes_from_cursor(vendor, cursor, attrs):
                 data = parse_dsn(dsn)
                 attrs[SpanAttributes.DB_NAME] = data.get("dbname")
                 attrs[SpanAttributes.NET_PEER_NAME] = data.get("host")
-                attrs[SpanAttributes.NET_PEER_PORT] = int(data.get("port"))
+                # parse_dsn may omit port when connecting via unix socket
+                if data.get("port"):
+                    attrs[SpanAttributes.NET_PEER_PORT] = int(data["port"])
```
lonewolf3739, line 168: If it is a unix socket path we can add different span attributes 
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/fbb677a01d82b6e4602bed5a555ee38474cd9046/instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py#L44-L47.

owais: Makes sense. Let's add NET_PEER_TRANSPORT conditionally when connecting via unix socket.

mattoberle: @owais that's a great point.

It looks like the `TypeError` could apply to non-socket cases too.

```py
from psycopg2.extensions import parse_dsn
from sqlalchemy import create_engine

engine = create_engine('postgresql://postgres@localhost/postgres')
conn = engine.raw_connection()
cur = conn.cursor()

cur.connection.dsn  # 'host=localhost user=postgres dbname=postgres
parse_dsn(cur.connection.dsn)  # {'user': 'postgres', 'dbname': 'postgres', 'host': 'localhost'}
```

However, there is `cur.connection.info` which provides:

```
cur.connection.info.port  # 5432
cur.connection.info.host  # 'localhost'
```

When connecting to a socket it provides:

```
cur.connection.info.port  # 5432
cur.connection.info.host  # '/var/run/postgresql/'
``

`psycopg2` makes the assumption that every socket is named `.s.PGSQL.{port}`, meaning it only stores the absolute path of the socket directory. Mirroring that assumption here seems risky.

The `info` object seems like the safer bet since it's always provides certain values.
If you initialize a SQLAlchemy engine with `engine = create_engine('postgresql://')` the `.dsn` attribute will be an empty string.

I think at the very least the presence of a filesystem path in `info.host` is a clue we are using sockets... but I'll dig around the `psycopg2` source a bit to see if I can find any better guarantees.

Comment (owais) [reply instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py:167]: Makes sense. Checking if the host starts with `unix://`, `unix:///` or `/` makes more sense to me than checking for absence of port.

File: instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py [hunk 2]
```diff
@@ -154,14 +155,25 @@ def _get_attributes_from_url(url):
...
-                data = parse_dsn(dsn)
-                attrs[SpanAttributes.DB_NAME] = data.get("dbname")
-                attrs[SpanAttributes.NET_PEER_NAME] = data.get("host")
-                attrs[SpanAttributes.NET_PEER_PORT] = int(data.get("port"))
+        info = getattr(getattr(cursor, "connection", None), "info", None)
+        if not info:
+            return attrs
+
+        attrs[SpanAttributes.DB_NAME] = info.dbname
+        is_unix_socket = info.host and info.host.startswith("/")
```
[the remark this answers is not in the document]
owais: We should add this as a comment to line 163

owais (approved):

PR #767, opened 2021-10-22
django: Fix carrier usage on ASGI requests

adamantike:
# Description

For ASGI requests, we must use `request.scope` instead of `request.META`. This is because `ASGIGetter` retrieves the `headers` key from the carrier [0], which is only present in `request.scope`.

[0] https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L133

## Type of change

# How Has This Been Tested?

- [x] Tested Django ASGI application along with the `aws` propagator and requests including header: `X-Amzn-Trace-Id: Root=1-5759e988-bd862e3fe1be46a994272793; Parent=53995c3f42cd8ad8; Sampled=1`

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
+++ b/instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
@@ -31,10 +31,16 @@ from opentelemetry.instrumentation.propagators import (
 )
 from opentelemetry.sdk import resources
 from opentelemetry.sdk.trace import Span
+from opentelemetry.sdk.trace.id_generator import RandomIdGenerator
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.test.test_base import TestBase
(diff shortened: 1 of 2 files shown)


owais:
would be nice to add a small test case so we detect breakage with ASGI context extraction.

owais (approved):

adamantike:
Happy to provide a test, but didn't find an easy way to do it, without adding a propagator as dependency, or mocking attach/extract and validating what is received there.

If you have a better test scenario in mind, I can definitely implement it and add it to this PR.

adamantike:
@owais, I could add a test by taking advantage of how the default propagator works, so we check the provided `traceparent` is retrieved from the carrier both for WSGI and ASGI.

PR #815, opened 2021-11-22
add option to disable asgi intern trace

raphaelauv:
# Description

Add an option to disable intern trace of the asgi instrumentation

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/831

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [X] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
@@ -94,14 +94,20 @@ API
 ---
 """
 
+import os
 import typing
 import urllib
+from distutils.util import strtobool
 from functools import wraps
 from typing import Tuple
 
 from asgiref.compatibility import guarantee_single_callable
 
 from opentelemetry import context, trace
+
+from opentelemetry.instrumentation.asgi.environment_variables import (
+    OTEL_PYTHON_ASGI_EXCLUDE_SPAN_RECEIVE, OTEL_PYTHON_ASGI_EXCLUDE_SPAN_SEND
+)
 from opentelemetry.instrumentation.asgi.version import __version__  # noqa
 from opentelemetry.instrumentation.propagators import (
     get_global_response_propagator,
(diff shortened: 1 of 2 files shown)


owais:
@raphaelauv could you please add some explanation about this or create an issue? It would be nice to know what problem this solves or why having those spans can be a problem in the first place. Normally a custom processor/exporter or Otel collector should be used to drop specific spans.

raphaelauv:
The current ASGI instrumentation produce 3 spans :

Screenshot from 2021-12-30 13-44-17

So I was proposing to add an option to disable the two internal spans. Like ask with this issue -> #831

`custom processor/exporter or Otel collector should be used to drop specific spans.`

never heard about that before , do you have any example ? Thank you

owais:
Thanks. Yes, I understand you'd like to disable generation of the two spans but what I'm trying to understand is why those spans are a problem in the first place?

raphaelauv:
` why those spans are a problem in the first place?`
I do not need them and I propose to add an option to disable theses internal spans.

lzchen:
@raphaelauv 
Is it possible to create a custom sampler that drop spans based off of `name` or `kind`? You can also create a custom span processor that drops the spans that you don't need.

shwetasah-cape:
> @raphaelauv Is it possible to create a custom sampler that drop spans based off of `name` or `kind`? You can also create a custom span processor that drops the spans that you don't need.

I am trying to create a custom span processor by overriding the `on_end` method in SpanProcessor, but it doesn't seem to be dropping the span. Is there any example in the docs for a customSpanProcessor to drop spans or prevent them from being exported?

```
class SpanDroppingProcessor(SpanProcessor):
    def on_end(self, span: ReadableSpan) -> None:
        if span.kind == SpanKind.INTERNAL and span.attributes.get("type") == "http.response.body":
            return
```

bsmedberg-xometry:
@shwetasah-cape did you ever come across a solution to suppress these spans? Context for our organization is that for larger requests or responses, we are sometimes seeing dozens or more little "send" and "receive" spans coming from the ASGI integration, and this is causing is cost+load issues in our opentelemetry collector for little actual value, so I'd like to ASAP figure out a way to suppress these at the application level before they get to our collector.

shwetasah-cape:
I am sorry I couldn't figure out how to suppress them. We too have the exact same context as yours.

PR #823, opened 2021-12-03
Handle null statuses in http_status_to_status_code

### LINKED ISSUE #820: urllib instrumentation fails for local file access
When reading local files the status code is not specified and is None. This isn't handled by the instrumentation and causes an exception.

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/444e0a13127304d3a04ccd44445b2e6caed3f770/instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py#L212-L217

gregbuehler:
# Description

This handles None statuses passed to the http_status_to_status_code utility function. This treats None statuses as `StatusCode.UNSET`.

Fixes #820 

## Type of change

# How Has This Been Tested?

This adds new unit tests.
- [x] None status codes are interpreted as StatusCode.UNSET.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


oxeye-nikolay (approved):

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -46,6 +46,8 @@ def http_status_to_status_code(
         status (int): HTTP status code
     """
     # See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#status
+    if status is None:
+        return StatusCode.UNSET
```
owais, line 50: Not sure about encoding knowledge about a special case in this function. IMO we should either have a `isinstance(status, int)` check or wrap in a try/except block to cover all cases of invalid input

gregbuehler: Agreed.

Would it also make sense to also try to coerce status to an integer in case the status was passed as a string representation while we're looking at this?

lzchen:
@gregbuehler 
Please sign the CLA

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
owais: Not sure about that. We don't necessarily want to work with a wide range of inputs. We just don't ever want to crash. I think the caller should be responsible for converting types.

lonewolf3739: Is this something common with other http client libraries as well? I personally prefer not to have the input sanitisation part here. And the caller would have the more context when the status is unusual and it can decide to what to do instead of we returning the UNSET here for all such scenarios.

lzchen: Perhaps we should just include the check only for this specific instrumentation.

ocelotl:
@gregbuehler please fix the lint issues.

Only your latest commit shows the "verified" label. I wonder if that is why the CLA check is not passing. Maybe merging them all in all in one single commit will make this test pass?

owais (approved):

saez0pub:
Hello,

Do you know when this fix will be included in a release ?

Thanks

lzchen:
@saez0pub 
We release on a [monthly](https://github.com/open-telemetry/opentelemetry-python-contrib#releasing) cadence so this release will probably go out some time next week.

PR #825, opened 2021-12-07
lambda instrumentation - support for flush timeout

kubawach:
# Description

Adds support for configurable flush timeout to AWS lambda instrumentation. Property name aligned with other implementations (Java).

## Type of change

# How Has This Been Tested?

- unit test
- previously implemented in the otel-lambda module (before instrumentation was moved here)

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
@@ -97,6 +97,9 @@ logger = logging.getLogger(__name__)
 _HANDLER = "_HANDLER"
 _X_AMZN_TRACE_ID = "_X_AMZN_TRACE_ID"
 ORIG_HANDLER = "ORIG_HANDLER"
+OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_TIMEOUT = (
+    "OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_TIMEOUT"
+)
 
 
 def _default_event_context_extractor(lambda_event: Any) -> Context:
@@ -167,6 +170,7 @@ def _determine_parent_context(
 def _instrument(
     wrapped_module_name,
     wrapped_function_name,
+    flush_timeout,
     event_context_extractor: Callable[[Any], Context],
     tracer_provider: TracerProvider = None,
 ):
@@ -222,7 +226,7 @@ def _instrument(
             # NOTE: `force_flush` before function quit in case of Lambda freeze.
             # Assumes we are using the OpenTelemetry SDK implementation of the
             # `TracerProvider`.
-            _tracer_provider.force_flush()
+            _tracer_provider.force_flush(flush_timeout)
         except Exception:  # pylint: disable=broad-except
             logger.error(
                 "TracerProvider was missing `force_flush` method. This is necessary in case of a Lambda freeze and would exist in the OTel SDK implementation."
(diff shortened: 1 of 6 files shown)


instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -262,9 +266,16 @@ def _instrument(self, **kwargs):
             self._wrapped_function_name,
         ) = lambda_handler.rsplit(".", 1)
 
+        flush_timeout = int(
```
owais, line 269: This can throw if the value cannot be converted to `int` so should be guarded against

owais (commented):
May be we should allow configuring timeout at the provided level instead of a specific instrumentation? @open-telemetry/python-approvers thoughts?

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
kubawach: Good catch!

kubawach:
> May be we should allow configuring timeout at the provided level instead of a specific instrumentation? @open-telemetry/python-approvers thoughts?

I believe that in this particular case it makes sense - force flush upon function completion is specific to AWS lambda and aligns Python with Java (and others in the future) - hence prop name `OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_TIMEOUT`.

owais:
It is not really harmful but I think we'll have to add support for something like `OTEL_PYTHON_SPAN_FLUSH_TIMEOUT` which would cover this and any other similar cases, and once we have that, we'll have two competing knobs in case of lambda so probably worth it to add support directly to the processor?

kubawach:
> It is not really harmful but I think we'll have to add support for something like `OTEL_PYTHON_SPAN_FLUSH_TIMEOUT` which would cover this and any other similar cases, and once we have that, we'll have two competing knobs in case of lambda so probably worth it to add support directly to the processor?

Hard to say :) Other distros (say - Java) didn't implement generic timeout for forceFlush. On the other side in Java it's implemented as an async call and the caller can decide to wait (or not) for completion of the call. If you see potential uses besides lambda for configurable timeout then it definitely makes sense to move this to SpanProcessor. Let me know (in the meantime I've added try/catch for int conversion ;) ).

owais:
This has come up a couple of times on related issues so I think it could be useful. Personally I'm fine with merging this as is in order to not block a useful feature. @open-telemetry/python-approvers any preferences?

owais (approved):

owais (changes requested):
I don't see the try..catch block around env var

lonewolf3739:
I thought there was already some env to support this. I also agree that we should have something like `OTEL_PROVIDER_FLUSH_TIMEOUT` instead of instrumentation specific env vars.

kubawach:
> I don't see the try..catch block around env var

Yeah, forgot to update the commit 🤦 Now it's where supposed to be.

owais (approved):

NathanielRN (commented):
I also think it would be beneficial to have the environment variable at the SDK layer, but I can also see [the Java SDK has this similar variable](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-lambda-1.0/library/README.md#using-wrappers).

Either way this looks useful, thanks for the change!

kubawach:
@open-telemetry/python-maintainers can we get this merged please? :)

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -262,9 +266,22 @@ def _instrument(self, **kwargs):
             self._wrapped_function_name,
         ) = lambda_handler.rsplit(".", 1)
 
+        flush_timeout_env = os.environ.get(
+            OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_TIMEOUT, ""
+        )
+        flush_timeout = 30000
+        try:
+            flush_timeout = int(flush_timeout_env)
```
NathanielRN, line 274: Just pass-by thinking that if gave a default in the function interface as `flush_timeout=3000` then you could something like this:

```suggestion
            flush_timeout = int(
                os.environ.get(
                    OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_TIMEOUT, flush_timeout
                )
            )
```

Not saying it is necessarily easier to read, but just another way to write it! The log statement would have to be updated to fetch from the env again.

kubawach: Thank, let's leave it as it is - IMO reading flow is good now and clearly separates types (string vs int).

Issue #448, opened 2021-04-15
Django: Conditionally create SERVER spans


owais:
Part of #445

sanketmehta28:
Hi @owais : I would like to work on this as my first contribution to Opentelemetry.
Can you please assign it to me?

owais:
Assigned. @ashu658 was planning to work on #445 as well. Please coordinate so you don't work on the same tickets.

sanketmehta28:
Sure, We are from the same team. we will coordinate with each other.

sanketmehta28:
You can close this issue as the PR mentioned above is merged into main branch

Issue #445, opened 2021-04-15
Server instrumentations should look for parent spans in current context before extracting context from carriers


owais:
Today server instrumentations such as Django always extract the parent span from the incoming request headers. This may not always be ideal. For example, there may be other instrumented components that wrap an instrumented web framework such as WSGI. In such cases WSGI would already have generated a server span and used the remote span as parent. If Django did the same, traces would not make a lot of sense. Depending on whether a remote span is present in the incoming request's carrier, Django and WSGI spans would be completely different traces or be siblings instead of parent and child. 

To avoid situations like this, all server instrumentations should first check if an active span is present in the current context by calling `opentelemetry.api.trace.get_current_span()` and use it as the parent when present. When no active span is found in the current context, the instrumentation should try to extract remote span context from the incoming request and use that as the parent.

This needs to be done for all instrumentations that generate server spans.

Changes that need to be made:

- If a span is found in the current context, then:
   - create a new span with kind set to INTERNAL
   - use the span found in current context as parent 
- If no span is in the current context, then:
   - create a new span with kind set to SERVER
   - extract remote span context from incoming request and use it as parent

lonewolf3739:
Just the other day I have seen somebody using both WSGI middleware and Django instrumentation. This makes sense in such scenarios but I am wondering if there are any cases where active span is not from remote context that we end up using it instead of actual remote context? I don't know any but just adding here for visibility.

lonewolf3739:
It could also be possible that there is a wrapper which has some instrumentation but not necessary that Django instrumentation spans should be child spans of them. Some raw thought; I don't know if this makes sense.

owais:
I think it is possible but even if it happens, I think it should be considered an error on part of the instrumentation/user who activates such a span in Django's request/response cycle context. 

For added safety, Django span can check if parent span is a SERVER span for additional safety but I'm not sure if it is worth it and it is possible WSGI can create another internal child span in future. 


FWIW, both Node and Java follow this pattern today in order to enhance parent spans. Some Node/Java grab parent spans and add attributes or update names. This would be akin Django instrumentation not generating a new span and instead adding additional information to WSGI span. I don't think we can or should do that given how Python ecosystem is different from Node/Java but it does at least set the precedent for server instrumentations looking at parent spans. 

(Using django as a placeholder for any python web framework)

lzchen:
[...]

PR #832, opened 2021-12-14
Develop/condition server span django

sanketmehta28:
# Description

there may be other instrumented components that wrap an instrumented web framework such as WSGI. In such cases WSGI would already have generated a server span and used the remote span as parent. If Django did the same, traces would not make a lot of sense. Depending on whether a remote span is present in the incoming request's carrier, Django and WSGI spans would be completely different traces or be siblings instead of parent and child.

Fixes # https://github.com/open-telemetry/opentelemetry-python-contrib/issues/448

## Type of change

# How Has This Been Tested?

- I have tested it locally using a sample app with django and wsgi instrumentation enabled with jaeger exporter.
- I have sent a request to the app and observe on jaeger that there are two span created: WSGI span as SERVER span and Django span as INTERNEL span.

- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/open-telemetry/opentelemetry-python/compare/v1.7.0-0.26b0...HEAD)
 
-## [1.7.0-0.26b0](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.7.0-0.26b0) - 2021-11-11
+## [1.7.1-0.26b1](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.7.0-0.26b0) - 2021-11-11
 
 - `opentelemetry-instrumentation-aws-lambda` Add instrumentation for AWS Lambda Service - pkg metadata files (Part 1/2)
   ([#739](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/739))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased](https://github.com/open-telemetry/opentelemetry-python/compare/v1.7.0-0.26b0...HEAD)
+## [Unreleased](https://github.com/open-telemetry/opentelemetry-python/compare/v1.8.0-0.27b0...HEAD)
+
+## [1.8.0-0.27b0](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.8.0-0.27b0) - 2021-12-17
+
+
 
 ### Added
 
@@ -21,6 +25,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `opentelemetry-instrumentation-django` Django: Conditionally create SERVER spans
   ([#812](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/832))
+
+- `opentelemetry-instrumentation-flask` Flask: Conditionally create SERVER spans
+  ([#828](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/828))
 
   
 ## [1.7.1-0.26b1](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.7.0-0.26b0) - 2021-11-11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.8.0-0.27b0](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.8.0-0.27b0) - 2021-12-17
 
+### Added
 
+- `opentelemetry-instrumentation-asgi` now returns a `traceresponse` response header.
+  ([#817](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/817))
+### Fixed
+
+- `opentelemetry-instrumentation-flask` Flask: Conditionally create SERVER spans
+  ([#828](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/828))
+
+## [1.8.0-0.27b0](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.8.0-0.27b0) - 2021-12-17
 
 ### Added
 
 - `opentelemetry-instrumentation-aws-lambda` Adds support for configurable flush timeout  via `OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_TIMEOUT` property. ([#825](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/825))
+- `opentelemetry-instrumentation-pika` Adds support for versions between `0.12.0` to `1.0.0`. ([#837](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/837))
 
 ### Fixed
 
@@ -26,11 +36,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-instrumentation-tornado` Add support instrumentation for Tornado 5.1.1
   ([#812](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/812))
 
-- `opentelemetry-instrumentation-flask` Flask: Conditionally create SERVER spans
-  ([#828](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/828))
-
-  
 ## [1.7.1-0.26b1](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.7.0-0.26b0) - 2021-11-11
+
+### Added
 
 - `opentelemetry-instrumentation-aws-lambda` Add instrumentation for AWS Lambda Service - pkg metadata files (Part 1/2)
   ([#739](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/739))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `opentelemetry-instrumentation-tornado` Add support instrumentation for Tornado 5.1.1
   ([#812](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/812))
+
+- `opentelemetry-instrumentation-django` Django: Conditionally create SERVER spans
+  ([#812](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/832))
+
   
 ## [1.7.1-0.26b1](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.7.0-0.26b0) - 2021-11-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `opentelemetry-instrumentation-django` Django: Conditionally create SERVER spans
+  ([#832](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/832))
+  
 - `opentelemetry-instrumentation-flask` Flask: Conditionally create SERVER spans
   ([#828](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/828))
 
-- `opentelemetry-instrumentation-django` Django: Conditionally create SERVER spans
-  ([#832](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/832))
+- `opentelemetry-instrumentation-asgi` ASGI: Conditionally create SERVER spans
+  ([#843](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/843))
+
 
 ## [1.8.0-0.27b0](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.8.0-0.27b0) - 2021-12-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/open-telemetry/opentelemetry-python/compare/v1.8.0-0.27b0...HEAD)
 
+- `opentelemetry-instrumentation-django` Django: Conditionally create SERVER spans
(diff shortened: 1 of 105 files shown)


sanketmehta28:
I need some input in the pytest I have added in this PR.
I have enabled django as well as WSGI instrumentation in TestDjangoWithOtherFramework class but I am getting only 1 span when I call self.exporter.get_finished_spans().

Strange thing is the span I am getting is of type INTERNEL and it has parent_id available. It means it does have a root span of type SERVER. 

Unfortunately I could not find a way to get the root span info and test is getting failed in the last statement:
self.assertEqual(trace.SpanKind.SERVER, span_list[1].kind)

Can someone help me on this?

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
```diff
@@ -430,3 +433,67 @@ def test_tracer_provider_traced(self):
...
+        
+        # resp = Client().get("/span_name/1234/")
+        # self.assertEqual(200, resp.status_code)
+
+        # span_list = self.memory_exporter.get_finished_spans()
+        span_list = self.exporter.get_finished_spans()
+        # print(span_list)
+        self.assertEqual(trace.SpanKind.INTERNAL, span_list[0].kind)
+
+        #Below line give me error "index out of the range" as there is only one span created where it should be 2.
```
lonewolf3739, line 498: It appears that your server span from middleware started but never ended. Please try changing the line 488 to `next(application(environ, start_response))` and check.

sanketmehta28:
changes are made according to Srikanth's comments. Unit tests are running fine now. Please do review the PR

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
sanketmehta28: Sure. Done and changes are pushed. Please do review them

lonewolf3739: I only intend to suggest that hack to validate the hypothesis. You should use test client instead.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
```diff
@@ -183,11 +189,16 @@ def process_request(self, request):
             carrier_getter = wsgi_getter
             collect_request_attributes = wsgi_collect_request_attributes
 
-        token = attach(extract(carrier, getter=carrier_getter))
-
+        token = context = None
+        span_kind = SpanKind.INTERNAL
+        if get_current_span() is INVALID_SPAN:
+            context = extract(request_meta, getter=wsgi_getter)
```
owais, line 195: earlier we were using `carrier_getter` which could be either a wsgi or asgi getter. Now we are always using `wsgi_getter`. is this intentional?

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
```diff
@@ -220,7 +231,8 @@ def process_request(self, request):
 
         request.META[self._environ_activation_key] = activation
         request.META[self._environ_span_key] = span
-        request.META[self._environ_token] = token
+        if token:
+            request.META[self._environ_token] = token
```
owais, line 235: since this is conditional now, do we need to update places where this is being read to make sure everthing continues to work?

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
```diff
@@ -432,3 +438,22 @@ def test_tracer_provider_traced(self):
         self.assertEqual(
             span.resource.attributes["resource-key"], "resource-value"
         )
+
+    def test_django_with_wsgi_instrumented(self):
```
owais, line 442: this is fine but it would be nicer if we could test this in a generic way without wsgi instrumentation. At least lets add a comment documenting what this test does.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
sanketmehta28: Yes Without this check, It was raising an exception while detaching the token from request object

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
sanketmehta28: Sure I will add the comment.
Also what do you mean by generic way? can you please give me an example?

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
sanketmehta28: With Client() it was not creating the WSGI span. That is the reason I have to go this way

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
sanketmehta28: Oh my bad. corrected to `carrier_getter`.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
owais: I mean earlier we were always setting this value in the dict even if it was `None` but now we won't be setting it at all in some cases. This can result in lookup error if code in other places assumes the key will always be present so we should make sure this doesn't break anything else.

sanketmehta28: Added a check before detaching the token

CHANGELOG.md
```diff
@@ -23,8 +23,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-instrumentation-tornado` Add support instrumentation for Tornado 5.1.1
   ([#812](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/812))
 
+- `opentelemetry-instrumentation-django` Django: Conditionally create SERVER spans
+  ([#812](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/832))
```
lonewolf3739, line 27:
```suggestion
  ([#832](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/832))
```

Please move this to unreleased section

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
```diff
@@ -183,11 +189,16 @@ def process_request(self, request):
             carrier_getter = wsgi_getter
             collect_request_attributes = wsgi_collect_request_attributes
 
-        token = attach(extract(carrier, getter=carrier_getter))
-
+        token = context = None
+        span_kind = SpanKind.INTERNAL
+        if get_current_span() is INVALID_SPAN:
+            context = extract(request_meta, getter=carrier_getter)
```
lonewolf3739, line 195:
```suggestion
            context = extract(carrier, getter=carrier_getter)
```

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
```diff
@@ -295,7 +307,8 @@ def process_response(self, request, response):
             else:
                 activation.__exit__(None, None, None)
 
-        if self._environ_token in request.META.keys():
+        # if self._environ_token in request.META.keys():
+        if request.META.get(self._environ_token, None) is not None:
```
lonewolf3739, line 311: How is this different from previous check?

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
```diff
@@ -432,3 +438,28 @@ def test_tracer_provider_traced(self):
...
+            PATH_INFO="/span_name/1234/",
+            CONTENT_TYPE="text/html; charset=utf-8",
+            REQUEST_METHOD="GET"
+        )
+        response_data = {}
+        def start_response(status, headers):
+            response_data["status"] = status
+            response_data["headers"] = headers
+
+        resp = next(application(environ, start_response))
```
lonewolf3739, line 461: Please use the test client here.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
```diff
@@ -183,11 +189,16 @@ def process_request(self, request):
...
+        span_kind = SpanKind.INTERNAL
+        if get_current_span() is INVALID_SPAN:
+            context = extract(request_meta, getter=carrier_getter)
+            token = attach(context)
+            span_kind = SpanKind.SERVER
         span = self._tracer.start_span(
             self._get_span_name(request),
-            kind=SpanKind.SERVER,
+            context,
+            kind=span_kind,
```
lonewolf3739, line 201: @owais do we always want the span kind to be INTERNAL when some other span already exists or do we want it only when there is a SERVER span?

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
sanketmehta28: I have tried but it is not able to create multiple spans when test client is being used

CHANGELOG.md
sanketmehta28: Done

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
sanketmehta28: Done

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
sanketmehta28: this will not fail even if the _environ_token is present in the request.META with None value.
Previously it will create an exception while calling detach() if the token is present with None value

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
owais: I still see `request_meta` being used @sanketmehta28

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
owais: I think if any span exists that we use as a parent, we should use INTERNAL as SERVER span must have a CLIENT span as the parent IMO

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
owais: I was thinking of just injecting a span into the execution context so the instrumentation would find it and trigger the code path. That way the test wouldn't depend on wsgi instrumentation but it is not a big issue for me.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
sanketmehta28: Done

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
sanketmehta28: Done. Please do check and provide your comments.

lzchen (approved):

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
```diff
@@ -432,3 +434,25 @@ def test_tracer_provider_traced(self):
...
+            resp = Client().get("/span_name/1234/")
+            span_list = self.exporter.get_finished_spans()
+            print(span_list)
+            self.assertEqual(span_list[0].attributes[SpanAttributes.HTTP_STATUS_CODE], 200)
+            self.assertEqual(trace.SpanKind.INTERNAL, span_list[0].kind)
+            self.assertEqual(parent_span.get_span_context().span_id, span_list[0].parent.span_id)
+                
+            
+        
+
```
lonewolf3739, line 458: I think you don't need this description anymore since you changed the implementation and issue #445 gives enough context.

```suggestion
        tracer = self.tracer_provider.get_tracer(__name__)
        with tracer.start_as_current_span("test", kind=SpanKind.SERVER) as parent_span:
            resp = Client().get("/span_name/1234/")
            span_list = self.exporter.get_finished_spans()
            self.assertEqual(span_list[0].attributes[SpanAttributes.HTTP_STATUS_CODE], 200)
            self.assertEqual(trace.SpanKind.INTERNAL, span_list[0].kind)
            self.assertEqual(parent_span.get_span_context().span_id, span_list[0].parent.span_id)
```

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py
sanketmehta28: Yes

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
sanketmehta28: @lonewolf3739 : Done

CHANGELOG.md
```diff
@@ -7,11 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/open-telemetry/opentelemetry-python/compare/v1.8.0-0.27b0...HEAD)
 
+- `opentelemetry-instrumentation-django` Django: Conditionally create SERVER spans
```
lonewolf3739, line 10: Please fix the CHANGELOG. Remove the duplicate previous release entry and move your entry to fixed list below.

Screenshot 2022-01-06 at 6 58 10 PM

lonewolf3739 (changes requested):

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-instrumentation-flask` Flask: Conditionally create SERVER spans
   ([#828](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/828))
 
+- `opentelemetry-instrumentation-django` Django: Conditionally create SERVER spans
+  ([#812](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/832))
```
lonewolf3739, line 21:
```suggestion
  ([#832](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/832))
```

lonewolf3739 (approved):

PR #837, opened 2021-12-19
Support older pika versions

nozik:
# Description

Currently, `opentelemetry-instrumentation-pika` supports only `pika` versions above `1.0.0`. This PR adds support to `pika >= 0.12.0`

## Type of change

# How Has This Been Tested?

I've updated `tox.ini` to include tests for the older `pika` versions.

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/pika_instrumentor.py
+++ b/instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/pika_instrumentor.py
@@ -14,7 +14,9 @@
 from logging import getLogger
 from typing import Any, Collection, Dict, Optional
 
+import pkg_resources
 import wrapt
+from packaging import version
 from pika.adapters import BlockingConnection
 from pika.adapters.blocking_connection import BlockingChannel
 
@@ -41,8 +43,12 @@ class PikaInstrumentor(BaseInstrumentor):  # type: ignore
         consume_hook: utils.HookT = utils.dummy_callback,
     ) -> Any:
         for consumer_tag, consumer_info in channel._consumer_infos.items():
+            callback_attr = (
+                PikaInstrumentor._consumer_callback_attribute_name()
+            )
+            consumer_callback = getattr(consumer_info, callback_attr)
             decorated_callback = utils._decorate_callback(
-                consumer_info.on_message_callback,
+                consumer_callback,
                 tracer,
                 consumer_tag,
                 consume_hook,
@@ -51,9 +57,9 @@ class PikaInstrumentor(BaseInstrumentor):  # type: ignore
             setattr(
                 decorated_callback,
                 "_original_callback",
-                consumer_info.on_message_callback,
+                consumer_callback,
             )
-            consumer_info.on_message_callback = decorated_callback
+            setattr(consumer_info, callback_attr, decorated_callback)
 
     @staticmethod
     def _instrument_basic_publish(
@@ -126,11 +132,26 @@ class PikaInstrumentor(BaseInstrumentor):  # type: ignore
             return
 
         for consumers_tag, client_info in channel._consumer_infos.items():
-            if hasattr(client_info.on_message_callback, "_original_callback"):
+            callback_attr = (
+                PikaInstrumentor._consumer_callback_attribute_name()
+            )
+            consumer_callback = getattr(client_info, callback_attr)
+            if hasattr(consumer_callback, "_original_callback"):
                 channel._consumer_infos[
                     consumers_tag
-                ] = client_info.on_message_callback._original_callback
+                ] = consumer_callback._original_callback
         PikaInstrumentor._uninstrument_channel_functions(channel)
+
+    @staticmethod
+    def _consumer_callback_attribute_name() -> str:
+        pika_version = version.parse(
+            pkg_resources.get_distribution("pika").version
+        )
+        return (
+            "on_message_callback"
+            if pika_version >= version.parse("1.0.0")
+            else "consumer_cb"
+        )
 
     def _decorate_channel_function(
         self,
--- a/instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/pika_instrumentor.py
+++ b/instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/pika_instrumentor.py
@@ -55,7 +55,9 @@ class PikaInstrumentor(BaseInstrumentor):  # type: ignore
     ) -> Any:
         for consumer_tag, consumer_info in channel._consumer_infos.items():
             callback_attr = PikaInstrumentor.CONSUMER_CALLBACK_ATTR
-            consumer_callback = getattr(consumer_info, callback_attr)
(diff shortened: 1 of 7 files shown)


instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/pika_instrumentor.py
```diff
@@ -126,12 +132,27 @@ def uninstrument_channel(channel: BlockingChannel) -> None:
...
+    @staticmethod
+    def _consumer_callback_attribute_name() -> str:
+        pika_version = version.parse(
+            pkg_resources.get_distribution("pika").version
+        )
+        return (
+            "on_message_callback"
+            if pika_version >= version.parse("1.0.0")
+            else "consumer_cb"
+        )
```
oxeye-nikolay, line 154: I think that if the pika package ever has a breaking change in the future (like they apparently had between `1.0.0` and `1.1.0`) and change `on_message_callback` to `on_message_cb` or whatever, this will break, and since we don't have an actual test to detect this, this will go undetected.
This function should check `hasattr` on the object, rather than checking the version of imported pika.

instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/pika_instrumentor.py
```diff
@@ -126,12 +132,27 @@ def uninstrument_channel(channel: BlockingChannel) -> None:
...
         PikaInstrumentor._uninstrument_channel_functions(channel)
 
+    @staticmethod
+    def _consumer_callback_attribute_name() -> str:
+        pika_version = version.parse(
+            pkg_resources.get_distribution("pika").version
+        )
+        return (
+            "on_message_callback"
+            if pika_version >= version.parse("1.0.0")
```
oxeye-nikolay, line 152: Another thought I had: Since we can't have a mixture of versions, maybe we should just check this once when loading the library instead of when instrumenting and uninstrumenting each object.

instrumentation/opentelemetry-instrumentation-pika/tests/test_pika_instrumentation.py
```diff
@@ -72,8 +73,11 @@ def test_instrument_consumers(
         self, decorate_callback: mock.MagicMock
     ) -> None:
         tracer = mock.MagicMock(spec=Tracer)
+        callback_attr = PikaInstrumentor._consumer_callback_attribute_name()
```
oxeye-nikolay, line 76: add a test for `_consumer_callback_attribute_name`

instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/pika_instrumentor.py
nozik: If a newer version is released with a breaking change, unit tests will fail - `getattr` throws `AttributeError` same as the dot operator. Also, I'm not changing the current behavior.

instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/pika_instrumentor.py
owais: +1. Also doesn't pika have a version module that we can import and check instead of using `pkg_resources`? Ideally code should not rely on packaging systems to function.

nozik: @owais It does, I'll use that instead. Thanks

instrumentation/opentelemetry-instrumentation-pika/tests/test_pika_instrumentation.py
nozik: @oxeye-nikolay What should it test? If it returns the wrong value, other tests will definitely fail.

oxeye-nikolay: mock the return value of `pkg_resources.get_distribution("pika").version` and ensure the value returned is good. The tests may fail also without it, but you want the tests to identify the source

instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/pika_instrumentor.py
```diff
@@ -43,9 +54,7 @@ def _instrument_blocking_channel_consumers(
         consume_hook: utils.HookT = utils.dummy_callback,
     ) -> Any:
         for consumer_tag, consumer_info in channel._consumer_infos.items():
-            callback_attr = (
-                PikaInstrumentor._consumer_callback_attribute_name()
-            )
+            callback_attr = PikaInstrumentor.CONSUMER_CALLBACK_ATTR
             consumer_callback = getattr(consumer_info, callback_attr)
```
oxeye-nikolay, line 58: Check if `consumer_info` has `callback_attr`. This will raise otherwise, which is not the wanted behavior.

instrumentation/opentelemetry-instrumentation-pika/tests/test_pika_instrumentation.py
nozik: @oxeye-nikolay I'm longer using `pkg_resources` but rather accessing `pika.__version__` directly, as suggested by owais. If I override the value exported by the package directly it might have unexpected implications and I'd rather avoid it

instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/pika_instrumentor.py
nozik: That's the current behavior - if `on_message_callback` isn't there, an exception would be thrown.
I don't mind simply skipping the instrumentation (skipping the `decorate_callback` part) in case that happens

instrumentation/opentelemetry-instrumentation-pika/tests/test_pika_instrumentation.py
oxeye-nikolay: Here, try this:
```python
import pika
from unittest import mock
print(pika.__version__)
with mock.patch("pika.__version__", "0.1.1"):
     print(pika.__version__)
print(pika.__version__)
```
the output should be:
```
1.2.0
0.1.1
1.2.0
```

instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/pika_instrumentor.py
oxeye-nikolay: Well then, this should also be fixed now then. We don't want to raise. Your solution sounds great for this

nozik: Done

oxeye-nikolay (approved):

owais (approved):

PR #847, opened 2022-01-09
Support partial timelimit for celery tasks

### LINKED ISSUE #846: Partial celery task time limit is not support
Python 3.8

**Steps to reproduce**
- Create celery task with only one of the time limit soft/hard
Or use this test to reproduce:

`instrumentation/opentelemetry-instrumentation-celery/tests/test_utils.py`

```python
    def test_set_attributes_partial_timelimit_hard_limit(self):
        context = {
            "correlation_id": "44b7f305",
            "delivery_info": {"eager": True},
            "eta": "soon",
            "expires": "later",
            "hostname": "localhost",
            "id": "44b7f305",
            "reply_to": "44b7f305",
            "retries": 4,
            "timelimit": ("now", None),
            "custom_meta": "custom_value",
            "routing_key": "celery",
        }
        span = trace._Span("name", mock.Mock(spec=trace_api.SpanContext))
        utils.set_attributes_from_context(span, context)
        self.assertEqual(span.attributes.get("celery.timelimit"), ("now", ""))
```
The time limit that was specify should be instrumented 

Exception is raised and no time limit is instrumented
```
[__init__.py:_translate_attributes:164] _translate_key_values(key, value)
exception.trace.1	
[exporter.py:_translate_key_values:126] return KeyValue(key=key, value=_translate_value(value))
exception.trace.2	
[ exporter.py:_translate_value:104] array_value=ArrayValue(values=[_translate_value(v) for v in value])
exception.trace.3	
[ exporter.py:<listcomp>:104] array_value=ArrayValue(values=[_translate_value(v) for v in value])
exception.trace.4	
[ exporter.py:_translate_value:119] "Invalid type {} of value {}".format(type(value), value)
```

OHeimlich:
# Description

Adding support in partial time limit for celery tasks.
Instrument celery task time limit even if only one limit was set (soft/hard limit).
Fixes #846 

## Type of change

# How Has This Been Tested?

Running unit test

- [ ] Yes. - https://github.com/open-telemetry/opentelemetry-python-contrib/pull/847

# Checklist:


codeboten (commented):
hey @OHeimlich, thanks for the contribution! Please make sure to sign the CLA and update the changelog to include a note about the fix.

instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/utils.py
```diff
@@ -64,6 +64,12 @@ def set_attributes_from_context(span, context):
         if key == "timelimit" and value in [(None, None), [None, None]]:
             continue
 
+        if key == "timelimit":
+            if value in [(None, None), [None, None]]:
+                continue
```
lonewolf3739, line 69: Looks like this is already covered in above lines 64-65

OHeimlich: Thank you @lonewolf3739 for your comment, The case I want to handle here is when we set only one of the soft/hard limits in the celery task so we will get here `[None, Value]` or `[Value, None]` and this case isn't covered.

OHeimlich:
Hey @codeboten,
Thank you for your response, I signed the CLA and updated the changelog :)

instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/utils.py
lonewolf3739: Correct, I see that below. What I am saying is `if value in [(None, None), [None, None]]: continue` is a redundant check.

OHeimlich: Not sure I understand why, we can have three cases:
1. soft and hard limit was set - we will instrument them both.
2. Only one of the limits was set and we will instrument that one and the other will set to empty string.
3. No time limit was set - then we don't want to instrument them at all.

If I'll remove this check we will instrument the time limit with empty string even if they weren't set.

lonewolf3739: I am saying that case is already covered [above](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/847/files#diff-672f16686553f36e3138aeaf3d146c467f3b6792448e59f393ac87c1f676bf23L62-L65)
```py
        # Skip `timelimit` if it is not set (it's default/unset value is a
        # tuple or a list of `None` values
         if key == "timelimit" and value in [(None, None), [None, None]]:
             continue
```

OHeimlich: Oh sorry @lonewolf3739 I missed these lines, I wanted to merge it to on `if` and forgot to remove the original ones.
I fixed that here 37e0a296cf33beb228f42ef600e832c36e907fcc.

lonewolf3739 (approved):

codeboten (approved):

PR #851, opened 2022-01-10
Unblock builds by removing botocore test

lzchen:
Instrumentation builds are failing for all version of Python due to a botocore test. Strangely enough this has never happened before, seems like the response body is empty? Removing this for now to unblock builds.

@mariojonke 
I'm not too familiar with botocore instrumentation. Where in the instrumentation does it modify the RESPONSE body to add trace header information?

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_lambda.py
+++ b/instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_lambda.py
@@ -126,6 +126,7 @@ class TestLambdaExtension(TestBase):
             Publish=True,
         )
 
+    @mark.skip(reason="Docker error, unblocking builds for now.")
     @mark.skipif(
         sys.platform == "win32",
         reason="requires docker and Github CI Windows does not have docker installed by default",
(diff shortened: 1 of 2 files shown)


lonewolf3739 (approved):

codeboten (commented):
Does pinning the version of the library solve the problem?

ocelotl:
> Does pinning the version of the library solve the problem?

Yeah, @lzchen there was a release of botocore [4 hours ago](https://pypi.org/project/botocore/#history)... :thinking:

ocelotl:
I don't think we should merge before we check the pinning of the version, marking it as draft to prevent accidental merging :v:

lzchen:
@codeboten  @ocelotl 
The last successful build run was [this](https://github.com/open-telemetry/opentelemetry-python-contrib/runs/4747016142?check_suite_focus=true#step:6:807) which used botocore v1.23.31. Pinning the [version](https://github.com/open-telemetry/opentelemetry-python-contrib/runs/4769425309?check_suite_focus=true#step:6:659) to v1.23.31 did not work. Also using the latest release a couple of hours ago does not work [either](https://github.com/open-telemetry/opentelemetry-python-contrib/runs/4769295211?check_suite_focus=true#step:6:807). Very strange.

ocelotl:
Hm, weird. Ok, marked as ready for review ir order not to block everything ✌️

I'll open an issue to track this mystery 😉

ocelotl:
Wait a minute... botocore tests are also failing in this PR :thinking:

owais:
We can add a skip marker to this test if this is blocking others for now.

codeboten (approved):

instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_lambda.py
```diff
@@ -152,7 +153,7 @@ def test_invoke(self):
             span = self.assert_invoke_span(function_name)
             span_context = span.get_span_context()
 
-            # assert injected span
+            # # assert injected span
```
codeboten, line 156: was this intentional?

Issue #449, opened 2021-04-15
Pyramid: Conditionally create SERVER spans


owais:
Part of #445

PR #865, opened 2022-01-17
[instrumentation/pyramid] Conditionally create SERVER spans

sanketmehta28:
# Description

This code change resolves the conditional span creation for pyramid framework in case of other framework is used as well. This fix will use the existing span and make the new span its child and create a parent child relation which is the right behavior

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/449

## Type of change

# How Has This Been Tested?

I have added the unit test cases in test_automatic.py for this code change where I have added a span in current context and send a request using test client which will create the child span and I have checked that parent.id of the newly created span is equal to the id of the existing span.

- [ ] No.

# Checklist:


codeboten (commented):
Thanks for the contribution @sanketmehta28! Please make sure to sign the CLA to get this change reviewed and merged.

PR #867, opened 2022-01-18
Conditionally create server spans for falcon

### LINKED ISSUE #447: Falcon: Conditionally create SERVER spans
Part of #445

ashu658:
# Description

Adds support to make spans as INTERNAL if a span is already present in current context in falcon.

Fixes #447

## Type of change

    Currently falcon only makes SERVER spans. With this PR it can make INTERNAL spans in presence of a span in current context.

# How Has This Been Tested?

Added unit test for the changes.
Manually tested the scenario.
To reproduce the issue: Add two instrumentors i.e. FalconInstrumentor and OpenTelemetryMiddleware in the same app. This creates two separate traces instead of one.

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
@@ -109,6 +109,7 @@ from opentelemetry.instrumentation.propagators import (
 from opentelemetry.instrumentation.utils import (
     extract_attributes_from_object,
     http_status_to_status_code,
+    get_token_context_span_kind
 )
 from opentelemetry.propagate import extract
 from opentelemetry.semconv.trace import SpanAttributes
@@ -195,16 +196,8 @@ class _InstrumentedFalconAPI(getattr(falcon, _instrument_app)):
 
         start_time = _time_ns()
 
-        token = ctx = span_kind = None
-
-        if trace.get_current_span() is trace.INVALID_SPAN:
-            ctx = extract(env, getter=otel_wsgi.wsgi_getter)
-            token = context.attach(ctx)
-            span_kind = trace.SpanKind.SERVER
-        else:
-            ctx = context.get_current()
-            span_kind = trace.SpanKind.INTERNAL
-
+        token, ctx, span_kind = get_token_context_span_kind(env, getter=otel_wsgi.wsgi_getter)
+  
         span = self._tracer.start_span(
             otel_wsgi.get_default_span_name(env),
             context=ctx,
--- a/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
@@ -108,10 +108,9 @@ from opentelemetry.instrumentation.propagators import (
 )
 from opentelemetry.instrumentation.utils import (
     extract_attributes_from_object,
+    get_token_context_span_kind,
     http_status_to_status_code,
-    get_token_context_span_kind
 )
-from opentelemetry.propagate import extract
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.trace.status import Status
 from opentelemetry.util._time import _time_ns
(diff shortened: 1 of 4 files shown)


instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -195,12 +195,23 @@ def __call__(self, env, start_response):
 
         start_time = _time_ns()
 
-        token = context.attach(extract(env, getter=otel_wsgi.wsgi_getter))
+        token = ctx = span_kind = None
+
```
owais, line 199: this snippet is becoming pretty common. probably could take it out as a re-usable function.

owais (approved):

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
ashu658: Added re-usable getter function for the three variables. Could not think of better name. Please suggest.

lzchen (approved):

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -195,12 +195,17 @@ def __call__(self, env, start_response):
 
         start_time = _time_ns()
 
-        token = context.attach(extract(env, getter=otel_wsgi.wsgi_getter))
+        token, ctx, span_kind = get_token_context_span_kind(
```
owais, line 198: I was thinking more like a wrapper function for `start_span` which internally takes care of checking server spans something like

```python
span = _start_internal_or_server_span(name, start_time, context_getter)
```

this function would accept whatever is necessary to figure out if an existing span is present or not and decide internally whether to create a server span or an internal span, and which context to use as parent.

ashu658: This function will need to return token also along with span.
```
start_internal_or_server_span(tracer, name, start_time, env, context_getter):
    ..
    create internal/server span
    ..
    
    return span, token
``` 
I am not sure if this is okay.
thinking of putting this function in `instrumentation/utils.py`


Also, we can shorten the recurring snippet if we want to skip the function
as the else part in the snippet is not necessary. Something like this:
```
token = context = None
span_kind = SpanKind.INTERNAL
if get_current_span() is INVALID_SPAN:
       context = extract(carrier, getter=carrier_getter)
       token = attach(context)
       span_kind = SpanKind.SERVER
```
Please share your thoughts.

owais: > This function will need to return token also along with span.

That sounds OK. It can return the span + token and let the caller store a reference to the token and detach it at a later time.

owais: > Also, we can shorten the recurring snippet if we want to skip the function

The main issue is that this is a very specific piece of logic strictly followed by all server instrumentation so I think it makes sense to encapsulate the logic in a re-usable function that is easily discoverable and changeable in a single place. Number of lines of code is not a big deal.

ashu658: Makes sense
Created wrapper for start_span.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -72,27 +72,40 @@ def unwrap(obj, attr: str):
         setattr(obj, attr, func.__wrapped__)
 
 
-def get_token_context_span_kind(env, getter):
-    """Based on presence of active span, extracts context and initializes token and span_kind
+def start_internal_or_server_span(
```
owais, line 75: should be private function so prefix with `_`

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -72,27 +72,40 @@ def unwrap(obj, attr: str):
...
+    tracer, span_name, start_time, env, context_getter
```
owais, line 76: `env` should be called `context`

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
ashu658: done

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
ashu658: changed to `context_carrier` as `context` is already imported module and context is extracted from `env` object.
Let me know if you want me to rename it.

owais:
Thanks @ashu658! Would be great to update other implementations to use the same function.

ashu658:
> Thanks @ashu658! Would be great to update other implementations to use the same function.

Sure @owais I will update other implementations in my next PR for tornado/other framework.

PR #871, opened 2022-01-20
add trailing_metadata to _OpenTelemetryServicerContext

jbraswell:
# Description

The trailing_metadata method was added to grpc._server.ServicerContext in gRPC v1.38.0, and we would like to use it with OpenTelemetry in our custom interceptors. To maintain support for earlier versions of gRPC, we throw a NotImplementedError if the subclass has not implemented `trailing_metadata`.

## Type of change

# How Has This Been Tested?

Manually validated this change works with our custom gRPC interceptors.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_server.py
```diff
@@ -114,6 +114,13 @@ def send_initial_metadata(self, *args, **kwargs):
     def set_trailing_metadata(self, *args, **kwargs):
         return self._servicer_context.set_trailing_metadata(*args, **kwargs)
 
+    def trailing_metadata(self):
+        # This method was added to grpc.ServicerContext in gRPC v1.38.0. We do
+        # this hasattr check to maintain support for earlier gRPC versions.
+        if hasattr(self._servicer_context, "trailing_metadata"):
+            return self._servicer_context.trailing_metadata()
+        raise NotImplementedError()
```
owais, line 122: If this method does not exist on the instrumented object, I think we should be raising AttributeError instead otherwise instrumentation here would result in changing behavior of the library. In other words, calling the method should result in the same error (may be with slightly different traceback) irrespective of whether the the library is instrumented or not. May be we should not even have hasattr check and just let it throw as it would normally. 

Another option is to sub-class this class for newer versions of grpc and use the appropriate version to wrap it based on grpc version being used.

jbraswell: Makes complete sense! I went ahead and removed the hasattr check.

owais (approved):

PR #872, opened 2022-01-24
fix: safe kafka partition extraction

nozik:
# Description

In some cases, which aren't consistently reproducible, the partition extraction of the kafka-python instrumentation fails. Since this isn't a crucial part of the instrumentation, we simply protect it with try/except to avoid a crash.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

A simple try/catch to prevent an issue that happened in our production environment.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
+++ b/instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
@@ -55,27 +55,36 @@ class KafkaPropertiesExtractor:
     @staticmethod
     def extract_send_partition(instance, args, kwargs):
         """extract partition `send` method arguments, using the `_partition` method in KafkaProducer class"""
-        topic = KafkaPropertiesExtractor.extract_send_topic(args)
-        key = KafkaPropertiesExtractor.extract_send_key(args, kwargs)
-        value = KafkaPropertiesExtractor.extract_send_value(args, kwargs)
-        partition = KafkaPropertiesExtractor._extract_argument(
-            "partition", 4, None, args, kwargs
-        )
-        key_bytes = instance._serialize(
-            instance.config["key_serializer"], topic, key
-        )
-        value_bytes = instance._serialize(
-            instance.config["value_serializer"], topic, value
-        )
-        valid_types = (bytes, bytearray, memoryview, type(None))
-        if (
-            type(key_bytes) not in valid_types
-            or type(value_bytes) not in valid_types
-        ):
+        try:
+            topic = KafkaPropertiesExtractor.extract_send_topic(args)
+            key = KafkaPropertiesExtractor.extract_send_key(args, kwargs)
+            value = KafkaPropertiesExtractor.extract_send_value(args, kwargs)
+            partition = KafkaPropertiesExtractor._extract_argument(
+                "partition", 4, None, args, kwargs
+            )
+            key_bytes = instance._serialize(
+                instance.config["key_serializer"], topic, key
+            )
+            value_bytes = instance._serialize(
+                instance.config["value_serializer"], topic, value
+            )
+            valid_types = (bytes, bytearray, memoryview, type(None))
+            if (
+                type(key_bytes) not in valid_types
+                or type(value_bytes) not in valid_types
+            ):
+                return None
+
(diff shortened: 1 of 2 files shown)


instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
```diff
@@ -55,27 +55,36 @@ def extract_send_headers(args, kwargs):
...
+                return None
+
+            all_partitions = instance._metadata.partitions_for_topic(topic)
+            if all_partitions is None or len(all_partitions) == 0:
+                return None
+
+            return instance._partition(
+                topic, partition, key, value, key_bytes, value_bytes
+            )
+        except Exception as exception:  # pylint: disable=W0703
```
owais, line 85: How does this affect users? In case this case does trigger, will we end up not recording a span, omitting some info from the span or will this affect the instrumented service or kafka client in some way?

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
```diff
@@ -55,27 +55,36 @@ def extract_send_headers(args, kwargs):
...
+            _LOG.debug("Unable to extract partition: %s", exception)
```
owais, line 86: should this be info/warn so we can actually find and fix the issue?

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
nozik: It will affect the instrumented service - hence the need to protect it with try/except

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
nozik: Since the impact of not collecting this data is low, I would not want to "bother" the user with a warning. I can go with info, but debug is also ok - whatever you think is best.

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
owais: I understand that. I'm asking how does this fix affect users? Will it result in missing spans, missing attributes on spans or something else entirely?

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
owais: It would be nice to know what exactly in this function is brittle so we can fix it later instead of wrapping it in a try/except and forgetting about it forever.

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
nozik: Missing attribute - the partition

owais:
Please update the changelog.

owais (approved):
I think we should find out the real reason for exceptions and add defensive checks around that but in the spirit of getting this into the next release, we can merge as is.

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
nozik: I definitely agree - that line that I've added`all_partitions = instance._metadata.partitions_for_topic(topic)` would've solved the crash. But I wanted to be on the safe side

nozik:
@owais Do I need to update the changelog, even though this instrumentation wasn't released yet?

owais:
Yes, please update changelog

nozik:
@owais Done

PR #877, opened 2022-01-26
add note about monthly cadence

### LINKED ISSUE #457: Can you add a release schedule to the readme?
When compared with the base `opentelemetry` repository, this one contains a lot of instrumentations and subpackages, and the rate at which bugs are fixed will keep being relatively bigger over time (more instrumentations, more bugs). After a fix for a specific instrumentation is out, there is no way to know when it will be in a release.

Define (or share) a schedule for releases. It does not have to be that often. The added clarity is valuable for people using the package.

Installing the instrumentation I need directly from github.

**Additional context**
This is mostly a documentation issue (if there is a release schedule in place already). If there is not a set schedule, we could define something.

codeboten:
Update the release section of the readme to include the monthly cadence of releases.

Fixes #457

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ on:
     - 'release/*'
   pull_request:
 env:
-  CORE_REPO_SHA: f0d559f5a8c40185e4cada6ed1914642b937f256
+  CORE_REPO_SHA: 15e457dd17bd1d9750a29fb0481c81e545e1bb71
 
 jobs:
   build:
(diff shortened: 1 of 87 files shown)


ocelotl (approved):

owais (approved):

lzchen (approved):

codeboten:
Closing this PR as it was done from a protected branch. See https://github.com/open-telemetry/opentelemetry-python-contrib/pull/879

Issue #862, opened 2022-01-13
Prototype retry/redirect logic for http instrumentations


lzchen:
As part of the effort to make http semantic conventions stable, this [otep](https://github.com/open-telemetry/oteps/pull/174) covers a list of topics required to be in scope for the first 1.0. Part of this is to support redirect logic as defined by the otep. To help get the approvals for this design, it is necessary to prototype this design to see if it is feasible in Python.

lzchen:
Closing as prototyping has finished.

PR #883, opened 2022-01-28
Prototype retry/redirect logic for urllib3 instrumentation

lzchen:
Part of https://github.com/open-telemetry/opentelemetry-python-contrib/issues/862

As part of an effort to push http semantic conventions to stable, this [otep](https://github.com/open-telemetry/oteps/pull/174) covers the outstanding http topics that are still being discussed and in scope for 1.0 release of http semantic conventions. The most notable topic is supporting creation of spans upon automatic redirect/retry for instrumentations that support them (for instrumentations that do not support automatic retry/redirect, our guidance to users is that the feature is not supported and they would have to add "semantic convention related attributes/spans" manually). Hence the "SHOULD key words in the specs (this was discussed in the instrumentation SIG).

This PR is a prototype implementation for the urllib3 library to make the instrumentation comply with the spec [pr](https://github.com/open-telemetry/opentelemetry-specification/pull/2078). Note a couple of things with urllib3:

1. Users can use [urllib3.PoolManager](https://urllib3.readthedocs.io/en/latest/reference/urllib3.poolmanager.html) or [urllib3.ConnectionPool](https://urllib3.readthedocs.io/en/latest/reference/urllib3.connectionpool.html#urllib3.HTTPConnectionPool) to make their requests. This implementation works for both ways.
2. redirects are essentially retries, but the code path for retries is done RECURSIVELY, whereas redirects are handled separately AFTER the initial response returns a 3XX. 
3. EDIT: From the [discussion](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/883#discussion_r796036493), all created redirect/retry spans will be siblings of the original span of the request. They will be correlated purely with LINKs.
4. This will be complicated if users want to use a higher level library such as `requests` that builds on top of urllib3 (as the implementation exists within urllib3 instrumentation and not requests instrumentation). In the future if this is implemented for requests, the logic will probably have to be duplicated, and suppression would probably have to be used in the underlying urllib3 library in the case both instrumentations are used.

Since this is a prototype, I am open to alternative suggestions in how to implement this. :)

Example cases exported to Jaeger:

Redirect scenario:
image

Retry scenario:
image

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
@@ -82,8 +82,9 @@ from opentelemetry.instrumentation.utils import (
 )
 from opentelemetry.propagate import inject
 from opentelemetry.semconv.trace import SpanAttributes
-from opentelemetry.trace import Span, SpanKind, get_tracer
+from opentelemetry.trace import INVALID_SPAN_CONTEXT, Link, Span, SpanKind, get_current_span, get_tracer
 from opentelemetry.trace.status import Status
+from opentelemetry.util.http import is_redirect
 from opentelemetry.util.http.httplib import set_ip_on_next_http_connection
 
 # A key to a context variable to avoid creating duplicate spans when instrumenting
@@ -156,7 +157,8 @@ def _instrument(
     url_filter: _UrlFilterT = None,
 ):
     def instrumented_urlopen(wrapped, instance, args, kwargs):
-        if _is_instrumentation_suppressed():
+        retry = kwargs.get("retries")
+        if _is_instrumentation_suppressed(retry and len(retry.history)):
             return wrapped(*args, **kwargs)
 
         method = _get_url_open_arg("method", args, kwargs).upper()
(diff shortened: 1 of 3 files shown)


owais:
> This means the trace tree for retries will look like a consecutive tree, in which each subsequent retry is child of the previous retry, but the trace tree for redirects will look like separate traces. 

In this case does the eventually successful request span become the only CLIENT span?

instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
```diff
@@ -170,26 +172,84 @@ def instrumented_urlopen(wrapped, instance, args, kwargs):
...
+            if prev_span_context is INVALID_SPAN_CONTEXT:
+                # If current span is root but we are in a retry scenario, this
+                # is a redirect scenario
+                prev_span_context = getattr(retry, "_OT_retry_prev_span_context", None)
+            if prev_span_context is not INVALID_SPAN_CONTEXT:
+                links = (Link(prev_span_context),)
+
         with tracer.start_as_current_span(
-            span_name, kind=SpanKind.CLIENT, attributes=span_attributes
+            span_name, kind=SpanKind.CLIENT, attributes=span_attributes, links=links,
```
owais, line 191: won't every retry span become a CLIENT span in this case? This would be fine if all the client spans were siblings but the PR description suggests in some cases the retry spans can be children of the previous retry attempts. In that case having a CLIENT span being a child of another CLIENT span could be a problem.

lzchen: That is a good point. I'd be open to setting the children retry attempts to `INTERNAL` maybe?

owais: That would be better than nested CLIENT spans but still not ideal. Ideally a CLIENT span should have a SERVER span as a child but not sure if Otel mandates that. Taking a step back, I feel the a better model might be to always have one INTERNAL span representing the overall request and then have each attempt as a child with kind set to CLIENT.  For example, in case where the request succeeds in the first attempt the trace would look like:


```
request (internal)
     |
     -- GET 200 (client)
         |
         --- server (SERVER)
```

and multiple attempts could look like:


```
request (internal)
     |
     -- GET / - 500 (client)
     |
     -- GET / - 500 (client)
     |
     -- GET / - 200 (client)
         |
         --- server (SERVER)
```

and a redirect could include server span/response as well



```
request (internal)
     |
     -- GET / - 302 (client)
     |
     -- GET /hello - 200 (client)
         |
         --- server (SERVER)
```

lzchen: Interesting. It looks like the dot net prototype is using CLIENT [spans](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2756/files) for each retry. Each span created from a retry/redirect is essentially a duplicate of the original request span (besides from the attributes/link defined in the spec). @denisivan0v any input on this parent/child relationship with these span kinds?

denisivan0v: > I feel the a better model might be to always have one INTERNAL span representing the overall request and then have each attempt as a child with kind set to CLIENT.

This is the valid point. There was a similar discussion within Instrumentation SIG https://github.com/open-telemetry/opentelemetry-specification/pull/2078#discussion_r755521314. Essentially, we have 4 options here:

- retry/redirect span being a child
- retry/redirect span created for a "physical"/CLIENT request only
- retry/redirect span created for a "logical"/INTERNAL request only
- retry/redirect span created for both "physical" and "logical" (+ adding the "logical" span in case there was a retry/redirect)

The idea of having spans for "physical" requests only + adding links to previous try/redirect span allows us to model it in a simplest way and avoid duplication - there will be no additional "logical"/INTERNAL span if case the initial request finished with 2xx (which is backward-compatible with the current behaivor). That said, it seems spans for "physical" requests _should_ be created as siblings to keep request-response semantics.

I'm in progress of adding these clarifications to the https://github.com/open-telemetry/opentelemetry-specification/pull/2078, please let me know if sounds reasonable to you.

The [PR for .NET](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2756/files) implements the mentioned proposal (so no duplicate of the original request span).

lzchen: From understanding, this proposal means that spans created from retry/redirect will have no parent/child relationship with each other or the original request, but will be LINKed together. @owais Thoughts?

owais: > From understanding, this proposal means that spans created from retry/redirect will have no parent/child relationship with each other or the original request, but will be LINKed together. @owais Thoughts?

That makes sense to me. Just to clarify, they'll still all be children of whatever the active span was before the first request attempt was made, right? What if there was no parent? Would we end up with N number of traces with N being number of attempts? May be we should create an encompassing span representing the whole operation to overcome this?

lzchen: > Would we end up with N number of traces with N being number of attempts

I think this would indeed be the case. Is there a huge need for this? I can foresee the implementation of that being kind of ugly, especially because we don't know that it is a retry/redirect scenario until after the second call is made.

owais: Imagine an auto-instrumented script or cronjob that is invoked and first (and possibly only) script it does is make an HTTP call to an external service. In case of retries/redirects, we'll end up with multiple traces with linked spans between them while as logically it was a single operation and should be a single trace.

lzchen: Would the encompassing span be of type INTERNAL?

lzchen: That is a good point. Personally I'm not fond of the idea of "automatically" creating spans that don't represent operations, it makes the already "magic-like" implementation have even more hidden behaviour. @denisivan0v has brought up some good points about how maybe an encompassing span is not totally necessary. Perhaps he can chime in more in this discussion. I was wondering if this was behaviour we had to decide/define now, or perhaps we can have the first iteration of this be a simple implementation and we can iterate on it later so as to push the spec along?

owais: I think we can totally iterate and leave these cases out as long as the feedback about these cases bubbles up to the spec level and it is addressed there eventually.

denisivan0v: Many thanks for this discussion, it helped to understand what should be more clarified in [the PR to the spec](https://github.com/open-telemetry/opentelemetry-specification/pull/2078).
I've re-worded the whole section to address these points, please have a look here https://github.com/open-telemetry/opentelemetry-specification/blob/bb265215d6b89dc0cfdae37f41123d6bb7986f9e/specification/trace/semantic_conventions/http.md#http-request-retries-and-redirects.

As for the encompassing span representing the whole operation, I've put it as 
```
No span is created corresponding to the "logical" (encompassing) request.
```
since it was a consensus within Instrumentation SIG folks to keep it as simple as possible and avoid any kind of hidden and automatic span creation.

lzchen: Nice! This makes sense to me. I have made the proposed changes as part of this PR.

srikanthccv:
@lzchen is this ready for review? or was it just a simple POC for something?

lzchen:
@srikanthccv 
This was simply a POC. I will close this for now until we decide we want to implement this.

Issue #454, opened 2021-04-15
WSGI:  Conditionally create SERVER spans


owais:
Part of #445

sanketmehta28:
@owais : can you please assign it to me?

sanketmehta28:
Hi @owais: you can close this one as [PR ](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/903) has been merged for the same

PR #903, opened 2022-02-05
code changes to resolve conditional server span creation for WSGI and adding 'attributes' parameter to util function

sanketmehta28:
…tps://github.com/open-telemetry/opentelemetry-python-contrib/issues/454)

# Description

This PR contains code changes for conditional server span creation for WSGI framework in case of another framework is present and it has already created SERVER span. In that case WSGI instrumentation will create a span of kind INTERNAL and make it a child of the SERVER span mentioned above.
Also I have added 'attributes' parameter to a util function _start_internal_or_server_span() util function to make it more compatible with multiple frameworks

Fixes # ([issue](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/454))

## Type of change

# How Has This Been Tested?

- [ ] Test A: Provide an unit test where a SERVER span was already exists in the context and check if the newly created span is of kind INTERNAL and is it a child of SERVER span mentioned before.

- [ ] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
@@ -280,11 +280,17 @@ class OpenTelemetryMiddleware:
             start_response: The WSGI start_response callable.
         """
 
-        token = context.attach(extract(environ, getter=wsgi_getter))
+        token = ctx = None
+        span_kind = trace.SpanKind.INTERNAL
+        if trace.get_current_span() is trace.INVALID_SPAN:
+            ctx = extract(environ, getter=wsgi_getter)
+            token = context.attach(ctx)
+            span_kind = trace.SpanKind.SERVER
 
         span = self.tracer.start_span(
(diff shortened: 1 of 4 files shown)


instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -280,11 +280,17 @@ def __call__(self, environ, start_response):
             start_response: The WSGI start_response callable.
         """
 
-        token = context.attach(extract(environ, getter=wsgi_getter))
```
owais, line 283: We aren't using the helper function we used for other packages here?

sanketmehta28: you mean _start_Internal_or_server_span() function?
No we cant as there is a missing argument in that function 'attributes' which is being used by start_span() function in wsgi framework.

owais (approved):

sanketmehta28:
@owais Can you please run the benchmark job again as this is flaky issue?

PR #908, opened 2022-02-09
feat(instrumentation-dbapi): add experimental sql commenter capability

lonewolf3739:
# Description

This PR adds support for an experimental and optional feature for enabling sql commenter capability to dbapi instrumentation. 

## Type of change

# How Has This Been Tested?

- [x] `tox -e test-instrumentation-dbapi`

- [x] No.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
@@ -45,11 +45,11 @@ import wrapt
 
 from opentelemetry import trace as trace_api
 from opentelemetry.instrumentation.dbapi.version import __version__
-from opentelemetry.instrumentation.utils import unwrap
+from opentelemetry.instrumentation.utils import unwrap, _generate_sql_comment
 from opentelemetry.semconv.trace import SpanAttributes
-from opentelemetry.trace import SpanKind, TracerProvider, get_tracer
+from opentelemetry.trace import Span, SpanKind, TracerProvider, get_tracer
 
-logger = logging.getLogger(__name__)
+_logger = logging.getLogger(__name__)
 
 
 def trace_integration(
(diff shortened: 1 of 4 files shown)


opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -115,3 +116,38 @@ def _start_internal_or_server_span(
         attributes=attributes,
     )
     return span, token
+
+
+_KEY_VALUE_DELIMITER = ","
+
+
+def _generate_sql_comment(**meta):
```
ocelotl, line 124: Is it really necessary to add this to `utils`? This is a single-use function, its contents can just be used directly where the function is called.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -115,3 +116,38 @@ def _start_internal_or_server_span(
         attributes=attributes,
     )
     return span, token
+
+
+_KEY_VALUE_DELIMITER = ","
```
ocelotl, line 121: Single-use constant

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -374,6 +397,16 @@ def traced_execution(
...
+            if args and self._commenter_enabled:
+                try:
+                    comment = self._generate_comment(span)
+                    if isinstance(args[0], bytes):
+                        comment = comment.encode("utf8")
+                    args_list = list(args)
+                    args_list[0] += comment
+                    args = tuple(args_list)
+                except Exception as exc:
+                    _logger.warning("Exception while generating sql comment: %s", exc)
```
ocelotl, line 409:
```suggestion
                    _logger.exception("Exception while generating sql comment: %s", exc)
```
Is [this](https://docs.python.org/3/library/logging.html#logging.Logger.exception) more appropriate?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
lonewolf3739: I think so, there will be many sql db instrumentations that will rely on this.

ocelotl (approved):

PR #925, opened 2022-02-17
Capture custom request/response headers for wsgi and change in passing response_headers in django, pyramid

### LINKED ISSUE #918: WSGI: Capture request/response headers as span attributes
part of #906

ashu658:
# Description
- This PR implements the [HTTP request/response headers as span attributes semantic convention](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md) for wsgi frameworks.
User can configure environment variable `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST ` and `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_RESPONSE ` to capture request/response headers respectively.
- Changes `django` Instrumentation to pass response headers instead of response object.
- Changes `pyramid` instrumentation to pass response headers as list.

Fixes #918 

## Type of change

# How Has This Been Tested?

Added unit tests for the changes. Also, tested captured headers manually with wsgi based framework by observing the output on jaeger backend.

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -204,10 +211,61 @@ def collect_request_attributes(environ):
...
+    )
+    for header_name in custom_request_headers:
+        wsgi_env_var = header_name.upper().replace("-", "_")
+        header_values = environ.get(f"HTTP_{wsgi_env_var}")
+
+        if header_values:
+            key = normalise_request_header_name(header_name)
+            attributes[key] = [
+                header_value.strip()
+                for header_value in header_values.split(",")
```
lonewolf3739, line 238: How does this work when there is , in header value?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -117,7 +117,14 @@ def response_hook(span: Span, environ: WSGIEnvironment, status: str, response_he
 from opentelemetry.propagators.textmap import Getter
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.trace.status import Status, StatusCode
-from opentelemetry.util.http import remove_url_credentials
+from opentelemetry.util.http import (
+    OTEL_PYTHON_CAPTURE_REQUEST_HEADERS,
+    OTEL_PYTHON_CAPTURE_RESPONSE_HEADERS,
```
lonewolf3739, line 122: Do you know if any other SIG is doing something similar `OTEL_{LANG}_CAPTURE_REQUEST_HEADERS`? I think we should be able to get consistent `OTEL_CAPTURE_REQUEST_HEADERS` accepted easily.

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
```diff
@@ -98,3 +101,23 @@ def remove_url_credentials(url: str) -> str:
     except ValueError:  # an unparseable url was passed
         pass
     return url
+
+
+def normalise_request_header_name(header):
```
lonewolf3739, line 106:
```suggestion
def normalise_request_header_name(header: str) -> str:
```

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
ashu658: Yes, in this case we won't get the correct value of the header. I chose ',' as delimiter because it is mentioned in the http spec that multiple header values should be concatenated with comma. ([link](https://stackoverflow.com/questions/3096888/standard-for-adding-multiple-values-of-a-single-http-header-to-a-request-or-resp))
But, I agree client can pass ',' in single header value. 
In case we don't want to split the header_values into multiple values we can just pass the whole string as a single element in the list. ([otel-spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers) allows that).
Please share your thoughts.

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
ashu658: Novice question, when is it recommended to use type hints?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
lonewolf3739: > either multiple header values as an array of strings or a single-item array containing a possibly comma-concatenated string, depending on the way the HTTP library provides access to headers.

Since the wsgi environ object gives the single value only, I think it's better we use the whole header value as attribute.

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
lonewolf3739: I wish there was some consistency for configuring the common use-cases of instrumentations across languages.

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
lonewolf3739: I would recommend we do this for all public apis. There are many reasons but I personally find them useful as a  documentation.

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
ashu658: Change done.

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -204,10 +211,48 @@ def collect_request_attributes(environ):
...
         result[SpanAttributes.HTTP_FLAVOR] = flavor
+    capture_custom_request_headers(
+        environ,
+        result,
+    )
 
     return result
 
 
+def capture_custom_request_headers(environ, attributes):
```
sanketmehta28, line 222: Any specific reason to pass the attribute dict as a argument and modifying it with new headers rather than returning the headers to calling function?
I think it will be better from readability perspective to update the result dict in a single function rather than passing to called function and update there

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -204,10 +211,48 @@ def collect_request_attributes(environ):
...
+    for header_name in custom_request_headers:
+        wsgi_env_var = header_name.upper().replace("-", "_")
+        header_values = environ.get(f"HTTP_{wsgi_env_var}")
+
+        if header_values:
+            key = normalise_request_header_name(header_name)
+            attributes[key] = [header_values]
+
+
+def capture_custom_response_headers(response_headers, span):
```
sanketmehta28, line 237: same comment as I have given in capture_custom_request_headers()

sanketmehta28 (changes requested):

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/open-telemetry/opentelemetry-python/compare/v1.9.1-0.28b1...HEAD)
 
+- `opentelemetry-instrumentation-wsgi` Capture custom request/response headers
```
lzchen, line 10:
```suggestion
- `opentelemetry-instrumentation-wsgi` Capture custom request/response headers in span attributes
```

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
```diff
@@ -15,9 +15,12 @@
 from os import environ
 from re import compile as re_compile
 from re import search
-from typing import Iterable
+from typing import Iterable, List
 from urllib.parse import urlparse, urlunparse
 
+OTEL_CAPTURE_REQUEST_HEADERS = "OTEL_CAPTURE_REQUEST_HEADERS"
```
lzchen, line 21: Do we want to further split this to client/server like Java does?

instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py
```diff
@@ -161,7 +161,7 @@ def trace_tween(request):
                 otel_wsgi.add_response_attributes(
                     span,
                     response_or_exception.status,
-                    response_or_exception.headers,
+                    response_or_exception.headerlist,
```
owais, line 164: Is this safe for all supported version of django?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
owais: Agree. It is not clear what the function does without reading the function body.

ashu658: Added suggested changes to improve readability. Please have a look.

instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py
ashu658: I was planning to add tests for all wsgi based frameworks in different PR.

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
ashu658: I think it is good to split as this will make the feature consistent between different sdks. Also, will keep the env vars same as java sdk.

Not sure but, when I first read the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers) it felt like its only talking about the server side. Please share your thoughts :)

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
ashu658: Changed env var for consistency with java sdk.

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
lzchen: Seems like we are going with the general approach for being consistent with other language sdks like [here](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/925#discussion_r814056927) so it would make sense to follow suit.

lzchen (approved):

instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py
owais: I was referring to replacing `headers` with `headerlist`. Is `headerlist` attribute available on all supported versions of django?

lzchen:
@sanketmehta28 
Have your comments been addressed?

instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py
ashu658: > available on all supported versions of django?

I think there is some confusion, the change `header` to `headerlist` is for pyramid :)
I checked [pyramid docs](https://docs.pylonsproject.org/projects/pyramid/en/1.7-branch/api/response.html#pyramid.response.Response.headerlist) `headerlist` attribute is present in all versions of pyramid we support (>=1.7).

sanketmehta28:
> 

@lzchen:yes I have checked it. They are addressed. I will approve this PR.

sanketmehta28:
> 

@lzchen:yes I have checked it. They are addressed. I will approve this PR.

sanketmehta28 (approved):

ashu658:
I forgot to add the condition to check for `SERVER` span before adding custom headers. Had to do some refactoring to add this condition. Please have a look. Apologies for late changes. 😓

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -208,6 +215,46 @@ def collect_request_attributes(environ):
...
+                key = normalise_request_header_name(header_name)
+                attributes[key] = [header_values]
+        span.set_attributes(attributes)
+
+
+def add_custom_response_headers(span, response_headers):
+    """Adds custom HTTP response headers into the sapn which are configured by the user from the
+    PEP3333-conforming WSGI environ as described in the specification
+    https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers"""
+    if span.kind == trace.SpanKind.SERVER:
```
sanketmehta28, line 240: This check should be outside of the add_custom_response_headers() and in the calling function before you call add_custom_response_headers(). 
i.e. 
if span.kind == SERVER:
            add_custom_response_headers()

And inside the function it should just add it to span attributes

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -208,6 +215,46 @@ def collect_request_attributes(environ):
...
+def add_custom_request_headers(span, environ):
+    """Adds custom HTTP request headers into the span which are configured by the user
+    from the PEP3333-conforming WSGI environ to be used as span creation attributes as described
+    in the specification https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers"""
+    if span.kind == trace.SpanKind.SERVER:
+        attributes = {}
+        custom_request_headers_name = get_custom_headers(
+            OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST
+        )
+        for header_name in custom_request_headers_name:
```
sanketmehta28, line 227: just same as response headers mentioned below

sanketmehta28 (changes requested):

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
sanketmehta28: for this use case you are right. but normally if there is an if condition which covers the whole function then that if condition should be outside of the function where that function is getting called, to make the function loosely coupled with that condition

ashu658: Okay. Will make the required changes.

ashu658: Changes done.

sanketmehta28 (approved):

instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
```diff
@@ -444,5 +461,119 @@ def test_mark_span_internal_in_presence_of_span_from_other_framework(self):
...
+            wsgi_with_custom_response_headers
+        )
+        response = app(self.environ, self.start_response)
+        self.iterate_response(response)
+        span = self.memory_exporter.get_finished_spans()[0]
+        expected = {
+            "http.response.header.content_type": (
+                "text/plain; charset=utf-8",
+            ),
+            "http.response.header.content_length": ("100",),
```
srikanthccv, line 546: Hmm, why is everything a sequence? it doesn't make sense that content_length is list, right?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -289,6 +336,8 @@ def __call__(self, environ, start_response):
             context_getter=wsgi_getter,
             attributes=collect_request_attributes(environ),
         )
+        if span.kind == trace.SpanKind.SERVER:
```
srikanthccv, line 339: Why is this needed? I thought feature is applicable to server instrumentations irrespective of span kind.

srikanthccv (commented):
mostly LGTM; few questions.

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
ashu658: In case of multiple server instrumentation (an app instrumented with both wsgi and flask/django) this check will avoid collection of custom headers in multiple spans.

instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
ashu658: This is part of the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers) actually.
>The attribute value MUST consist of either multiple header values as an array of strings or a single-item array containing a possibly comma-concatenated string, depending on the way the HTTP library provides access to headers.

srikanthccv: but that is for the multiple header values?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
srikanthccv: What should be the behaviour if the upstream server instrumentation injects some new headers which should be captured in downstream server instrumentation?

instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
ashu658: I think its for single values also. There is an example in the spec.
> http.request.header.content_type=["application/json"]

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
ashu658: Thats a good point. I didn't know if we allow user to add new headers during instrumentation.
If that is allowed, I think in that case we should let both spans to collect headers to avoid loss of info. 
There was a brief discussion with on slack for adding SERVER span check a while back [link](https://cloud-native.slack.com/archives/C01PD4HUVBL/p1636545699054600?thread_ts=1630559003.013900&cid=C01PD4HUVBL)
Please share your thoughts @srikanthccv @owais for the same.

srikanthccv: I am just thinking out loud here. We can revisit this if this becomes a problem for user.

instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
srikanthccv: That's strange but not a block for me.

srikanthccv (approved):

lzchen (approved):

Issue #902, opened 2022-02-04
lambdas triggered from dynamodb are failing


cowfoodbagman:
When lambdas are triggered from dyanmodb messages the event will look like `{"Records": [{"eventSource": "aws:dynamodb" ...}]}` 
https://docs.aws.amazon.com/lambda/latest/dg/with-ddb.html
As a result, I am seeing these error messages:
```
[ERROR] UnboundLocalError: local variable 'span_kind' referenced before assignment
Traceback (most recent call last):
  File "/opt/python/opentelemetry/instrumentation/aws_lambda/__init__.py", line 203, in _instrumented_lambda_handler_call
    kind=span_kind,
```
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py#L192-L196

PR #926, opened 2022-02-18
adding additional event sources

brett-bim:
# Description

Fixes [902](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/902).  Current behavior is not setting a default for SpanKind in the lambda library.  When an event arrives that is not of type SQS, no exception is triggered and `span_kind` is accessed later when not set.  

This PR does two things:

1. Sets a default of SpanKind.SERVER which seems to be the intent of the existing code.
2. Expands the concept of consumer beyond SQS to include any event source that might send a message to a lambda.  This seems consistent with the idea that the lambda is responding to an event, e.g. consuming that event even though it's not a classic queue based consumer.

Fixes # (issue)
https://github.com/open-telemetry/opentelemetry-python-contrib/issues/902

## Type of change

# How Has This Been Tested?

- [ ] Create an S3 bucket.  Create a Lambda that utilizes the lambda layer for OTel.  Subscribe the lambda to the bucket via an event that occurs when files are created in the bucket.  Upload a file to the bucket.  Verify that telemetry does not occur correctly and instead messages related to `span_kind` being accessed before it is set.  

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

- [ NA] Followed the style guidelines of this project
- [ Yes] Changelogs have been updated. 
- [ Yes] Unit tests have been added
- [ NA] Documentation has been updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased](https://github.com/open-telemetry/opentelemetry-python/compare/v1.9.1-0.28b1...HEAD)
+## [Unreleased](https://github.com/open-telemetry/opentelemetry-python/compare/v1.10.0-0.29b0...HEAD)
+
+## [1.10.0-0.29b0](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.10.0-0.29b0) - 2022-03-10
+
+
 
 - `opentelemetry-instrumentation-wsgi` Capture custom request/response headers in span attributes
   ([#925])(https://github.com/open-telemetry/opentelemetry-python-contrib/pull/925)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-instrumentation-wsgi` Capture custom request/response headers in span attributes
   ([#925])(https://github.com/open-telemetry/opentelemetry-python-contrib/pull/925)
 
+- `opentelemetry-instrumentation-flask` Flask: Capture custom request/response headers in span attributes
+  ([#952])(https://github.com/open-telemetry/opentelemetry-python-contrib/pull/952)
+
 ### Added
 
-
 - `opentelemetry-instrumentation-aws-lambda` `SpanKind.SERVER` by default, add more cases for `SpanKind.CONSUMER` services. ([#926](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/926))
-
+- `opentelemetry-instrumentation-sqlalchemy` added experimental sql commenter capability
+   ([#924](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/924))
 - `opentelemetry-instrumentation-dbapi` add experimental sql commenter capability
   ([#908](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/908))
 - `opentelemetry-instrumentation-requests` make span attribute available to samplers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+
+- `opentelemetry-instrumentation-aws-lambda` Added additional consumers. Fixes issue 902  
+  ([#926](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/926))
+
 - `opentelemetry-instrumentation-dbapi` add experimental sql commenter capability
   ([#908](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/908))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-instrumentation-requests` make span attribute available to samplers
   ([#931](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/931))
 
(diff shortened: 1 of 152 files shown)


instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -189,8 +189,14 @@ def _instrumented_lambda_handler_call(
 
         # See more:
         # https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html
+        # https://docs.aws.amazon.com/lambda/latest/dg/with-sns.html
+        # https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html
+        # https://docs.aws.amazon.com/lambda/latest/dg/with-ddb.html
+        span_kind = SpanKind.SERVER
```
srikanthccv, line 195: is there any change when this default is incorrect?

srikanthccv:
ping @NathanielRN

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
brett-bim: Right now the default is implicitly set to null which doesn't allow the code to even proceed past the usage of span_kind downstream.  Seems like the original intent was to look for SQS and call it a consumer if it didn't exist in the map.  I'm interested to hear @NathanielRN 's opinion though.

NathanielRN: Yes this was an oversight on my part, appreciate your deep dive into the code @brett-bim! In general we assumed that we wanted all spans to be Server spans.

(This is important for AWS X-Ray because only Server spans get converted to segments).

NathanielRN, line 195: Let me know what you think of these suggestions. We definitely want to fix all code branches even the default branch, but I like using the code to explain that.

```suggestion
span_kind = None
```

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -189,8 +189,14 @@ def _instrumented_lambda_handler_call(
...
+        valid_consumers = ["aws:sqs", "aws:s3", "aws:sns", "aws:dynamodb"]
+
         try:
-            if lambda_event["Records"][0]["eventSource"] == "aws:sqs":
+            if lambda_event["Records"][0]["eventSource"] in valid_consumers:
                 span_kind = SpanKind.CONSUMER
         except (IndexError, KeyError, TypeError):
             span_kind = SpanKind.SERVER
```
NathanielRN, line 202: What about this?

```python
try:
    if lambda_event["Records"][0]["eventSource"] in set(["aws:dynamodb", "aws:s3", "aws:sns", "aws:sqs"]):
        # See more:
        # https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html
        # https://docs.aws.amazon.com/lambda/latest/dg/with-sns.html
        # https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html
        # https://docs.aws.amazon.com/lambda/latest/dg/with-ddb.html
        span_kind = SpanKind.CONSUMER
    else:
        span_kind = SpanKind.SERVER
except (IndexError, KeyError, TypeError):
    span_kind = SpanKind.SERVER
```

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+
+- `opentelemetry-instrumentation-aws-lambda` Added additional consumers. Fixes issue 902
```
NathanielRN, line 13: I don't think we normally include the issue number in the CHANGELOG? The PR description will have it. Probably best to remove in this case.

```suggestion
- `opentelemetry-instrumentation-aws-lambda` `SpanKind.SERVER` by default,  add more cases for `SpanKind.CONSUMER` services
```

NathanielRN (approved):
Thanks for this! Just small comments but it's good as is for the fix.

Do you think you could add a test? I thought I had one but I guess I never did for that code path.

The `instrumentation/opentelemetry-instrumentation-aws-lambda/tests/test_aws_lambda_instrumentation_manual.py` file should be completely set up for your to do that. You can copy a test like [the `test_parent_context_from_lambda_event` test](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/1080b04902ff689e7a3e5838c99e90f68119b273/instrumentation/opentelemetry-instrumentation-aws-lambda/tests/test_aws_lambda_instrumentation_manual.py#L167-L174) to create a SQS/DynamoDB/etc. test event.

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
brett-bim: I like this.

brett-bim:
I think I've addressed @NathanielRN 's comments and suggestions.

brett-bim:
Seems that several of the actions failed after I added a test.  I was able to execute the test locally with pytest both before my changes and after to prove it worked.  As I and several others have noted in the Slack channel, I haven't been able to follow the set up documentation to run tox locally.  Not sure how to proceed.

srikanthccv:
There were some failures outside your changes. That should be fixed now.

srikanthccv (approved):
@brett-bim please sign the cla

brett-bim:
CLA is signed.

ocelotl:
@brett-bim please fix conflicts so we can merge

brett-bim:
@ocelotl fixed in my fork.

PR #900, opened 2022-02-03
datadog-apm: deprecate datadog exporter

mabdinur:
# Description

Adds deprecation note to the Datadog Span Exporter.

## Type of change

Updates Readme

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

N/A

- [ ] Test A

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [ ] Followed the style guidelines of this project
- [ ] Changelogs have been updated
- [ ] Unit tests have been added
- [ ] Documentation has been updated

PR #932, opened 2022-02-28
Make the DatadogExportSpanProcessor re-initialize itself after fork

phillipuniverse:
# Description

The `DatadogExportSpanProcessor ` suffered from the same problems as the [`BatchSpanProcessor`](https://github.com/open-telemetry/opentelemetry-python/pull/2242). When using the `DatadogSpanProcessor` I would see my background workers initialized with `os.fork()` deadlock. I observed this in RQ but I'm sure the same types of problems would happen in Celery.

## Type of change

# How Has This Been Tested?

I effectively copied over the test from https://github.com/open-telemetry/opentelemetry-python/pull/2242. Test passes when re-initializing on fork, fails without it.

I haven't yet put this live into my project because there are a few questions I want to ask about the particulars of this implementation.

- [x] No

# Checklist:

### DIFF
--- a/exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/spanprocessor.py
+++ b/exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/spanprocessor.py
@@ -14,6 +14,7 @@
 
 import collections
 import logging
+import os
 import threading
 import typing
 
@@ -67,7 +68,7 @@ class DatadogExportSpanProcessor(SpanProcessor):
         self.traces_spans_count = collections.Counter()
         self.traces_spans_ended_count = collections.Counter()
 
-        self.worker_thread = threading.Thread(target=self.worker, daemon=True)
+        self.worker_thread = self._create_worker_thread()
 
         # threading conditions used for flushing and shutdown
         self.condition = threading.Condition(threading.Lock())
@@ -81,6 +82,24 @@ class DatadogExportSpanProcessor(SpanProcessor):
         self.schedule_delay_millis = schedule_delay_millis
         self.done = False
         self.worker_thread.start()
+        # Only available in *nix since py37.
+        if hasattr(os, "register_at_fork"):
+            os.register_at_fork(
+                after_in_child=self._at_fork_reinit
+            )  # pylint: disable=protected-access
+
+    def _at_fork_reinit(self):
+        self.condition = threading.Condition(threading.Lock())
+        self.flush_condition = threading.Condition(threading.Lock())
+        self.traces_lock = threading.Lock()
+        self.check_traces_queue.clear()
+        self.worker_thread = self._create_worker_thread()
+        self.worker_thread.start()
+
+    def _create_worker_thread(self) -> threading.Thread:
+        return threading.Thread(
+            name="DatadogProcessor", target=self.worker, daemon=True
+        )
 
     def on_start(
         self, span: Span, parent_context: typing.Optional[Context] = None
--- a/exporter/opentelemetry-exporter-datadog/tests/test_datadog_exporter.py
+++ b/exporter/opentelemetry-exporter-datadog/tests/test_datadog_exporter.py
@@ -676,7 +676,6 @@ class TestDatadogSpanExporter(unittest.TestCase):
 
 
 class TestDatadogSpanProcessorConcurrency(ConcurrencyTestBase):
-
     @unittest.skipUnless(
         hasattr(os, "fork") and sys.version_info >= (3, 7),
         "needs *nix and minor version 7 or later",
--- a/exporter/opentelemetry-exporter-datadog/tests/test_datadog_exporter.py
+++ b/exporter/opentelemetry-exporter-datadog/tests/test_datadog_exporter.py
@@ -14,6 +14,8 @@
 
 import itertools
 import logging
+import multiprocessing
+import os
 import sys
 import time
 import unittest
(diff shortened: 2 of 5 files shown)


exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/spanprocessor.py
```diff
@@ -81,6 +82,24 @@ def __init__(
...
+                after_in_child=self._at_fork_reinit
+            )  # pylint: disable=protected-access
+
+    def _at_fork_reinit(self):
+        self.condition = threading.Condition(threading.Lock())
+        self.flush_condition = threading.Condition(threading.Lock())
+        self.traces_lock = threading.Lock()
+        self.check_traces_queue.clear()
+        self.worker_thread = self._create_worker_thread()
+        self.worker_thread.start()
```
phillipuniverse, line 97: I really wasn't too sure here what I should/shouldn't re-initialize. I figured I would initialize all of it just in case

exporter/opentelemetry-exporter-datadog/tests/test_datadog_exporter.py
```diff
@@ -667,3 +673,50 @@ def test_service_name_fallback(self):
 
         span = datadog_spans[0]
         self.assertEqual(span["service"], "fallback_service_name")
+
+    @unittest.skipUnless(
+        hasattr(os, "fork") and sys.version_info >= (3, 7),
+        "needs *nix and minor version 7 or later",
+    )
+    def test_exports_with_fork(self):
```
phillipuniverse, line 681: This test is effectively the same sort of one from https://github.com/open-telemetry/opentelemetry-python/pull/2242 with as few changes as possible.

exporter/opentelemetry-exporter-datadog/tests/test_datadog_exporter.py
```diff
@@ -667,3 +673,50 @@ def test_service_name_fallback(self):
...
+            self.run_with_many_threads(_target, 100)
+
+            time.sleep(0.5)
+
+            spans = get_spans(tracer, exporter, shutdown=False)
+            conn.send(len(spans) == 200)
+            conn.close()
+
+        parent_conn, child_conn = multiprocessing.Pipe()
+        fork_context = multiprocessing.get_context("fork")
```
phillipuniverse, line 716: I'm running this code on macOS where [the default is `spawn`](https://github.com/open-telemetry/opentelemetry-python/pull/2242). With that default I got a 'Can't pickle local object' when the created `Process` linked to this inner `child` function. I didn't really want to move out `child` but figured this small detail didn't matter much since the test should behave the same on all systems.

tox.ini
```diff
@@ -360,7 +360,7 @@ commands_pre =
 
   aiopg: pip install {toxinidir}/instrumentation/opentelemetry-instrumentation-dbapi pip install {toxinidir}/instrumentation/opentelemetry-instrumentation-aiopg[test]
 
-  datadog: pip install flaky {toxinidir}/exporter/opentelemetry-exporter-datadog[test]
```
phillipuniverse, line 363: Made a bit more sense to me to keep this `flaky` dependency in the `extras_require` of the module instead of here in `tox.ini`.

owais:
@phillipuniverse I'm pretty sure this package has been deprecated in favor of DataDog's official library (https://github.com/DataDog/dd-trace-py/) so not sure if we should invest considerable effort in maintaining/reviewing this package. 

@open-telemetry/opentelemetry-python-contrib-approvers  Perhaps we should add a deprecation notice to this package and re-direct to the official DataDog solution.

@majorgreys could you please confirm if this package is deprecated and if so, is `DataDog/dd-trace-py` the right repo to redirect users to?

phillipuniverse:
@owais interesting context but dd-trace-py does not have any OpenTelemetry exporters which would take an otel trace and convert it to a valid DataDog trace. But there is code there that takes an OpenTracing trace and converts it to Datadog - maybe this exporter belongs there too 🤷 .

If this package should truly be deprecated then we should probably also remove these issues as they probably don't apply anymore and should likely be closed:

- https://github.com/DataDog/dd-trace-py/issues/2269
- https://github.com/open-telemetry/opentelemetry-python/issues/2066

Since I started using OpenTelemetry and Datadog a few months ago, I have seen that the [official docs](https://docs.datadoghq.com/tracing/setup_overview/open_standards/) recommend using the OpenTelemetry collector to do the transformation into the Datadog format via an exporter. It would be great to clear up the discrepancy - I came into the ecosystem a few months ago and thought that using the service exporter directly into the Datadog Agent was the "blessed" path. Deprecating this opentelemetry-python-contrib exporter and instead guiding users to user the datadog-specific exporter at the OpenTelemetry collector level would make things a lot clearer.

Thanks for the consideration!

owais:
@phillipuniverse it seems DataDog recommends sending native OTLP to DataDog agent instead of using exporters such as these. 

https://docs.datadoghq.com/tracing/setup_overview/open_standards/#otlp-ingest-in-datadog-agent

https://github.com/open-telemetry/opentelemetry-python-contrib/pull/900
https://github.com/open-telemetry/opentelemetry-python/pull/2517

phillipuniverse:
@owais thanks for your note about the `DatadogExportSpanProcessor` being effectively deprecated. I have since converted my services to use the `BatchSpanProcessor` and that has fixed my issues.

I am currently still utilizing the Datadog agent to receive traces, but this I run it with the `DD_OTLP_GRPC_PORT` which allows me to receive OpenTelemetry traces:

```
docker run -e DD_OTLP_GRPC_PORT=4317 -e DD_APM_CONFIG_AMP_NON_LOCAL_TRAFFIC=1 -e DD_API_KEY=<<some-api-key>> -p 8125:8125/udp -p 8126:8126 -p 4317:4317 --rm datadog/agent:latest
```

> @opentelemetry-python-contrib-approvers Perhaps we should add a deprecation notice to this package and re-direct to the official DataDog solution.

This would be really nice and would definitely help avoid confusion in the future.

Thanks again!

PR #942, opened 2022-03-03
Aiohttp server instrumentation

dmanchon:
# Description

Instrument aiohttp server. Right now there is only instrumentation for the client.

## Type of change

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+async def middleware(request, handler):
+    """Middleware for aiohttp implementing tracing logic"""
+    if (
+        context.get_value("suppress_instrumentation")
+        or context.get_value(_SUPPRESS_HTTP_INSTRUMENTATION_KEY)
+        or _excluded_urls.url_disabled(request.url)
+    ):
+        return await handler(request)
+
+    token = context.attach(extract(request, getter=getter))
```
kennytrytek-wf, line 145: `context.attach` here is incorrect. The request being handled is from an external service, and the context contained within that request is a representation of the state of the external service. It doesn't make sense to set the external service's context as the current context for the service running this middleware.

The context should still be extracted from the request and supplied to `tracer.start_as_current_span`, but the `attach` and `detach` can be removed.

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+    span_name, additional_attributes = get_default_span_details(request)
+
+    with tracer.start_as_current_span(
+        span_name,
+        kind=trace.SpanKind.SERVER,
+    ) as span:
+        if span.is_recording():
```
kennytrytek-wf, line 152: I think this `if` statement should be removed. In the case that this is a non-recording span, `set_attribute` is a no-op. Additionally, this middleware should _always_ call the request handler (line 158).

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+            attributes = collect_request_attributes(request)
+            attributes.update(additional_attributes)
+            for key, value in attributes.items():
+                span.set_attribute(key, value)
```
kennytrytek-wf, line 156: Replace with `span.setattributes(attributes)`.

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+            try:
+                resp = await handler(request)
+                set_status_code(span, resp.status)
+            finally:
+                context.detach(token)
```
kennytrytek-wf, line 161:
```suggestion
            resp = await handler(request)
            set_status_code(span, resp.status)
```

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+getter = AiohttpGetter()
+
+
+@web.middleware
+async def middleware(request, handler):
+    """Middleware for aiohttp implementing tracing logic"""
+    if (
+        context.get_value("suppress_instrumentation")
+        or context.get_value(_SUPPRESS_HTTP_INSTRUMENTATION_KEY)
+        or _excluded_urls.url_disabled(request.url)
```
kennytrytek-wf, line 141: `request.url` is not a `str`. This should be `request.url.path`.

bmbouter:
@dmanchon I'm interested in testing and maybe helping to finish this PR? How can I test this PR after it's checked out? I have an aiohttp based app to test with. Also what do you know that needs finishing? The reviewer comments look all pretty straightforward, so after that what else is left?

instrumentation/opentelemetry-instrumentation-aiohttp-server/setup.cfg
```diff
@@ -0,0 +1,58 @@
...
+[options]
+python_requires = >=3.6
+package_dir=
+    =src
+packages=find_namespace:
+install_requires =
+    opentelemetry-api ~= 1.3
+    opentelemetry-semantic-conventions == 0.28b1
+    opentelemetry-instrumentation == 0.28b1
+    opentelemetry-util-http == 0.28b1
```
kennytrytek-wf, line 46: These requirements should be updated to the latest version

instrumentation/opentelemetry-instrumentation-aiohttp-server/setup.cfg
```diff
@@ -0,0 +1,58 @@
...
+
+[options.packages.find]
+where = src
+
+[options.extras_require]
+test =
+    opentelemetry-test-utils == 0.28b1
```
kennytrytek-wf, line 53: Update to latest version

dralley:
> I remember one of my concerns was that aiohttp is both server and client, given that client already exists i wonder if this should live in the same module as the client or as an independent module, i guess if we want autoinstrumentation based on the installed packages it will end up installing both server and client modules.

Both server and client in one package is probably the best option by virtue of being simple + not real upsides of separating it.  It's ultimately not that much extra code.

decko:
@dmanchon correct me if I'm wrong, but I didn't see any code for metrics. Is that right? Would be interesting to have it in this PR?
Also, would be possible for contributing to this PR specifically? Or we need to open a PR to your fork for then to me merged into this one?

dralley:
@decko I would just open a new PR from your own branch which has these commits

decko:
> @decko I would just open a new PR from your own branch which has these commits

What if we open it if we don't get a reply in two days?
In any case, I'm continuing the work from where @dmanchon left.

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
```diff
@@ -0,0 +1,184 @@
...
+
+    span_name, additional_attributes = get_default_span_details(request)
+
+    with tracer.start_as_current_span(
+        span_name,
+        kind=trace.SpanKind.SERVER,
+    ) as span:
+        attributes = collect_request_attributes(request)
+        attributes.update(additional_attributes)
+        span.setattributes(attributes)
```
aleuk0, line 153: little misspelling 

```suggestion
        span.set_attributes(attributes)
```

lzchen:
Duplicate of https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1800

Issue #881, opened 2022-01-27
New Pika 0.28b0 instrumentation requires packaging, not declared in transitive dependencies


phillipuniverse:
**Describe your environment**

I just upgraded `opentelemetry-instrumentation-pika` in my environment to 0.28b0. The commit at 26aa17f8e93c43e80a204cde14d6bac680918c27 and #837 introduced [this line](https://github.com/open-telemetry/opentelemetry-python-contrib/commit/26aa17f8e93c43e80a204cde14d6bac680918c27#diff-366eb17355e97103aefd16dc014927d69f78b2e4650c362b950c206ecaa34465R19):

```
from packaging import version
```

This results in the following in my environment:

```
ModuleNotFoundError: No module named 'packaging'
```

I assume the problem is that I don't have any of the `setuptools` or related dependencies in my environment since I'm deploying a webapp on Docker. But I would have expected at least a new transitive dependency or something in order to use this functionality.

My environment is Python `3.7.12` and I am using the Docker image `FROM python:3.7.12-buster`.

This fortunately doesn't break my whole environment, but I do get the `ModuleNotFoundError: Instrumenting of pika failed` message.

phillipuniverse:
The fix is to add `packaging` into your venv.

PR #1000, opened 2022-03-15
Add falcon version 1.4.1 support to opentelemetry-instrumentation-falcon

rjduffner:
# Description

The largest change between falcon 1.x and 2.x was the changing the placement of some exception args. This PR adds some logic to place those args in the correct places for all versions of falcon.

## Type of change

# How Has This Been Tested?
- Made sure tests pass
- Forked and ran against falcon instance internally.

- [X] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
@@ -326,14 +326,13 @@ class _TraceMiddleware:
 
             # Falcon 1 does not support response headers. So
             # send an empty dict.
-            if _falcon_version == 1:
-                _response_headers_dict = {}.items()
-            else:
-                _response_headers_dict = resp.headers.items()
+            response_headers = {}
+            if _falcon_version > 1:
+                response_headers = resp.headers
 
             if span.is_recording() and span.kind == trace.SpanKind.SERVER:
                 otel_wsgi.add_custom_response_headers(
-                    span, _response_headers_dict
+                    span, response_headers.items()
                 )
         except ValueError:
             pass
--- a/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
@@ -95,6 +95,7 @@ from sys import exc_info
 from typing import Collection
 
 import falcon
+from packaging import version as package_version
 
 import opentelemetry.instrumentation.wsgi as otel_wsgi
 from opentelemetry import context, trace
@@ -126,12 +127,19 @@ _ENVIRON_EXC = "opentelemetry-falcon.exc"
 
 _response_propagation_setter = FuncSetter(falcon.Response.append_header)
 
-if hasattr(falcon, "App"):
+_parsed_falcon_version = package_version.parse(falcon.__version__)
+if _parsed_falcon_version >= package_version.parse("3.0.0"):
     # Falcon 3
     _instrument_app = "App"
-else:
+    _falcon_version = 3
+elif _parsed_falcon_version >= package_version.parse("2.0.0"):
     # Falcon 2
     _instrument_app = "API"
+    _falcon_version = 2
+else:
+    # Falcon 1
+    _instrument_app = "API"
+    _falcon_version = 1
 
 
 class _InstrumentedFalconAPI(getattr(falcon, _instrument_app)):
@@ -163,12 +171,30 @@ class _InstrumentedFalconAPI(getattr(falcon, _instrument_app)):
         super().__init__(*args, **kwargs)
 
     def _handle_exception(
-        self, req, resp, ex, params
+        self, arg1, arg2, arg3, arg4
     ):  # pylint: disable=C0103
         # Falcon 3 does not execute middleware within the context of the exception
         # so we capture the exception here and save it into the env dict
+
+        # Translation layer for handling the changed arg position of "ex" in Falcon > 2 vs
+        # Falcon < 2
+        if _falcon_version == 1:
+            ex = arg1
+            req = arg2
+            resp = arg3
+            params = arg4
+        else:
+            req = arg1
+            resp = arg2
+            ex = arg3
+            params = arg4
+
         _, exc, _ = exc_info()
         req.env[_ENVIRON_EXC] = exc
+
+        if _falcon_version == 1:
+            return super()._handle_exception(ex, req, resp, params)
+
         return super()._handle_exception(req, resp, ex, params)
 
     def __call__(self, env, start_response):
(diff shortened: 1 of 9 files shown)


rjduffner:
Do Not Merge:

Breaks tests in falcon 3.x

In falcon 3, not setting remote_addr leads to an error where net.peer.ip is not set in the span.

It seems related to a change where ASGI is used in tests in falcon3 and WSGI is used in falcon < 3. I am hoping this is a test only error related to the simulated request but unsure.

instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
```diff
@@ -101,17 +99,20 @@ def _test_method(self, method):
...
-                SpanAttributes.NET_PEER_IP: "127.0.0.1",
                 SpanAttributes.NET_PEER_PORT: "65133",
                 SpanAttributes.HTTP_FLAVOR: "1.1",
                 "falcon.resource": "HelloWorldResource",
                 SpanAttributes.HTTP_STATUS_CODE: 201,
             },
         )
+        if SpanAttributes.NET_PEER_IP in span.attributes:
+            self.assertEqual(
+                span.attributes[SpanAttributes.NET_PEER_IP], "127.0.0.1"
```
ocelotl, line 110: Curious, if `remote_addr="127.0.0.1"` was removed from the call to `simulate_request` then why should it appear here? Is this a default value? If so, it would be good to test also a non-default value.

rjduffner: @ocelotl,

In falcon 1.4.1 (the new version), the remote_addr param doesn't exist. This to me wasn't a huge deal as in the tests, it was being tested with the default value anyways, '127.0.0.1'.

However in falcon 3.x, there was an update to the code that was made which doesn't set the remote_addr if the default value is used. (See my comment above) or

https://github.com/falconry/falcon/blob/2.0.0/falcon/testing/helpers.py#L176
vs
https://github.com/falconry/falcon/blob/3.0.1/falcon/testing/helpers.py#L1168-L1172


 This is different to how it was done in falcon 2.x where the remote_addr was always set, just with the default value.
 
 The if statement in the test here was a compromise to show that there is something strange going on between the different versions.
 
 If you have any thoughts on this, please let me know. I am open to suggestions.

rjduffner:
Other note, I see the lint is failing, I will work on this tomorrow.

lzchen:
@owais 
PTAL

instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
ocelotl: Ok, that's fine, can you please add this explanation to the test as a comment? :v:

rjduffner: Added. Thank you

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -297,9 +323,17 @@ def process_response(
...
                 )
             )
+
+            # Falcon 1 does not support response headers. So
+            # send an empty dict.
+            if _falcon_version == 1:
+                _response_headers_dict = {}.items()
+            else:
+                _response_headers_dict = resp.headers.items()
+
```
srikanthccv, line 333: nit: assign default empty list and replace when version > 1

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -297,9 +323,17 @@ def process_response(
...
             if span.is_recording() and span.kind == trace.SpanKind.SERVER:
                 otel_wsgi.add_custom_response_headers(
-                    span, resp.headers.items()
+                    span, _response_headers_dict
```
srikanthccv, line 336: nit: this is list
```suggestion
                    span, _response_headers_list
```

instrumentation/opentelemetry-instrumentation-falcon/tests/app.py
```diff
@@ -46,12 +47,17 @@ def on_get(self, _, resp):
...
         # Falcon 3
         app = falcon.App()
-    else:
+    elif _parsed_falcon_version >= package_version.parse("2.0.0"):
         # Falcon 2
         app = falcon.API()
+    else:
+        # Falcon 1
+        app = falcon.API()
+
```
srikanthccv, line 60: do you need this branch? v1 and v2 have the same `falcon.API()`.

srikanthccv (commented):
Some comments, LGTM.

instrumentation/opentelemetry-instrumentation-falcon/tests/app.py
rjduffner: Nice call, carry over from earlier code. Changed.

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
rjduffner: Refactored here. Its technically not a dict or list. Its a dict_list. Moved the .items() back to where it was to make it more clear. (hopefully)

srikanthccv (approved):

rjduffner:
@srikanthccv Not sure what to do here. Lint locally seems to be failing due to this commit, https://github.com/open-telemetry/opentelemetry-python-contrib/commit/40c1805b1123072bfec7d5348bef0ae0a6055a2a

Specifically this

```************* Module tests.test_system_metrics
instrumentation/opentelemetry-instrumentation-system-metrics/tests/test_system_metrics.py:25:0: E0611: No name 'InMemoryMetricReader' in module 'opentelemetry.sdk._metrics.export' (no-name-in-module)```


And lint on the runner seems to have a different issue

``` TypeError: __init__() got an unexpected keyword argument 'styles'```

is there a re-run command for lint?

**Comment by srikanthccv** [comment]
If it's not related to your changes please ignore. We will get that fixed in separate PR.

**Comment by rjduffner** [comment]
Thanks @srikanthccv

File: instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py [hunk 5]
```diff
@@ -146,6 +146,7 @@ def response_hook(span, req, resp):
 from typing import Collection
 
 import falcon
+from packaging import version as package_version
```
owais, line 149: Is this part of all python distributions? If not, we'll need to add it as a dependency to setup.cfg

owais (approved):

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
srikanthccv: Good catch https://github.com/open-telemetry/opentelemetry-python-contrib/issues/881

srikanthccv (changes requested):
Blocking till we resolve the transitive dep comment

srikanthccv:
@rjduffner Please add `packaging` dependency and we are good to merge.

rjduffner:
@srikanthccv packaging dependency has been added to the setup.cfg.

srikanthccv (approved):

PR #1001, opened 2022-03-15
Pyramid handle non-HTTPException

### LINKED ISSUE #888: Pyramid tween only handles HTTPException
**Describe your environment** 
Seeing this exhibited in Pyramid 1.10 and as far back as Pyramid 1.7.

**Steps to reproduce**
Raise a non-HTTPException. 

Tweens should process normally.

An error for use of `response_or_exception` before assignment is raised.

**Additional context**
The problem exists here https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py#L146-L157. It seems there should be an additional exception case for `BaseException` to initialize `response_or_exception`.

gregbuehler:
# Description

When raising non-HTTPExceptions, treat them as internal server errors for the purposes of tracing.

Fixes #888

## Type of change

# How Has This Been Tested?

Adds a new unit test that raises a non-HTTPException.

- [x ] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py
+++ b/instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py
@@ -139,7 +139,7 @@ def trace_tween_factory(handler, registry):
 
         response = None
         status = None
-        
+
         try:
             response = handler(request)
         except HTTPException as exc:
--- a/instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py
+++ b/instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py
@@ -12,12 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from lib2to3.pytree import Base
 from logging import getLogger
 from xmlrpc.client import ResponseError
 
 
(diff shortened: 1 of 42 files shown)


instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py
```diff
@@ -158,10 +166,16 @@ def trace_tween(request):
...
+                        status = response.status
+                    if hasattr(response, "headerList"):
+                        headerList = response.headerList
+
                 otel_wsgi.add_response_attributes(
                     span,
-                    response_or_exception.status,
-                    response_or_exception.headerlist,
+                    status,
+                    headerList,
```
srikanthccv, line 178: Can these `status` and `headerList` ever be `None`? I would be defensive and make sure to call `add_response_attributes` only when status is set to non null value.

lzchen (approved):

instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py
gregbuehler: Agreed status should be guarded. I believe `headerList` can feasibly be `None`, but as of right now `add_response_attributes` does not actually handle the header list.

I'm updating things to be a bit more defensive overall.

srikanthccv (approved):

Issue #919, opened 2022-02-11
ASGI: Capture request/response headers as span attributes


ashu658:
part of #906

sanketmehta28:
Hi @owais , Can you assign this to me? I had a discussion with @ashu658 . I will look into this issue.

sanketmehta28:
Hi @ocelotl : Can you please close this as PR is meged?

Issue #998, opened 2022-03-15
AttributeError: 'NonRecordingSpan' object has no attribute 'kind'


seanpile:
Recent PR #952 seems to have introduced a regression in our code; when using a non-recording span, the flask instrumentation is breaking.  See the stacktrace below:

```
  File "lib/python3.9/site-packages/werkzeug/wrappers/response.py", line 632, in __call__
    start_response(status, headers)
  File "lib/python3.9/site-packages/opentelemetry/instrumentation/flask/__init__.py", line 156, in _start_response
    if span.kind == trace.SpanKind.SERVER:
AttributeError: 'NonRecordingSpan' object has no attribute 'kind'
```

My expectation is that using a non-recording span should be fine to use safely, and not cause downstream errors.  I'm using the following libraries:

- opentelemetry-api==1.10.0                                                                                                                                                      
- opentelemetry-exporter-otlp-proto-grpc==1.10.0                                                                                                                                 
- opentelemetry-instrumentation-flask==0.29b0                                                                                                                                    
- opentelemetry-instrumentation-requests==0.29b0                                                                                                                                 
- opentelemetry-instrumentation-sqlalchemy==0.29b0                                                                                                                               
- opentelemetry-sdk==1.10.0                                                                                                                                                      
- opentelemetry-semantic-conventions==0.29b0

srikanthccv:
Note to people who run into this issue: did a hot fix release for affected packages.

Issue #958, opened 2022-03-10
Audit all packages to ensure NoOpTracerProvider works


codeboten:
An issue was found in the wsgi instrumentation by integration tests in the core repository. The issue should have been caught by unit tests in the instrumentation using the NoOpTracerProvider.

The outcome of this issue is to ensure each instrumentation contains a test using the NoOpTracerProvider and that it does not rely on the OpenTelemetry SDK:

- [ ] Audit and test opentelemetry-instrumentation-aiohttp-client
- [ ] Audit and test opentelemetry-instrumentation-aiopg
- [ ] Audit and test opentelemetry-instrumentation-asgi
- [ ] Audit and test opentelemetry-instrumentation-asyncpg
- [ ] Audit and test opentelemetry-instrumentation-aws-lambda
- [ ] Audit and test opentelemetry-instrumentation-boto
- [ ] Audit and test opentelemetry-instrumentation-botocore
- [ ] Audit and test opentelemetry-instrumentation-celery
- [ ] Audit and test opentelemetry-instrumentation-dbapi
- [ ] Audit and test opentelemetry-instrumentation-django
- [ ] Audit and test opentelemetry-instrumentation-elasticsearch
- [ ] Audit and test opentelemetry-instrumentation-falcon
- [ ] Audit and test opentelemetry-instrumentation-fastapi
- [ ] Audit and test opentelemetry-instrumentation-flask
- [ ] Audit and test opentelemetry-instrumentation-grpc
- [ ] Audit and test opentelemetry-instrumentation-httpx
- [ ] Audit and test opentelemetry-instrumentation-jinja2
- [ ] Audit and test opentelemetry-instrumentation-kafka-python
- [ ] Audit and test opentelemetry-instrumentation-logging
- [ ] Audit and test opentelemetry-instrumentation-mysql
- [ ] Audit and test opentelemetry-instrumentation-pika
- [ ] Audit and test opentelemetry-instrumentation-psycopg2
- [ ] Audit and test opentelemetry-instrumentation-pymemcache
- [ ] Audit and test opentelemetry-instrumentation-pymongo
- [ ] Audit and test opentelemetry-instrumentation-pymysql
- [ ] Audit and test opentelemetry-instrumentation-pyramid
- [ ] Audit and test opentelemetry-instrumentation-redis
- [ ] Audit and test opentelemetry-instrumentation-requests
- [ ] Audit and test opentelemetry-instrumentation-sklearn
[...]

PR #1004, opened 2022-03-17
code change to add custom http and websocket request and response hea…

sanketmehta28:
# Description
This PR will add a feature to add custom request and response headers of http as well as websocket application to span as attributes.

Fixes # https://github.com/open-telemetry/opentelemetry-python-contrib/issues/919

## Type of change

# How Has This Been Tested?

I have added a new sample http_app_with_custom_headers and new websocket_app_with_custom_headers which will gives us sample apps to test and I have created a test class which will check expected as well as not expected span attributes in both type of applications

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
@@ -115,7 +115,15 @@ from opentelemetry.propagators.textmap import Getter, Setter
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.trace import Span, set_span_in_context
 from opentelemetry.trace.status import Status, StatusCode
-from opentelemetry.util.http import remove_url_credentials
+from opentelemetry.util.http import (
+    OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST,
+    OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_RESPONSE,
+    get_custom_headers,
+    normalise_request_header_name,
+    normalise_response_header_name,
+    remove_url_credentials,
+)
+
 
 _ServerRequestHookT = typing.Optional[typing.Callable[[Span, dict], None]]
 _ClientRequestHookT = typing.Optional[typing.Callable[[Span, dict], None]]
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -342,6 +384,13 @@ async def __call__(self, scope, receive, send):
                     for key, value in attributes.items():
                         current_span.set_attribute(key, value)
 
+                    if span.kind == trace.SpanKind.SERVER:
```
ashu658, line 387: `if current_span.kind == trace.SpanKind.SERVER:`

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -395,6 +444,17 @@ async def otel_send(message):
                         set_status_code(server_span, 200)
                         set_status_code(send_span, 200)
                     send_span.set_attribute("type", message["type"])
+                    if (
+                        server_span.kind == trace.SpanKind.SERVER
```
ashu658, line 448: We should check if server_span is a recording span before checking its kind. NonRecordingSpan do not contain kind attribute. (https://github.com/open-telemetry/opentelemetry-python-contrib/issues/998 was reported for wsgi based frameworks for the same)

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
codeboten, line 387: Please check for `span.is_recording()` before accessing `span.kind`. This causes exceptions to be raised if a `NonRecordingSpan` is received here.

It would be great to add a test to this instrumentation to ensure it doesn't crash when the SDK isn't loaded, see https://github.com/open-telemetry/opentelemetry-python-contrib/issues/958 for more details.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -223,6 +230,41 @@ def collect_request_attributes(scope):
...
+    for header in custom_request_headers:
+        values = asgi_getter.get(scope, header)
+        if values:
+            key = normalise_request_header_name(header)
+            attributes.setdefault(key, []).extend(values)
+
+    return attributes
+
+
+def collect_custom_response_headers_attributes(message):
```
ocelotl, line 251: This can be ok, but please make sure any new public (non-underscore prefixed) function or symbol is really necessary and meant to be used directly by the user. If not, please prefix them with underscores.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
sanketmehta28: I have changed `span` to `current_span` which is already being recording. So there will not be any exceptions.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
sanketmehta28: This functions will be used by other frameworks like django or FastAPI to fetch the custom headers and add them to spans as attributes.

sanketmehta28:
@ocelotl @codeboten Please do review and provide your approval on this PR.

ashu658 (approved):

codeboten (approved):

PR #1014, opened 2022-03-21
make span attribute available to urllib

### LINKED ISSUE #938: opentelemetry-instrumentation-urllib

DiptoChakrabarty:
# Description

Fixes #938 

## Type of change

Makes use of span attribute in the urllib instrumentation

# How Has This Been Tested?

- [ ] Test A

- [ ] No.

# Checklist:


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `opentelemetry-instrumentation-urllib` make spam attribute available to urllib
```
srikanthccv, line 12:
```suggestion
- `opentelemetry-instrumentation-urllib` make span attribute available to sampler
```

instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
```diff
@@ -183,15 +183,17 @@ def _instrumented_open_call(
             SpanAttributes.HTTP_URL: url,
         }
 
+        span_attributes = {
```
lzchen, line 186: There is already a dict with the exact same fields called `labels`. Maybe we can just use that.

lzchen (approved):

instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
```diff
@@ -184,14 +184,11 @@ def _instrumented_open_call(
         }
 
         with tracer.start_as_current_span(
-            span_name, kind=SpanKind.CLIENT
+            span_name, kind=SpanKind.CLIENT, attributes=labels
         ) as span:
             exception = None
             if callable(request_hook):
                 request_hook(span, request)
-            if span.is_recording():
```
ocelotl, line 192: This changes some behavior (no longer `is_recording` is being checked before setting the attributes). Please add a test case that validates this is correct.

ocelotl (changes requested):

instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
DiptoChakrabarty: @ocelotl what should I do exactly is it something like this to , make a new test function to check ```is_recording``` in a mock_span https://github.com/open-telemetry/opentelemetry-python-contrib/blob/1bb3dcf3852dd31ffc1be6c4d7425b47c628e50c/instrumentation/opentelemetry-instrumentation-urllib/tests/test_urllib_integration.py#L201

ocelotl: What I mean is that before this change we checked `span.is_recording` before setting the attributes. After this change we will set the attributes without checking. The test should make sure that this change of behavior does not break anything.

srikanthccv: @ocelotl How is that relevant here? Author removed the `set_attribute` and passing the attributes while creating the span.

ocelotl: We used to set those attributes only if `span.is_recording()`. Are those attributes being set now even if not `span.is_recording()`? Does that have any consequence? This is a change in code behavior, it should have a test.

srikanthccv: They are now passed as an arguments during the span creation. If the sampling decision turns out be non-recording then there is no change, right?

ocelotl: This is actually a minor situation and since no other tests are breaking I'm approving this PR in order not to block it. Nevertheless, this is a change in code, it needs a test. If the intention is having those attributes regardless of sampling decision, then we should have a test that shows those attributes regardless of sampling decision.

ocelotl (approved):

srikanthccv (approved):

PR #1022, opened 2022-03-29
Pyramid: Capture custom request/response headers

### LINKED ISSUE #914: Pyramid: Capture request/response headers as span attributes
part of #906

ashu658:
# Description
This PR implements the [HTTP request/response headers as span attributes semantic convention](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md) for pyramid framework.

Fixes #914 

## Type of change

# How Has This Been Tested?

Added unit tests for the changes. Also, tested by capturing headers manually with instrumented pyramid app and observing the output on jaeger backend.
Added unit test for non-recording span.

- [x] No.

# Checklist:


ashu658:
Can someone re-run the failing builds?

owais (approved):

instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py
```diff
@@ -104,6 +104,8 @@ def _before_traversal(event):
             ] = request.matched_route.pattern
         for key, value in attributes.items():
```
lzchen, line 105: Could you add an example of this usage under `Configuration` in `__init__.py`?

ashu658: Added docstring.

ashu658:
I have some small questions.
1) Should we add usage for each instrumentation or a single usage doc? As the env vars are framework independent.

2) In http request headers provided by WSGI based frameworks `-` is replaced with `_` (tested with pyramid, flask) . This would mean `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST="custom_request_header"` will capture `custom-request-header` (even though header name are not same) whereas in asgi request headers provided by framework do not replace `-` with `_` (for fastapi). Similarly, response headers can preserve casing and `-` is not replaced with `_`. 
This creates some inconsistencies in capturing headers among different frameworks.
I think we should leave it to the user to provide correct name of the header to be captured. (although giving wrong names might be able to capture header for some framework e.g. wsgi frameworks). Please share your thoughts.
cc @lzchen @srikanthccv @owais

lzchen:
@ashu658 

For (1), we should probably stick with separate usage docs per instrumentation for the reason you listed above.
For (2), doesn't [asgi](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L245) instrumentation normalize the headers as well?

instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/__init__.py
```diff
@@ -90,6 +90,48 @@
...
+
+For example,
+
+::
+
+    export OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST="content_type,custom_request_header"
+
+will extract content_type and custom_request_header from request headers and add them as span attributes.
+
+Name of the added span attribute will follow the format ``http.request.header.<header_name>`` where ``<header_name>`` being the normalized HTTP header name (lowercase, with - characters replaced by _ ).
```
lzchen, line 110:
```suggestion
The name of the added span attribute will follow the format ``http.request.header.<header_name>`` where ``<header_name>`` being the normalized HTTP header name (lowercase, with - characters replaced by _ ).
```

instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/__init__.py
```diff
@@ -90,6 +90,48 @@
...
+Value of the attribute will be single item list containing all the header values.
```
lzchen, line 111:
```suggestion
The value of the attribute will be single item list containing all the header values.
```

ashu658:
@lzchen 
I think there is some confusion.
In [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L245)  we normalise the header name to be added to the span attributes. This is same for both wsgi and asgi. 
But I was talking about the difference when we are extracting the headers from the request. The request headers that we get from the framework itself is normalised in case of wsgi frameworks (as we get wsgi headers as [wsgi env vars](https://peps.python.org/pep-0333/#environ-variables) where it replaces `-` with `_`) whereas in asgi it preserves `-`.

So, in case a user sends a request with `custom-request-header`. He will be able to capture it with `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST="custom_request_header"` in wsgi but not in asgi (as asgi will preserve `-` and the header name won't match).

ashu658:
In 31 March SIG meet discussion it is decided to document the [above mentioned behaviour](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1022#issuecomment-1083466937) as an example in instrumentation docs. 
Something like, in wsgi instrumentation docs we will mention that `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST="custom_request_header"` can caputre `custom-request-header` also and we will a recommended way to the user to configure these environment variables.
Examples of the behaviour and recommended way to configure env var should be added for each instrumentation.

instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/__init__.py
```diff
@@ -90,6 +90,57 @@
...
+
+    export OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_RESPONSE="content-type,custom_response_header"
+
+will extract ``content-type`` and ``custom_response_header`` from response headers and add them as span attributes.
+
+It is recommended that you should give the correct names of the headers to be captured in the environment variable. 
+Response header names captured in pyramid are case insensitive. So, giving header name as ``CUStomHeader`` in environment variable will be able capture header with name ``customheader``. 
+
+The name of the added span attribute will follow the format ``http.response.header.<header_name>`` where ``<header_name>`` being the normalized HTTP header name (lowercase, with - characters replaced by _ ).
+The value of the attribute will be single item list containing all the header values.
```
ashu658, line 136: Request and response headers we get are different so, I have to give separate behaviour examples for request and response headers as the headers.

instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/__init__.py
```diff
@@ -90,6 +90,57 @@
...
+For example,
+
+::
+
+    export OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_RESPONSE="content-type,custom_response_header"
+
+will extract ``content-type`` and ``custom_response_header`` from response headers and add them as span attributes.
+
+It is recommended that you should give the correct names of the headers to be captured in the environment variable. 
+Response header names captured in pyramid are case insensitive. So, giving header name as ``CUStomHeader`` in environment variable will be able capture header with name ``customheader``.
```
ashu658, line 133: Request and response headers we get are different so, I have to give separate behaviour examples for request and response headers as the headers.

instrumentation/opentelemetry-instrumentation-pyramid/tests/test_automatic.py
```diff
@@ -109,3 +117,146 @@ def test_with_existing_span(self):
...
+        resp = self.client.get("/hello/123", headers=headers)
+        self.assertEqual(200, resp.status_code)
+        span = self.memory_exporter.get_finished_spans()[0]
+        expected = {
+            "http.request.header.custom_test_header_1": ("Test Value 1",),
+            "http.request.header.custom_test_header_2": (
+                "TestValue2,TestValue3",
+            ),
+        }
+        not_expected = {
```
sanketmehta28, line 160: you dont need not_expected here. you are anyway testing it in the next test case.

instrumentation/opentelemetry-instrumentation-pyramid/tests/test_automatic.py
```diff
@@ -109,3 +117,146 @@ def test_with_existing_span(self):
...
+            "http.request.header.custom_test_header_3": ("TestValue4",),
+        }
+        self.assertEqual(span.kind, SpanKind.SERVER)
+        self.assertSpanHasAttributes(span, expected)
+        for key, _ in not_expected.items():
+            self.assertNotIn(key, span.attributes)
+
+    def test_custom_request_header_not_added_in_internal_span(self):
+        tracer = trace.get_tracer(__name__)
+        with tracer.start_as_current_span("test", kind=SpanKind.SERVER):
```
sanketmehta28, line 170: do you really need to start a SERVER span here? I guess by default it will create 1 SERVER span like it does in above use case

instrumentation/opentelemetry-instrumentation-pyramid/tests/test_automatic.py
```diff
@@ -109,3 +117,146 @@ def test_with_existing_span(self):
...
+        expected = {
+            "http.response.header.content_type": (
+                "text/plain; charset=utf-8",
+            ),
+            "http.response.header.content_length": ("7",),
+            "http.response.header.my_custom_header": (
+                "my-custom-value-1,my-custom-header-2",
+            ),
+        }
+        not_expected = {
```
sanketmehta28, line 201: same as in first test case, not expected is not required here. you are anyway testing it in the next use case

srikanthccv (approved):

srikanthccv:
@ashu658 fix conflicts and address the unresolved comments.

sanketmehta28 (approved):

PR #1056, opened 2022-04-15
Release 1.11.0-0.30b0 

lzchen:


ocelotl (approved):

exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.29b0"
+__version__ = "0.30b0"
```
srikanthccv, line 15: This package is deprecated, right? We don't want to release this anymore?

opentelemetry-contrib-instrumentations/src/opentelemetry/contrib-instrumentations/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.28b1"
+__version__ = "0.30b0"
```
srikanthccv, line 15: Did we miss this in `0.29b0`?

srikanthccv (approved):

exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/version.py
lzchen: Good point. I will undo this change and also create a task to exclude it from the release script.

opentelemetry-contrib-instrumentations/src/opentelemetry/contrib-instrumentations/version.py
lzchen: I believe it was set to `0.28b1` when the PR was created but not merged until `0.29b0` was released already. So this would be its first release. (pypi does not return anything for `opentelemetry-contrib-instrumentations`)

exporter/opentelemetry-exporter-datadog/src/opentelemetry/exporter/datadog/version.py
lzchen: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1058

PR #1081, opened 2022-05-03
Feature/add boto3 sqs instrumentation

oxeye-nikolay:
# Description

This PR instruments the SQS client as part of python's boto3 package. Unlike the [instrumentation of boto3](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-botocore), this instrumentation propagates the context and baggage over the sent messages, following the [messaging systems spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md). 

## Type of change

# How Has This Been Tested?

- [X] I have created two services using the same queue. The client was sending both batch and single messages and the producer was polling on the queue and iterating over the messages.

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:


owais:
I see this is for `boto3` while as the old one is for `boto`. Makes sense.

owais:
I'll review the code in more detail shortly but at first glance, any reason we don't follow same pattern as boto instrumentation where we have a single boto/botocore package with extensions for things like sqs and s3?

NathanielRN:
> I'll review the code in more detail shortly but at first glance, any reason we don't follow same pattern as boto instrumentation where we have a single boto/botocore package with extensions for things like sqs and s3?

+1 to this. I personally liked how OTel JS does it where they have a single instrumentation package (they actually have one pkg both v3 and v2, but it's fine that we how one for `boto3` spearately) where they have a package for `v3` that includes several extensions for the multiple services.

They have [a `ServiceExtensions.ts` file where they register the different services](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/ServicesExtensions.ts#L31-L36)

```
constructor() {
    this.services.set('SQS', new SqsServiceExtension());
    this.services.set('SNS', new SnsServiceExtension());
    this.services.set('DynamoDB', new DynamodbServiceExtension());
    this.services.set('Lambda', new LambdaServiceExtension());
  }
```

and their [file organization by service](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-aws-sdk/src/services) looks very clean too.

It would be nice for users to have `opentelemetry-instrumentation-boto3` and find immediate tracing for all the services that they do end up using without having to install more packages 🙂. If size were an issue (like it is in Lambda) then we can always mini-fy the code for services we don't use by removing directories.


As another point, what about calling this `opentelemetry-instrumentaiton-boto3-sqs` since there is no `boto3sqs` package? I don't know how this will break our scripts that count on that divider 😅

oxeye-nikolay:
> I'll review the code in more detail shortly but at first glance, any reason we don't follow same pattern as boto instrumentation where we have a single boto/botocore package with extensions for things like sqs and s3?

I would really love for it to have been like this. Unfortunatly, as I state in the documentation, the SQS classes are loaded in runtime from a template via the `boto3.client` function. This means I cannot instrument it until created, and that's why I wrap the `boto3.client` function.



> > I'll review the code in more detail shortly but at first glance, any reason we don't follow same pattern as boto instrumentation where we have a single boto/botocore package with extensions for things like sqs and s3?
> 
> +1 to this. I personally liked how OTel JS does it where they have a single instrumentation package (they actually have one pkg both v3 and v2, but it's fine that we how one for `boto3` spearately) where they have a package for `v3` that includes several extensions for the multiple services.
> 
> They have [a `ServiceExtensions.ts` file where they register the different services](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/ServicesExtensions.ts#L31-L36)
> 
> ```
> constructor() {
>     this.services.set('SQS', new SqsServiceExtension());
>     this.services.set('SNS', new SnsServiceExtension());
>     this.services.set('DynamoDB', new DynamodbServiceExtension());
>     this.services.set('Lambda', new LambdaServiceExtension());
>   }
> ```
> 
> and their [file organization by service](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-aws-sdk/src/services) looks very clean too.
> 
> It would be nice for users to have `opentelemetry-instrumentation-boto3` and find immediate tracing for all the services that they do end up using without having to install more packages 🙂. If size were an issue (like it is in Lambda) then we can always mini-fy the code for services we don't use by removing directories.
> 
> As another point, what about calling this `opentelemetry-instrumentaiton-boto3-sqs` since there is no `boto3sqs` package? I don't know how this will break our scripts that count on that divider 😅

We can create an `opentelemetry-instrumentation-boto3` package which will depend on both `opentelemetry-instrumentation-botocore` and `opentelemetry-instrumentation-boto3sqs`, and just use their instrumentations. This way users will just install `opentelemetry-instrumentation-boto3` but we will be able to keep separation for ease of use. WDYT?

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+
+        def __getitem__(
+            self, *args: List[Any], **kwargs: Dict[str, Any]
+        ) -> Any:
+            retval = super(
+                Boto3SQSInstrumentor.ContextableList, self
+            ).__getitem__(*args, **kwargs)
+            if not isinstance(retval, dict):
+                return retval
+            if not (receipt_handle := retval.get("ReceiptHandle", None)):
```
sanketmehta28, line 123: this could create a problem in python < 3.8 as walrus operator (':=') is available only from python 3.8. Its better to make it more generic by separating assignment and checking of receipt_handle

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+        if conversation_id:
+            span.set_attribute(
+                SpanAttributes.MESSAGING_CONVERSATION_ID, conversation_id
+            )
+        if message_id:
+            span.set_attribute(SpanAttributes.MESSAGING_MESSAGE_ID, message_id)
+
+    @staticmethod
+    def _safe_end_processing_span(receipt_handle: str) -> None:
+        if started_span := Boto3SQSInstrumentor.received_messages_spans.pop(
```
sanketmehta28, line 179: same comment as before for walrus operator (':=')

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+        ):
+            for item in ctx.values():
+                if hasattr(item, "get_span_context"):
+                    links.append(Link(context=item.get_span_context()))
+        span = self._tracer.start_span(
+            name=f"{queue_name} process",
+            links=links,
+            kind=SpanKind.CONSUMER,
+        )
+        with trace.use_span(span):
```
sanketmehta28, line 211: You can combine start_span and use_span in one API: start_as_current_span() with with statement

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+            span = self._tracer.start_span(
+                name=f"{queue_name} send",
+                kind=SpanKind.PRODUCER,
+            )
+            Boto3SQSInstrumentor._enrich_span(span, queue_name)
+            with trace.use_span(span, end_on_exit=True):
+                attributes = kwargs.pop("MessageAttributes", {})
+                propagate.inject(attributes, setter=boto3sqs_setter)
+                retval = wrapped(*args, MessageAttributes=attributes, **kwargs)
+                if message_id := retval.get("MessageId"):
```
sanketmehta28, line 240: you can pass message_id in enrich_span() so that you dont have to repeat the line 242 and 243 again as its already present in enrich_span()

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+                        entry["MessageAttributes"] = {}
+                    propagate.inject(
+                        entry["MessageAttributes"], setter=boto3sqs_setter
+                    )
+            retval = wrapped(*args, **kwargs)
+            for successful_messages in retval["Successful"]:
+                message_identifier = successful_messages["Id"]
+                if message_span := ids_to_spans.get(message_identifier):
+                    if message_span.is_recording():
+                        message_span.set_attribute(
```
sanketmehta28, line 288: same as before for message_id

sanketmehta28 (changes requested):

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: I think it's better to add it later, because if the call of `retval = wrapped(*args, MessageAttributes=attributes, **kwargs) ` will raise (by a  wrong queue for example), the span will still have the available information - helping the user to trace and debug. @sanketmehta28 WDYT?

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: Similarly to [this case](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1081#discussion_r873561456) I add the information after because if I set all the information after, a failure will create a span with lacking information that can be provided.

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: Nice! Done

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: Oh, my bad. For some reason, I remembered it was a python 3.6 feature. Changed

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: Done

instrumentation/opentelemetry-instrumentation-boto3sqs/tests/test_setter.py
```diff
@@ -0,0 +1,44 @@
+# Copyright The OpenTelemetry Authors
```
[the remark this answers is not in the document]
oxeye-nikolay: Done

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -0,0 +1,428 @@
...
+
+
+    Boto3SQSInstrumentor().instrument()
+"""
+import logging
+from typing import Any, Collection, Dict, Generator, List, Optional
+
+import boto3
+import botocore.client
+import wrapt
```
ocelotl, line 36: `boto3` and `wrapt` should be dependencies of this package.

ocelotl: For consistency with the rest of imports, please use `from ... import` whenever possible.

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -0,0 +1,428 @@
...
+    MessagingDestinationKindValues,
+    MessagingOperationValues,
+    SpanAttributes,
+)
+from opentelemetry.trace import Link, Span, SpanKind, Tracer, TracerProvider
+
+from .package import _instruments
+from .version import __version__
+
+logger = logging.getLogger(__name__)
```
ocelotl, line 55:
```suggestion
_logger = logging.getLogger(__name__)
```
Everything should be private unless it is intended to be exposed to the user.

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -0,0 +1,428 @@
...
+# We use this prefix so we can request all instrumentation MessageAttributeNames with a wildcard, without harming
+# existing filters
+OPENTELEMETRY_ATTRIBUTE_IDENTIFIER: str = "otel."
```
ocelotl, line 58: Is this intended to be importable by the user? If not, please make it private.

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -0,0 +1,428 @@
...
+            return None
+        return [value.get("StringValue", None)]
+
+    def keys(self, carrier: CarrierT) -> List[str]:
+        return [
+            key[len(OPENTELEMETRY_ATTRIBUTE_IDENTIFIER) :]
+            if key.startswith(OPENTELEMETRY_ATTRIBUTE_IDENTIFIER)
+            else key
+            for key in carrier.keys()
+        ]
```
ocelotl, line 74: Nit: move the call to `len` outside of the for loop.

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -0,0 +1,428 @@
...
+        Since the classic way to process SQS messages is using a `for` loop, without a well defined scope like a
+        callback - we are doing something similar to the instrumentaiton of Kafka-python and instrumenting the
+        `__iter__` functions and the `__getitem__` functions to set the span context of the addressed message. Since
+        the return value from an `SQS.ReceiveMessage` returns a builtin list, we cannot wrap it and change all of the
+        calls for `list.__iter__` and `list.__getitem__` - therefore we use ContextableList. It is bound to the
+        received_messages_spans dict
+        """
+
+        def __getitem__(
+            self, *args: List[Any], **kwargs: Dict[str, Any]
```
ocelotl, line 113: Having `*args` and `**kwargs` in `__getitem__` is unusual, usually only a [single `key`](https://docs.python.org/3/reference/datamodel.html#object.__getitem__) argument is used.

Having `*args` in the arguments of `__getitem__` allows you to do this:
``` python
class Test:

    def __getitem__(self, *args, **kwargs):

        print(args)

        return 4


print(Test()[32, 23])
```
```
((32, 23),)
4
```

Nevertheless, I think the Python syntax forbids named parameters to be passed inside `[]`:
``` python
class Test:

    def __getitem__(self, *args, **kwargs):

        print(args)
        print(kwargs)

        return 4


print(Test()[32, 23, a=2, b=5])
```
```
  File "/home/ocelotl/sandbox/tato.py", line 11
    print(Test()[32, 23, a=2, b=5])
                          ^
SyntaxError: invalid syntax
```
It also seems like passing `*args` to `list.__getitem__` can cause trouble:
``` python
class Test(list):

    def __getitem__(self, *args, **kwargs):

        super().__getitem__(*args, **kwargs)

        print(args)

        return 4


print(Test()[32, 23])
```
```
  File "/home/ocelotl/sandbox/tato.py", line 25, in <module>
    print(Test()[32, 23])
  File "/home/ocelotl/sandbox/tato.py", line 18, in __getitem__
    super().__getitem__(*args, **kwargs)
TypeError: list indices must be integers or slices, not tuple
```

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -0,0 +1,428 @@
...
+        ) -> Any:
+            retval = super(
+                Boto3SQSInstrumentor.ContextableList, self
+            ).__getitem__(*args, **kwargs)
+            if not isinstance(retval, dict):
+                return retval
+            receipt_handle = retval.get("ReceiptHandle", None)
```
ocelotl, line 120: Nit: `None` is the default return value of `dict.get`:
```suggestion
            receipt_handle = retval.get("ReceiptHandle")
```

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -0,0 +1,428 @@
...
+            if not receipt_handle:
+                return retval
+            started_span = Boto3SQSInstrumentor.received_messages_spans.get(
+                receipt_handle, None
+            )
+            if not started_span:
```
ocelotl, line 126:
```suggestion
            if started_span is not None:
```
As specified [here](https://peps.python.org/pep-0008/#programming-recommendations).

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -0,0 +1,428 @@
...
+                return retval
+            if Boto3SQSInstrumentor.current_context_token:
+                context.detach(Boto3SQSInstrumentor.current_context_token)
+            Boto3SQSInstrumentor.current_context_token = context.attach(
+                trace.set_span_in_context(started_span)
+            )
+            Boto3SQSInstrumentor.current_span_related_to_token = started_span
+            return retval
+
+        def __iter__(self) -> Generator:
```
ocelotl, line 136: Just curious, why this? Do you want to use a generator for performance reasons?

ocelotl (changes requested):

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: I have added wrapt, but isn't boto3 added as part of `_instruments` from `package.py`?

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: Done

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: Done

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: Ah, nice! Done

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: Yeah, you're right. I wanted to create a proxy function as easily as possible - but this way I introduce a new functionality. Changed

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: Done

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: Done

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: In general when it comes to SQS queues, unlike other libraries where you define a callback that handles each message, here we don't have an actual context (like a callback) that starts and ends for handling a single message. That means we can't know which message span is the currently active one.
This is why I change the response to be of this proxy object type, where every access to an item (via `__getitem__`) or just iteration over the response (via `__iter__`) sets the context.
When I implement `__iter__` calling `__getitem__`, I maintain the context management only in that `__getitem__` 😄

sanketmehta28 (approved):

ocelotl (approved):

PR #1089, opened 2022-05-13
Use a shorter timeout for AWS EC2 metadata requests

### LINKED ISSUE #1088: EC2 resource detector hangs for a long time outside of an EC2 instance

The environment I initially saw this in was a container running in Docker compose on an AWS EC2 instance but I've been able to reproduce it on my laptop as well. I think it will show up in anything not directly running in AWS.

**Steps to reproduce**

The following code reproduced the issue on my laptop:

```python
from opentelemetry.sdk.extension.aws.resource.ec2 import AwsEc2ResourceDetector
from opentelemetry.sdk.resources import get_aggregated_resources

resource = get_aggregated_resources(
    detectors=[AwsEc2ResourceDetector()]
)
```

It should complete quickly (this is the behavior I see running on an EC2 instance).

What did you see instead?

On my laptop, it will hand ~indefinitely.

Note: one solution is just to remove the resource detector but we'd like to be able to include it and just have it fail silently, which is the behavior we've seen in other resource detectors.

**Additional context**

I think the problem is here: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/80969a06da77d1e616124de178d12a1ebe3ffe7f/sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/ec2.py#L37

It looks like the request is using a 1000 _second_ timeout which I suspect is intended to be a 1000 _millisecond_ timeout. At least with the server program I've been working on that will block the startup of the program until the request completes.

You can verify by running:

```
curl http://169.254.169.254/latest/api/token
```

Which is one of the requests that the resource detector makes -- it should hang indefinitely as well.

jemisonf:
Fix #1088 

According to the docs, the value for `timeout` is in seconds: https://docs.python.org/3/library/urllib.request.html#urllib.request.urlopen. 1000 seconds seems slow and in some cases can block the startup of the program being instrumented (see #1088 as an example), because the request will hang indefinitely in non-AWS environments. Using a much shorter 1 second timeout seems like a reasonable workaround for this.

# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

### DIFF
--- a/sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/ec2.py
+++ b/sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/ec2.py
@@ -34,7 +34,7 @@ def _aws_http_request(method, path, headers):
         Request(
             "http://169.254.169.254" + path, headers=headers, method=method
         ),
-        timeout=1000,
+        timeout=1,
     ) as response:
         return response.read().decode("utf-8")
 


srikanthccv (approved):

srikanthccv:
Bug fixes require a changelog entry even if the fix is simple. Please add entry.

NathanielRN:
Thanks for this change @jemisonf ! This looks like the right move to me, OTel JS had a similar issue but in the other direction; [a timeout of 1 second was not enough, they needed at least 5 seconds](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/722).

In that case, do you think changing to be `5` would be reasonable here as well?

Could you also update the timeout in the `EKS` detector to be `5`?

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/10659f897086fe6037f14caece570a75b932a3ba/sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/eks.py#L40

jemisonf:
Quick note on the EKS change -- 2000 is probably higher than required so it definitely seems worth changing but I wouldn't expect the same issue to show up there since it's making a request to an internal k8s DNS name rather than using a public AWS IP like the EC2 detector

PR #1097, opened 2022-05-19
Refactor TracedConnectionProxy

### LINKED ISSUE #1077: dbapi instrumentation breaks psycopg2 when registering types.
**Describe your environment** 
I'm using dbapi instrumentation to instrument postgresql.  When I attempt to register jsonb, psycopg2 crashes.  This issue is nearly identical to #143, with the exception that I'm not using `Psycopg2Instrumentor`.

The reason I'm using dbapi instrumentation vs `Psycopg2Instrumentor` is because the latter does not support capturing parameters as span attributes.

**Steps to reproduce**

```python
from opentelemetry.instrumentation import dbapi as dbapi_instrumentor
import psycopg2
import psycopg2.extras

db_connection_attributes = {
    "database": "info.dbname",
    "port": "info.port",
    "host": "info.host",
    "user": "info.user",
}
dbapi_instrumentor.trace_integration(
    psycopg2,
    "connect",
    "postgresql",
    db_connection_attributes,
    capture_parameters=True,
)

cnx = psycopg2.connect(host='localhost', port='5432', dbname='postgres', password='1234', user='postgres')
psycopg2.extras.register_default_jsonb(conn_or_curs=cnx, loads=lambda x: x)
```
results in 
```
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/psycopg2/_json.py", line 156, in register_default_jsonb
    loads=loads, oid=JSONB_OID, array_oid=JSONBARRAY_OID, name='jsonb')
  File "/usr/local/lib/python3.7/site-packages/psycopg2/_json.py", line 125, in register_json
    register_type(JSON, not globally and conn_or_curs or None)
TypeError: argument 2 must be a connection, cursor or None
```

For psycopg2 to not crash when used with dbapi instrumentation.

It crashes when psycopg2 enforces connection or cursor types.

**Additional context**

ocelotl:
Fixes #1077

Fixing this issue is a bit complicated:

[Here](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/143#issuecomment-742477481) it is suggested that we may monkey patch the `register_types` function instead of wrapping the connection/cursor. The problem with this approach is that it may be necessary to do this kind of patching in private modules of the Psycopg2 library that may change paths unexpectedly (since they are private), causing our instrumentation to be breakable with any release of Psycopg2.

@owais suggests [here](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/143#issuecomment-708099012) that we should patch the classes themselves instead of wrapping to avoid problems with type checking. I understand the point but I think this approach also has the problem that it may require importing from private modules to patch effectively.

The particular issue here happens because [`PyObject_TypeCheck`](https://docs.python.org/3/c-api/object.html#c.PyObject_TypeCheck) is used in Psycopg2 `register_types` and this type checking implemented in C is more restrictive than `isinstance` (`wrapt.ObjectProxy` objects [passes an `isinstance` check](https://wrapt.readthedocs.io/en/latest/wrappers.html#object-proxy) against the type of the wrapped object).

The solution implemented here is able to pass the `PyObject_TypeCheck` because the connection created dynamically inherits from the class of the original connection, something `wrapt.ObjectProxy` does not.


instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -214,8 +214,8 @@ def uninstrument_connection(connection):
     Returns:
         An uninstrumented connection.
     """
-    if isinstance(connection, wrapt.ObjectProxy):
-        return connection.__wrapped__
+    if isinstance(connection, _TracedConnectionProxy):
```
lzchen, line 217: Any reason why we need to use `_TracedConnectionProxy` for type check and not just `TracedConnectionProxy`?

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -300,28 +300,42 @@ def get_connection_attributes(self, connection):
...
 
         def __enter__(self):
-            self.__wrapped__.__enter__()
+            self._connection.__enter__()
             return self
 
-        def __exit__(self, *args, **kwargs):
-            self.__wrapped__.__exit__(*args, **kwargs)
+        def __exit__(self, exc_type, exc_value, traceback):
+            self._connection.__exit__(exc_type, exc_value, traceback)
```
oxeye-nikolay, line 325: If you already have defined `def __getattr__(self, name):`, why is this needed without additional logic?

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
ocelotl: Yes, we need to check the type of the connection and the `TracedConnectionProxy` is inside a function making it inaccessible for type comparison.

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
ocelotl: Good point, removing.

lzchen (approved):

phillipuniverse:
@ocelotl I think this might have caused a regression in the latest release for SQLIte at #1179

PR #1066, opened 2022-04-20
Prune autoinstrumentation sitecustomize module directory from PYTHONPATH immediately

aabmass:
# Description

This prevents auto-instrumentation of subprocesses launched with `exec*` as soon as possible. The risk of autoinstrumenting these subprocesses is creating a loop of autoinstrumentation if the initialization code creates the subprocess.

Fixes #1050

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

Tested manually

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [x] Followed the style guidelines of this project
- [x] Changelogs have been updated
- [ ] Unit tests have been added
- [x] Documentation has been updated

PR #1100, opened 2022-05-20
Fixing inescapable character bug for Windows path string

jeremydvoss:
# Description

Added an escape call in sitecustomize to fix unescapable character bug in auto-instrumentation on Windows.
Fixes [#2703 on opentelemetry-python](open-telemetry/opentelemetry-python-contrib#1102)

## Type of change

# How Has This Been Tested?

Tested with server_uninstrumented.py script. Confirmed spans are printed on Windows.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

- [x] Unit tests have been added.
- [ ] Documentation has been updated.


jeremyvoss-microsoft:
I expected my PR # to be 1098, but it is apparently 1100. Since this is my first PR, I am not sure how to write the CHANGELOG knowing the PR number that will be created in the future. I can fix the CHANGELOG in another commit

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/open-telemetry/opentelemetry-python/compare/v1.12.0rc1-0.31b0...HEAD)
 
+- Adding escape call to fix [auto-instrumentation not producing spans on Windows](https://github.com/open-telemetry/opentelemetry-python/issues/2703).
+  ([#1098](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1098))
```
srikanthccv, line 11:
```suggestion
  ([#1100](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1100))
```

srikanthccv: @jeremydvoss Either use this commit suggestion UI or commit from your machine to update the PR number. In the end we squash all your commits before merging so the multiple commits is not a problem if you are wondering about that.

jeremydvoss: I was wondering about that. Do you mean that the commits are squashed automatically? I didn't know squashing commits was possible once they are pushed to a public remote.

srikanthccv: Before we merge all your commits are squashed.

aabmass:
Is this the cause of https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1102? I don't understand how this is fixing the issue, all that #1066 did was move the stripping of python path earlier in the file.

Or was the stripping code broken on windows all along and now it's causing an exception to be raised earlier?

ramthi (approved):

jeremydvoss:
> Is this the cause of #1102? I don't understand how this is fixing the issue, all that #1066 did was move the stripping of python path earlier in the file.
> 
> Or was the stripping code broken on windows all along and now it's causing an exception to be raised earlier?

That code block was always broken on Windows, But since it was later in the finally block, it didn't cause an issue every time. The underlying problem is that we passed the re.sub function a raw string, when what it needs is an escaped pattern. This isn't an issue on Linux because linux paths use "/" which do not need to be escaped. But Windows paths use "\" and therefore do need to be escaped. By adding an escape, the string should satisfy re.sub on Linux and Windows

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
```diff
@@ -112,7 +112,7 @@ def _load_configurators():
 def initialize():
     # prevents auto-instrumentation of subprocesses if code execs another python process
     environ["PYTHONPATH"] = sub(
-        rf"{dirname(abspath(__file__))}{pathsep}?",
+        escape(rf"{dirname(abspath(__file__))}{pathsep}?"),
```
aabmass, line 115: This will now escape the `?` at the end which we don't want. The escaping should only apply to the path part

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
```diff
@@ -112,7 +112,7 @@ def _load_configurators():
 def initialize():
     # prevents auto-instrumentation of subprocesses if code execs another python process
     environ["PYTHONPATH"] = sub(
```
aabmass, line 114: Can you also refactor the substitution into a function and add some windows and unix test cases for it?

aabmass (changes requested):
I read what you wrote in #1102 closer and understand 👍 thanks

lzchen (approved):

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
jeremydvoss: I have pushed the refactor. However, since it seems there are no tests for auto-instrumentation and sitecustomize, I propose creating a separate ticket to add such tests.

jeremydvoss: Here's the ticket I created.
https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1104

aabmass: Thanks for opening the ticket, I agree we could use more tests here overall. I think it's still worthwhile to add tests for the change in this PR, e.g. if you add a test case for when the PYTHONPATH doesn't have pathsep (just a single directory in the path), it would catch the bug I mentioned in my other comment

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
```diff
@@ -109,14 +109,18 @@ def _load_configurators():
             raise exc
 
 
-def initialize():
+def _remove_current_directory_from_python_path():
```
aabmass, line 112: It would be easier to test this as a pure function, like if it takes in a string and returns a new string

sanketmehta28 (approved):

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
jeremydvoss: That would basically just be the re.sub function itself. The logic lies iun the rf"" string substitution of the current path. So, I don't think it makes sense to refactor a new function without that logic in it. Let me know if you mean something else.

aabmass: I just meant remove the part that modifies the global `os.environ` out, so that you can test just the substitution/escaping bit. Something like

```py
def strip_directory_from_path(path: str, dir: str, path_sep: str) -> str:
  return re.sub(...)
```

```py
# unix
self.assertEqual(strip_current_directory_from_path("/foo/bar/baz", "/foo/bar/baz", ":"), "")
self.assertEqual(strip_current_directory_from_path("/foo/bar/baz:/some/other/path", "/foo/bar/baz", ":"), "/some/other/path")
# windows
self.assertEqual(strip_current_directory_from_path(r"C:\foo\bar", r"C:\foo\bar", ";"), "")
self.assertEqual(strip_current_directory_from_path(r"C:\foo\bar;C:\some\other\path", r"C:\foo\bar", ";"), "C:\some\other\path")
# other cases like dir not in path, path is empty string etc.
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -163,3 +164,11 @@ def _generate_opentelemetry_traceparent(span: Span) -> str:
     _traceparent = _version + "-" + _trace_id + "-" + _span_id + "-" + _flags
     meta.update({"traceparent": _traceparent})
     return meta
+
+
+def _python_path_without_current_directory(python_path, dir, path_separator):
```
aabmass, line 169: nit 
```suggestion
def _python_path_without_directory(python_path, dir, path_separator):
```

opentelemetry-instrumentation/tests/test_utils.py
```diff
@@ -111,3 +111,28 @@ def test_http_status_to_status_code_server(self):
...
+        dir = r"c:\users\trayvonmartin\workplace\opentelemetry-python-contrib\opentelemetry-instrumentation\src\opentelemetry\instrumentation\auto_instrumentation"
+        path_separator = r";"
+        python_path = r"c:\users\trayvonmartin\workplace\opentelemetry-python-contrib\opentelemetry-instrumentation\src\opentelemetry\instrumentation\auto_instrumentation;C:\Users\trayvonmartin\workplace"
+        actual_python_path = _python_path_without_current_directory(
+            python_path,
+            dir,
+            path_separator
+        )
+        expected_python_path = r"C:\Users\trayvonmartin\workplace"
+        assert(expected_python_path == actual_python_path
```
aabmass, line 125: nit `self.assertEqual` for consistency

opentelemetry-instrumentation/tests/test_utils.py
```diff
@@ -111,3 +111,28 @@ def test_http_status_to_status_code_server(self):
                     int(status_code), server_span=True
                 )
                 self.assertEqual(actual, expected, status_code)
+
+    def test_remove_current_directory_from_python_path_windows(self):
```
aabmass, line 115: can you add test cases for when the python_path is just the directory (where the `?` in the regex is needed)

aabmass (approved):
Thanks for bearing with me. LGTM!

srikanthccv:
@jeremydvoss Fix the lint and we are good to merge this.

CHANGELOG.md
```diff
@@ -8,14 +8,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
   ([1109](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1109))
 - cleanup type hints for textmap `Getter` and `Setter` classes
   ([1106](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1106))
+<<<<<<< HEAD
   
 ### Added
 - `opentelemetry-instrumentation-remoulade` Initial release
   ([#1082](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1082))
+=======
+>>>>>>> b4440e8d (updated type hints for getters and setters (#1106))
```
srikanthccv, line 23:
```suggestion
  
### Added
- `opentelemetry-instrumentation-remoulade` Initial release
  ([#1082](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1082))
```

opentelemetry-instrumentation/tests/test_utils.py
```diff
@@ -111,3 +114,42 @@ def test_http_status_to_status_code_server(self):
...
+
+    def test_remove_current_directory_from_python_path_linux_only_path(self):
+        dir = r"/home/sandranland/workplace/opentelemetry-python-contrib/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation"
+        path_separator = r":"
+        python_path = r"/home/SandraBland/workplace/opentelemetry-python-contrib/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation"
+        actual_python_path = _python_path_without_directory(
+            python_path, dir, path_separator
+        )
+        self.assertEqual(actual_python_path, python_path)
+
```
srikanthccv, line 155:
```suggestion
```

PR #1128, opened 2022-06-11
Add metric instrumentation for WSGI

### LINKED ISSUE #1126: [instrumentation-wsgi] add metric instrumentation for wsgi
Add support for the [`http.server`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#http-server) metric semantic conventions in WSGI.

srikanthccv:
# Description

Fixes #1126 

## Type of change

# How Has This Been Tested?

- [x] `tox -e test-instrumentation-wsgi`

- [x] Yes. - Link to PR: https://github.com/open-telemetry/opentelemetry-python/pull/2757

# Checklist:

### DIFF
--- a/instrumentation/README.md
+++ b/instrumentation/README.md
@@ -1,43 +1,43 @@
 
-| Instrumentation | Supported Packages |
-| --------------- | ------------------ |
-| [opentelemetry-instrumentation-aiohttp-client](./opentelemetry-instrumentation-aiohttp-client) | aiohttp ~= 3.0 |
-| [opentelemetry-instrumentation-aiopg](./opentelemetry-instrumentation-aiopg) | aiopg >= 0.13.0, < 1.3.0 |
-| [opentelemetry-instrumentation-asgi](./opentelemetry-instrumentation-asgi) | asgiref ~= 3.0 |
-| [opentelemetry-instrumentation-asyncpg](./opentelemetry-instrumentation-asyncpg) | asyncpg >= 0.12.0 |
-| [opentelemetry-instrumentation-aws-lambda](./opentelemetry-instrumentation-aws-lambda) | aws_lambda |
-| [opentelemetry-instrumentation-boto](./opentelemetry-instrumentation-boto) | boto~=2.0 |
-| [opentelemetry-instrumentation-boto3sqs](./opentelemetry-instrumentation-boto3sqs) | boto3 ~= 1.0 |
-| [opentelemetry-instrumentation-botocore](./opentelemetry-instrumentation-botocore) | botocore ~= 1.0 |
-| [opentelemetry-instrumentation-celery](./opentelemetry-instrumentation-celery) | celery >= 4.0, < 6.0 |
-| [opentelemetry-instrumentation-confluent-kafka](./opentelemetry-instrumentation-confluent-kafka) | confluent-kafka ~= 1.8.2 |
-| [opentelemetry-instrumentation-dbapi](./opentelemetry-instrumentation-dbapi) | dbapi |
-| [opentelemetry-instrumentation-django](./opentelemetry-instrumentation-django) | django >= 1.10 |
-| [opentelemetry-instrumentation-elasticsearch](./opentelemetry-instrumentation-elasticsearch) | elasticsearch >= 2.0 |
-| [opentelemetry-instrumentation-falcon](./opentelemetry-instrumentation-falcon) | falcon >= 1.4.1, < 4.0.0 |
-| [opentelemetry-instrumentation-fastapi](./opentelemetry-instrumentation-fastapi) | fastapi ~= 0.58 |
-| [opentelemetry-instrumentation-flask](./opentelemetry-instrumentation-flask) | flask >= 1.0, < 3.0 |
-| [opentelemetry-instrumentation-grpc](./opentelemetry-instrumentation-grpc) | grpcio ~= 1.27 |
-| [opentelemetry-instrumentation-httpx](./opentelemetry-instrumentation-httpx) | httpx >= 0.18.0 |
-| [opentelemetry-instrumentation-jinja2](./opentelemetry-instrumentation-jinja2) | jinja2 >= 2.7, < 4.0 |
-| [opentelemetry-instrumentation-kafka-python](./opentelemetry-instrumentation-kafka-python) | kafka-python >= 2.0 |
-| [opentelemetry-instrumentation-logging](./opentelemetry-instrumentation-logging) | logging |
-| [opentelemetry-instrumentation-mysql](./opentelemetry-instrumentation-mysql) | mysql-connector-python ~= 8.0 |
-| [opentelemetry-instrumentation-pika](./opentelemetry-instrumentation-pika) | pika >= 0.12.0 |
-| [opentelemetry-instrumentation-psycopg2](./opentelemetry-instrumentation-psycopg2) | psycopg2 >= 2.7.3.1 |
-| [opentelemetry-instrumentation-pymemcache](./opentelemetry-instrumentation-pymemcache) | pymemcache >= 1.3.5, < 4 |
-| [opentelemetry-instrumentation-pymongo](./opentelemetry-instrumentation-pymongo) | pymongo >= 3.1, < 5.0 |
-| [opentelemetry-instrumentation-pymysql](./opentelemetry-instrumentation-pymysql) | PyMySQL < 2 |
-| [opentelemetry-instrumentation-pyramid](./opentelemetry-instrumentation-pyramid) | pyramid >= 1.7 |
-| [opentelemetry-instrumentation-redis](./opentelemetry-instrumentation-redis) | redis >= 2.6 |
-| [opentelemetry-instrumentation-remoulade](./opentelemetry-instrumentation-remoulade) | remoulade >= 0.50 |
-| [opentelemetry-instrumentation-requests](./opentelemetry-instrumentation-requests) | requests ~= 2.0 |
-| [opentelemetry-instrumentation-sklearn](./opentelemetry-instrumentation-sklearn) | scikit-learn ~= 0.24.0 |
-| [opentelemetry-instrumentation-sqlalchemy](./opentelemetry-instrumentation-sqlalchemy) | sqlalchemy |
-| [opentelemetry-instrumentation-sqlite3](./opentelemetry-instrumentation-sqlite3) | sqlite3 |
-| [opentelemetry-instrumentation-starlette](./opentelemetry-instrumentation-starlette) | starlette ~= 0.13.0 |
-| [opentelemetry-instrumentation-system-metrics](./opentelemetry-instrumentation-system-metrics) | psutil >= 5 |
-| [opentelemetry-instrumentation-tornado](./opentelemetry-instrumentation-tornado) | tornado >= 5.1.1 |
-| [opentelemetry-instrumentation-urllib](./opentelemetry-instrumentation-urllib) | urllib |
-| [opentelemetry-instrumentation-urllib3](./opentelemetry-instrumentation-urllib3) | urllib3 >= 1.0.0, < 2.0.0 |
-| [opentelemetry-instrumentation-wsgi](./opentelemetry-instrumentation-wsgi) | wsgi |
+| Instrumentation | Supported Packages | Metrics support |
+| --------------- | ------------------ | --------------- |
+| [opentelemetry-instrumentation-aiohttp-client](./opentelemetry-instrumentation-aiohttp-client) | aiohttp ~= 3.0 | [ ]
+| [opentelemetry-instrumentation-aiopg](./opentelemetry-instrumentation-aiopg) | aiopg >= 0.13.0, < 1.3.0 | [ ]
+| [opentelemetry-instrumentation-asgi](./opentelemetry-instrumentation-asgi) | asgiref ~= 3.0 | [ ]
+| [opentelemetry-instrumentation-asyncpg](./opentelemetry-instrumentation-asyncpg) | asyncpg >= 0.12.0 | [ ]
+| [opentelemetry-instrumentation-aws-lambda](./opentelemetry-instrumentation-aws-lambda) | aws_lambda | [ ]
+| [opentelemetry-instrumentation-boto](./opentelemetry-instrumentation-boto) | boto~=2.0 | [ ]
+| [opentelemetry-instrumentation-boto3sqs](./opentelemetry-instrumentation-boto3sqs) | boto3 ~= 1.0 | [ ]
+| [opentelemetry-instrumentation-botocore](./opentelemetry-instrumentation-botocore) | botocore ~= 1.0 | [ ]
+| [opentelemetry-instrumentation-celery](./opentelemetry-instrumentation-celery) | celery >= 4.0, < 6.0 | [ ]
+| [opentelemetry-instrumentation-confluent-kafka](./opentelemetry-instrumentation-confluent-kafka) | confluent-kafka ~= 1.8.2 | [ ]
+| [opentelemetry-instrumentation-dbapi](./opentelemetry-instrumentation-dbapi) | dbapi | [ ]
+| [opentelemetry-instrumentation-django](./opentelemetry-instrumentation-django) | django >= 1.10 | [ ]
+| [opentelemetry-instrumentation-elasticsearch](./opentelemetry-instrumentation-elasticsearch) | elasticsearch >= 2.0 | [ ]
+| [opentelemetry-instrumentation-falcon](./opentelemetry-instrumentation-falcon) | falcon >= 1.4.1, < 4.0.0 | [ ]
+| [opentelemetry-instrumentation-fastapi](./opentelemetry-instrumentation-fastapi) | fastapi ~= 0.58 | [ ]
+| [opentelemetry-instrumentation-flask](./opentelemetry-instrumentation-flask) | flask >= 1.0, < 3.0 | [ ]
+| [opentelemetry-instrumentation-grpc](./opentelemetry-instrumentation-grpc) | grpcio ~= 1.27 | [ ]
+| [opentelemetry-instrumentation-httpx](./opentelemetry-instrumentation-httpx) | httpx >= 0.18.0 | [ ]
+| [opentelemetry-instrumentation-jinja2](./opentelemetry-instrumentation-jinja2) | jinja2 >= 2.7, < 4.0 | [ ]
+| [opentelemetry-instrumentation-kafka-python](./opentelemetry-instrumentation-kafka-python) | kafka-python >= 2.0 | [ ]
+| [opentelemetry-instrumentation-logging](./opentelemetry-instrumentation-logging) | logging | [ ]
+| [opentelemetry-instrumentation-mysql](./opentelemetry-instrumentation-mysql) | mysql-connector-python ~= 8.0 | [ ]
+| [opentelemetry-instrumentation-pika](./opentelemetry-instrumentation-pika) | pika >= 0.12.0 | [ ]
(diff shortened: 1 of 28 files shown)


instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -352,18 +382,39 @@ class OpenTelemetryMiddleware:
...
         self.tracer = trace.get_tracer(__name__, __version__, tracer_provider)
+        self.meter = get_meter(__name__, __version__, meter_provider)
+        self.duration_recorder = self.meter.create_histogram(
+            "http.server.duration",
+            "ms",
+            "measures the duration of the inbound HTTP request",
+        )
+        self.active_requests_counter = self.meter.create_up_down_counter(
+            "http.server.active_requests",
+            "1",
```
ocelotl, line 402: Should this be [`requests`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#http-server)?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -181,6 +183,26 @@ def response_hook(span: Span, environ: WSGIEnvironment, status: str, response_he
 _CARRIER_KEY_PREFIX = "HTTP_"
 _CARRIER_KEY_PREFIX_LEN = len(_CARRIER_KEY_PREFIX)
 
+# List of recommended attributes
```
ocelotl, line 186: Just curious, why do we have different recommended attributes for request duration and active requests?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -352,18 +382,39 @@ class OpenTelemetryMiddleware:
...
+        wsgi,
+        request_hook=None,
+        response_hook=None,
+        tracer_provider=None,
+        meter_provider=None,
     ):
         self.wsgi = wsgi
         self.tracer = trace.get_tracer(__name__, __version__, tracer_provider)
+        self.meter = get_meter(__name__, __version__, meter_provider)
+        self.duration_recorder = self.meter.create_histogram(
```
aabmass, line 395: any reason for this to be `duration_recorder` instead of `duration_histogram`?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -352,18 +382,39 @@ class OpenTelemetryMiddleware:
...
+            "http.server.duration",
+            "ms",
+            "measures the duration of the inbound HTTP request",
```
aabmass, line 398: nit this would be more readable with named arguments

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -304,6 +326,14 @@ def collect_custom_response_headers_attributes(response_headers):
     return attributes
 
 
+def _parse_status_code(resp_status):
+    status_code, _ = resp_status.split(" ", 1)
```
aabmass, line 330: Status code [is always 3 digits](https://datatracker.ietf.org/doc/html/rfc2616.html#page-39:~:text=The%20Status%2DCode%20element%20is%20a%203%2Ddigit%20integer%20result%20code%20of%20the), so `status_code[:3]` should work?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -383,13 +434,24 @@ def __call__(self, environ, start_response):
             environ: A WSGI environment.
             start_response: The WSGI start_response callable.
         """
+        req_attrs = collect_request_attributes(environ)
+        active_requests_count_attrs = {}
+        for attr_key in _active_requests_count_attrs:
+            if req_attrs.get(attr_key) is not None:
+                active_requests_count_attrs[attr_key] = req_attrs[attr_key]
```
aabmass, line 441: I think we should be setting the same attribute keys for a given instrument regardless if the value is available or not. If the value is not available, we should have some default like empty string or 0.

This is what [Prometheus/OM expects](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric:~:text=Metrics%20with%20the%20same%20name%20for%20a%20given%20MetricFamily%20SHOULD%20have%20the%20same%20set%20of%20label%20names%20in%20their%20LabelSet.)

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -352,18 +382,39 @@ class OpenTelemetryMiddleware:
...
     @staticmethod
-    def _create_start_response(span, start_response, response_hook):
+    def _create_start_response(
+        span, start_response, response_hook, duration_attrs
+    ):
         @functools.wraps(start_response)
         def _start_response(status, response_headers, *args, **kwargs):
             add_response_attributes(span, status, response_headers)
+            status_code = _parse_status_code(status)
+            if status_code is not None:
```
aabmass, line 416: Same comment as below about always setting attributes

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -405,28 +467,32 @@ def __call__(self, environ, start_response):
         if response_hook:
             response_hook = functools.partial(response_hook, span, environ)
 
+        start = default_timer()
```
aabmass, line 470: Hmm checking the docs, `default_timer()` is `time.perf_counter()`. How does this compare to `time.time_ns()` wall time? I believe we use the latter for span duration right

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -405,28 +467,32 @@ def __call__(self, environ, start_response):
...
         except Exception as ex:
             if span.is_recording():
                 span.set_status(Status(StatusCode.ERROR, str(ex)))
             span.end()
             if token is not None:
                 context.detach(token)
             raise
+        finally:
+            duration = max(default_timer() - start, 0)
+            self.duration_recorder.record(duration, duration_attrs)
```
aabmass, line 488: Looks like duration here would be [in fractional seconds](https://docs.python.org/3/library/time.html#time.perf_counter:~:text=Return%20the%20value-,(in%20fractional%20seconds,-)%20of%20a%20performance) but the instrument is configured with milliseconds.

instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
```diff
@@ -230,6 +241,30 @@ def test_wsgi_internal_error(self):
...
+        for resource_metric in metrics_list.resource_metrics:
+            self.assertTrue(len(resource_metric.scope_metrics) != 0)
+            for scope_metric in resource_metric.scope_metrics:
+                self.assertTrue(len(scope_metric.metrics) != 0)
+                for metric in scope_metric.metrics:
+                    self.assertIn(metric.name, _expected_metric_names)
+                    data_points = list(metric.data.data_points)
+                    self.assertEqual(len(data_points), 1)
+                    for point in data_points:
+                        if isinstance(point, HistogramDataPoint):
```
aabmass, line 261: What about the Gauge point for active requests?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
srikanthccv: Because not all attributes are available at the beginning of request ex status code

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
srikanthccv: Yes, that should work too

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
srikanthccv: Good point. My original intention was to filter out the attribute keys which are not available at all. For example I couldn't find `NET_HOST_NAME` in the collected requested attributes. Should we be setting it 0 all the time or just don't set it?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
srikanthccv: We don't really calculate the duration there. We needed a span start_time and end_time in nanos so `time.time_ns` was a good choice. Here we are interested in the which deals with the relative time. And docs say
> a clock with the highest available resolution to measure a short duration

So this seemed like better choice for measuring duration.

instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
srikanthccv: I wanted to ask what's the idea/purpose of active requests metrics since its aggregated value turns out to be zero most of times.

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
ocelotl: We should use a value that will never, under any circumstances be also a valid value. Maybe `None`?

aabmass: @srikanthccv maybe I misread the code but it looks like the intention here is to handle it dynamically if they are or aren't available? If the options are "setting it 0 all the time or just don't set it" i would say don't set it in that case 👍 I guess my point is we should keep the full attribute key set as if there was a strict type for it imo

> Maybe None?

None isn't allowed attribute value though? https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#:~:text=Attribute%20values%20of%20null%20are%20not%20valid%20and%20attempting%20to%20set%20a%20null%20value%20is%20undefined%20behavior

instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
aabmass: I think if server is relatively constant load, it should be nonzero. But ya, I had a discussion with a coworker about this as well and he had some ideas to improve things, I think we should raise it to the spec at some point.

I was mostly just asking for purposes of testing that the Gauge is emitted at all

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -158,13 +158,15 @@ def response_hook(span: Span, environ: WSGIEnvironment, status: str, response_he
 import functools
 import typing
 import wsgiref.util as wsgiref_util
+from timeit import default_timer
 
 from opentelemetry import context, trace
 from opentelemetry.instrumentation.utils import (
     _start_internal_or_server_span,
     http_status_to_status_code,
 )
```
lzchen, line 167: Can you add example usage of these metrics in the doc strings (maybe in flask or django)?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
lzchen, line 330: Could we use some of this logic for `add_response_attributes` for span as well?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
aabmass: Checked with coworkers offline about this. Seems like perf counter/CLOCK_MONOTONIC is the way to go here as it is meant to be wall time with no reference.

Side note, apparently for OTel Java there's a specific clock per-span that measures time-since-span-start using monotonic time

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
srikanthccv: This is not an opt-in feature. The existing example already cover it. What do you want to see in the doc string?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
srikanthccv: That logic looks bit different because it sets error status with status string even when it's invalid.

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
srikanthccv: Talked about this during 16th Jun SIG meeting. This requires some broader discussion because we setting zero-values might not be obvious to users and can't be generalised to non-existence of attribute value because there can be cases where zero-values are the legitimate input. And only setting the attributes that will always exist might be useful given the very little info from tag set. We agreed to get this merged and continue discussion for long term solution.

instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
srikanthccv: Added the test.

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
lzchen: Similar to [here](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1110/files)

srikanthccv: We didn't want to include any SDK setup in the instrumentation docs. That's why all of the instrumentation inline docs for tracing show how to setup just the instrumentation.

lzchen: Right. We should have some way that this should be exposed to the user though. Maybe in examples or getting started?

srikanthccv: >We should have some way that this should be exposed to the user though

That the instrumentation supports metrics? I am not sure if it's worth adding example each time we add metric instrumentation and it's no different from existing example. What I think would be helpful is if there is table which shows if instrumentation has metrics support yet.

lzchen: I like this idea. Do you want to create one as part of this PR or create an issue for it?

lzchen (approved):

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
srikanthccv: I can add that.

ocelotl (approved):

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
srikanthccv: Done here https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1128/files#diff-50998f4ff4326f7db47af0fe368ec0a8c2ddbae864cb40384355766c564910d6

instrumentation/README.md
```diff
@@ -1,43 +1,44 @@
...
-| [opentelemetry-instrumentation-system-metrics](./opentelemetry-instrumentation-system-metrics) | psutil >= 5 |
-| [opentelemetry-instrumentation-tornado](./opentelemetry-instrumentation-tornado) | tornado >= 5.1.1 |
-| [opentelemetry-instrumentation-urllib](./opentelemetry-instrumentation-urllib) | urllib |
-| [opentelemetry-instrumentation-urllib3](./opentelemetry-instrumentation-urllib3) | urllib3 >= 1.0.0, < 2.0.0 |
-| [opentelemetry-instrumentation-wsgi](./opentelemetry-instrumentation-wsgi) | wsgi |
\ No newline at end of file
+| Instrumentation | Supported Packages | Metrics support |
+| --------------- | ------------------ | --------------- |
+
+| [opentelemetry-instrumentation-aiohttp-client](./opentelemetry-instrumentation-aiohttp-client) | aiohttp ~= 3.0 | [ ]
```
aabmass, line 5: This doesn't render as is, missing the `|` at the end maybe?
```suggestion
| [opentelemetry-instrumentation-aiohttp-client](./opentelemetry-instrumentation-aiohttp-client) | aiohttp ~= 3.0 | [ ] |
```

scripts/generate_instrumentation_readme.py
```diff
@@ -62,11 +63,14 @@ def main():
             exec(fh.read(), pkg_info)
 
         instruments = pkg_info["_instruments"]
+        supports_metrics = pkg_info.get("_supports_metrics")
         if not instruments:
             instruments = (name,)
 
+        metric_colum = "[x]" if supports_metrics else "[ ]"
```
aabmass, line 70:
```suggestion
        metric_column = "[x]" if supports_metrics else "[ ]"
```

scripts/generate_instrumentation_readme.py
```diff
@@ -62,11 +63,14 @@ def main():
...
+
         table.append(
-            f"| [{instrumentation}](./{instrumentation}) | {','.join(instruments)} |"
+            f"| [{instrumentation}](./{instrumentation}) | {','.join(instruments)} | {metric_colum}"
```
aabmass, line 73:
```suggestion
            f"| [{instrumentation}](./{instrumentation}) | {','.join(instruments)} | {metric_column}"
```

aabmass (approved):

instrumentation/README.md
srikanthccv: Good catch. The suggestion  didn't change anything and couldn't figure why. I changed it to text Yes/No.

PR #1186, opened 2022-07-07
Metrics instrumentation flask

### LINKED ISSUE #1155: Metrics instrumentation flask
HTPP metrics: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md

TheAnshul756:
# Description

Fixes #1155 

## Type of change

# How Has This Been Tested?

- [x] `tox -e test-instrumentation-flask`

- [x] No.

# Checklist:


ocelotl:
@TheAnshul756 is this PR ready for review?

If yes, please remove the [WIP] from its name.
If not, please mark it as draft.

srikanthccv:
Good part of this is same as in WSGI instrumentation. I am wondering if we can base all wsgi compatible frameworks on it.

TheAnshul756:
Yes I think we can implement all the wsgi based frameworks taking inspiration from WSGI Instrumentation but I don't think we can have a common function to do this job in all frameworks. Correct me if I misunderstood what you are saying.

srikanthccv:
Many of WSGI complaint instrumentations ([flask](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/ee4083982f5919e1366de62b9cf4dac28cc8a314/instrumentation/opentelemetry-instrumentation-flask/setup.cfg#L45), [falcon](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/ee4083982f5919e1366de62b9cf4dac28cc8a314/instrumentation/opentelemetry-instrumentation-falcon/setup.cfg#L43), [pyramid](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/ee4083982f5919e1366de62b9cf4dac28cc8a314/instrumentation/opentelemetry-instrumentation-pyramid/setup.cfg#L46),  etc..) already take dep on `opentelemetry-instrumentation-wsgi` and use it.

srikanthccv:
And when both WSGI and instrumentation are enabled they produce idetical metrics with very little changes in the time series.

srikanthccv:
>because although the frameworks take dependency on wsgi but they do not use it for metric or trace instrumentation.

What do you mean? Why else do you think we are taking dep on wsgi instrumentation. Because we use parts of it.

>They have their own implementation.

Right, I am not saying they will not have any implementation on their own. But I don't see the reason each of them maintaining their own HTTP server semantic attributes, parsing the request response data from wsgi environ and filtering based on values and testing exact same thing unless you add some new tests (which is not the case with this PR).

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -327,7 +359,22 @@ def __init__(self, *args, **kwargs):
...
 
+        meter = get_meter(
+            __name__, __version__, _InstrumentedFlask._meter_provider
+        )
+        duration_histogram = meter.create_histogram(
+            name="http.server.duration",
+            unit="ms",
+            description="measures the duration of the inbound HTTP request",
+        )
+        active_requests_counter = meter.create_up_down_counter(
```
lzchen, line 370: If we have separate metric instruments for if user uses `_instrument` and `instrument_app`, does that mean that they will get different metric streams representing "metrics for all flask apps" and "metrics for a certain flask app"?

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -290,7 +307,21 @@ def _teardown_request(exc):
             # a way that doesn't run `before_request`, like when it is created
             # with `app.test_request_context`.
             return
+        start = flask.request.environ.get(_ENVIRON_STARTTIME_KEY)
+        duration = max(round((default_timer() - start) * 1000), 0)
```
lzchen, line 311: Shouldn't all this logic be in `_start_response`?

TheAnshul756: yes, that is also a possible way. But we are starting span in before_request and ending it in teardown_request so I did the same for metrics too. Also [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py#L184) it says that, it is discouraged, I'm also not sure what to prefer. If doing it in start_response is more acceptable then I can make the changes.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -290,7 +307,21 @@ def _teardown_request(exc):
             # a way that doesn't run `before_request`, like when it is created
             # with `app.test_request_context`.
             return
+        start = flask.request.environ.get(_ENVIRON_STARTTIME_KEY)
```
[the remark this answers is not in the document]
TheAnshul756: Yes. It returns a float value.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -184,7 +185,7 @@ def _wrapped_app(wrapped_app_environ, start_response):
         # In theory, we could start the span here and use
         # update_name later but that API is "highly discouraged" so
         # we better avoid it.
-        wrapped_app_environ[_ENVIRON_STARTTIME_KEY] = _time_ns()
+        wrapped_app_environ[_ENVIRON_STARTTIME_KEY] = default_timer()
```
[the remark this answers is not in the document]
TheAnshul756: fixed.

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -326,14 +326,30 @@ def collect_custom_response_headers_attributes(response_headers):
...
-def _parse_status_code(resp_status):
+def parse_status_code(resp_status):
     status_code, _ = resp_status.split(" ", 1)
     try:
         return int(status_code)
     except ValueError:
         return None
 
 
+def parse_active_request_count_attrs(req_attrs):
```
sanketmehta28, line 337: dont you need to remove it as you are using otel_wsgi.parse_active_request_count_attrs() ?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -436,15 +452,10 @@ def __call__(self, environ, start_response):
...
-        active_requests_count_attrs = {}
-        for attr_key in _active_requests_count_attrs:
-            if req_attrs.get(attr_key) is not None:
-                active_requests_count_attrs[attr_key] = req_attrs[attr_key]
-
-        duration_attrs = {}
-        for attr_key in _duration_attrs:
-            if req_attrs.get(attr_key) is not None:
-                duration_attrs[attr_key] = req_attrs[attr_key]
+        active_requests_count_attrs = parse_active_request_count_attrs(
```
sanketmehta28, line 455: this should be replaced with otel_wsgi.parse_active_request_count_attrs()

sanketmehta28 (changes requested):

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
TheAnshul756: This change is from wsgi file. I created a common function in wsgi file to be used in other wsgi-based frameworks.

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -326,14 +326,30 @@ def collect_custom_response_headers_attributes(response_headers):
     return attributes
 
 
-def _parse_status_code(resp_status):
+def parse_status_code(resp_status):
```
srikanthccv, line 329: I would still keep this and other functions private. They are not intended to be user accessible. If there is really need we can always change from private to public but not vice versa.

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
srikanthccv: No, this is function exists in the same file

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
sanketmehta28: ok got it.

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
sanketmehta28: ok got it

sanketmehta28 (approved):

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
TheAnshul756: done.

ocelotl:
Marking as draft to prevent accidental merging since there are enough approvals but @srikanthccv has still open comments. @srikanthccv feel free to mark as ready for review/merge when your comments have been addressed.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -178,13 +179,26 @@ def get_default_span_name():
...
+    response_hook=None,
+    excluded_urls=None,
+):
     def _wrapped_app(wrapped_app_environ, start_response):
         # We want to measure the time for route matching, etc.
         # In theory, we could start the span here and use
         # update_name later but that API is "highly discouraged" so
         # we better avoid it.
         wrapped_app_environ[_ENVIRON_STARTTIME_KEY] = _time_ns()
+        start = default_timer()
```
lzchen, line 195: I'm wondering if this is really the correct place to start the timer. I'm not too sure when `_wrapped_app` is called but is it invoked everytime a request is made or is only `_start_response` called upon each request?

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
```diff
@@ -250,6 +268,34 @@ def test_exclude_lists_from_explicit(self):
...
+            self.assertTrue(len(resource_metric.scope_metrics) != 0)
+            for scope_metric in resource_metric.scope_metrics:
+                self.assertTrue(len(scope_metric.metrics) != 0)
+                for metric in scope_metric.metrics:
+                    self.assertIn(metric.name, _expected_metric_names)
+                    data_points = list(metric.data.data_points)
+                    self.assertEqual(len(data_points), 1)
+                    for point in data_points:
+                        if isinstance(point, HistogramDataPoint):
+                            self.assertEqual(point.count, 3)
```
lzchen, line 289: Can we have a test for the actual values in the histogram? This will test that what we are recording for duration value is actually accurate.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
srikanthccv: Yes, it is invoked every time.

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
TheAnshul756: I'm sorry, I do not understand how can we get the value of time duration to test. Can you please give me an example?

srikanthccv: `point.sum` gives you the total duration.

TheAnshul756: Yes, we can get the actual value, but what will be the expected value? For three get requests `point.sum` value will be random time taken. We can not compare it to some hardcoded value. One possible solution I can think of is to mock the `default_timer()` function that we use to get the timestamp, and with that, we test the time duration captured. But I'm not sure if it is the correct way to test?

srikanthccv: You do not have to make it exact hardcoded value. It can be `assertGreaterEqual` or `assertAlmostEqual` with delta based on how much time a view/endpoint takes to respond back.

srikanthccv: trigger a request -> operation takes some time -> expected duration should be within x+/-∆

srikanthccv (changes requested):
blocking to prevent accidental merge to make sure tests are updated

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
TheAnshul756: I've added the test. Please have a look.

lzchen (approved):

srikanthccv (commented):
Does `uninstrument` stop producing the measurements? Please add some tests for `instruement`/`uninstrument` and other HTTP verbs.

TheAnshul756:
> Does `uninstrument` stop producing the measurements? Please add some tests for `instruement`/`uninstrument` and other HTTP verbs.

@srikanthccv I've added test for uninstrument.

What do you mean by other HTTP verbs?

srikanthccv:
I meant something other than `get`. Could be something like `client.{post,delete}` etc...

TheAnshul756:
> I meant something other than `get`. Could be something like `client.{post,delete}` etc...

I've changed get to post in one of the test. 

I'm just curious why it was needed? Because I think instrumentation logic is written independent of the type of request so if it is working fine with get then it should work fine with post too? Correct me if I'm wrong.

srikanthccv:
What I really wanted is a something like post request that sends form data and see the duration results.

lzchen:
Please update the metrics support for this instrumentation [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/README.md)

srikanthccv:
Or add [`_supports_metrics = True`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/package.py#L18) in package.py and run `tox -e generate`.

srikanthccv (approved):

PR #1198, opened 2022-07-19
Feature/metrics instrumentation urllib3

shalevr:
# Description
Add metrics instrumentation for urllib3

Fixes #1144 

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] tox -e test-instrumentation-urllib3

# Does This PR Require a Core Repo Change?

- [x] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [x] Followed the style guidelines of this project
- [x] Changelogs have been updated
- [x] Unit tests have been added

PR #1199, opened 2022-07-19
Metric instrumentation fastapi

### LINKED ISSUE #1156: Metrics instrumentation fastapi
HTTP metrics : https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md

TheAnshul756:
# Description

Fixes #1156

## Type of change

# How Has This Been Tested?

- [x] No.

# Checklist:


lzchen:
Depends on https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1197

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -454,10 +509,14 @@ async def __call__(self, scope, receive, send):
                     span_name,
                     scope,
                     send,
+                    duration_attrs,
                 )
 
                 await self.app(scope, otel_receive, otel_send)
```
lzchen, line 515: I'm wondering if we want to start the timer right before the send? Otherwise we will be counting the time it takes to do the OpenTelemetry stuff like set attributes, request/response hook, etc.?

TheAnshul756: added changes.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -161,6 +178,34 @@ def test_fastapi_excluded_urls_not_env(self):
...
+            self.assertTrue(len(resource_metric.scope_metrics) != 0)
+            for scope_metric in resource_metric.scope_metrics:
+                self.assertTrue(len(scope_metric.metrics) != 0)
+                for metric in scope_metric.metrics:
+                    self.assertIn(metric.name, _expected_metric_names)
+                    data_points = list(metric.data.data_points)
+                    self.assertEqual(len(data_points), 1)
+                    for point in data_points:
+                        if isinstance(point, HistogramDataPoint):
+                            self.assertEqual(point.count, 3)
```
lzchen, line 199: Can we test out the values as well? (sum, value, etc.)

TheAnshul756: addes changes.

lzchen (approved):

lzchen:
Please update the metrics support for this instrumentation [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/README.md)

srikanthccv:
Or Add _supports_metrics=True in the package.py file like [this](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/package.py#L18) and run tox -e generate

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -246,6 +250,7 @@ def __init__(self, *args, **kwargs):
             client_request_hook=_InstrumentedFastAPI._client_request_hook,
             client_response_hook=_InstrumentedFastAPI._client_response_hook,
             tracer_provider=_InstrumentedFastAPI._tracer_provider,
+            meter_provider=_InstrumentedFastAPI._meter_provider,
```
srikanthccv, line 253: This shouldn't pass the `meter_provider` because the instrumentation scope be will the asgi's `__name__`, and `__version__` which is incorrect. Create a meter here and pass it.

srikanthccv (changes requested):

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
TheAnshul756: How will we pass meter as asgi middleware accepts meter_provider. This change will require changes in Asgi middleware too. Also this same thing is happening in case of tracer_provoder too. Should we change that too?

srikanthccv: >Also this same thing is happening in case of tracer_provoder too. Should we change that too?

It's a bug with trace provider too but that can be addressed in separate issue. 

>How will we pass meter as asgi middleware accepts meter_provider. This change will require changes in Asgi middleware too.

Yes, it should be changed.

TheAnshul756: added changes as suggested. Please review.

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-instrumentation-sqlalchemy` Added span for the connection phase ([#1133](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1133))
 - Add metric instrumentation in asgi
   ([#1197](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1197))
+- Add metric instrumentation in fastapi
```
srikanthccv, line 24: Move this to unreleased section

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -183,6 +183,26 @@ def client_response_hook(span: Span, message: dict):
...
+]
+
+_active_requests_count_attrs = [
+    SpanAttributes.HTTP_METHOD,
+    SpanAttributes.HTTP_HOST,
+    SpanAttributes.HTTP_SCHEME,
+    SpanAttributes.HTTP_FLAVOR,
+    SpanAttributes.HTTP_SERVER_NAME,
+]
+
```
srikanthccv, line 205: I believe you added these to `opentelemetry-util-http` lib. Please import from there.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -365,6 +385,22 @@ def get_default_span_details(scope: dict) -> Tuple[str, dict]:
...
+            active_requests_count_attrs[attr_key] = req_attrs[attr_key]
+    return active_requests_count_attrs
+
+
+def _parse_duration_attrs(req_attrs):
+    duration_attrs = {}
+    for attr_key in _duration_attrs:
+        if req_attrs.get(attr_key) is not None:
+            duration_attrs[attr_key] = req_attrs[attr_key]
+    return duration_attrs
```
srikanthccv, line 401: Same

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
srikanthccv: This is still not fixed fully.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -21,7 +22,15 @@
 
 import opentelemetry.instrumentation.fastapi as otel_fastapi
 from opentelemetry import trace
-from opentelemetry.instrumentation.asgi import OpenTelemetryMiddleware
+from opentelemetry.instrumentation.asgi import (
+    OpenTelemetryMiddleware,
+    _active_requests_count_attrs,
+    _duration_attrs,
+)
```
srikanthccv, line 29: Same as above

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -161,6 +179,106 @@ def test_fastapi_excluded_urls_not_env(self):
...
+                            self.assertEqual(point.count, 1)
+                            self.assertAlmostEqual(
+                                duration, point.sum, delta=10
+                            )
+                        if isinstance(point, NumberDataPoint):
+                            self.assertEqual(point.value, 0)
+
+    def test_metric_uninstruemnt(self):
+        self._client.get("/foobar")
+        self._instrumentor.uninstrument_app(self._app)
```
srikanthccv, line 270: Also please add tests for `uninstrument`.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -183,6 +183,26 @@ def client_response_hook(span: Span, message: dict):
...
+    SpanAttributes.HTTP_HOST,
+    SpanAttributes.HTTP_SCHEME,
+    SpanAttributes.HTTP_STATUS_CODE,
+    SpanAttributes.HTTP_FLAVOR,
+    SpanAttributes.HTTP_SERVER_NAME,
+    SpanAttributes.NET_HOST_NAME,
+    SpanAttributes.NET_HOST_PORT,
+]
+
+_active_requests_count_attrs = [
```
ocelotl, line 198: These lists should be sets, order is not meaningful for these keys:

``` python
_duration_attrs = {
    "SpanAttributes.HTTP_METHOD",
    "SpanAttributes.HTTP_HOST",
    "SpanAttributes.HTTP_SCHEME",
    "SpanAttributes.HTTP_STATUS_CODE",
    "SpanAttributes.HTTP_FLAVOR",
    "SpanAttributes.HTTP_SERVER_NAME",
    "SpanAttributes.NET_HOST_NAME",
    "SpanAttributes.NET_HOST_PORT",
}


def _parse_duration_attrs(req_attrs):
    print(
        {
            key: req_attrs[key] for key in _duration_attrs.intersection(
                req_attrs.keys()
            )
        }
    )


req_attrs = {
    "SpanAttributes.HTTP_METHOD": 1,
    "SpanAttributes.HTTP_HOST": 2,
    "SpanAttributes.HTTP_SCHEME": 3,
    "SpanAttributes.HTTP_X": 5,
    "SpanAttributes.HTTP_Y": 6,
}

_parse_duration_attrs(req_attrs)
```
```
{'SpanAttributes.HTTP_HOST': 2, 'SpanAttributes.HTTP_METHOD': 1, 'SpanAttributes.HTTP_SCHEME': 3}
```

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -161,6 +179,106 @@ def test_fastapi_excluded_urls_not_env(self):
...
+    def test_fastapi_metrics(self):
+        self._client.get("/foobar")
+        self._client.get("/foobar")
+        self._client.get("/foobar")
+        metrics_list = self.memory_metrics_reader.get_metrics_data()
+        number_data_point_seen = False
+        histogram_data_point_seen = False
+        self.assertTrue(len(metrics_list.resource_metrics) != 0)
+        for resource_metric in metrics_list.resource_metrics:
+            self.assertTrue(len(resource_metric.scope_metrics) != 0)
```
ocelotl, line 191: Is it possible to check equality against an exact value instead?

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -161,6 +179,106 @@ def test_fastapi_excluded_urls_not_env(self):
...
+    def test_metric_uninstruemnt(self):
+        self._client.get("/foobar")
+        self._instrumentor.uninstrument_app(self._app)
+        self._client.get("/foobar")
+        metrics_list = self.memory_metrics_reader.get_metrics_data()
+        for resource_metric in metrics_list.resource_metrics:
+            for scope_metric in resource_metric.scope_metrics:
+                for metric in scope_metric.metrics:
+                    for point in list(metric.data.data_points):
+                        if isinstance(point, HistogramDataPoint):
```
ocelotl, line 277: Can these tests be made exact? Do we expect more than one element in `resource_metric.scope_metrics`, for example?

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
TheAnshul756: I'm sorry I didn't get what you mean by test for uninstrument. In this test I uninstrumented the app and tested if the metric is still present or not.

srikanthccv: There is both `uninstrument` and `uninstrument_app`. I was asking to include tests for former where an instrumentation is enabled with call `instrument` instead of `instruemt_app`.

TheAnshul756: added changes. Please review.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -161,6 +177,125 @@ def test_fastapi_excluded_urls_not_env(self):
...
+        # instrumenting class and creating app to send request
+        self._instrumentor.instrument()
+        app = self._create_fastapi_app()
+        client = TestClient(app)
+        client.get("/foobar")
+        # uninstrumenting class and creating the app again
+        self._instrumentor.uninstrument()
+        app = self._create_fastapi_app()
+        client = TestClient(app)
+        client.get("/foobar")
```
srikanthccv, line 287: You shouldn't create a new app. You should use the same app instance from above.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -161,6 +177,125 @@ def test_fastapi_excluded_urls_not_env(self):
...
+        self._client.get("/foobar")
+        metrics_list = self.memory_metrics_reader.get_metrics_data()
+        number_data_point_seen = False
+        histogram_data_point_seen = False
+        self.assertTrue(len(metrics_list.resource_metrics) == 1)
+        for resource_metric in metrics_list.resource_metrics:
+            self.assertTrue(len(resource_metric.scope_metrics) == 1)
+            for scope_metric in resource_metric.scope_metrics:
+                self.assertTrue(len(scope_metric.metrics) == 2)
+                print(scope_metric.metrics)
```
srikanthccv, line 192:
```suggestion
```

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -161,6 +177,125 @@ def test_fastapi_excluded_urls_not_env(self):
...
+        self._client.get("/foobar")
+        self._client.get("/foobar")
+        metrics_list = self.memory_metrics_reader.get_metrics_data()
+        number_data_point_seen = False
+        histogram_data_point_seen = False
+        self.assertTrue(len(metrics_list.resource_metrics) == 1)
+        for resource_metric in metrics_list.resource_metrics:
+            self.assertTrue(len(resource_metric.scope_metrics) == 1)
+            for scope_metric in resource_metric.scope_metrics:
+                self.assertTrue(len(scope_metric.metrics) == 2)
```
srikanthccv, line 191: What are the two scopes here? Shouldn't there be only one scope?

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
TheAnshul756: We can not use the same app instance as the earlier app object was created by patched fastapi library. Once we uninstrument, original fastapi class is restored, and we need to create a new fastapi app object to see the changes. Correct me if I'm wrong.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
TheAnshul756: There will be two scope metrics, one is duration and another is active requests.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
srikanthccv: No, the test is to ensure the patching is undone correctly when uninstrumented and the same app instance doesn't produce any more telemetry.

TBBle: Is that behaviour/expectation of `uninstrument` documented anywhere?

Poking around, most of the web API instrumentation libraries I looked at don't do this, e.g. Flask, aiohttp-client, Starlette, FastAPI (i.e. this one) all have their `uninstrument` (and `instrument`) APIs only affect _new_ instances, not existing instances.

Some _do_ unhook existing hooks, e.g., Tornado. I'll note that the Tornado instrumentation can _only_ globally instrument/uninstrument, there isn't an API like `instrument_app` as there is for the ASGI and WSGI code, because it relies on patching the underlying class, rather than hooking `__init__` to insert a middleware in the request flow.

TheAnshul756: > the test is to ensure the patching is undone correctly when uninstrumented and the same app instance doesn't produce any more telemetry.

`uninstrument_app` will uninstrument the existing app and I've added a test for the same but `uninstrument` will just restore the class and all the existing app instances that are already created will not be affected by the change and it'll keep on sending the telemetry data. I don't think we can write test for what you are suggesting.

TBBle: I actually meant specifically for this repo. My list of comparisons was other instrumentation packages here. It might be worth documenting the generic instrumentation behaviours expected in this repo for all instruments, as what should have been a near-trivial PR (add metrics to the existing tracing hooks, because the instrumentation hook setup and ASGI intercept was already done) has turned into a significant refactoring and if that's going to be the case for all the Python instrumentation packages, then it might be worth doing that refactoring _first_ across the repo for the instrumentation packages that need it.

That will also help ensure that any new instrumentation contributions behave correctly from first PR, rather than needing months of reworking _away_ from the existing behaviours. It would also possibly allow some more-generic tests (or at least test-templates) for things like the instrument/uninstrument behaviours being iterated on here.

srikanthccv: We may have missed this for some instrumentation earlier but we try to ensure that `uninstrument` tests are added and work as expected. For instance, the last two PRs adding metrics for [django](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1208#discussion_r943744431) and [urllib3](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1198/commits/fcb2d42e121f24b68a0ebb4d67ce68abf35e1568) didn't initially have them but later added in review. And same applies to the new instrumentation that gets added (most recent one https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1082) The general guideline help for new instrumentation specifically for this repo is written here https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md#guideline-for-instrumentations. Changes across the repo require some proposal/discussion and followed by implementation. Feel free to volunteer to take that up and help us.

TheAnshul756: Can we merge this PR for now and address the changes in other PR so that it do not get blocked for long?

srikanthccv (approved):

PR #1234, opened 2022-08-18
boto3sqs: Make propagation compatible with other instrumentations and add 'messaging.url' span attribute

### LINKED ISSUE #1202: boto3sqs: Do not override propagator-determined key
The boto3sqs instrumentation prepends "otel." to all attributes it adds to SQS messages (relevant code: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/7c75b3867bb81d59037e7a3f5ab5c3dfe1bb5733/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py#L56-L82). This will make it impossible to write a propagator that works with instrumentations (of OTel or other vendors) that don't use this prefix.

The reasons cited at https://www.oxeye.io/blog/diving-into-opentelemetrys-specs:

* Prevents propagators setting keys that start with AWS. or Amazon., which would be rejected -- a rather weak reason IMHO as this is very unlikely to happen and there are other, perhaps more likely, rules that keys must adhere to that are not addressed.
* Keys needed for extraction, or prefix thereof, must be known before calling the receive API which requires a list of attributes to receive -- this can be solved by using [`TextMapPropagator.fields`](https://github.com/open-telemetry/opentelemetry-python/blob/v1.11.1/opentelemetry-api/src/opentelemetry/propagators/textmap.py#L185-L196), although it is meant for listing injected fields, these will usually be the same as extracted ones. This is also what Node.Js does: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/1db1fecc16ecb3dbad530de530418260e54c087a/plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/sqs.ts#L71-L75

So I suggest to remove the `otel.` prefix and instead use  [`TextMapPropagator.fields`](https://github.com/open-telemetry/opentelemetry-python/blob/v1.11.1/opentelemetry-api/src/opentelemetry/propagators/textmap.py#L185-L196) to ensure the required attributes are fetched.

CC @oxeye-nikolay

mariojonke:
# Description

When propagating context via message attributes the instrumentation used `otel.` prefix for keys injected or extracted from an carrier. This makes it incompatible with other instrumentations (e.g. aws-sdk instrumentation for Node.Js). 
Instead of prefixed keys use plain keys from the propagator.
SQS `recieve_message` calls require to provide a list with the names of message attributes that are to be delivered with the message. Instead of using the `otel.*` wildcard use the `fields` attribute provided by the propagator.

Fixes #1202

* Additionally adds the `messaging.url` as span attribute for the created spans since this is the main identifier for SQS when interacting with a queue.
* Removes setting the `messaging.temp_destination` span attribute which is only intended for temporary queues and doesn't really make sense for SQS.

### Fixes also the following issues to make the instrumentation unit testable
* add the instrumentation to the `tox.ini` file to actually run the unit tests in CI
* add missing required dependencies to `setup.cfg`

Additionally the current instrumentation only worked for the first instance of a `boto3.SQS` client since it was instrumenting only the first SQS like sub class of the `botocore.client.BaseClient` class. boto3 (or more precisely botocore) however generates a new subclass for each new instance of a client.
For unit testing (and also production) this is sub-optimal as it means to either have one unit test or reuse the same client over multiple tests (and hope that no other test created an SQS client instance).
Thus, this PR includes also the fix to make it possible to instrument multiple instances of an SQS client.

## Type of change

# How Has This Been Tested?

Adapted existing Getter and Setter tests to check injected propagation keys.
Added additional unit tests to actually check created spans, span attributes and that context is injected/extracted accordingly.

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
@@ -415,6 +415,8 @@ class Boto3SQSInstrumentor(BaseInstrumentor):
         unwrap(sqs_class, "delete_message")
         unwrap(sqs_class, "delete_message_batch")
 
+        setattr(sqs_class, _IS_SQS_INSTRUMENTED_ATTRIBUTE, False)
+
     def _instrument(self, **kwargs: Dict[str, Any]) -> None:
         self._tracer_provider: Optional[TracerProvider] = kwargs.get(
             "tracer_provider"
--- a/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
@@ -29,7 +29,7 @@ Usage
     Boto3SQSInstrumentor().instrument()
 """
 import logging
-from typing import Any, Collection, Dict, Generator, List, Optional
+from typing import Any, Collection, Dict, Generator, List, Mapping, Optional
 
 import boto3
 import botocore.client
@@ -53,33 +53,31 @@ from .package import _instruments
 from .version import __version__
 
 _logger = logging.getLogger(__name__)
-# We use this prefix so we can request all instrumentation MessageAttributeNames with a wildcard, without harming
-# existing filters
-_OPENTELEMETRY_ATTRIBUTE_IDENTIFIER: str = "otel."
-_OTEL_IDENTIFIER_LENGTH = len(_OPENTELEMETRY_ATTRIBUTE_IDENTIFIER)
+
+_IS_SQS_INSTRUMENTED_ATTRIBUTE = "_otel_boto3sqs_instrumented"
 
 
 class Boto3SQSGetter(Getter[CarrierT]):
     def get(self, carrier: CarrierT, key: str) -> Optional[List[str]]:
-        value = carrier.get(f"{_OPENTELEMETRY_ATTRIBUTE_IDENTIFIER}{key}", {})
-        if not value:
+        msg_attr = carrier.get(key)
+        if not isinstance(msg_attr, Mapping):
             return None
-        return [value.get("StringValue")]
+
+        value = msg_attr.get("StringValue")
+        if value is None:
+            return None
+
+        return [value]
 
     def keys(self, carrier: CarrierT) -> List[str]:
-        return [
-            key[_OTEL_IDENTIFIER_LENGTH:]
-            if key.startswith(_OPENTELEMETRY_ATTRIBUTE_IDENTIFIER)
-            else key
-            for key in carrier.keys()
-        ]
+        return list(carrier.keys())
 
 
 class Boto3SQSSetter(Setter[CarrierT]):
     def set(self, carrier: CarrierT, key: str, value: str) -> None:
         # This is a limitation defined by AWS for SQS MessageAttributes size
         if len(carrier.items()) < 10:
-            carrier[f"{_OPENTELEMETRY_ATTRIBUTE_IDENTIFIER}{key}"] = {
+            carrier[key] = {
                 "StringValue": value,
                 "DataType": "String",
             }
@@ -145,6 +143,7 @@ class Boto3SQSInstrumentor(BaseInstrumentor):
     def _enrich_span(
         span: Span,
         queue_name: str,
+        queue_url: str,
         conversation_id: Optional[str] = None,
         operation: Optional[MessagingOperationValues] = None,
         message_id: Optional[str] = None,
@@ -157,12 +156,12 @@ class Boto3SQSInstrumentor(BaseInstrumentor):
             SpanAttributes.MESSAGING_DESTINATION_KIND,
             MessagingDestinationKindValues.QUEUE.value,
         )
+        span.set_attribute(SpanAttributes.MESSAGING_URL, queue_url)
+
         if operation:
             span.set_attribute(
                 SpanAttributes.MESSAGING_OPERATION, operation.value
             )
-        else:
-            span.set_attribute(SpanAttributes.MESSAGING_TEMP_DESTINATION, True)
         if conversation_id:
             span.set_attribute(
                 SpanAttributes.MESSAGING_CONVERSATION_ID, conversation_id
@@ -190,15 +189,19 @@ class Boto3SQSInstrumentor(BaseInstrumentor):
         return queue_url.split("/")[-1]
 
     def _create_processing_span(
-        self, queue_name: str, receipt_handle: str, message: Dict[str, Any]
+        self,
+        queue_name: str,
+        queue_url: str,
+        receipt_handle: str,
+        message: Dict[str, Any],
     ) -> None:
         message_attributes = message.get("MessageAttributes", {})
         links = []
         ctx = propagate.extract(message_attributes, getter=boto3sqs_getter)
-        if ctx:
-            for item in ctx.values():
-                if hasattr(item, "get_span_context"):
-                    links.append(Link(context=item.get_span_context()))
+        parent_span_ctx = trace.get_current_span(ctx).get_span_context()
+        if parent_span_ctx.is_valid:
+            links.append(Link(context=parent_span_ctx))
+
         span = self._tracer.start_span(
             name=f"{queue_name} process", links=links, kind=SpanKind.CONSUMER
         )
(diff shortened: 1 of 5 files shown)


instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -375,52 +379,56 @@ def _wrap_client_creation(self) -> None:
...
-            unwrap(self._sqs_class, "send_message_batch")
-            unwrap(self._sqs_class, "receive_message")
-            unwrap(self._sqs_class, "delete_message")
-            unwrap(self._sqs_class, "delete_message_batch")
-            self._did_decorate = False
+        if not hasattr(sqs_class, "send_message_batch"):
+            return
+
+        if getattr(sqs_class, _IS_SQS_INSTRUMENTED_ATTRIBUTE, False):
+            return
```
oxeye-nikolay, line 397: I think this caching mechanism is not needed. There cannot be two `botocore.client.BaseClient` anyway, so if you instrumented once, you will not be able to instrument again (because of the instrumentor's base class caching mechanism)

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -53,33 +53,31 @@
...
-            else key
-            for key in carrier.keys()
-        ]
+        return list(carrier.keys())
 
 
 class Boto3SQSSetter(Setter[CarrierT]):
     def set(self, carrier: CarrierT, key: str, value: str) -> None:
         # This is a limitation defined by AWS for SQS MessageAttributes size
         if len(carrier.items()) < 10:
```
oxeye-nikolay, line 79: AWS does not accept keys that start with `AWS` or `Amazon`. Can we add another check here that enforces that, and if this is the case a warning will be printed?

Oberon00: I don't think guarding against this very specific case is worth it, unless you know any propagator that actually uses such keys? There are rules about periods/dots in the name as well and a maximum length which we also don't check.

Oberon00 (approved):

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: You're right. We should also enforce these. Otherwise, the behavior would be a failure to send SQS messages *because* the user added OpenTelemetry. Adding OpenTelemtry must not alter the original behavior of the application

Oberon00: That's an ideal, not something we can always guarantee. Checking for startswith AWS/Amazon sounds at least reasonably easy & fast, but should we really run a regex over each request? It just feels like going too far to me.

oxeye-nikolay: I think it's something that cannot be ignored. The instrumentation must comply with the amazon API

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
mariojonke: While it is true that there is only one `botocore.client.BaseClient`, there can be multiples of any class that is generated by botocore, like `botocore.client.SQS`.  When calling `boto3.client(...)`, `botocore` will internally [generate](https://github.com/boto/botocore/blob/1.27.56/botocore/client.py#L174) the concrete service class.  It will do so every time even if the class for an AWS service was already created before. This means, when you call `boto3.client("sqs")` two or more times you will get instances that are type `botocore.client.SQS` but their actual `botocore.client.SQS` type is not the same.
The `_IS_SQS_INSTRUMENTED_ATTRIBUTE` attribute is an indicator for the instrumentor to know which class was already instrumented. It is only set onto the concrete/generated class returned by the `boto3.client(...)` call. The `botocore.client.BaseClient` is never modified and does not have this attribute at any time.

Your comment made me think that `_IS_SQS_INSTRUMENTED_ATTRIBUTE` should be reset/unset in the `_un_decorate_sqs` function as otherwise the concrete class will not instumented again in a instrument -> uninstrument -> instrument call sequence.

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
Oberon00: We also don't check that for any HTTP instrumentation. There are rules for HTTP headers as well. So it surely _can_ be ignored 😃

mariojonke:
Only partly related to this PR, but was there a special reason why the SQS instrumentation was implemented as separate package and not integrated into the [botocore instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-botocore)?

From what i saw, everything in this package should also be doable with an [extension](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions) in the botocore instrumentation package (e.g. similar to what the [DynamoDB extension](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/v0.33b0/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/dynamodb.py#L349) is doing). There is already even a more or less empty [SQS extension](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/sqs.py) where this package could be integrated.

Integrating this package into the `botocore instrumentation` would have some advantages imo:
* no need for users to install two separate packages
* simplifies/solves suppression of downstream spans (boto3sqs -> botocore) if both instrumentations are installed
* fewer resources used in CI since only one package needs to be installed/tested
* single codebase for everything boto3/botocore instrumentation related
* no need to instrument generated subclasses since instrumentation is already taken care of by the botocore instrumenation.

oxeye-nikolay:
> Only partly related to this PR, but was there a special reason why the SQS instrumentation was implemented as separate package and not integrated into the [botocore instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-botocore)?
> 
> From what i saw, everything in this package should also be doable with an [extension](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions) in the botocore instrumentation package (e.g. similar to what the [DynamoDB extension](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/v0.33b0/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/dynamodb.py#L349) is doing). There is already even a more or less empty [SQS extension](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/sqs.py) where this package could be integrated.
> 
> Integrating this package into the `botocore instrumentation` would have some advantages imo:
> 
> * no need for users to install two separate packages
> * simplifies/solves suppression of downstream spans (boto3sqs -> botocore) if both instrumentations are installed
> * fewer resources used in CI since only one package needs to be installed/tested
> * single codebase for everything boto3/botocore instrumentation related
> * no need to instrument generated subclasses since instrumentation is already taken care of by the botocore instrumenation.

This instrumentation was created separately because it instruments `boto3`, while the [botocore instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-botocore) instruments `botocore`. These are two different libraries (despite their strong coupling).

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
oxeye-nikolay: You're right. I now understand your solution :)

oxeye-nikolay (approved):

Oberon00:
> This instrumentation was created separately because it instruments boto3, while the [botocore instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-botocore) instruments botocore.

Could this be changed to also instrument botocore instead?

PR #1247, opened 2022-08-27
Get rid of pymongo command in pymongo instrumentation span name

amitsetty:
As others have pointed out 
https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/341
by including the query within the span name we lead to very high cardinality which is too high to be useful and this can also lead to privacy concerns

This pull requests follows the logic found in the java and golang mongo instrumentations and does not include the entire command but the database name and command name within the span name

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/mongo/mongo-3.1/library/src/main/java/io/opentelemetry/instrumentation/mongo/v3_1

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/go.mongodb.org/mongo-driver/mongo/otelmongo/mongo.go

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] test_pymongo.py

- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `opentelemetry-instrumentation-boto3sqs` Make propagation compatible with other SQS instrumentations, add 'messaging.url' span attribute, and fix missing package dependencies.
-  ([#1234](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1234))
+  ([#1234](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1234)) 
+- `opentelemetry-instrumentation-pymongo` Change span names to not contain queries but only database name and command name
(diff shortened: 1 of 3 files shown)


srikanthccv (commented):
While at it could you please check what's the semconv recommendation for this? Also please add some tests.

amitsetty:
Hey I checked the semconv recommendation and it seems like this is more in line with [https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md](the conventions) 

While being in line with all stated conventions there is an example within the conventions which shows instrumentation for a Mongo database. There is a small diff here where the example instrumentation contains a collection name - however the [https://api.mongodb.com/python/3.1/api/pymongo/monitoring.html](the pymongo library monitoring tool) does not return the collection name

More work could be done here in the future to try to add the collection name. In the meantime it is imperative that the instrumentation is in line with the overall database specification as having a very high cardinality leads the instrumentation to become a lot less useful

amitsetty:
Hey would love if someone could review this

srikanthccv (approved):

PR #1258, opened 2022-09-01
uninstruemnt existing instances before uninstrumenting fastapi class

### LINKED ISSUE #1156: Metrics instrumentation fastapi

### LINKED ISSUE #1256: Uninstrument Existing fastapi instances
Existing instances shouldn't continue to send any telemetry after the `uninstrument`.

_Originally posted by @srikanthccv in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1199#discussion_r957094027_

`uninstrument` method of `FastAPIInstrumentor` restore the original `fastapi.FastAPI` class, but all the existing objects of the earlier instrumented class do not get uninstrumented, which is not the expected behaviour.

TheAnshul756:
# Description

Create a static list of all instances of InstrumentedFastAPI and fix uninstrument function.

Fixes #1256 

## Type of change

# How Has This Been Tested?

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -218,14 +228,33 @@ def _instrument(self, **kwargs):
...
         fastapi.FastAPI = _InstrumentedFastAPI
+        for instance in _InstrumentedFastAPI._instrumented_fastapi_apps:
+            self.instrument_app(
+                instance,
+                server_request_hook=_InstrumentedFastAPI._server_request_hook,
+                client_request_hook=_InstrumentedFastAPI._client_request_hook,
+                client_response_hook=_InstrumentedFastAPI._client_response_hook,
+                tracer_provider=_InstrumentedFastAPI._tracer_provider,
+                excluded_urls=_InstrumentedFastAPI._excluded_urls,
+            )
```
srikanthccv, line 248: I am not following this logic. Can you elaborate what this PR is doing?

TheAnshul756: So, as discussed, we wanted to uninstrument existing instrumented apps when we call `instrumentor.uninstrument()` so we created a static list named `_instrumented_fastapi_apps` in `_InstrumentedFastAPI` class, and when we make any instance of that class, then we add it to the list. We use the list in the following ways:
- On `instrumentor.uninstrument()` call, we call `uninstrument_app()` for all the existing instances in the list.
- In the case of `instrumentor.instrument()` call again after uninstrument, we again instrument the app instances of the list.

Apart from this, we also add those app instances, if any, that were manually instrumented and add them to the list. But in the case of `instrumentor.uninstrument()` call, we remove them from that list as they were instances of original `fastapi.FastAPI` class and not `_InstrumentedFastAPI`  class, so that they do not get instrumented again on call of `instrumentor.instrument()`. We can change this logic and keep them there if we want.

Also, apart from all this, we added a check for parsed `excluded_urls` and do not call `parse_excluded_urls()` method for already parsed URLs as it throws error.

srikanthccv: > In the case of instrumentor.instrument() call again after uninstrument, we again instrument the app instances of the list.

Why? Shouldn't it be only concerned about the instance after the `instrument` call?

TheAnshul756: Yes, that is also one possible way. But I proceeded with this approach because these apps are instances of the `_InstrumentedFastAPI` class, so I thought if we are instrumenting again, then all those instances which are of the `_instrumentedFastAPI` class should also get instrumented back.

I can revert the changes if you feel this is not expected behaviour.

srikanthccv: I don't have a strong opinion on this. As a user, I would uninstument to stop the application to produce all telemetry but the old instances getting re-instrumeted would be a surprise. For example I would do some cleanup stuff where uninstrument is called and then the app instance is deleted but since you keep the reference here it's actually not the behavior one would expect. What do you think?

sanketmehta28: Hi @srikanthccv : If we delete the instance after uninstrument(), it will remove that app instance from the list, So it will not get instrumented again in any case. 
I guess it makes sense if we allow to re-instrument in case the `_instrumentedFastAPI ` app instance is not deleted and user call instrument() on them. If user does not want to re-instrument again, he should not call instrument() after uninstrument() at all.

srikanthccv: > If user does not want to re-instrument again, he should not call instrument() after uninstrument() at all.

No, we can't say this to the user. They can and should be able to call `instrument` and `uninstrument` any number of times. The current implementation is even inconsistent with itself in regards to `instrument_app`/`instrument`. I think we should be only concerned after the call to `instrument` and do not automatically re-insturment the prev instances of `_instrumentedFastAPI`.

srikanthccv: > I think we should be only concerned after the call to instrument and do not automatically re-insturment the prev instances of _instrumentedFastAPI.

And why I think this should be the expected behaviour is, that if a user were to use this instrumentation and invoke `instrument` the first time the prior instances to this call are not instrumented and it should be the same when they do it with `uninstrument` and then `instrument`.

TheAnshul756: Yes, it makes sense. I'll revert the changes and make a commit.

srikanthccv (approved):

lzchen:
@TheAnshul756  @srikanthccv 
> So, as discussed, we wanted to uninstrument existing instrumented apps when we call instrumentor.uninstrument()

When was this discussed? As a user this makes sense to me but we'll have to do this for other instrumentations as [well](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py#L500). Could we make tracking issues for those?

srikanthccv:
This was briefly discussed in one of the SIG meetings when the topic was brought up by I believe Sanket. Some instrumentations haven't been fixed for this case but generally, we are trying to ensure the `uninstrument` works as expected. Sure, I will create an issue for the instrumentation we have missed this.

sanketmehta28 (approved):

sanketmehta28:
Resolve the generate test issue

PR #1324, opened 2022-09-11
asyncpg: Use only the first word from query as a span name

### LINKED ISSUE #1170: AsyncPG Instrumentation Span Names are too long when using query string
Not a problem per se however, the `asyncpg` instrumentation uses sets span names as the query string which results in some very messing looking trace names in jaeger, datadog, etc and outright doesn't work with promscale due to long queries exhaust the available bytes for btree indexes.

- The ability to change the name of the span with a hook or something similar. The `httpx` instrumentation provides hooks that receive the span and the name can be updated there.
- Just use a shorter or truncated version of the query as the name.

Which alternative solutions or features have you considered?
Not using the `asyncpg` instrumentation and manually instrumenting specific queries.

krnr:
# Description

to reduce high cardinal value for a span name in the DB instrumentation use only the first word from the query.

Other DB instrumentations (e.g. psycopg2) do likewise.

Fixes #1170 

## Type of change

# How Has This Been Tested?

tested locally with a simple endpoint which selects data from a local db

before:
image
after:
image

- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 
 
+- `opentelemetry-instrumentation-asyncpg` Fix high cardinality in the span name
+  ([#1170](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1324))
+
 ### Added
 
 - `opentelemetry-instrumentation-grpc` add supports to filter requests to instrument. ([#1241](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1241))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 
 - `opentelemetry-instrumentation-asyncpg` Fix high cardinality in the span name
-  ([#1170](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1324))
+  ([#1324](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1324))
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
```diff
@@ -133,6 +133,7 @@ async def _do_execute(self, func, instance, args, kwargs):
         exception = None
         params = getattr(instance, "_params", {})
         name = args[0] if args[0] else params.get("database", "postgresql")
+        name = name.split()[0]
```
srikanthccv, line 136: Maybe wrap this in `try...except IndexError...`?

CHANGELOG.md
```diff
@@ -41,9 +41,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - restoring metrics in django framework
   ([#1208](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1208))
 - `opentelemetry-instrumentation-aiohttp-client` Fix producing additional spans with each newly created ClientSession
-- ([#1246](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1246))
+  ([#1246](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1246))
 - Add _is_openetlemetry_instrumented check in _InstrumentedFastAPI class
   ([#1313](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1313))
+- `opentelemetry-instrumentation-asyncpg` Fix high cardinality in the span name
```
srikanthccv, line 47: Move this under Unreleased section

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
krnr: good idea! i'll use `name = ""` in except block, since this may happen for both "" and " " (`\s*` in general).

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 
 
+- `opentelemetry-instrumentation-asyncpg` Fix high cardinality in the span name
+  ([#1170](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1324))
```
srikanthccv, line 19:
```suggestion
  ([#1324](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1324))
```

srikanthccv (approved):

PR #1344, opened 2022-09-20
fix flaky test in fastapi

TheAnshul756:
# Description

fix the flaky test in fastapi.

## Type of change

# How Has This Been Tested?

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
+++ b/instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
@@ -236,7 +236,7 @@ class TestFastAPIManualInstrumentation(TestBase):
                         dict(point.attributes),
                     )
                     self.assertEqual(point.count, 1)
-                    self.assertAlmostEqual(duration, point.sum, delta=20)
+                    self.assertAlmostEqual(duration, point.sum, delta=30)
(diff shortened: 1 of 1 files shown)


instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -236,7 +236,7 @@ def test_basic_metric_success(self):
                         dict(point.attributes),
                     )
                     self.assertEqual(point.count, 1)
-                    self.assertAlmostEqual(duration, point.sum, delta=20)
+                    self.assertAlmostEqual(duration, point.sum, delta=30)
```
srikanthccv, line 239: You could also use something like `@flaky(max_runs=..., min_sucess=...)` if this is small but eventually passes with a re-run.

srikanthccv (approved):

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
TheAnshul756: Hey @srikanthccv 
It sounds good as even `delta = 30` could be flaky. Should I create a PR and add this in all the frameworks where this kind of test is written??

PR #1365, opened 2022-09-30
Manual instrumentation falcon

### LINKED ISSUE #1342: Manual instrumentation in Falcon Framework
Currently Falcon only has support for automatic instrumentation and not for manual instrumentation. We can provide support for manual instrumentation in falcon by creating static methods `instrument_app` and `uninsturment_app` in instrumentor class.

TheAnshul756:
# Description

Manual Instrumentation in Falcon

Fixes #1342 

## Type of change

# How Has This Been Tested?

- [ ] `tox -e test-instrumentation-falcon1`
- [ ] `tox -e test-instrumentation-falcon2`
- [ ] `tox -e test-instrumentation-falcon3`

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
@@ -455,6 +455,46 @@ class FalconInstrumentor(BaseInstrumentor):
 
     def instrumentation_dependencies(self) -> Collection[str]:
         return _instruments
+    
+    @staticmethod
+    def instrument_app(
+        app, 
+        request_hook=None, 
+        response_hook=None, 
+        tracer_provider=None, 
+        meter_provider=None, 
+        excluded_urls=None,
+        **kwargs
+    ):
+        if not hasattr(app, "_is_instrumented_by_opentelemetry"):
+            app._is_instrumented_by_opentelemetry = False
+        
+        if not getattr(app, "_is_instrumented_by_opentelemetry", False):
+            app._otel_excluded_urls = excluded_urls if excluded_urls is None else get_excluded_urls("FALCON")
+            app._otel_tracer = trace.get_tracer(__name__, __version__, tracer_provider)
+            app._otel_meter = get_meter(__name__, __version__, meter_provider)
+            app.duration_histogram = self._otel_meter.create_histogram(
+                name="http.server.duration",
+                unit="ms",
+                description="measures the duration of the inbound HTTP request",
+            )
+            app.active_requests_counter = app._otel_meter.create_up_down_counter(
+                name="http.server.active_requests",
+                unit="requests",
+                description="measures the number of concurrent HTTP requests that are currently in-flight",
+            )
+            trace_middleware = _TraceMiddleware(
+                app._otel_tracer,
+                kwargs.pop(
+                    "traced_request_attributes", get_traced_request_attrs("FALCON")
+                ),
+                request_hook,
+                response_hook,
+            )
+            # TODO: implement middleware addition logic for falcon 1,2 & 3
+            app._is_instrumented_by_opentelemetry = True
+            app.__class__ = _InstrumentedFalconAPI
+
 
     # pylint:disable=no-self-use
     def _remove_instrumented_middleware(self, app):
--- a/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
@@ -146,6 +146,7 @@ from sys import exc_info
 from time import time_ns
 from timeit import default_timer
 from typing import Collection
+from xxlimited import new
 
 import falcon
 from packaging import version as package_version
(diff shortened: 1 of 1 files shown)


sanketmehta28 (approved):

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -461,8 +522,76 @@ class FalconInstrumentor(BaseInstrumentor):
...
+                        description="measures the duration of the inbound HTTP request",
+                    )
+                    self.active_requests_counter = self._otel_meter.create_up_down_counter(
+                        name="http.server.active_requests",
+                        unit="requests",
+                        description="measures the number of concurrent HTTP requests that are currently in-flight",
+                    )
+                    self._is_instrumented_by_opentelemetry = False
+
+            app = FalconAPI()
```
srikanthccv, line 564: I didn't understand this. The `app` arg passed is getting replaced by this; how doe this work?

TheAnshul756: So, here we wanted to add more attributes in the original falcon api, but falcon do not let us add anything to already created app because of the use of [`__slots__`](https://towardsdatascience.com/understand-slots-in-python-e3081ef5196d#:~:text=However%2C%20__slots__%20will%20fix%20the%20attributes%20when%20you%20create%20the%20class.%20So%2C%20it%E2%80%99s%20not%20possible%20to%20add%20new%20attributes%20later.). So, what we did is, we created a new child class of Falcon API and copied all the attributes in the new objects with the additional attributes that we wanted to add and returned the new app instead of the original app.

srikanthccv: Does it contain everything? What happens to the rest of the attributes/properties/basically everything other than what `__slots__` have on `app` instance?

srikanthccv (changes requested):
To prevent accidental merge before the comments are resolved.

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
TheAnshul756: Yes. Falcon can not have additional instance variables because of `__slots__`  and because the new object is of child class so all the methods of falcon class will be inherited.

srikanthccv: I am a little skeptical of this. I will try to find an example where this is not true.

srikanthccv: I think this is kind of sloppy. Try customising the base falcon and adding some instance variables, which will fail. For example, try this, and your manual instrumentation test fails with `AttributeError: 'FalconAPI' object has no attribute 'class_var'`

```patch
diff --git a/instrumentation/opentelemetry-instrumentation-falcon/tests/app.py b/instrumentation/opentelemetry-instrumentation-falcon/tests/app.py
index 3e4c62ec..ccac7113 100644
--- a/instrumentation/opentelemetry-instrumentation-falcon/tests/app.py
+++ b/instrumentation/opentelemetry-instrumentation-falcon/tests/app.py
@@ -64,11 +64,17 @@ class CustomResponseHeaderResource:
 def make_app():
     _parsed_falcon_version = package_version.parse(falcon.__version__)
     if _parsed_falcon_version < package_version.parse("3.0.0"):
+        class MyAPI(falcon.API):
+            class_var = "class_var"
+
         # Falcon 1 and Falcon 2
-        app = falcon.API()
+        app = MyAPI()
     else:
+        class MyApp(falcon.App):
+            class_var = "class_var"
+
         # Falcon 3
-        app = falcon.App()
+        app = MyApp()

     app.add_route("/hello", HelloWorldResource())
     app.add_route("/ping", HelloWorldResource())
diff --git a/instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py b/instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
index df6f77c4..77eda353 100644
--- a/instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
+++ b/instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
@@ -272,6 +272,7 @@ class TestFalconInstrumentation(TestFalconBase, WsgiTestBase):
         self.client().simulate_get(path="/hello")
+        self.assertEqual(self.app.class_var, "class_var")

     def test_falcon_metrics(self):
         self.client().simulate_get("/hello/756")
@@ -394,6 +395,7 @@ class TestFalconManualInstrumentation(TestFalconInstrumentation):
         self.client().simulate_get(path="/hello")
+        self.assertEqual(self.app.class_var, "class_var")

     def test_metric_uninstrument(self):
         self.client().simulate_request(method="POST", path="/hello/756")
```

TheAnshul756: Yes this is one possibility but this case is also not difficult to handle. I've added changes for that with the test you did. Please review.

srikanthccv: It's not about the difficulty of handling the case. We don't want to break people's application to make instrumentation work. The new try-except looks hacky to just address this comment.

TheAnshul756: Yes, I understand your point. What are the other possibilities? Do you have any suggestion? I can not think of anything else.

srikanthccv: I would move the tracer/meter/instruments/config state to `_TraceMiddleware` since it's the one that's actually using them and then add it to middleware list.

TheAnshul756: Not just `_TraceMiddleware` but `__call__` function from `_InstrumentedFalconAPI` also uses those states. If we move all these states to `_TraceMiddleware` then we'll need to move all the instrumentation logic to middleware. Are you suggesting that?

srikanthccv: I think `_InstrumentedFalconAPI` is a simple WSGI middleware. It's not necessary they maintain separate tracer/meter/instruments on it's own. It can use what will be used for the TraceMiddleware.

TheAnshul756: Sorry I'm not able to understand. Can you please elaborate?

ocelotl:
Just for the record, discussed in today's SIG, @sanketmehta28 will look into other SIGs implementations maybe they have a way to solve this problem that can work for us :v:

ocelotl:
> @ocelotl : I guess you have mentioned that in wrong issue. Please delete your comment from here and paste it on [this issue](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1372)

Right, moved :v:

PR #1373, opened 2022-10-05
Use compatible runtime for the dependency in Opentelemetry-distro

### LINKED ISSUE #1375: Recurring python version conflict error after every release
**Describe your environment** 
This issue is created to kick off the discussion and lead to possible solution for the version conflict error. 

After a new release in `otel-python` and `otel-python-contrib` the usage of the instrumentation in previous releases is leading to version conflct error. While using the instrumentations of previous release `1.12.0` has broken when using the current Release `1.13.0`  
it is suspected that pinned version requirement dependency of opentelemetry-sdk in [opentelemetry-distro](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-distro/pyproject.toml#L28) might be the reason.

Use compatible Runtime `~=` for dependencies like [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/sdk-extension/opentelemetry-sdk-extension-aws/pyproject.toml#L27) ? in opentelemetry-distro and any other similar places. 

No Version conflict and ease to use the instrumentation of previous release version.
 
image

**Additional context**
For more context [requirements.txt](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/python/src/otel/otel_sdk/requirements.txt)
[Relevant slack thread](https://cloud-native.slack.com/archives/C01PD4HUVBL/p1661927606492329)

vasireddy99:
# Description

This PR aims to fix #1375 

`skip change log`would be needed

## Type of change


opentelemetry-distro/pyproject.toml
```diff
@@ -25,7 +25,7 @@ classifiers = [
 dependencies = [
   "opentelemetry-api ~= 1.12",
   "opentelemetry-instrumentation == 0.34b0",
-  "opentelemetry-sdk == 1.13.0",
+  "opentelemetry-sdk ~= 1.13.0",
```
srikanthccv, line 28: This is equivalent to: opentelemetry-sdk >= 1.13.0, == 1.13.* but what ideally should be ~= 1.13
```suggestion
  "opentelemetry-sdk ~= 1.13",
```

vasireddy99: good call out! :-)

srikanthccv (approved):

PR #1382, opened 2022-10-14
[#1353]fix dbapi connection instrument wrapper has no _sock member

### LINKED ISSUE #1353: conn.commit() raise InterfaceError because connection instrument wrapper has no _sock member

instrumentation version: opentelemetry-instrumentation-pymysql    0.33b0

**Steps to reproduce**
When using `opentelemetry-instrument` command to automatically instrument pymysql

```
conn = pymysql.connect(...)

with conn.cursor() as cursor:
  cursor.execute(...) # run any insert statement
  conn.commit() # this line will trigger InterfaceError
```

Current bandaid solution:
```
conn = pymysql.connect(...)
conn._sock = conn._connection._sock # add this line to work around

with conn.cursor() as cursor:
  cursor.execute(...)
  conn.commit() # this line will no longer trigger InterfaceError
```

Tracing down the stack will show that on line 792 of the `pymysql/connections/.py`, the Exception will be thrown because self now refers to the wrapper instead of the internal connection object:
```
    def _execute_command(self, command, sql):
        """
        :raise InterfaceError: If the connection is closed.
        :raise ValueError: If no username was specified.
        """
        if not self._sock:
            raise err.InterfaceError(0, "")
        ...
```
If I do not instrument pymysql then this will not happen.

jae-jae:
Fixes #1353

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
@@ -324,6 +324,10 @@ def get_traced_connection_proxy(
     # pylint: disable=abstract-method
     class TracedConnectionProxy(type(connection), _TracedConnectionProxy):
         def __init__(self, connection):
+            self._sock = connection._sock
+            self._auth_plugin_name = connection._auth_plugin_name
+            self._closed = connection._closed
+            self._secure = connection._secure
             self._connection = connection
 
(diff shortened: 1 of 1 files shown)


rogersd:
@jae-jae I'm not a reviewer, but they're probably going to want a `CHANGELOG` entry for this.

Also, thanks for opening this PR, I'd spent far too much time trying to figure out why newer versions completely broke my app.

srikanthccv:
Please add tests

srikanthccv:
Preceded by #1424. It has unit tests, and the author is responsive.

PR #1410, opened 2022-10-27
Cherrypy instrumentation

### LINKED ISSUE #1130: Auto Instrumentation support for CherryPy Python Framework
Before opening a feature request against this repo, consider whether the feature should/could be implemented in the [other OpenTelemetry client libraries](https://github.com/open-telemetry/). If so, please [open an issue on opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/issues/new) first.

Unable to set up open-telemetry auto-instrumentation for python CherrPy APP.

TheAnshul756:
# Description

Fixes #1130 

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-cherrypy/src/opentelemetry/instrumentation/cherrypy/__init__.py
```diff
@@ -0,0 +1,140 @@
...
+    """An instrumentor for FastAPI
+
+    See `BaseInstrumentor`
+    """
+
+    def instrumentation_dependencies(self) -> Collection[str]:
+        return _instruments
+
+    def _instrument(self, **kwargs):
+        self._original_cherrypy_application = cherrypy._cptree.Application
```
srikanthccv, line 35: There should be a better way to do this. We can't use the private symbols because they are not guaranteed to remain backward compatible.

instrumentation/opentelemetry-instrumentation-cherrypy/src/opentelemetry/instrumentation/cherrypy/__init__.py
```diff
@@ -0,0 +1,140 @@
...
+        cherrypy._cptree.Application = _InstrumentedCherryPyApplication
+        cherrypy.Application = _InstrumentedCherryPyApplication
```
srikanthccv, line 37: https://docs.cherrypy.dev/en/latest/extend.html#per-request-functions

instrumentation/opentelemetry-instrumentation-cherrypy/src/opentelemetry/instrumentation/cherrypy/package.py
```diff
@@ -0,0 +1,3 @@
+_instruments = ("cherrypy >= 1.0",)
```
srikanthccv, line 1: Please also pin the upper version, there can be breaking changes b/w major version changes.

instrumentation/opentelemetry-instrumentation-cherrypy/src/opentelemetry/instrumentation/cherrypy/__init__.py
```diff
@@ -0,0 +1,140 @@
...
+                status_code = otel_wsgi._parse_status_code(status)
+                if status_code is not None:
+                    duration_attrs[SpanAttributes.HTTP_STATUS_CODE] = status_code
+                if span.is_recording() and span.kind == trace.SpanKind.SERVER:
+                    custom_attributes = otel_wsgi.collect_custom_response_headers_attributes(response_headers)
+                    if len(custom_attributes) > 0:
+                        span.set_attributes(custom_attributes)
+            
+            if self.response_hook:
+                self.response_hook(span, status, response_headers)
```
srikanthccv, line 118: Please add tests. If the PR is not ready then please mark it a draft.

TheAnshul756: Hey, I'm really sorry. I forgot to mark this PR draft. There's still lot of work required in this PR.

instrumentation/opentelemetry-instrumentation-cherrypy/src/opentelemetry/instrumentation/cherrypy/__init__.py
TheAnshul756: Issue with these hooks are that they needed to be called for every paths that we want to use them for. They do not apply directly to everywhere. [source](https://docs.cherrypy.dev/en/latest/extend.html#per-request-functions:~:text=Using%20that%20tool%20is%20as%20simple%20as%20follows%3A)

TheAnshul756:
I've came up with three different approaches for instrumentation in cherrypy and I wanted to discuss all three of them with there pros and cons.
### Instrumentation with CherryPy Hooks
Unlike other frameworks, CherryPy does not have support for custom middleware. Instead, CherryPy provides some [hook points](https://docs.cherrypy.dev/en/latest/extend.html#id16) where you can add your functions. [Datadog](https://github.com/DataDog/dd-trace-py/blob/1.x/ddtrace/contrib/cherrypy/middleware.py) uses this approach for instrumentation. But all the endpoints for which tools need to be used must be decorated by that tool.
Eg.
```
class Root(object):
    @cherrypy.expose
    @cherrypy.tools.logit()
    def index(self):
        return "hello world"
```
Because of it, this approach can only be used for manual instrumentation, not auto-instrumentation. Although there are some workarounds that I came up with (like monkey patching the expose function) but they were also not covering all the possible cases (what if the function is not exposed with expose decorator `index.exposed = true`).

### Instrumentation with CherryPy Publisher Subscriber
> CherryPy’s backbone consists of a bus system implementing a simple publish/subscribe messaging pattern. Simply put, in CherryPy everything is controlled via that bus.

Various channels are available in CherryPy, but only two were helpful to us, i.e. `before_request` and `after_request`. [Scout Python APM Agent](https://github.com/scoutapp/scout_apm_python/blob/master/src/scout_apm/cherrypy.py) use this approach of instrumentation. But the challenge with this approach is the `before_request` channel triggers at a very early stage of the request handling cycle, even before the `request` object parses the `environ` dictionary from the HTTP message, because of which we can not do much in before_request (including span creation). Thus this approach leaves us with a significant limitation.

### Instrumentation by Patching CherryPy Application class
This is how we have instrumented various frameworks, and this approach can also be done here. [New Relics](https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/hooks/framework_cherrypy.py) and Appdynamics agent also uses this approach. The issue with this approach is `Application` class is implemented in [`_cptree` module](https://github.com/cherrypy/cherrypy/blob/5964c48b466448a14703eecc062223fda46caf4e/cherrypy/_cptree.py#L10) in CherryPy and imported in `__init__` module. We can patch the class imported in `__init__`, but this will not enable instrumentation as CherryPy creates the [object of the `Application` class](https://github.com/cherrypy/cherrypy/blob/5964c48b466448a14703eecc062223fda46caf4e/cherrypy/_cptree.py#L231) in  `_cptree` and uses the `Application` from that module directly, which forces us to instrument `Application` class both in `__init__` and `_cptree`. I have used this approach for instrumenting in my PR but @srikanthccv believes 
> We can't use the private symbols because they are not guaranteed to remain backwards compatible.

I've implemented all three approaches and tested them to understand their pros and cons. The first two approaches will also come with other challenges, such as handling exceptions. And the third approach, although being backward incompatible, is tried and tested and guarantees to work. 
I need help knowing what to do next and which approach to proceed with. Also, if anyone has other ideas, I'm more than interested to know.

instrumentation/opentelemetry-instrumentation-cherrypy/README.rst
```diff
@@ -0,0 +1,40 @@
+OpenTelemetry CherryPy Tracing
```
sanketmehta28, line 1: Please add an example how to use this instrumentation

instrumentation/opentelemetry-instrumentation-cherrypy/src/opentelemetry/instrumentation/cherrypy/__init__.py
```diff
@@ -0,0 +1,167 @@
+from logging import getLogger
+from time import time_ns
+from timeit import default_timer
+from typing import Collection
+
+from opentelemetry.util.http import parse_excluded_urls, get_excluded_urls, get_traced_request_attrs
+import cherrypy
```
sanketmehta28, line 7: organize the import statement properly

instrumentation/opentelemetry-instrumentation-cherrypy/src/opentelemetry/instrumentation/cherrypy/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+from opentelemetry.instrumentation.cherrypy.version import __version__
+from opentelemetry.metrics import get_meter
+
+
+_logger = getLogger(__name__)
+_excluded_urls_from_env = get_excluded_urls("CHERRYPY")
+
+
+class CherryPyInstrumentor(BaseInstrumentor):
+    """An instrumentor for FastAPI
```
sanketmehta28, line 26: Change the docstring from FastAPI to CherryPy. Also please add an example how to use it

instrumentation/opentelemetry-instrumentation-cherrypy/src/opentelemetry/instrumentation/cherrypy/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+            return super().__call__(environ, _start_response)
+        except Exception as exc:
+            exception = exc
+            raise
+        finally:
+            if exception is None:
+                activation.__exit__(None, None, None)
+            else:
+                activation.__exit__(
+                    type(exc),
```
sanketmehta28, line 159: you should use exception instead of exc just to make it more readable

instrumentation/opentelemetry-instrumentation-cherrypy/tests/test_cherrypy.py
```diff
@@ -0,0 +1,513 @@
...
+        self.assertEqual(len(span_list), 1)
+
+        server_span = [
+            span for span in span_list if span.kind == trace.SpanKind.SERVER
+        ][0]
+
+        for key, _ in not_expected.items():
+            self.assertNotIn(key, server_span.attributes)
+
+class TestNonRecordingSpanWithCustomHeaders(TestBase, helper.CPWebCase):
```
sanketmehta28, line 464: why dont you inherit this class from the TestCherryPyBase and override all functions? This will remove some redundant code

sanketmehta28 (changes requested):

sanketmehta28:
Also resolve the build checks

TheAnshul756:
> Overall, I don't think it's a good idea to patch the private/internals of the library. We have seen this fail with a few instrumentations, so they either had to be pinned to some minor version which works or rewritten to not patch private symbols and get it to work in general. I would like to hear others thoughts on this.

@open-telemetry/opentelemetry-python-contrib-approvers Please share your thoughts on this.

PR #1434, opened 2022-11-09
Strip leading comments from SQL queries when generating the span name.

### LINKED ISSUE #1433: Strip leading comments from SQL queries when generating the span name.
Currently an SQL statement that begins with a comment, such as

```
/* Get foo from bar */ SELECT foo FROM bar
```

Will have `span.name` set to `/*`, which makes it harder to tell what's going on in the waterfall view for a request.

Instead, the leading comment should be removed so that the operation type is used instead.

rogersd:
# Description

Currently an SQL statement that begins with a comment, such as

```
/* Get foo from bar */ SELECT foo FROM bar
```

Will have `span.name` set to `/*`, which makes it harder to tell what's going on in the waterfall view for a request.

Instead, this removes the leading comment so that the operation type is used instead.

Fixes #1433 

## Type of change

# How Has This Been Tested?

- [X] Added additional unit tests.
- [X] Tested against my company's code.

- [X] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#1350](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1350))
 - `opentelemetry-instrumentation-starlette` Add support for regular expression matching and sanitization of HTTP headers.
   ([#1404](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1404))
+- Strip leading comments from SQL queries when generating the span name.
+  ([#1434](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1434))
 
 ### Fixed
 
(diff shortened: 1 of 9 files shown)


srikanthccv:
We also have similar logic in a few other places and need to update them too. If you don't want to address it in this PR, Please update the description from Fixes #XX -> Part of #XX.

rogersd:
I updated it in `psycopg2` but I don't see any other places.  Let me know if I missed a spot.

srikanthccv:
[here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py#L110-L123) and [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py#L131-L140)

srikanthccv (approved):

PR #1460, opened 2022-11-18
Fix httpx version

### LINKED ISSUE #1459: httpx tests failing on httpx==0.23.1

ocelotl:
Fixes #1459

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-httpx/pyproject.toml
+++ b/instrumentation/opentelemetry-instrumentation-httpx/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
-  "httpx >= 0.18.0",
+  "httpx >= 0.18.0, <= 0.23.0",
 ]
 test = [
   "opentelemetry-instrumentation-httpx[instruments]",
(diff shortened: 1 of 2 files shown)


aabmass (commented):
Can we leave https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1459 open until we actually fix the instrumentation to work with that version?

ocelotl:
> Can we leave #1459 open until we actually fix the instrumentation to work with that version?

Opened #1463 :v:

lzchen (approved):

phillipuniverse:
@lzchen @ocelotl this is the 2nd time a version change like this has bit me, the first one being changed FastAPI requirements. I think that these need to be highlighted much more explicitly in new releases of the otel instrumentation since there is no other way to notify downstream users.

In my case, I am using httpx 0.23.3 and requests were being instrumented ok, but then this version changed and now httpx is silently uninstrumented. I don't configure instrumentation dependencies explicitly as I instead pull in everything via the `opentelemetry-instrumentations` package.

Could we get some better messaging around when these version requirements change between releases? I'm not sure what the right process is but silently failing can be very frustrating as often you don't realize it until you go and look for the data and it's not there.

phillipuniverse:
I added more color to a first-class issue report at #1745

PR #1512, opened 2022-12-20
mongo db - fix db statement capturing

### LINKED ISSUE #1462: mongo DB instrumentation doesn't capture `db.statement` properly
According to the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes) - `db.statement` should capture the entire request, and not only the command name.

**Steps to reproduce**
Instrument a client using PymongoInstrumentor.
Send a request to the db.

Produce a span with a `db.statement` value containing the whole request.

Produce a span with `db.statement` containing only the request name.

**Additional context**
Another discussion [here](https://github.com/open-telemetry/opentelemetry-specification/issues/2907)

Another thing that needs to be fixed is adding the ability to hide the `db.statement` value on demand, in order to hide sensitive data when needed.

If you can, assign this to me - I would like to solve it :)

avzis:
# Description
Fix capturing the DB_STATEMENT parameter correctly in mongo db.
Also added an optional value to enable capturing the database statement that is being executed.

Fixes #1462 

## Type of change

# How Has This Been Tested?

- [x] the tests now validates the db.statement for every type of request: find, update, delete, insert
- [x] added a test that validates that db.statement is not being captured when 'capture_statement' is disabled

- [x] No.


instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
```diff
@@ -106,29 +110,28 @@ def __init__(
...
         ):
             return
-        command = event.command.get(event.command_name, "")
-        name = event.database_name
-        name += "." + event.command_name
-        statement = event.command_name
-        if command:
-            statement += " " + str(command)
+        command_name = event.command_name
+        span_name = event.database_name + "." + command_name
```
shalevr, line 130: Consider use python format strings instead of concatenate strings

shalevr (approved):

srikanthccv (approved):

Issue #1491, opened 2022-12-18
Missing readthedocs instrumentation documentation


nemoshlag:
Several instrumentations are missed within the [read the docs](https://opentelemetry-python-contrib.readthedocs.io/en/latest/index.html) documentation. This is auto-generated and requires a format for the following missing docs instrumentations:

- [x] #1493 
- [x] #1494 
- [x] #1495 
- [x] #1496 
- [x] #1497 
- [x] #1498 
- [x] #1499 
- [x] #1500 
- [x] #1501 
- [x] #1502 
- [x] #1503

PR #1535, opened 2022-12-25
add docs/instrumentation component owner

nemoshlag:
I've noticed many instrumentations are missing from the readthedocs documentation. I add myself as the component owner for verifying new instrumentations are added and to keep a similar documentation structure. 
See issue #1491


srikanthccv (approved):
Thank you

PR #1399, opened 2022-10-23
Fix exception in Urllib3 when dealing with filelike body.

isra17:
### Description

Only use `len` on `body` if it's not a filelike object.

Fixex https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1235

### How Has This Been Tested?

tox -e test-instrumentation-urllib3

PR #1553, opened 2022-12-28
Add metric instrumentation for urllib

### LINKED ISSUE #1041: [instrumentation-urllib] Restore metrics
The urllib instrumentation used to produce metrics: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/metrics/instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py

This issue is to restore that functionality.

shalevr:
# Description
Add metrics instrumentation for urllib accroding to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#http-server

Fixes #1041

## Type of change

# How Has This Been Tested?

- [x] tox -e test-instrumentation-urllib

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-urllib/tests/test_metrics_instrumentation.py
```diff
@@ -0,0 +1,248 @@
...
+    def sorted_metrics(metrics):
+        """
+        Sorts metrics by metric name.
+        """
+        return sorted(
+            metrics,
+            key=lambda m: m.name,
+        )
+
+    def assert_metric_expected(
```
shalevr, line 76: I used utils function for metrics assert
they can be deleted after https://github.com/open-telemetry/opentelemetry-python/pull/3092 merged

avzis (approved):
looking great

instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
```diff
@@ -214,6 +225,10 @@ def _instrumented_open_call(
                         SpanAttributes.HTTP_FLAVOR
                     ] = f"{ver_[:1]}.{ver_[:-1]}"
 
+                _record_histograms(
+                    histograms, labels, request, result, elapsed_time
+                )
```
srikanthccv, line 230: Will there be a case where the result is None, but we should record the timing regardless of its return value?

instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
```diff
@@ -248,3 +263,44 @@ def _uninstrument_from(instr_root, restore_as_bound_func=False):
...
+
+def _record_histograms(
+    histograms, metric_attributes, request, response, elapsed_time
+):
+    histograms[MetricInstruments.HTTP_CLIENT_DURATION].record(
+        elapsed_time, attributes=metric_attributes
+    )
+
+    data = getattr(request, "data", None)
+    request_size = 0 if data is None else len(data)
```
srikanthccv, line 298: Will this be a compressed size? Does data always return an iterable?

instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
```diff
@@ -248,3 +263,44 @@ def _uninstrument_from(instr_root, restore_as_bound_func=False):
...
+    histograms[MetricInstruments.HTTP_CLIENT_REQUEST_SIZE].record(
+        request_size, attributes=metric_attributes
+    )
+
+    response_size = int(response.headers.get("Content-Length", 0))
+    histograms[MetricInstruments.HTTP_CLIENT_RESPONSE_SIZE].record(
```
srikanthccv, line 304: Same question about the compression as the instrument description make it explicit

instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
srikanthccv: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1399

shalevr: Thanks for the comment, I checked this issue will I wrote the code and the difference is the urllib api works only with bytes, and urllib3 works with bytes and IO(like BytesIO)

instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
shalevr: You are right this line should be outside the condition

shalevr: I changed it so if the result is None we add records only for http_client_duration and http_client_request size,
the status code and HTTP flavor are missing from attributes because we take them from the response.
how does it sound?

instrumentation/opentelemetry-instrumentation-urllib/tests/test_metrics_instrumentation.py
srikanthccv: I have some reservations about the PR on the main repo. I will leave a review there about it.

shalevr: Thank you!

Issue #1478, opened 2022-12-07
Instrumentation raises uncaught exception on non utf-8 request header sequences


marianogappa:
**Describe your environment** 
Poetry (version 1.2.2)
Python 3.10.4
opentelemetry-instrumentation-fastapi = ">=0.32b0,<1.0dev"

**Steps to reproduce**
Send an HTTP request to a Python ASGI application with `opentelemetry-instrumentation-fastapi` instrumentation, which contains non utf-8 sequences in its request headers.

**What is the expected behavior?**
REST request does not fail without the instrumentation, and thus should not fail with the instrumentation either.

**What is the actual behavior?**

```
ERROR 2022-12-07 00:47:10,938 run_asgi Exception in ASGI application
 
Traceback (most recent call last):
  File "/root/.cache/pypoetry/virtualenvs/--REDACTED--/lib/python3.10/site-packages/opentelemetry/instrumentation/asgi/__init__.py", line 560, in __call__
    collect_custom_request_headers_attributes(scope)
  File "/root/.cache/pypoetry/virtualenvs/--REDACTED--/lib/python3.10/site-packages/opentelemetry/instrumentation/asgi/__init__.py", line 345, in collect_custom_request_headers_attributes
    headers = {
  File "/root/.cache/pypoetry/virtualenvs/--REDACTED--/lib/python3.10/site-packages/opentelemetry/instrumentation/asgi/__init__.py", line 346, in <dictcomp>
    _key.decode("utf8"): _value.decode("utf8")
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xbf in position 0: invalid start byte

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/root/.cache/pypoetry/virtualenvs/--REDACTED--/lib/python3.10/site-packages/uvicorn/protocols/http/httptools_impl.py", line 372, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/root/.cache/pypoetry/virtualenvs/--REDACTED--/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 75, in __call__
    return await self.app(scope, receive, send)
  File "/root/.cache/pypoetry/virtualenvs/--REDACTED--/lib/python3.10/site-packages/fastapi/applications.py", line 270, in __call__
    await super().__call__(scope, receive, send)
[...]

PR #1564, opened 2023-01-06
Fix UnboundLocalError in asgi handler

mike-baker-on:
# Description
Currently the duration calculation in the ASGI instrumentation fails if anything outside the `self.app` call raises an error because start will not have been instantiated.

This hides any instansation errors are actually occurring since they will be replaced by the `UnboundLocalError`

This PR moves the duration tracking into its own try/finally block so it will not be ran unless self.app was called.

It would have also been possible to set `start` to None at the top of the function and then have an if statement inside the finally block, however I decided that was too close to overloading the concept of the `start` variable as a `should record duration` control flow flag.

The actual issue is a unicode decoding error in the handling of headers, however since https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1492 potentially fixes this issue too I have restricted the scope of this PR to be focused on error propagation.

Fixes [(#1478)](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1478)

## Type of change

# How Has This Been Tested?

Added new unit test (`test_failed_instrumentation_errors_propagate`) it currently uses the `server_request_hook` functionality to force an instrumentation error without relying on any specific bugs.

The test does also rely on setting self.communicator to None in order to prevent the error being re-raised in teardown, relevant code links:

https://github.com/open-telemetry/opentelemetry-python/blob/b184dc95844561f0dce3f5c0a496bc89d0fb9a6d/tests/opentelemetry-test-utils/src/opentelemetry/test/asgitestutil.py#L38

https://github.com/django/asgiref/blob/36022636d3390c1c1cbb952336a34d652a6bfb26/asgiref/testing.py#L8

- [ ] Yes.
- [ x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
@@ -572,15 +572,17 @@ class OpenTelemetryMiddleware:
                     duration_attrs,
                 )
                 start = default_timer()
-
-                await self.app(scope, otel_receive, otel_send)
+                try:
+                    await self.app(scope, otel_receive, otel_send)
+                finally:
+                    if scope["type"] == "http":
+                        target = _collect_target_attribute(scope)
+                        if target:
+                            duration_attrs[SpanAttributes.HTTP_TARGET] = target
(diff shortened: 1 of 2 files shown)


thomasleveil:
I guess this may be closed since #1889 is merged

PR #1574, opened 2023-01-11
add a test for urllib3 using NoOpTracerProvider

### LINKED ISSUE #960: Audit and test opentelemetry-instrumentation-urllib3

avzis:
# Description

Add a test for urllib3 instrumentation to ensure that NoOpTracerProvider works

Fixes #960 

# How Has This Been Tested?

- Instrument the library using a NoOpTracerProvider. the test validates that no spans are being produced.

- No.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-urllib3/tests/test_urllib3_integration.py
+++ b/instrumentation/opentelemetry-instrumentation-urllib3/tests/test_urllib3_integration.py
@@ -330,3 +330,12 @@ class TestURLLib3Instrumentor(TestBase):
 
         self.assertIn("request_hook_body", span.attributes)
         self.assertEqual(span.attributes["request_hook_body"], body)
+
+    def test_no_op_tracer_provider(self):
+        URLLib3Instrumentor().uninstrument()
+        tracer_provider = trace.NoOpTracerProvider
+        URLLib3Instrumentor().instrument(tracer_provider=tracer_provider)
+
+        response = self.perform_request(self.HTTP_URL)
+        self.assertEqual(b"Hello!", response.data)
+        self.assert_span(num_spans=0)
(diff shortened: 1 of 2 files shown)


shalevr (approved):

instrumentation/opentelemetry-instrumentation-urllib3/tests/test_urllib3_integration.py
```diff
@@ -330,3 +330,12 @@ def request_hook(span, request, headers, body):
 
         self.assertIn("request_hook_body", span.attributes)
         self.assertEqual(span.attributes["request_hook_body"], body)
+
+    def test_no_op_tracer_provider(self):
+        URLLib3Instrumentor().uninstrument()
+        tracer_provider = trace.NoOpTracerProvider
```
mariojonke, line 336: shouldn't the `NoOpTracerProvider` be instantiated here instead of using the class?

avzis: @mariojonke what do you mean?

mariojonke: `trace.NoOpTracerProvider` itself is just a class so you probably have to use `trace.NoOpTracerProvider()` since `URLLib3Instrumentor().instrument(tracer_provider=...)` expects an instance of the `TracerProvider` class.

I guess the test doesn't fail because the eventual call to `tracer_provider.get_tracer(...)` doesn't pass any arguments by name.  When passing the `NoOpTracerProvider` to `instrument` the `tracer_provider.get_tracer(...)` call becomes  `NoOpTracerProvider.get_tracer(...)`, so argument that is passed to the `self` parameter the `NoOpTracerProvider.get_tracer(self, instrumentation_module_name, ...) will actually hold the value of `instrumentation_module_name`. 
This currently seems to work because the `NoOpTracerProvider` isn't doing anything with the arguments passed to `get_tracer` and because the last parameter in the signature is optional.

avzis: oh i see, thanks!

mariojonke (approved):

srikanthccv (approved):

PR #1576, opened 2023-01-12
pymemcache query sanitization

avzis:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-pymemcache/src/opentelemetry/instrumentation/pymemcache/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-pymemcache/src/opentelemetry/instrumentation/pymemcache/__init__.py
@@ -13,10 +13,6 @@
 # limitations under the License.
 
 """
-
-Usage
------
-
 The OpenTelemetry ``pymemcache`` integration traces pymemcache client operations
 
 Usage
@@ -34,6 +30,20 @@ Usage
 
 API
 ---
+The `instrument` method accepts the following keyword args:
+
+tracer_provider (TracerProvider) - an optional tracer provider
+sanitize_query (Boolean) - default True, enable the pymemcache query sanitization
+
+for example:
+
+.. code-block:: python
+
+    from opentelemetry.instrumentation.pymemcache import PymemcacheInstrumentor
+
+    # Instrument pymemcache without query sanitization
+    PymemcacheInstrumentor().instrument(sanitize_query=False)
+
 """
 # pylint: disable=no-value-for-parameter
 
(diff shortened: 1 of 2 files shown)


srikanthccv (commented):
>Until now the default behavior was to present the original query,
and now the default behavior is to present a sanitized query.

Is this generally the thing that users prefer?

instrumentation/opentelemetry-instrumentation-pymemcache/tests/test_pymemcache.py
```diff
@@ -59,7 +59,7 @@ class PymemcacheClientTestCase(
 
     def setUp(self):
         super().setUp()
-        PymemcacheInstrumentor().instrument()
+        PymemcacheInstrumentor().instrument(sanitize_query=False)
```
shalevr, line 62: The default behavior shouldn't be False? to keep the same behavior.
this could cause different results for different versions
And according to the specs the Requirement Level for db.statement is "if not explicitly disabled via instrumentation configuration." I think the sanitizing should be the same

avzis:
> > Until now the default behavior was to present the original query,
> > and now the default behavior is to present a sanitized query.
> 
> Is this generally the thing that users prefer?

actually this is a good question, i opened an issue in the spec repository in order to clarify this:
https://github.com/open-telemetry/opentelemetry-specification/issues/3104

instrumentation/opentelemetry-instrumentation-pymemcache/tests/test_pymemcache.py
srikanthccv: @shalevr can you link the spec where it says this

shalevr: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes

The third note referred to the db.statement, but it is not explicitly written what to do in case of sanitization.
In general, I think that it is better to sanitize the value by default but I'm not sure if we can change the default behavior like this

CHANGELOG.md
```diff
@@ -44,6 +44,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#1592](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1592))
 - `opentelemetry-instrumentation-django` Allow explicit `excluded_urls` configuration through `instrument()`
   ([#1618](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1618))
+- Added query sanitization for pymemcache
+  ([#1576](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1576))
```
srikanthccv, line 48: Move this to Unreleased section

instrumentation/opentelemetry-instrumentation-pymemcache/src/opentelemetry/instrumentation/pymemcache/__init__.py
```diff
@@ -89,27 +99,29 @@ def _set_connection_attributes(span, instance):
...
 @_with_tracer_wrapper
-def _wrap_cmd(tracer, cmd, wrapped, instance, args, kwargs):
+def _wrap_cmd(tracer, cmd, sanitize_query, wrapped, instance, args, kwargs):
     with tracer.start_as_current_span(
         cmd, kind=SpanKind.CLIENT, attributes={}
     ) as span:
         try:
             if span.is_recording():
-                if not args:
+                if not args or sanitize_query:
```
srikanthccv, line 124: I don't think this is the approach we follow elsewhere. The values get replaced with something like `?` or similar.

PR #1581, opened 2023-01-15
Fix sqlalchemy uninstrument

### LINKED ISSUE #1163: Sqlalchemy uninstrument function doesnt work as expected
**Describe your environment** Describe any aspect of your environment relevant to the problem, including your Python version: 3.9

**Steps to reproduce**
 Just create a SQLAlchemy instrumentation and call `uninstrument` function
 Example:
```python
db_url = os.environ['DATABASE_URI']

engine = create_engine(db_url)

instrumentation = SQLAlchemyInstrumentor()
instrumentation.instrument(
    engine=engine
)

def set_interval(func, sec):
    def func_wrapper():
        # set_interval(func, sec)
        func()

    t = threading.Timer(sec, func_wrapper)
    t.start()
    return t


def instrument():
    print("instrument")
    instrumentation.instrument_app(app, engine)


def uninstrument():
    print("uninstrument")
    instrumentation.uninstrument()


set_interval(instrument, 1)
set_interval(uninstrument, 20)

```

Don't collect spans 

Still recording and collecting spans

**Additional context**

shalevr:
Bug fix for sqlalchemy uninstrument, inside the EngineTracer we call listen on events but no one removes the event listeners after the uninstrument

Fixes #1163 

## Type of change

# How Has This Been Tested?

- [x] tox -e test-instrumentation-sqlalchemy14

- [x] No.

# Checklist:


avzis:
Will it work if you create the engine separately?
like this:
```
SQLAlchemyInstrumentor().instrument(enable_commenter=True)
from sqlalchemy import create_engine
engine = create_engine("sqlite:///:memory:")
```

When you instrument like this the engine will not be saved as part of the instrumentation, and the engine wont be un-wrapped as well (I am not sure if this is a problem, just raising a discussion)

There are tests that create the engine separately, maybe you can check the uninstrument() on them

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -111,6 +114,11 @@ def __init__(
         listen(engine, "after_cursor_execute", _after_cur_exec)
         listen(engine, "handle_error", _handle_error)
 
+    def remove_event_listeners(self):
```
devopsmetis, line 117: That's exactly how i solved it on my project

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
```diff
@@ -158,17 +160,22 @@ def _instrument(self, **kwargs):
                 "create_async_engine",
                 _wrap_create_async_engine(tracer_provider, enable_commenter),
             )
+
+        self.engines = []
```
liorzmetis, line 164: Is there any reason to keep list of engines, why is it necessary to keep all the engines in array?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
liorzmetis, line 117: its a private method should start with `_` prefix

instrumentation/opentelemetry-instrumentation-sqlalchemy/tests/test_sqlalchemy.py
```diff
@@ -248,6 +248,7 @@ def test_uninstrument(self):
 
         self.memory_exporter.clear()
         SQLAlchemyInstrumentor().uninstrument()
```
liorzmetis, line 250: Please dont call `uninstrument` on a new SQLAlchemyInstrumentor instance,
suggested code:
```js
def test_uninstrument(self):
    engine = create_engine("sqlite:///:memory:")
    instrumentor = SQLAlchemyInstrumentor()
    instrumentor.instrument(
        engine=engine,
        tracer_provider=self.tracer_provider,
    )
    cnx = engine.connect()
    cnx.execute("SELECT	1 + 1;").fetchall()

    # first span - the connection to the db
    self.assertEqual(spans[0].name, "connect")
    self.assertEqual(spans[0].kind, trace.SpanKind.CLIENT)
    # second span - the query itself
    self.assertEqual(spans[1].name, "SELECT :memory:")
    self.assertEqual(spans[1].kind, trace.SpanKind.CLIENT)

    self.memory_exporter.clear()
    instrumentor.uninstrument()
    cnx.execute("SELECT	2 + 2;").fetchall()
```

liorzmetis (changes requested):

shalevr:
> 

You need to pass engine/engines to SQLAlchemyInstrumentor to create EngineTracer, if not there is no EngineTracer to create so my fix will not change anything

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
shalevr: You can create one engine or engines, so if I keep them in array it will be easier to clean up

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
shalevr: It's not a private method, I need to use this function from sqlalchemy

instrumentation/opentelemetry-instrumentation-sqlalchemy/tests/test_sqlalchemy.py
shalevr: I'm not sure, can you explain what's the difference?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
devopsmetis: No problem, We should consider when we will remove the engine from that list as well, otherwise potentially a memory leak could happen

liorzmetis: No problem, We should consider when we will remove the engine from that list as well, otherwise potentially a memory leak could happen

instrumentation/opentelemetry-instrumentation-sqlalchemy/tests/test_sqlalchemy.py
liorzmetis: We dont want to create additional instrumentor class with new connection and new engine,
we would like to see that we remove listener events for specific instrumentation with specific engine

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
liorzmetis: i wonder if we should implement call uninstrument logic when the garbage collector trying to collect the engine 
https://docs.python.org/3.6/library/weakref.html#finalizer-objects

instrumentation/opentelemetry-instrumentation-sqlalchemy/tests/test_sqlalchemy.py
shalevr: It will not create another sqlalchemy instrumentor, if I understand this correctly it's like Singelton:
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/c92ba14316a90db2a8e8a7702aa21ce114a581a5/opentelemetry-instrumentation/src/opentelemetry/instrumentation/instrumentor.py#L50
What do you think?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
shalevr: I thought about this but the problem is after the instrumentation the EngineTracer returns to the user and we don't know when the garbage collector works

shalevr: Thank you for the comments

avzis:
> > Will it work if you create the engine separately? like this:
> > ```
> > SQLAlchemyInstrumentor().instrument(enable_commenter=True)
> > from sqlalchemy import create_engine
> > engine = create_engine("sqlite:///:memory:")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > When you instrument like this the engine will not be saved as part of the instrumentation, and the engine wont be un-wrapped as well (I am not sure if this is a problem, just raising a discussion)
> > There are tests that create the engine separately, maybe you can check the uninstrument() on them
> 
> You need to pass engine/engines to SQLAlchemyInstrumentor to create EngineTracer, if not there is no EngineTracer to create so my fix will not change anything

Not necessarily, 
when you call `instrument()` you create wrappers for all the functions, including the `create_engine()` function,
So if you run the code like in my example, it will still be wrapped and work as intended - that's the reason that the engine is optional in `instrument()`

shalevr:
> > > Will it work if you create the engine separately? like this:
> > > ```
> > > SQLAlchemyInstrumentor().instrument(enable_commenter=True)
> > > from sqlalchemy import create_engine
> > > engine = create_engine("sqlite:///:memory:")
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > When you instrument like this the engine will not be saved as part of the instrumentation, and the engine wont be un-wrapped as well (I am not sure if this is a problem, just raising a discussion)
> > > There are tests that create the engine separately, maybe you can check the uninstrument() on them
> > 
> > 
> > You need to pass engine/engines to SQLAlchemyInstrumentor to create EngineTracer, if not there is no EngineTracer to create so my fix will not change anything
> 
> Not necessarily, when you call `instrument()` you create wrappers for all the functions, including the `create_engine()` function, So if you run the code like in my example, it will still be wrapped and work as intended - that's the reason that the engine is optional in `instrument()`

You are right the create engine wrapper creates EngineTracer and no one manages this instance, I will fix this

shalevr:
I refactor the code so when EngineTracer registers to the event listener those params will be saved in list of tuples,
so we will know how to remove them from event listeners when uninstument.
and another test function was added for the case that create_engine called after the instrumentation
If the code is not clear enough let me know and I will document this change more

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -95,6 +98,8 @@ def _wrap_connect_internal(func, module, args, kwargs):
 
 
 class EngineTracer:
+    _remove_event_listener_params = []
```
avzis, line 101: suggestion: maybe drop the 'remove'
like, call it `event_listener_params`, or even `event_listeners`

shalevr: but the params are for the remove function
I called the list like this so it will be clear that I keep the params only for remove

devopsmetis: i agree with the `event_listeners` name convention because that what you store in this list

shalevr: Not exactly... I don't store the event listeners params, its without args and kw for exmaple its without(retval=True)
I store the remove event listernes params

devopsmetis:
> I refactor the code so when EngineTracer registers to the event listener those params will be saved in list of tuples, so we will know how to remove them from event listeners when uninstument. and another test function was added for the case that create_engine called after the instrumentation If the code is not clear enough let me know and I will document this change more

@avzis @shalevr 

As a user i didnt its not clear at all how SQLAlchemy Implementation actually works.
why i can pass engine to the instrumentation if in any case *all my engines* will be instrumented ?
image

Its not related directly to your PR but it should be clear if SQLAlchemy Instrumentation will be created then all the created engines will be instrumented even if i specified what Engine to instrument .

shalevr:
> > I refactor the code so when EngineTracer registers to the event listener those params will be saved in list of tuples, so we will know how to remove them from event listeners when uninstument. and another test function was added for the case that create_engine called after the instrumentation If the code is not clear enough let me know and I will document this change more
> 
> @avzis @shalevr
> 
> As a user i didnt its not clear at all how SQLAlchemy Implementation actually works. why i can pass engine to the instrumentation if in any case _all my engines_ will be instrumented ? image
> 
> Its not related directly to your PR but it should be clear if SQLAlchemy Instrumentation will be created then all the created engines will be instrumented even if i specified what Engine to instrument .

I agree with you It wasn't clear to me too, but @avzis is right it's possible to create an engine after sqlalchemy instrumentation, we don't want to miss engines that were created after the instrumentation.

avzis:
> Its not related directly to your PR but it should be clear if SQLAlchemy Instrumentation will be created then all the created engines will be instrumented even if i specified what Engine to instrument .

@devopsmetis I agree that this is a weird behavior, 
one thing I will mention, is that if you import the create_engine function before instrumenting, then it wont happen..

And I also agree that this discussion is not related to this PR, if you think that it is problematic I suggest you open another issue and discuss it

srikanthccv (approved):

devopsmetis (approved):

PR #1599, opened 2023-01-25
botocore: Add enhanced support for SQS

mariojonke:
# Description

Adds enhanced support for SQS by setting relevant span attributes, and injecting/extracting the trace context into/from the MessageAttributes of SQS messages.

This PR adds the same functionality to the `botocore` instrumentation that already exists in the [Node.Js AWS-SDK instrumentation](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-aws-sdk) and in the separate [boto3sqs instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-boto3sqs) package.

The way how SQS processing spans are handled is different though between the `boto3sqs` instrumentation and the approach in this PR.
The `boto3sqs` instrumentation tries to automatically create the processing spans while the `Messages` list (the return value of the `sqs.receive_message` call) is iterated. This is rather error prone since e.g. the list might be iterated multiple times, and the processing spans are tracked by the instrumentation in a more or less global dictionary which might lead to memory leaks.
Since for an instrumentation it is not reasonably possible to determine when to start/end a processing span, the automatic detection "feature" was not considered in this PR.
Instead the instrumentation provides a context manager and a decorator that need to be added manually, which in turn create the processing span for a corresponding SQS message.

## Type of change

# How Has This Been Tested?

Added multiple unit tests to verify span attributes and trace context propagation for SQS and also SNS to SQS scenarios.

- [x] No.

# Checklist:

### DIFF
--- a/docs/instrumentation/botocore/botocore.rst
+++ b/docs/instrumentation/botocore/botocore.rst
@@ -5,3 +5,11 @@ OpenTelemetry Botocore Instrumentation
     :members:
     :undoc-members:
     :show-inheritance:
+
+Submodules
+----------
+
+.. toctree::
+    :maxdepth: 1
+
+    sqs
--- a/docs/instrumentation/botocore/sqs.rst
+++ b/docs/instrumentation/botocore/sqs.rst
@@ -1,0 +1,7 @@
+OpenTelemetry Botocore Instrumentation - SQS
+============================================
+
+.. automodule:: opentelemetry.instrumentation.botocore.extensions.sqs
+    :members:
+    :undoc-members:
+    :show-inheritance:
--- a/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
@@ -47,15 +47,15 @@ API
 
 The `instrument` method accepts the following keyword args:
 
-tracer_provider (TracerProvider) - an optional tracer provider
-request_hook (Callable) - a function with extra user-defined logic to be performed before performing the request
-this function signature is:  def request_hook(span: Span, service_name: str, operation_name: str, api_params: dict) -> None
-response_hook (Callable) - a function with extra user-defined logic to be performed after performing the request
-this function signature is:  def request_hook(span: Span, service_name: str, operation_name: str, result: dict) -> None
+* tracer_provider (TracerProvider) - an optional tracer provider
+* request_hook (Callable) - a function with extra user-defined logic to be performed before performing the request.
+  This function signature is:  def request_hook(span: Span, service_name: str, operation_name: str, api_params: dict) -> None
+* response_hook (Callable) - a function with extra user-defined logic to be performed after performing the request.
+  This function signature is:  def request_hook(span: Span, service_name: str, operation_name: str, result: dict) -> None
 
 for example:
 
-.. code: python
+.. code:: python
 
     from opentelemetry.instrumentation.botocore import BotocoreInstrumentor
     import botocore
@@ -103,7 +103,7 @@ from opentelemetry.instrumentation.utils import (
 )
 from opentelemetry.propagate import inject
 from opentelemetry.semconv.trace import SpanAttributes
-from opentelemetry.trace import get_tracer
+from opentelemetry.trace import Tracer, get_tracer
 from opentelemetry.trace.span import Span
 
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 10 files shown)


mariojonke:
Hi @oxeye-nikolay, i'm aware about the boto3sqs instrumentation which is currently a separate package that exists in parallel to the botocore instrumentation.

The reason for this PR is to integrate the functionality of the boto3sqs instrumentation into the botocore instrumentation. There are mainly two reasons for this:
1. Currently there are two instrumentations that basically instrument the same thing. boto3sqs instruments SQS specific parts and the botocore instrumentation also already includes some limited SQS support by capturing certain attributes. When both instrumentations are active at the same time (e.g. wanting to trace SQS + some other AWS service) this will lead to double tracing for the operations that are instrumented by both instrumentations.
2. Since the botocore instrumentation is the de-facto instrumentation for the Python AWS SDK (boto3/botocore) it makes it easier for users to require just a single package for everything AWS SDK related.

Regarding the boto3sqs issues mentioned in the PR description, those are all about the handling of processing spans and should only point out why the approach taken in this PR is different to the implementation in the boto3sqs instrumentation.

From what i saw while moving the bits over to the botocore instrumentation i think these issues cannot be or only partially be fixed from an instrumentation's point of view.
The start and end of a processing span depends on user code that is completely unknown to the instrumentation. The boto3sqs instrumentation tries to solve this by pre-starting the span for each message in the `sqs.receive_message` function and then activate the corresponding span when the user code iterates over or tries to get a message from the list of messages.
There are some issues with this approach:
* since all processing spans are started in `sqs.receive_message` the timings of the spans do not match with the actual time when processing of a message starts.
* to determine the end of a span the instrumentation relies that user code calls `sqs.delete_message` API. In case of an exception in the user code `delete_message` might never be called, leaving behind an unended span.
* To correlate spans created in `receive_message` and `delete_message` the instrumentation keeps them in a 'global' dictionary identified by the message's `ReceiptHandle`. The `ReceiptHandle` (afaik) will be different on every `receive_message` call even if the ID of a message is the same. So if user code would raise an exception and `receive_message` would receive the same message again it would have a different `ReceiptHandle` leaving an orphaned entry in the 'global' dictionary with the previously created span. (This could probably be solved by a weakref dictionary).
* multiprocessing, e.g. when `delete_message` / processing of a message happens in different process as `receive_message`.
* user code might create a new list / dictionary from the list of received messages and iterate over this copy while processing SQS messages, which would bypass instrumented iterator of the original list.

From this list, I don't think that it is possible to reliably implement an instrumentation that can automatically handle processing spans without the user code explicitly telling the instrumentation where processing of an SQS message starts / ends.

oxeye-nikolay:
Hey @mariojonke, thanks for answering!
I agree with you. It isn't ideal that we have two separate instrumentations. We should unite them, however, as you said, the `boto3sqs` one instruments `boto3`, while the other one instruments `botocore`. This was raised when `boto3sqs` was in development and the maintainers agreed this was the best solution.
Regarding the "never closing process spans" issues you have raised, this is a real problem. But this is also a problem that exists in SQS. In fact, if you don't send a "delete_message" API call, the message will remain in the queue (it has a serving timeout to prevent two handlers from handling it).
There are some other issues with the existing instrumentation that you have rightfully raised (like span timing and changing ReceiptHandle) that should be fixed. Please open an issue for them :)
Regarding your solution - it is quite manual. I'm not sure we can expect the user to manually decorate each function just to continue the context propagation over the processing span. The `boto3sqs` supports auto-instrumentation, which is a requirement for the instrumentation modules (AFAIK).

mariojonke:
Actually `boto3` and `botocore` are two sides of the same coin. `botocore` is the base on which the `boto3` library is built upon (see [boto3 vs botocore](https://www.learnaws.org/2022/08/21/boto3-vs-botocore/), you could potentially even instrument the AWS CLI with the botocore instrumentation since according to this doc the AWS CLI also builds upon botocore).
This means that the botocore instrumentation also instruments boto3. I probably didn't explain it very well but i thought it was clear from the double tracing part (creating two similar spans for the same SQS operation) in my previous comment that the botocore instrumentation also does instrument boto3. 

I'm fully aware that the solution in this PR requires manual handling but imo this is the only option that reliably works. Trying to automatically derive when an SQS message is used by some user code (may it be by tracking the list iterator or some other means) is just an attempt to guess at what is happening. While it works in some cases there are (probably) many of other cases where it doesn't work. Imo, having an automatic instrumentation that works for x% of cases but produces incorrect results remaining % doesn't seem like a good solution, especially if the automatic creation of processing spans cannot be disabled.

Wrt to the mentioned issues in the `boto3sqs` instrumentation, i can create open an issue for them but i doubt that fixing it will solve the underlying problem that the automatic process span detection is just an heuristic that will work only for specific use cases.

PR #1618, opened 2023-02-02
 Add ability to exclude some routes explicitly with Django Instrumentation

lzchen:
We have exclude_lists via `OTEL_PYTHON_DJANGO_EXCLUDED_URLS` environment variable but not explicitly passed in through `instrument()`.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/__init__.py
@@ -255,9 +255,11 @@ from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
 from opentelemetry.metrics import get_meter
 from opentelemetry.semconv.metrics import MetricInstruments
 from opentelemetry.trace import get_tracer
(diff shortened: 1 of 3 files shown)


shalevr (approved):

shalevr:
Is there a test that moves excluded_urls directly to the instrumentation?

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/__init__.py
```diff
@@ -302,6 +305,11 @@ def _instrument(self, **kwargs):
         meter = get_meter(__name__, __version__, meter_provider=meter_provider)
         _DjangoMiddleware._tracer = tracer
         _DjangoMiddleware._meter = meter
+        _DjangoMiddleware._excluded_urls = (
+            _excluded_urls_from_env
+            if _excluded_urls is None
+            else parse_excluded_urls(_excluded_urls)
+        )
```
srikanthccv, line 312: Maybe add a new test that uses `instrument(excluded_urls=...)` if it doesn't exist already.

srikanthccv (approved):

Issue #1653, opened 2023-02-09
opentelemetry-instrumentation-flask error during _teardown_request if request spawns child threads and copy_current_request_context is used


hangonlyra:
**Describe your environment** 
Python 3.9

opentelemetry-instrumentation-flask==0.36b0
Flask==2.2.2 (but happens on older versions as well)

**Steps to reproduce**
```python3
from concurrent.futures import ThreadPoolExecutor, as_completed
from random import randint

from flask import Flask, copy_current_request_context

from opentelemetry.instrumentation.flask import FlaskInstrumentor

app = Flask(__name__)

def do_random_stuff():
    @copy_current_request_context
    def inner():
        return randint(0, 100)
    return inner

@app.route("/")
def hello_world():
    count = 4
    executor = ThreadPoolExecutor(count)
    futures = []
    for _ in range(count):
        futures.append(executor.submit(do_random_stuff()))

    for future in as_completed(futures):
        print(future.result())

    return "Hello, World!"

FlaskInstrumentor.instrument_app(app)
```

Run it with
```opentelemetry-instrument  --traces_exporter console --metrics_exporter console  flask run```

**What is the expected behavior?**
Normal Open Telemetry traces being outputted when accessed via HTTP

**What is the actual behavior?**
```
[2023-02-09 10:11:16,480] ERROR in app: Exception on / [GET]
Traceback (most recent call last):
  File "~/otelbug/env/lib/python3.9/site-packages/flask/app.py", line 2525, in wsgi_app
    response = self.full_dispatch_request()
  File "~/otelbug/env/lib/python3.9/site-packages/flask/app.py", line 1822, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "~/otelbug/env/lib/python3.9/site-packages/flask/app.py", line 1820, in full_dispatch_request
    rv = self.dispatch_request()
  File "~/otelbug/env/lib/python3.9/site-packages/flask/app.py", line 1796, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "~/otelbug/app.py", line 25, in hello_world
    print(future.result())
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/concurrent/futures/_base.py", line 391, in __get_result
[...]

Issue #1551, opened 2022-12-27
`flask.copy_current_request_context`+`gevent`+`FlaskInstrumentor` cause improper span lifecycles


matthewgrossman:
### high-level problem
To create a new request-aware greenlet within a flask request, we copy the existing flask-req-ctx and push a new req-ctx onto the req-ctx-stack. When this greenlet+req-ctx ends, it tries to call the patched [`_teardown_request`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py#L429) within `flask-instrumentor`, which throws errors because the greenlet doesn't "own" the span, and therefore shouldn't be allowed to end it.

**Describe your environment** 

I created a test repo with all of the source code for reproduction: https://github.com/matthewgrossman/otel-flask-repro. The important file is the [app.py](https://github.com/matthewgrossman/otel-flask-repro/blob/main/app.py).

From trying a few different versions, it doesn't seem that the pinned deps matter too much, but here's a `requirements.txt` on macOS 13.0.1:
```
gevent==22.10.2
flask==1.1.4
opentelemetry-instrumentation-flask
opentelemetry-sdk
opentelemetry-api
```
If you'd prefer, I could turn this env into a docker image to make reproing easier for others. They key deps are `flask`, `gevent`, and `opentelemetry-instrumentation-flask`, and I've noticed this problem  in every variation of those deps that I've tried.

**Steps to reproduce**
```bash
$ python3.8 -m venv venv
$ source venv/bin/activate
$ pip install -r requirements.txt
$ python -m flask run -p 8080
```
in other shell:
```bash
curl localhost:8080/test -H 'x-test-header: val`
```
In the running shell with the flask service, you should see an error like:
```
(venv)  ~/src/otel-flask-repro  » python -m flask run -p 8080
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
 * Running on http://127.0.0.1:8080/ (Press CTRL+C to quit)
Failed to detach context
Traceback (most recent call last):
  File "/Users/mgrossman/src/otel-flask-repro/venv/lib/python3.8/site-packages/opentelemetry/context/__init__.py", line 157, in detach
    _RUNTIME_CONTEXT.detach(token)  # type: ignore
[...]

PR #1654, opened 2023-02-09
Fix issue with Flask instrumentation when a request spawn children threads and copies the request context

hangonlyra:
# Description
In `opentelemetry.instrumentation.flask` record the thread ID in `_before_request` in addition to the span. Then in `_teardown_request` check that it's the same thread for that span before calling `activation.__exit__`

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1653
Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1551

## Type of change

# How Has This Been Tested?

The bug is reproducible with this simple Flask app:
```python3
from concurrent.futures import ThreadPoolExecutor, as_completed
from random import randint

from flask import Flask, copy_current_request_context

app = Flask(__name__)

def do_random_stuff():
    @copy_current_request_context
    def inner():
        return randint(0, 100)
    return inner

@app.route("/")
def hello_world():
    count = 4
    executor = ThreadPoolExecutor(count)
    futures = []
    for _ in range(count):
        futures.append(executor.submit(do_random_stuff()))

    for future in as_completed(futures):
        print(future.result())

    return "Hello, World!"

FlaskInstrumentor.instrument_app(app)
```
When run with `opentelemetry-instrument  --traces_exporter console --metrics_exporter console  flask run`  and accessed using HTTP it results in
```File "~/otelbug/env/lib/python3.9/site-packages/flask/ctx.py", line 399, in pop
    self.app.do_teardown_request(exc)
  File "~/otelbug/env/lib/python3.9/site-packages/flask/app.py", line 2373, in do_teardown_request
    self.ensure_sync(func)(exc)
  File "~/otelbug/env/lib/python3.9/site-packages/opentelemetry/instrumentation/flask/__init__.py", line 441, in _teardown_request
    activation.__exit__(None, None, None)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/contextlib.py", line 126, in __exit__
    next(self.gen)
ValueError: generator already executing
```

After the change, the issue no longer appears and the trace appears in the console.

- [x] No.

# Checklist:


shalevr (approved):

shalevr:
Please add a changelog entry

hangonlyra:
> Please add a changelog entry

Updated. Thanks for the help.

CHANGELOG.md
```diff
@@ -26,7 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#1618](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1618))
 
 ### Fixed
-
+- Fix Flask instrumentation to only close the span if it was created by the same thread. `ValueError: generator already executing` ([#1653](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1653))
```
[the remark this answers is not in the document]
hangonlyra: Fixed. Thanks!

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -397,7 +397,10 @@ def _before_request():
 
         activation = trace.use_span(span, end_on_exit=True)
         activation.__enter__()  # pylint: disable=E1101
-        flask_request_environ[_ENVIRON_ACTIVATION_KEY] = activation
+        flask_request_environ[_ENVIRON_ACTIVATION_KEY] = (
+            get_ident(),
```
shalevr, line 401: Consider adding a new environ key for the thread_id, It looks like the thread_id is not part of the activation

shalevr:
I am not familiar with flask requests
Is there a legitimate way that the request starts with some thread and ends up in another thread?

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
hangonlyra: Oh. That's a really good point. Let me update the PR.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -436,8 +438,9 @@ def _teardown_request(exc):
         if excluded_urls and excluded_urls.url_disabled(flask.request.url):
             return
 
-        activation = flask.request.environ.get(_ENVIRON_ACTIVATION_KEY)
-        if not activation:
+        thread_id, activation = flask.request.environ.get(_ENVIRON_ACTIVATION_KEY)
```
shalevr, line 441: The thread_id here has not been updated

hangonlyra:
> I am not familiar with flask requests, Is there a legitimate way that the request starts with some thread and ends up in another thread?

I can't think of any legitimate reason for that because it won't serve any useful purpose. It's definitely possible but it won't make any sense.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
hangonlyra: Thanks for the catch.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -437,7 +439,8 @@ def _teardown_request(exc):
             return
 
         activation = flask.request.environ.get(_ENVIRON_ACTIVATION_KEY)
-        if not activation:
+        thread_id = flask.request.environ.get(_ENVIRON_THREAD_ID_KEY)
+        if not activation or thread_id != get_ident():
             # This request didn't start a span, maybe because it was created in
             # a way that doesn't run `before_request`, like when it is created
             # with `app.test_request_context`.
```
matthewgrossman, line 446: can we update this comment? We now have another condition that causes us to not call `__exit__`

matthewgrossman: I think this could be the only/sole place that we have an in-code comment about what we're doing with all these threading shenanigans.

And I do think recording + comparing against the `get_ident()` is weird enough that it's worth documenting in code a bit via a comment somewhere

hangonlyra: Good call. Updated the comment.

matthewgrossman: 🎉 comment looks great, thanks

CHANGELOG.md
```diff
@@ -28,7 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#1618](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1618))
 
 ### Fixed
-
+- Fix Flask instrumentation to only close the span if it was created by the same thread. `ValueError: generator already executing` ([#1654](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1654))
```
srikanthccv, line 31:
```suggestion

- Fix Flask instrumentation to only close the span if it was created by the same thread. `ValueError: generator already executing`
  ([#1654](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1654))
```

hangonlyra: Thanks for the suggestion! Added to the commit.

aabmass (commented):
Thanks for the informative PR description! I'm not super familiar with this but the solution makes sense to me after reading through.

How difficult would it be to add a unit test for this?

hangonlyra:
> 

Good idea. Added a new test case.

hangonlyra:
@aabmass Good idea. Added new test case `TestMultiThreading.test_multithreaded`

aabmass (approved):
Looks great! Thanks again

CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix Flask instrumentation to only close the span if it was created by the same thread.
+  ([#1654](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1654))
```
srikanthccv, line 36: This should be moved to the new ### Fixed section under ## Unreleased.

srikanthccv (approved):
Thanks

Issue #1506, opened 2022-12-19
ConfluentKafkaInstrumentor's ProxiedConsumer does not support commit


jeehun-pas:
**Describe your environment**
Python 3.8
confluent-kafka[avro]==1.9.2
opentelemetry-instrumentation-confluent-kafka==0.35b0

**Steps to reproduce**
Call ProxiedConsumer.commit

**What is the expected behavior?**
Message is committed in Kafka cluster

**What is the actual behavior?**
Get an exception with message "Consumer closed"

**Additional context**
ProxiedConsumer is not an actual Consumer but using a private internal Consumer instance to do poll, consume, etc. It is missing an over-ridden version of ProxiedConsumer.commit which calls the private Consumer's commit.

alolita:
@srikanthccv @lzchen can you please assign @mrajashree to this issue :-) She's working on a PR for this. Thanks!

srikanthccv:
Can you ask them to leave a comment? Username doesn't show up in the assignees dropdown otherwise.

mrajashree:
Thanks @alolita @srikanthccv for assigning me this issue (hope my username will now show up :-) )

PR #1656, opened 2023-02-10
Add commit method for ConfluentKafkaInstrumentor's ProxiedConsumer

mrajashree:
# Description

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1506

The package confluent-kafka has a class Consumer, which has a method called commit (Ref: https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Consumer.commit)

The package opentelemetry-instrumentation-confluent-kafka offers a wrapper around confluent-kafka's Consumer, called ProxiedConsumer. However ProxiedConsumer does not support `commit` method. This leads to an exception with the message "Consumer closed" when ProxiedConsumer.commit is called.

This PR fixes this by adding a method `commit` on ProxiedConsumer, which accepts the same args as confluent-kafka's Consumer's commit method and makes a call to that method.

## Type of change

# How Has This Been Tested?

Previously calling `commit` on the ProxiedConsumer would lead to an exception. With the changes in this PR, calling `commit` ensures the message is committed since it calls the underlying Consumer's commit method.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/__init__.py
@@ -173,6 +173,9 @@ class ProxiedConsumer(Consumer):
     def committed(self, partitions, timeout=-1):
         return self._consumer.committed(partitions, timeout)
 
+    def commit(self, *args, **kwargs):
+        return self._consumer.commit(*args, **kwargs)
+
     def consume(
         self, num_messages=1, *args, **kwargs
     ):  # pylint: disable=keyword-arg-before-vararg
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/__init__.py
```diff
@@ -173,6 +173,9 @@ def __init__(self, consumer: Consumer, tracer: Tracer):
     def committed(self, partitions, timeout=-1):
         return self._consumer.committed(partitions, timeout)
 
+    def commit(self, message, asynchronous):
+        return self._consumer.commit(message=message, asynchronous=asynchronous)
```
shalevr, line 177: According to
https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Consumer.commit
There is offsets param in the commit function, Is there a reason not to pass it?

mrajashree: no reason other than I missed it! thanks for catching it, will update the PR

shalevr: Maybe it will be better to work with args and kwargs
so we won't assume the default values, and if the signature will change we still pass all the arguments

srikanthccv (approved):

PR #1660, opened 2023-02-10
Revert #1097

### LINKED ISSUE #1658: Revert fix for #1077
#1097 fixed #1077 but introduced another issue, reverting to find a fix that does not cause this problem.

ocelotl:
Several issues have arisen from this bugfix, reverting here until a better solution can be found.

Fixes #1658


ocelotl:
@phillipuniverse @rogersd @philipcwhite please take a look and let us know if this allows you to work around your issues.

phillipuniverse:
@ocelotl I have confirmed this fixes my originally reported issue with the coverage module!

PR #1673, opened 2023-02-15
Bring SQS Propagation in line with the spec

tsloughter:
Opening as a draft because 1 test still fails. I'm hoping someone more familiar with the code might notice what I did wrong easily :)

The failure is in `TestBoto3SQSInstrumentation.test_receive_message`, no links are found:

```
>           self.assertEqual(1, len(span.links))
E           AssertionError: 1 != 0
```

I did some simple `printf` debugging to check what the `MessageSystemAttributes` were before calling `extract` and the `ctx` after calling `extract`:

```
        message_system_attributes = message.get("MessageSystemAttributes", {})
        print(message_system_attributes)
        links = []
        ctx = AwsXRayPropagator().extract(
            message_system_attributes, getter=boto3sqs_getter
        )
        print(ctx)
```

And the header is there as expected and it is extracted properly but the ctx is then empty:

`message_system_attributes`:

```
{'AWSTraceHeader': {'StringValue': 'root=1-00000000-00000000000000000000000a;parent=0000000000000001;sampled=1', 'DataType': 'String'}}
```

Value in `Boto3SQSGetter.get`:

```
root=1-00000000-00000000000000000000000a;parent=0000000000000001;sampled=1
```

`ctx`:

```
{}
```

# Description

The instrumentation spec for SQS defines that propagation is done with X-Ray to the key `AWSTraceHeader` in `MessageSystemAttributes`. This patch brings the propagation in line with the spec from what it was doing before of using the global propagator and `MessageAttributes`. 

## Type of change

# How Has This Been Tested?

- [X] Existing tests were updated to extract/inject the propagated context with the new format. 

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-boto3sqs/pyproject.toml
+++ b/instrumentation/opentelemetry-instrumentation-boto3sqs/pyproject.toml
@@ -27,6 +27,7 @@ classifiers = [
 dependencies = [
   "opentelemetry-api ~= 1.12",
   "opentelemetry-instrumentation == 0.37b0.dev",
+  "opentelemetry-propagator-aws-xray == 1.0.1",
   "opentelemetry-semantic-conventions == 0.37b0.dev",
   "wrapt >= 1.0.0, < 2.0.0",
 ]
--- a/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
@@ -35,11 +35,15 @@ import boto3
 import botocore.client
 from wrapt import wrap_function_wrapper
 
-from opentelemetry import context, propagate, trace
+from opentelemetry import context, trace
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
 from opentelemetry.instrumentation.utils import (
     _SUPPRESS_INSTRUMENTATION_KEY,
     unwrap,
+)
+from opentelemetry.propagators.aws.aws_xray_propagator import (
+    TRACE_HEADER_KEY,
+    AwsXRayPropagator,
 )
 from opentelemetry.propagators.textmap import CarrierT, Getter, Setter
 from opentelemetry.semconv.trace import (
@@ -197,7 +201,11 @@ class Boto3SQSInstrumentor(BaseInstrumentor):
     ) -> None:
         message_attributes = message.get("MessageAttributes", {})
         links = []
-        ctx = propagate.extract(message_attributes, getter=boto3sqs_getter)
+        print(message_attributes)
+        ctx = AwsXRayPropagator().extract(
+            message_attributes, getter=boto3sqs_getter
+        )
+        print(ctx)
         parent_span_ctx = trace.get_current_span(ctx).get_span_context()
         if parent_span_ctx.is_valid:
             links.append(Link(context=parent_span_ctx))
@@ -233,7 +241,7 @@ class Boto3SQSInstrumentor(BaseInstrumentor):
             ) as span:
                 Boto3SQSInstrumentor._enrich_span(span, queue_name, queue_url)
                 attributes = kwargs.pop("MessageAttributes", {})
-                propagate.inject(attributes, setter=boto3sqs_setter)
+                AwsXRayPropagator().inject(attributes, setter=boto3sqs_setter)
                 retval = wrapped(*args, MessageAttributes=attributes, **kwargs)
                 message_id = retval.get("MessageId")
                 if message_id:
@@ -273,7 +281,7 @@ class Boto3SQSInstrumentor(BaseInstrumentor):
                 with trace.use_span(span):
                     if "MessageAttributes" not in entry:
                         entry["MessageAttributes"] = {}
-                    propagate.inject(
+                    AwsXRayPropagator().inject(
                         entry["MessageAttributes"], setter=boto3sqs_setter
                     )
             retval = wrapped(*args, **kwargs)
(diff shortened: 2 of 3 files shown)


tylerbenson:
@tsloughter does AWS support x-ray propagation on the message attribute level? I think in the Java instrumentation, the propagation is done via http headers on the outgoing request rather than the request body. I'm not sure if there's a difference in behavior.

mariojonke:
I think context propagation with `MessageSystemAttributes` will only work in a very limited way. If you search for `MessageSystemAttributes` in the [AWS SDK docs for SQS](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html), they are only defined for `send_message` and `send_message_batch`. For [receive_message](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Client.receive_message) they are not returned by AWS, meaning that the propagated context cannot be extracted anymore when the SDK is used at the receiver side.
Context propagation via the XRay propagator only seems to work if the SQS receiver is a lambda function where AWS internally converts the `MessageSystemAttributes` to an XRay environment variable that the lambda function then can access.
For other scenarios like SQS.send_message -> SQS.receive_message, SNS.publish -> (SNS subscription to SQS) -> SQS.receive (either via SDK or lambda), ... the `MessageSystemAttributes` will become unaccessible and break context propagation.
I guess this is the reason why context propagation in this repo and in other techs like [js](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/MessageAttributes.ts) or [ruby](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/message_attributes.rb) propagate context via `MessageAttributes` instead of `MessageSystemAttributes`.

tsloughter:
Ah. I thought I saw them in the response body but I guess not.

My reading of the SQS spec was that they shouldn't be on regular attributes because of user limits there and instead to use system attributes, but if those aren't propagated to the user that doesn't help :)

tylerbenson:
It's being done for all requests, not just SQS:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java#L138

tsloughter:
Ah. It does look like Python is doing the same, https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py#L116

So I could instead update that too always use XRay and remove the existing propagation in attributes..?

tsloughter:
Or wait, wouldn't putting it in headers break sending batches? Batches will have messages with separate trace contexts.

tylerbenson:
Good question.  I don't know the answer to that. In theory we should be doing something similar to what AWS is doing for their sdks: https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-python-patching.html

tsloughter:
@mariojonke I noticed `AWSTraceHeader` is in the `AttributeNames` for `receive_message` https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Queue.receive_messages

Might that be the value from `MessageSystemAttributes`?

mariojonke:
It looks like it is the same thing but i haven't tested it yet. The instrumentation probably would need inject into the `AttributeNames` to tell AWS that it wants to receive the `AWSTraceHeader` and then extract it from the `Attributes` instead of the `MessageAttributes`.
However, as far as i can tell this only works for SQS to SQS scenarios and i'm not sure if it would require to explicitly enable XRay in AWS. When looking at the [SNS docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#SNS.Client.publish) there doesn't seem to be any mention of `AWSTraceHeader` or `MessageSystemAttributes`, so tracing SNS to SQS messages probably isn't working unless the XRay header is somehow magically injected (maybe into the [underlying HTTP request](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/ext/botocore/patch.py#L26)). 
I guess this would need to be tested and also if it works with enabeling/disabeling [raw message delivery](https://docs.aws.amazon.com/sns/latest/dg/sns-large-payload-raw-message-delivery.html) on the SNS to SQS subscription.

Another thing to consider i guess is that the [XRay trace header](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader) only contains trace ID, span ID and sampling flag. This is more or less in line with the W3C traceparent but it doesn't seem to provide a tracestate or similar. So any vendor specific information will be lost when tracing an SQS message and using the XRay propagator.

Ma-Ba:
The proposed change will introduce an incompatibility with SNS and SQS tracing scenarios currently in use, that rely on the global propagator being called to inject/extract the context into message attributes.

See https://github.com/open-telemetry/opentelemetry-specification/pull/3212#pullrequestreview-1301429705 for a more detailed explanation.

I therefore suggest to keep the call to the global propagator's inject and extract methods in place and add the X-Ray propagator in a backwards compatible way.

tsloughter:
Experimenting with the x-ray sdk and rereading the existing spec I'll be updating this PR to no longer inject the `AWSTraceHeader` attribute. This is added by AWS when the HTTP x-ray header is found in the send request. Extraction will look for AWSTraceHeader in the `attributes` (not the message attributes).

I still need to get this working with otel python boto and boto3sqs to make sure it works as I expect based on use of x-ray python sdk. I'll be asking about that on slack tomorrow :)

I'm not sure what to do about backwards compatibility.

tsloughter:
I'm going to close this and open a new PR only focused on `botocore`.

I still think changes are needed to SQS instrumentation, but not the ones in this PR. I'll likely make one in the near future that adds support for checking for `AWSTraceHeader` in the System Attributes. Injection should not be done how it is in this PR but instead left to the PR I'm making on `botocore` where the context is injected into the HTTP headers with x-ray.

Oberon00:
@tsloughter 

> I'll likely make one in the near future that adds support for checking for AWSTraceHeader

This should not actually be needed: If you set the X-Ray HTTP header when sending an SQS message, AWS will automatically fill in the AWSTraceHeader message system attribute based on that.

tsloughter:
@Oberon00 right, it fills it in, but it needs to be checked for on the receiving side.

Issue #868, opened 2022-01-19
Calling `flask.copy_current_request_context` results in error logged from OTEL context.detach/Flask instrumentation


joshuahlang:
```
Flask: 2.0.2
Python: 3.9.8 (main, Nov 10 2021, 09:21:22)  [Clang 13.0.0 (clang-1300.0.29.3)]
OTEL Flask instrumentation: 0.27b0
```

**Steps to reproduce**
Install the latest requirements in a `venv`:
```
python3 -m venv otel-venv && otel-venv/bin/pip install gevent flask opentelemetry.instrumentation.flask opentelemetry.sdk
```

Use the following code to create a running Flask application. `curl http://localhost:5000/` and not the logged error of `Failed to detach context` [from here](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/context/__init__.py#L159). 
```
import sys

import gevent
from gevent.pywsgi import WSGIServer
import flask
from opentelemetry import trace
import opentelemetry.instrumentation.flask
from opentelemetry.sdk.trace.export import ConsoleSpanExporter, SimpleSpanProcessor
from opentelemetry.sdk.trace import TracerProvider


app = flask.Flask(__name__)

tracer_provider = TracerProvider(
    active_span_processor=SimpleSpanProcessor(ConsoleSpanExporter())
)
trace.set_tracer_provider(tracer_provider)

opentelemetry.instrumentation.flask.FlaskInstrumentor().instrument_app(app)


@app.route('/')
def index():
    @flask.copy_current_request_context
    def do_some_work():
        with trace.get_tracer(__name__).start_span('async-work'):
            print('did some async work')

    with trace.get_tracer(__name__).start_span('sync-work'):
        gevent.spawn(do_some_work)
    return '\n'.join(map(
        lambda args: '{}: {}'.format(*args),
        [
            ('Flask', flask.__version__),
            ('Python', sys.version.replace('\n', ' ')),
            ('OTEL Flask instrumentation', opentelemetry.instrumentation.flask.version.__version__),
        ]
    )) + '\n'


http_server = WSGIServer(('', 5000), app)
http_server.serve_forever()
```

**What is the expected behavior?**
No `Failed to detach context` error message logged. I'd expect the async work span(s) to be also have the `parent_id` set to a span in the originating HTTP request thread of execution (though perhaps my expectation is off on this point).

**What is the actual behavior?**
[...]

PR #1692, opened 2023-02-24
Use `request_ctx` to determine whether or not `_teardown_request` should end flask span

matthewgrossman:
# Description

This is effectively a continuation of PR https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1654.

In that PR, we added additional context to the flask-instrumentor that ensured we wouldn't try to close spans if we weren't in the same thread that created the span. These problems are derived from `flask.copy_current_request_context`-decorated functions, which are often used when running a func in a background thread/greenlet. Because a backgrounded-function is wrapped in a copied ctx, it calls `_teardown_request` upon return, which which will fail/error when attempting to end the span.

However, problems arise even in situations where we aren't running in a separate thread/greenlet. Here's an example:

```python
import logging
import flask

app = flask.Flask(__name__)
FlaskInstrumentor.instrument_app(app=app)

@app.get("/hello")
def hello_route() -> flask.Response:

    @flask.copy_current_request_context
    def _log_header():
        print(f"inside call_on_close, {flask.request.headers.get('x-test', 'noheader')}")

    resp = flask.make_response("hello", 200)
    resp.call_on_close(_log_header)
    return resp
```
In this instance, using the `thread_id` is insufficient to differentiate in `_teardown_request` who should try to end the span; they're both running in the same thread.

Instead, we should simply keep a (weak) reference to the `request_ctx` that we're confident _did_ create the span. When a `copy_current_request_context`-func is called, it'll push on a new `request_ctx` onto the stack which won't match the original . During `_teardown_request`, we can just compare the current `request_ctx` against the original one, and escape if they don't match.

This solution should work generically no-matter how the request is run (thread/greenlet/call_on_close/synchronously). The existing multithreaded tests created in the last PR still pass (and are useful to keep around IMO)

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/868

## Type of change

# How Has This Been Tested?

- [x] Unit test added as part of this PR, run via `tox -e 'py3{7,8,9,10,11}-test-instrumentation-flask{213,220}'`
- [x] Added new test parametrization in `tox.ini` for flask==2.1.3 and flask==2.2.0. This will exercise both conditionals of how we grab the `request_ctx`
- [x] The failing example in this PR I ran locally both on `main` and my test branch, and confirmed the issue was resolved.

<details>
  <summary>Full setup for that</summary>
  
 `app.py`:
```python
import logging
import flask

app = flask.Flask(__name__)
FlaskInstrumentor.instrument_app(app=app)

@app.get("/hello")
def hello_route() -> flask.Response:

    @flask.copy_current_request_context
    def _log_header():
        print(f"inside call_on_close, {flask.request.headers.get('x-test', 'noheader')}")

    resp = flask.make_response("hello", 200)
    resp.call_on_close(_log_header)
    return resp
```
`requirements.txt`:
```
-e ${HOME}/src/opentelemetry-python/opentelemetry-api
-e ${HOME}/src/opentelemetry-python/opentelemetry-sdk
-e ${HOME}/src/opentelemetry-python/opentelemetry-semantic-conventions
-e ${HOME}/src/opentelemetry-python-contrib/opentelemetry-instrumentation/
-e ${HOME}/src/opentelemetry-python-contrib/util/opentelemetry-util-http/
-e ${HOME}/src/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-wsgi/
-e ${HOME}/src/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-flask/
Flask
```

```bash
python3 -m venv venv
. venv/bin/activate
pip install -r requirements.txt
flask run
```

```bash
curl 0:5000/hello -H 'x-test: QQ'
```
</details>

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

- [x] ~Documentation has been updated~

# Implementation discussion

#### Why are you using a `weakref`?
Originally I was considering just storing the `id(request_ctx)` in the `environ`, but there's issues around potential `id` reuse [("Two objects with non-overlapping lifetimes may have the same `id()` value.")](https://docs.python.org/3/library/functions.html#id). I highly doubt there'd be strange overlapping lifetime issues that cause this to be a problem, but it just seemed better to grab a true reference to what we were interested in. This seems less brittle than effectively storing an `int` pointer to the underlying data.

The `weakref` guarantees we don't accidentally hold onto the memory simply due to this reference, as well.

#### What's with the `packaging` version checks on `Flask`?
For awhile, the blessed way of grabbing the request_ctx was `flask._request_ctx_stack.top`. This was however deprecated in version `2.2.0` ([link](https://flask.palletsprojects.com/en/2.2.x/changes/?highlight=_request_ctx_stack#version-2-2-0)). Now that we have slightly different functionality across flask versions, I also added a `flask{213,220)` section in the `tox.ini`, to ensure we test both versions

#### Why are we storing this data on the `environ` in the first place? Isn't that what `request` is for?
I'm a bit confused on this. If this file  persisted metadata using `flask.request._otel_flask_attrib`, then any new contexts spawned via `copy_current_request_context` wouldn't have inherited that data, which means teardown would do the "right thing" automatically simply by not having access to the `activation` key; itt would properly act like a stack. This _could_ be a refactor we do, but it wasn't something I wanted to do in this PR (nor am I sure if the maintainers are amenable to it).


instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -265,11 +266,17 @@ def response_hook(span: Span, status: str, response_headers: List):
...
 _ENVIRON_ACTIVATION_KEY = "opentelemetry-flask.activation_key"
-_ENVIRON_THREAD_ID_KEY = "opentelemetry-flask.thread_id_key"
+_ENVIRON_REQCTX_REF_KEY = "opentelemetry-flask.reqctx_ref_key"
 _ENVIRON_TOKEN = "opentelemetry-flask.token"
 
 _excluded_urls_from_env = get_excluded_urls("FLASK")
 
+if package_version.parse(flask.__version__) >= package_version.parse("2.2.0"):
+    def _request_ctx_ref() -> weakref.ReferenceType:
+        return weakref.ref(flask.globals.request_ctx._get_current_object())
```
matthewgrossman, line 276: Used these docs to inform `_get_current_object`: https://flask.palletsprojects.com/en/2.2.x/reqcontext/#notes-on-proxies

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -265,11 +266,17 @@ def response_hook(span: Span, status: str, response_headers: List):
 _ENVIRON_STARTTIME_KEY = "opentelemetry-flask.starttime_key"
 _ENVIRON_SPAN_KEY = "opentelemetry-flask.span_key"
 _ENVIRON_ACTIVATION_KEY = "opentelemetry-flask.activation_key"
-_ENVIRON_THREAD_ID_KEY = "opentelemetry-flask.thread_id_key"
+_ENVIRON_REQCTX_REF_KEY = "opentelemetry-flask.reqctx_ref_key"
 _ENVIRON_TOKEN = "opentelemetry-flask.token"
 
 _excluded_urls_from_env = get_excluded_urls("FLASK")
 
+if package_version.parse(flask.__version__) >= package_version.parse("2.2.0"):
```
matthewgrossman, line 274: In newer versions, we're supposed to access `flask.globals.request_ctx`: https://flask.palletsprojects.com/en/2.2.x/changes/?highlight=_request_ctx_stack#version-2-2-0

instrumentation/opentelemetry-instrumentation-flask/pyproject.toml
```diff
@@ -38,7 +39,8 @@ instruments = [
 ]
 test = [
   "opentelemetry-instrumentation-flask[instruments]",
-  "markupsafe==2.0.1",
+  "flask >= 2.2.0",
```
matthewgrossman, line 42: what would be a good way to test both pre-2.2.0 and post-2.2.0 flask in this repo?

dev-requirements.txt
```diff
@@ -14,7 +14,7 @@ bleach==4.1.0 # transient dependency for readme-renderer
 grpcio-tools==1.29.0
 mypy-protobuf>=1.23
 protobuf~=3.13
-markupsafe==2.0.1
+markupsafe>=2.0.1
```
matthewgrossman, line 17: newer flask2.2.0 requires newer jinja2, which requires newer markupsafe

matthewgrossman:
fyi @hangonlyra; modification/continuations of your PR

instrumentation/opentelemetry-instrumentation-flask/pyproject.toml
matthewgrossman: figured out tox.ini changes

hangonlyra:
> fyi @hangonlyra; modification/continuation of your PR

This is really great! Thanks for improving it.

shalevr:
Please fix the lint errors, run "tox -e lint"

matthewgrossman:
Thanks for the help; I got preoccupied with a bunch of things in work/life, but should come back to this PR and fix it up within the next few weeks. Will ping once its in a better state!

matthewgrossman:
hey @shalevr; got this working locally and now would appreciate some code-review/feedback. Thanks!

ocelotl (approved):

PR #1701, opened 2023-03-01
Add sqlalchemy db.statement sanitization flag

### LINKED ISSUE #1549: Implement sensitive data sanitization for sqlalchemy instrumentation
Aggregated by #1543

nemoshlag:
# Description

Added an optional query sanitizer to the SQLAlchemy instrumentation.
Usage
SQLAlchemyInstrumentor().instrument(sanitize_query=True)

This will affect the DB_STATEMENT value to contain the original query or sanitized one.

Fixes #1549    
Following the specification discussion [here](https://github.com/open-telemetry/opentelemetry-specification/issues/3104)

## Type of change

# How Has This Been Tested?

- [x] Test A db.statement has been sanitized
- [x] Test B no side affects occurred on a query without a db.statement attribute 

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
+++ b/instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
@@ -27,6 +27,16 @@ from opentelemetry.semconv.trace import NetTransportValues, SpanAttributes
 from opentelemetry.trace.status import Status, StatusCode
 
 
+def _sanitize_query(query):
+    """Remove query content, replace with sanitization symbol.
+    For example `SELECT * FROM table` will sanitize to SELECT ? ?`
+    """
+    sanitize_symbol = " ?"
+    if query and query.split():
+        return query.split()[0] + sanitize_symbol + sanitize_symbol
+    return ""
+
+
 def _normalize_vendor(vendor):
     """Return a canonical name for a type of database."""
     if not vendor:
@@ -42,7 +52,7 @@ def _normalize_vendor(vendor):
 
 
 def _wrap_create_async_engine(
-    tracer, connections_usage, enable_commenter=False
+    tracer, connections_usage, sanitize_query, enable_commenter=False
 ):
     # pylint: disable=unused-argument
     def _wrap_create_async_engine_internal(func, module, args, kwargs):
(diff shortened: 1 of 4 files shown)


shalevr (approved):
Looking great! Thank you for adding this

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `opentelemetry-instrumentation-sqlalchemy` Add optional db.statement query sanitization.
+  ([#1662](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1701))
```
srikanthccv, line 22:
```suggestion
  ([#1701](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1701))
```

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -27,6 +27,16 @@
 from opentelemetry.trace.status import Status, StatusCode
 
 
+def _sanitize_query(query):
+    """Remove query content, replace with sanitization symbol.
+    For example `SELECT * FROM table` will sanitize to SELECT ? ?`
```
srikanthccv, line 32: Is this how other SQL db instrumentations sanitize?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -27,6 +27,16 @@
...
 
 
+def _sanitize_query(query):
+    """Remove query content, replace with sanitization symbol.
+    For example `SELECT * FROM table` will sanitize to SELECT ? FROM ?`
+    """
+    sanitize_symbol = " ?"
+    if query and query.split():
+        return query.split()[0] + sanitize_symbol + " FROM" + sanitize_symbol
+    return ""
```
srikanthccv, line 37: This logic is incorrect for many cases, right? Just taking the first word and concatenating it with `? FROM ?` because statements may contain multi-word instructions such as `INSERT INTO` and `ALTER TABLE`, and it's not always guaranteed that there will be `FROM`.

srikanthccv (commented):
I don't think the `_sanitize_query` is correct.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -26,6 +26,87 @@
...
+    For example `SELECT * FROM table` will sanitize to SELECT ? FROM ?`
+    """
+    sanitized_query = ""
+    if not query:
+        return sanitized_query
+
+    for word in query.split():
+        if word.upper() not in sql_reserved_dict:
+            word = "?"
+        sanitized_query += word + " "
```
srikanthccv, line 107: While this is better than the last change I still find this naive because I can think of some `WHERE` clause cases this breaks. I wonder if there is a sqlalchemy does this sanitization work more reliably?

nemoshlag: The idea behind this is that anything which isn't a reserved word will be sanitized. Can you clarify with an example of where clause case which breaks this?

srikanthccv: Imagine a scenario `WHERE` clause filter containing spaces and one of these reserved words. How does it behave for the following query

```sql
SELECT * FROM table WHERE column_name="PRIMARY BANKING STOCK INDEX COLLAPSE". 
```

nemoshlag: Hmm yes, I see what you're saying. This will result with `SELECT ? FROM table WHERE ? ? ? INDEX ?`. I still think the `SELECT ? ?` option is the safest

arbiv:
@shalevr can this be merged? seems that comments were fixed

PR #1716, opened 2023-03-10
Allow access to grpc channel api with the client interceptor

### LINKED ISSUE #1697: Grpc channel API breaks when using grpc client instrumentor
**Describe your environment**
I'm using GrpcInstrumentorClient (https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/grpc/grpc.html#usage-client) version 0.30b1 in a python (3.7) flask application. When closing the channel, using a custom wrapper around channel, calls to the grpc.Channel._channel API fail.

**Steps to reproduce**
1. Instrument with GrpcInstrumentorClient().instrument()
2. In a grpc middleware, or anywhere with access to the grpc channel object, call a method on `channel._channel`, like `channel._channel.check_connectivity_state()`

the grpc `channel._channel` is an instance of `grpc._cython.cygrpc.Channel`

the grpc `channel._channel` is an instance of `grpc._channel.Channel`

**Additional context**
GRPCInstrumentorClient works fine with no middleware or other interceptors present.

ianmcnally:
This would close #1697. The interceptor channel as is breaks access to channel private attributes, like Channel._channel. To allow it, I wrapped the incoming channel with a getattr that tries Channel._channel if it isn't present in Channel.

# Description

It's described in #1697 , but the gist is:

I'm using GrpcInstrumentorClient (https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/grpc/grpc.html#usage-client) version 0.30b1 in a python (3.7) flask application. When closing the channel, using a custom wrapper around channel, calls to the grpc.Channel._channel API fail.

Fixes #1697 

## Type of change

# How Has This Been Tested?

Instrument with GrpcInstrumentorClient().instrument()
In a grpc middleware, or anywhere with access to the grpc channel object, call a method on channel._channel, like channel._channel.check_connectivity_state()

- [x] Test A

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


ianmcnally:
The spellcheck failure is in an untouched file. It's a typo on "instrumentaiton"

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py
```diff
@@ -260,9 +260,25 @@ def invoker(request_iterator, metadata):
...
+    """
+
+    def __init__(self, channel):
+        self._channel = channel
+
+    def __getattr__(self, name):
+        try:
+            return getattr(self._channel, name)
+        except AttributeError:
+            return getattr(self._channel._channel, name)
```
ocelotl, line 276: This is accessing a private attribute and may break at any moment.

ocelotl (changes requested):

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py
ianmcnally: accessing grpc.Channel's `_channel` is the point of this change. because the client interceptor wraps every grpc channel as _InterceptorChannel, it breaks access to any API (public or private) that grpc users may be relying on. Whether that's right or wrong is orthogonal to the issue here, that _InterceptorChannel wraps grpc.Channel, with no access to the original channel's API.

ianmcnally: @ocelotl i'd like to make progress on this. have you had a chance to think about this some more?

i do agree that private attributes should be private, though in this case 1) _InterceptorChannel breaks the grpc.Channel API and 2) that private API is needed to work around a known bug in our version of grpc. i hope we can find a pragmatic solution here. and i'm open to any and all feedback on how to accomplish that.

MikeGoldsmith: Hey @ocelotl - could you add some direction to improving the solution to get the channel? Would wrapping it in another try/catch if the field not longer exists be work?

Thanks 😄

ocelotl: Sorry, reviewing now...

ocelotl: I think we can find a solution:

If I understand the situation correctly, this is what is happening here:

1. We have a public [`intercept_channel`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py#L349) function that returns an `_InterceptorChannel` object.
2. Currently, when an `_InterceptorChannel` is instantiated, it receives a `grpc._channel.Channel` object which is stored in the `_channel` attribute.

@ianmcnally if the problem is that users can't access the `grpc._channel.Channel` object, then this PR does not fix that problem, since the `grpc._channel.Channel` object is still being assigned to the `_channel` attribute of `_InterceptorChannel` and that `_channel` attribute is private and users should not be accessing it. If what you want is to make that `grpc._channel.Channel` object in the `_InteceptorChannel._channel` attribute accessible, then it would be better to add a public `property` method (probably named `channel`) to `_InterceptorChannel` that allows users to access the `grpc._channel.Channel`. I am suggesting a `property` here to keep the `_channel` attribute read-only.

Now, accessing the `_channel` attribute of the `grpc._channel.Channel` object is a separate issue unrelated from the one discussed above.

Apparently the motivation to access this the private `grpc._channel.Channel._channel` attribute is a bug in some version of grpc. I understand that. For that particular case, I would suggest to add an `if` that checks the `gprc` version and if it finds the buggy one, then do the attribute check in the private `_channel` object. Also, please add a comment explaining why this is being done and mention this PR number in that comment for future reference.

ianmcnally: the problem is access to grpc.Channel._channel. Because InterceptorChannel is returned, and sets _channel to grpc.Channel (instead of grpc.Channel._channel), and subsequent grpc code can no longer access the original Channel._channel (which is a cygrpc channel).

my approach solves the problem by getting an attribute from grpc.Channel on InterceptorChannel._channel, then going into grpc.Channel._channel. it's one of several possible approaches. if i understand your proposal correctly, it's adding a new property, which would require downstream changes to access which doesn't seem quite right. and regarding the version check - we still need to do it in a newer grpc version so it's a moving target.

i hope my explanation further clarifies the problem. i've tried the solution proposed here in my company's systems and it's worked well. and i hope i can integrate it upstream here.

ocelotl: I am not opposing your approach of looking into a private `_channel` attributes if the requested attribue can't be found on the original channel object _as a workaround for a bug in grpc_. What I mentioned was:

1. The approach you implemented has the problem of providing the users a _private_ `_channel` attribute instead of a public one.
2. The approach you implemented does not mention this is a workaround and it is not constrained to the buggy versions of grpc. Without this it simply looks like we are accessing private attributes here when we shouldn't.

There is another problem with the implementation in this PR, the original `Channel` object is being wrapped in a `_ChannelWithAccess` object. If a user depended on the object keeping the original class of the object, this would fail.

To be more clear, this is pretty much what I suggest we do instead:

``` diff
diff --git a/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py b/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
index 9f4ede9c..c34be821 100644
--- a/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
@@ -101,7 +101,7 @@ class Boto3SQSInstrumentor(BaseInstrumentor):
     class ContextableList(list):
         """
         Since the classic way to process SQS messages is using a `for` loop, without a well defined scope like a
-        callback - we are doing something similar to the instrumentaiton of Kafka-python and instrumenting the
+        callback - we are doing something similar to the instrumentation of Kafka-python and instrumenting the
         `__iter__` functions and the `__getitem__` functions to set the span context of the addressed message. Since
         the return value from an `SQS.ReceiveMessage` returns a builtin list, we cannot wrap it and change all of the
         calls for `list.__iter__` and `list.__getitem__` - therefore we use ContextableList. It is bound to the
diff --git a/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py b/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py
index edd1d1ce..70294699 100644
--- a/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py
+++ b/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/grpcext/_interceptor.py
@@ -21,6 +21,7 @@
 import collections
 
 import grpc
+from wrapt import ObjectProxy
 
 from opentelemetry.instrumentation.grpc import grpcext
 
@@ -260,27 +261,36 @@ class _InterceptorStreamStreamMultiCallable(grpc.StreamStreamMultiCallable):
         )
 
 
-class _ChannelWithAccess:
+# It seems like this should be a transparent replacement for a
+# grpc._channel.Channel object, which is what ObjectProxy does.
+class _ChannelWithAccess(ObjectProxy):
     """
-    Wraps a grpc channel to allow access to its API before it is replaced
-    with _InterceptorChannel
+    This is a bugfix for ...
     """
 
     def __init__(self, channel):
-        self._channel = channel
+        super().__init__(channel)
+        self._self_channel = channel
 
     def __getattr__(self, name):
-        try:
-            return getattr(self._channel, name)
-        except AttributeError:
-            return getattr(self._channel._channel, name)
+        return getattr(self._self_channel._channel, name)
 
 
 class _InterceptorChannel(grpc.Channel):
     def __init__(self, channel, interceptor):
-        self._channel = _ChannelWithAccess(channel)
+        self._channel = channel
+        self._channel_with_access = _ChannelWithAccess(channel)
         self._interceptor = interceptor
 
+    @property
+    def channel(self):
+        # This is being done here because there is a bug in grpc from version
+        # X that causes...
+        # See #1716
+        if True:  # Instead of True, check that the grpc version is a buggy one
+            return self._channel_with_access
+        return self._channel
+
     def subscribe(self, *args, **kwargs):
         self._channel.subscribe(*args, **kwargs)
 
diff --git a/instrumentation/opentelemetry-instrumentation-grpc/tests/test_client_interceptor.py b/instrumentation/opentelemetry-instrumentation-grpc/tests/test_client_interceptor.py
index 01c322d9..1233c465 100644
--- a/instrumentation/opentelemetry-instrumentation-grpc/tests/test_client_interceptor.py
+++ b/instrumentation/opentelemetry-instrumentation-grpc/tests/test_client_interceptor.py
@@ -352,7 +352,14 @@ class TestClientProto(TestBase):
 
     def test_interceptor_channel_api(self):
         channel = self.channel._channel
-        original_channel_methods = [m for m in dir(cygrpc.Channel) if not m.startswith('_')]
+        original_channel_methods = [
+            m for m in dir(cygrpc.Channel) if not m.startswith("_")
+        ]
 
         for m in original_channel_methods:
-            self.assertTrue(getattr(channel._channel, m), 'method `{m}` is not available on channel._channel'.format(m=m))
+            self.assertTrue(
+                getattr(channel.channel, m),
+                f"method `{m}` is not available on channel.channel",
+            )
+
+        self.assertIsInstance(channel.channel, channel._channel.__class__)
```

Notice the additional test `self.assertIsInstance` at the very end.

I added some changes in the diff above to fix `lint` and `spellcheck`, just for the record. Feel free to use the code in the diff above to implement your solution.

Also, [here](https://wrapt.readthedocs.io/en/latest/wrappers.html#proxy-object-attributes) it explains why I am using `self._self_channel` above.

Hope this helps.

ianmcnally: this is very helpful, thank you

PR #1717, opened 2023-03-10
Implement request and response hooks in requests library instrumentation

### LINKED ISSUE #411: [Requests] Replace span name callback with request and response hooks
Requests instrumentation accepts a span name callback which should be replaced with more generic request/response callbacks (hooks).

Details: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/408

federicobond:
# Description

Replaces the non-standard `name_callback` and `span_callback` with `request_hook` and `response_hook` with standard parameters. Reduced the number of instrumented functions to only one, as only `Session.send` is required to provide the necessary functionality.

I had to remove one test for invalid URL which should not be covered by the OTel spans. This simplifies the implementation of the instrumentor and aligns it with how the other client instrumentations work.

Fixes #411

## Type of change

# How Has This Been Tested?

- [x] Added integration tests for request and response hooks, modified an existing test.

- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix httpx resource warnings
   ([#1695](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1695))
 
+### Changed
+
+- `opentelemetry-instrumentation-requests` Replace `name_callback` and `span_callback` with standard `response_hook` and `request_hook` callbacks
+  ([#670](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/670))
+
 ## Version 1.16.0/0.37b0 (2023-02-17)
 
 ### Added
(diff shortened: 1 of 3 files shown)


srikanthccv:
>I had to remove one test for invalid URL which should not be covered by the OTel spans.

Why should it not be covered? It's probably why we have instrumentation for `Session.request`. If other instrumentation do not have this either it is not possible to do the same in them, or they are missing this but capturing the invalid URL should be part of the instrumentation. This could be essential for some users.

federicobond:
Hi @srikanthccv I hadn't properly explained my reasoning, so thank you for bringing this up.

It's not that it should not be covered by the OTel spans, but rather that before there is a syntactically valid request object there is probably little benefit in starting a request trace. An error while building the request would still be logged to the parent span and most likely appear in an error tracking system, but there would not be a corresponding request span because one wouldn't have been attempted.

This has two added benefits: you can work with a well-built request object to append additional information during instrumentation (the use case for these hooks), and you have stricter guarantees about the values of the request span attributes that get sent to your OTel collector.

It also makes the instrumentation behavior more consistent, since invalid URL errors when manually building `PreparedRequest` objects are not currently logged.

srikanthccv:
I think those are good points. Is there anything significant users lose if we remove `Session.request` wrapper?

srikanthccv (approved):
Thanks!

cruisehall:
@federicobond just wanted to say THANK YOU for implementing this feature -- much needed & very powerful in the right hands :)

Issue #1371, opened 2022-10-04
test-instrumentation-celery failing and blocking PRs


jeremydvoss:
**Describe your environment**
I have a PR out: #1369 . 2 tests are failing. I sent out a plank PR which had the same failures, so I know they are unrelated:
[Contrib Repo Tests / build (py37, instrumentation, ubuntu-20.04) (pull_request)](https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/3178821162/jobs/5180695857)
```
  py37-test-instrumentation-remoulade: commands succeeded
ERROR:   py37-test-instrumentation-celery: commands failed
  py37-test-instrumentation-sklearn: commands succeeded
```
[Contrib Repo Tests / build (pypy3, instrumentation, ubuntu-20.04) (pull_request)](https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/3178821162/jobs/5180696946)
```
E   AttributeError: 'EntryPoints' object has no attribute 'get'
_ ERROR collecting instrumentation/opentelemetry-instrumentation-celery/tests/test_utils.py _
ImportError while importing test module '/home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-celery/tests/test_utils.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.7.14/x64/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test_utils.py:18: in <module>
    from celery import Celery
E   ImportError: cannot import name 'Celery' from 'celery' (/home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/.tox/py37-test-instrumentation-celery/lib/python3.7/site-packages/celery/__init__.py)
!!!!!!!!!!!!!!!!!!! Interrupted: 3 errors during collection !!!!!!!!!!!!!!!!!!!!
============================== 3 errors in 3.19s ===============================
/home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/.tox/py37-test-instrumentation-celery/lib/python3.7/site-packages/pytest_benchmark/logger.py:46: PytestBenchmarkWarning: Not saving anything, no benchmarks have been run!
  warner(PytestBenchmarkWarning(text))
[...]

PR #1725, opened 2023-03-20
Remove pinned version importlib-metadata

srikanthccv:
# Description

Because this conflicts with https://github.com/open-telemetry/opentelemetry-python/blob/d77ca16badd5197d66b9b4dfac1221a5fb09c028/opentelemetry-api/pyproject.toml#L32 and fails

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-celery/pyproject.toml
+++ b/instrumentation/opentelemetry-instrumentation-celery/pyproject.toml
@@ -37,7 +37,6 @@ instruments = [
 test = [
   "opentelemetry-instrumentation-celery[instruments]",
   "opentelemetry-test-utils == 0.38b0.dev",
-  "importlib-metadata==4.13.0",
   "pytest",
 ]
 


instrumentation/opentelemetry-instrumentation-celery/pyproject.toml
```diff
@@ -37,7 +37,6 @@ instruments = [
 test = [
   "opentelemetry-instrumentation-celery[instruments]",
   "opentelemetry-test-utils == 0.38b0.dev",
-  "importlib-metadata==4.13.0",
```
aabmass, line 40: Is this `importlib-metadata` not used at all in this package?

lzchen, line 40: Haha what a coincidence

srikanthccv: No, IIRC, this was pinned to fix the CI that was failing when `importlib-metadata` 5.x was released with breaking changes. This is also part of the `test` dependencies.

srikanthccv: The commit that added this https://github.com/open-telemetry/opentelemetry-python-contrib/commit/93b8398b6068728f999247ad53275d487374f116 to fix https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1371

srikanthccv: But the `opentelemetry-api` pinned version, however, feels very strict `importlib-metadata~= 6.0.0` and can potentially conflict with user programs?

lzchen: @srikanthccv 
it's possible. Maybe we can fix this on an ongoing basis if it comes up?

lzchen (approved):

instrumentation/opentelemetry-instrumentation-celery/pyproject.toml
aabmass: > this is also part of the test dependencies.

ack that makes sense

aabmass (approved):

instrumentation/opentelemetry-instrumentation-celery/pyproject.toml
aabmass: > But the opentelemetry-api pinned version, however, feels very strict importlib-metadata~= 6.0.0 and can potentially conflict with user programs?

I think there was a miscommunication with my comment on @ocelotl's PR. I agree we should do as wide as possible, I just wanted there to be *some* upper bound (which we can then expand as needed)

lzchen: I don't think this should be a blocker.

PR #1806, opened 2023-05-16
Skip requests tests for pypy3

shalevr:
# Description


srikanthccv:
Can you create an issue to address the original issue? We should find out if there is something we can do and enable the tests again.

srikanthccv (approved):

PR #1823, opened 2023-05-23
Distro selection

### LINKED ISSUE #1822: Allow users to specify which distro and configurator to use in auto instrumentation
Yes. Currently, if a user has multiple packages with distro or configurator entry points installed, they have no control over which are used for the auto instrumentation flow.

Copying what we do for exporters, I am adding 2 env vars where users can specify the entry point names of the distros and configurators they want used. I added warnings for each distro and configurator not used.

I considered requiring thes environment variables to be used but instead chose to make them optional. If the env vars are not set, then existing behavior continues: the first distro and configurators found will be used.

**Additional context**
With the way the package is set up, this also adds the --distro and --configurator command line args. Also, we currently have not tests for the sitecustomize file. I found it tricky to implement because importing the module triggers instrumentation. Open to suggestions.

jeremydvoss:
# Description

I added env vars and arguments to allow users to chose which distro and configurator they want to use in auto-instrumentations. I also refactored sitecustomize and added tests for the pre-existing and new functionality.

Fixes #1822 

## Type of change

# How Has This Been Tested?

Manually + unit tests

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


sanketmehta28 (commented):
@jeremydvoss : can you please add some test cases for this change?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/environment_variables.py
```diff
@@ -16,3 +16,13 @@
 """
 .. envvar:: OTEL_PYTHON_DISABLED_INSTRUMENTATIONS
 """
+
+OTEL_PYTHON_DISTRO = "OTEL_PYTHON_DISTRO"
```
sanketmehta28, line 20: are these vars added to specification as well?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/environment_variables.py
```diff
@@ -16,3 +16,13 @@
...
+"""
+.. envvar:: OTEL_PYTHON_DISTRO
+"""
+
+OTEL_PYTHON_CONFIGURATOR = "OTEL_PYTHON_CONFIGURATOR"
```
sanketmehta28, line 25: same as above

sanketmehta28 (changes requested):

opentelemetry-instrumentation/src/opentelemetry/instrumentation/environment_variables.py
jeremydvoss: > @jeremydvoss : can you please add some test cases for this change?

I would like to. Sitecustomize tests don't currently exist. I'm guessing this is because importing the module triggers autoinstrumentation. So I will need to figure out some workaround to enable testing...

jeremydvoss:
> @jeremydvoss : can you please add some test cases for this change?

I would like to. Sitecustomize tests don't currently exist. I'm guessing this is because importing the module triggers autoinstrumentation. So I will need to figure out some workaround to enable testing...

opentelemetry-instrumentation/src/opentelemetry/instrumentation/environment_variables.py
jeremydvoss: Actually, since this is python specific, a cross language spec update does not seem to make sense. But I can update the docs:
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/a6797541721af72e6feab400e24ceaabc08cb31e/opentelemetry-instrumentation/README.rst#L93

https://github.com/open-telemetry/opentelemetry.io/blob/54d9c9e97a7f4afda78a90939835e71bf263db79/content/en/docs/instrumentation/python/automatic/agent-config.md?plain=1#L159

jeremydvoss: https://github.com/open-telemetry/opentelemetry.io/pull/2788

opentelemetry-instrumentation/README.rst
```diff
@@ -18,12 +18,15 @@ This package provides a couple of commands that help automatically instruments a
...
 .. note::
     You need to install a distro package to get auto instrumentation working. The ``opentelemetry-distro``
-    package contains the default distro and automatically configures some of the common options for users.
+    package contains the default distro and configurator and automatically configures some of the common options for users.
     For more info about ``opentelemetry-distro`` check `here <https://opentelemetry-python.readthedocs.io/en/latest/examples/distro/README.html>`__
     ::
 
         pip install opentelemetry-distro[otlp]
 
+    When creating a custom distro and/or configurator, be sure to add entry points for each under opentelemetry_distro and opentelemetry_configurator respectfully.
```
ocelotl, line 27:
```suggestion
    When creating a custom distro and/or configurator, be sure to add entry points for each under `opentelemetry_distro` and `opentelemetry_configurator` respectfully.
```

opentelemetry-instrumentation/README.rst
```diff
@@ -18,12 +18,15 @@ This package provides a couple of commands that help automatically instruments a
...
+    If you have entry points for multiple distros or configurators present in your environment, you should specify the entry point name of the distro and configurator you want to be used via the OTEL_PYTHON_DISTRO and OTEL_PYTHON_CONFIGURATOR environment variables.
```
ocelotl, line 28:
```suggestion
    If you have entry points for multiple distros or configurators present in your environment, you should specify the entry point name of the distro and configurator you want to be used via the `OTEL_PYTHON_DISTRO` and `OTEL_PYTHON_CONFIGURATOR` environment variables.
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -0,0 +1,123 @@
...
+)
+from opentelemetry.instrumentation.distro import BaseDistro, DefaultDistro
+from opentelemetry.instrumentation.environment_variables import (
+    OTEL_PYTHON_CONFIGURATOR,
+    OTEL_PYTHON_DISABLED_INSTRUMENTATIONS,
+    OTEL_PYTHON_DISTRO,
+)
+from opentelemetry.instrumentation.version import __version__
+
+logger = getLogger(__name__)
```
ocelotl, line 31:
```suggestion
_logger = getLogger(__name__)
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def _load_distros() -> BaseDistro:
+    distro_name = environ.get(OTEL_PYTHON_DISTRO, None)
+    for entry_point in iter_entry_points("opentelemetry_distro"):
+        try:
+            if distro_name is None or distro_name == entry_point.name:
```
ocelotl, line 38: Should we call `entry_point.load()` if `distro_name is None`?

ocelotl: This seems to make the default behavior to be "return the first distro if there is no selected distro. This is a behavior we want to avoid, if the distro is not specifically selected a default fixed distro should be returned. Maybe this is what happens because a default distro is always set as the first distro but this function does not make that very obvious.

ocelotl (commented):
Leaving a comment regarding the default distro returned

opentelemetry-instrumentation/tests/auto_instrumentation/test_load.py
```diff
@@ -0,0 +1,299 @@
...
+from opentelemetry.instrumentation.auto_instrumentation import _load
+from opentelemetry.instrumentation.environment_variables import (
+    OTEL_PYTHON_CONFIGURATOR,
+    OTEL_PYTHON_DISABLED_INSTRUMENTATIONS,
+    OTEL_PYTHON_DISTRO,
+)
+from opentelemetry.instrumentation.version import __version__
+
+
+class TestLoad(TestCase):
```
aabmass, line 28: LGTM but it would be good to add some comments describing what the test scenarios are when not obvious. It's a little tricky to follow what's happening with the mocks

aabmass (approved):
Approval for the testing code discussed in the SIG.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
jeremydvoss: This is the current behavior. I figured it would be best to expand upon it so as to not break existing users. I can add a comment to make that clear in the code, but I feel like it is clearer in the README.

opentelemetry-instrumentation/README.rst
```diff
@@ -18,12 +18,15 @@ This package provides a couple of commands that help automatically instruments a
...
 .. note::
     You need to install a distro package to get auto instrumentation working. The ``opentelemetry-distro``
-    package contains the default distro and automatically configures some of the common options for users.
+    package contains the default distro and configurator and automatically configures some of the common options for users.
     For more info about ``opentelemetry-distro`` check `here <https://opentelemetry-python.readthedocs.io/en/latest/examples/distro/README.html>`__
     ::
 
         pip install opentelemetry-distro[otlp]
 
+    When creating a custom distro and/or configurator, be sure to add entry points for each under `opentelemetry_distro` and `opentelemetry_configurator` respectfully.
```
pmcollins, line 27:
```suggestion
    When creating a custom distro and/or configurator, be sure to add entry points for each under `opentelemetry_distro` and `opentelemetry_configurator` respectively.
```

pmcollins:
Hi all -- is there a doc or a spec somewhere that explains the differences between a "distro" and a "configurator"? (if not I'd be happy to write it) [These docs](https://opentelemetry.io/docs/instrumentation/python/distro/) say that the intent of a distro is to "automatically configure some of the more common options for users", which seems like an apt description for a configurator as well.

This change appears to expose to users that there is indeed a difference, but they may be confused as to what it is.

If the answer ends up being "it's a distinction without a difference" we may want to consider consolidating them before we expose their existence to users.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
jeremydvoss: Done manually.

opentelemetry-instrumentation/tests/auto_instrumentation/test_load.py
jeremydvoss: Added comments. Let me know if these are in line with what you mean.

aabmass: LGTM

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -0,0 +1,124 @@
...
+    OTEL_PYTHON_DISABLED_INSTRUMENTATIONS,
+    OTEL_PYTHON_DISTRO,
+)
+from opentelemetry.instrumentation.version import __version__
+
+_logger = getLogger(__name__)
+
+
+def _load_distros() -> BaseDistro:
+    distro_name = environ.get(OTEL_PYTHON_DISTRO, None)
```
pmcollins, line 35: Since we're only one level away from the bottom of the call stack here, what do you think about making `distro_name` an argument to this function and having the caller pass in the environment variable value?

Also, while we're in here, should the name of the function be `_load_distro()` since it only appears to load one?

jeremydvoss: Good point. Will change

jeremydvoss: Done.

jeremydvoss: Missed the first part of this. I changed the name to _load_distro(). @pmcollins Could you explain the purpose of changing function signature? I think since we cannot test sitecustomize itself, it is nice to have all the functionality in the function itself.

pmcollins: Ok, no worries.

My vision though, is to nudge the code in the direction of reading environment variables in one place at the beginning of a program's lifecycle (in main or the equivalent, which in this case is sitecustomize.py) and then passing those values as function arguments, rather than reading environment variables at specific points in the call stack only when they are needed.

To this end, environment variables could be read directly in `initialize()` and then passed as arguments to the functions that need those values. However, in the longer term, it would be nice to have a separate, independently testable loader for environment variables. This would allow us to modify the environment in our tests without affecting the global state, which could be beneficial for testing parts of the code that depend on these variables.

In any case, we can address this separately if folks are interested. I am open to other perspectives on this, and I am happy to discuss more if needed.

pmcollins: 👆 Not a blocking comment btw. Discussed this with Jeremy and environment variable management is something that we can consider separately.

ocelotl:
Filed #1880

ocelotl (approved):

PR #1825, opened 2023-05-25
[WIP] add our own build_absolute_uri method

### LINKED ISSUE #1781: DisallowedHost exception for Django instrumentation
**Describe your environment**
- Django 4.1.7
- Python 3.11.1
- opentelemetry-instrumentation-django 0.38b0

**Steps to reproduce**
Create basic django app:

```sh
mkdir django-test && cd django-test
python -m venv .venv
source .venv/bin/activate
pip install django
django-admin startproject django_test .
python manage.py migrate
cat <<EOF > gunicorn.config.py
from opentelemetry.instrumentation.django import DjangoInstrumentor

def post_fork(server, worker):
    DjangoInstrumentor().instrument()
EOF
```

Only allow hosts from a specific origin. This is best practice in terms of security.
Change the following variables in `django_test/settings.py`:

```python
DEBUG = False
ALLOWED_HOSTS = ['example.com']
```

Create a health check middleware. The middleware is added at the beginning of the middleware
array, since we don't want the allowed hosts rule to apply to the health check, since we'll
likely be receiving health checks from the load balancer, which does not have the correct host.

```sh
cat <<EOF > django_test/middleware.py
from django.http import HttpResponse


class HealthCheckMiddleware:
    def __init__(self, get_response):
        self.get_response = get_response

    def __call__(self, request):
        if request.path == "/api/health":
            return HttpResponse("ok\n")
        return self.get_response(request)
EOF
```

Add to the array in `django_test.settings.py`:

```python
MIDDLEWARE = [
    'django_test.middleware.HealthCheckMiddleware',
    # ... other middlewares here
]
```

Now the health check route works as expected when we run the server:
```sh
python manage.py runserver
```

```sh
❯ curl localhost:8000/api/health
ok
```

Now we need to add open telemetry django and gunicorn:

```sh
pip install opentelemetry-instrumentation-django
pip install gunicorn
```

Now we run django with instrumentation:

```sh
DJANGO_SETTINGS_MODULE=django_test.settings OTEL_SERVICE_NAME=TestApi gunicorn django_test.wsgi:application -c gunicorn.config.py
```

The healthcheck endpoint fails with 400 now, because of the DisallowedHost exception. More specifically, the error is raised [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/890e5dd9b8a4ec263f183fe9d0362ffb907847a4/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py#L195) on the `request.build_absolute_uri` call. 

The healthcheck endpoint should not fail when telemetry is enabled.

I'm not sure what the best solution for this is, but I guess we want to build the absolute URI without calling `request.build_absolute_uri`, since it will throw an exception.

**EDIT**: since I'm new to the library, I'm not sure what the best solution would be. But if you have any pointers feel free to say them that I can try to fix the issue as well.

TheAnshul756:
# Description

`request.build_absolute_uri` method when called internally checks for Alloowed Host which is unwanted behaviour and leads to bugs in some cases. We implemented our own `build_absolute_uri` and used that instead to avoid it.

Fixes #1781 

## Type of change

# How Has This Been Tested?

- [ ] `tox -e test-instrumentation-django1`
- [ ] `tox -e test-instrumentation-django2`
- [ ] `tox -e test-instrumentation-django3`

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


srikanthccv:
I wouldn't recommend this. I think It is a hacky way to get around the fact that `build_absolute_uri` calls some internal method `get_host` which throws the exception. I belive the ability to configure the OTEL middleware is the best approach.

srikanthccv:
> Checking for allowed_hosts while calling build_absolute_uri() is totally unexpected and out of scope for otel middleware.
> Using this solution, i believe we are providing the functionality of build_absolute_uri() without the unnecessary allowed_hosts check.

I think this is a hacky workaround and customising the OTEL middleware order to be a better solution. The allowed_hosts check is not unnecessary for the user. It's a security practice for them. This solution copies the original `build_absolute_uri` because it calls the internal `get_host ` that raises the Exception. By writing "our own `build_absolute_uri`" it just skips that part. 


> just to overcome the problem of one django method, it does not make sense to modify the order of the Otel middleware.

How are you sure that it is just one method? we realised this `get_host` because someone created an issue about it. If a new person comes up with a different request path where things break because instrumentation always adds it to the beginning of the middleware list, will we keep copying the Django methods? 

I also mentioned other use cases where adding OTEL middleware always at the beginning breaks people's applications. Say, an end user has custom middlewares that must execute first before anything, even before the request reaches the Telemetry. They can't be okay with the fact that OTEL middleware must be first in the beginning and there is no way to change it as of today. I also thought @TheAnshul756 was on board with the idea since they responded +1.

TheAnshul756:
> The allowed_hosts check is not unnecessary for the user. It's a security practice for them.

Yes, it is not unnecessary, but the user does not expect the middleware to throw an exception. The basic expectation from the Otel middleware is to capture necessary telemetry data without impacting the incoming request or outgoing response.

> How are you sure that it is just one method? we realised this `get_host` because someone created an issue about it. If a new person comes up with a different request path where things break because instrumentation always adds it to the beginning of the middleware list, will we keep copying the Django methods?

We need to thoroughly test any Django methods we use in the middleware to ensure they do not have any side effects. If they do, we should avoid using them and consider alternative methods or custom implementations.

> I also mentioned other use cases where adding OTEL middleware always at the beginning breaks people's applications. Say, an end user has custom middleware that must execute first before anything, even before the request reaches the Telemetry. They can't be okay with the fact that OTEL middleware must be first in the beginning and there is no way to change it as of today. I also thought @TheAnshul756 was on board with the idea since they responded +1.

I agree with your suggestion for the new feature, as it would be beneficial to have more flexibility in the positioning of OTEL middleware. However, I believe it does not address the current bug at hand. The exception being thrown by Otel middleware is not the expected behavior and needs to be addressed appropriately.

After our previous discussion, I pondered over this matter and raised my concerns during the SIG meeting to gather input from others regarding this issue.

srikanthccv:
I no longer attend the SIG, so I will go with what the majority decides there. My broader point is there will always be exceptions because of the incorrect order of execution. And letting users choose where to place the OTEL middleware solves them, including the linked issue.

TheAnshul756:
While I agree that allowing users to choose the placement of the OTEL middleware can address some exceptions and the linked issue, I still believe that the current bug should be handled separately.

Even if we provide the option to place the middleware at any specific position, it does not guarantee that this particular bug will not arise. Customers do not expect the Otel middleware to exhibit this behaviour, and they may not anticipate this issue when adding middleware at any specific position. I think it would be valuable to gather input from others to hear their opinions on this issue.

I agree with you that providing an option to specify the middleware location would be useful for customers. I will create a separate issue to address this feature request.

srikanthccv:
My comments are based on my past observations about the instrumentations. The vetting of methods to avoid side effects becomes obsolete because we don't control the libraries, and library authors change the code in surprising ways internally (this may not be as often with Django since it's a mature framework, but in general, this happens a non-trivial number of times) while keeping the API and Behaviour backwards compatible. You are using the private function `_get_raw_host`, which we strictly avoid, given how such attempts bit us in the past.  The custom solution might no longer do the right thing as the underlying library evolves. IIRC, the call taken was to integrate well with the library and rely on it as much as possible. 

Django developers make a conscious decision about the order of middlewares and my first thought was OTEL middleware shouldn't be any special. There is already the `DJANGO_SETTINGS_MODULE` we instruct, so adding another one for order specification will only make things clear and more upfront. 

I think I tried to convey what I have in my mind; it's enough discussion so I will rest my case here.

PR #1830, opened 2023-05-30
Fix async redis clients tracing

### LINKED ISSUE #1829: Redis asyncio clients not correctly traced
Currently the redis asyncio clients (Redis, Pipeline etc) are not being traced correctly. Every trace is a tiny fragment of the actual time, occurring around the start time of the redis command being called.

I narrowed this down to the instrumentor for redis not wrapping the async variants as expected (it uses the same sync approach that it uses for the sync redis - and as a result the span only traces the coroutine creation, but not the coroutine actually being awaited).

The result is a very short trace (< 1ms) at the start of the redis call that doesn't actually show the full time of the redis command.

I have fixed this in a follow up PR, which I will put up shortly and link.

**Describe your environment**
Python version: 3.11
redis-py version: 4.5.5
opentelemetry-instrumentation-redis version: current master or 0.39b0
OS: Ubuntu

**Steps to reproduce**

Here is an example script, which calls an async redis command, and an async redis pipeline.

For each async call:
- A parent span is started
- asyncio.sleep for 50 milliseconds (to create a time delta between the parent span and the actual redis span for visual clarity)
- call a redis command (debug-sleep) for 100 milliseconds

The data can then be visualised and the parent span compared to the redis-instrumentor generated span to see that the span doesn't encompass the full time of the command.

```
import asyncio

from opentelemetry.instrumentation.redis import RedisInstrumentor
from opentelemetry.sdk.resources import Resource
from redis.asyncio import Redis



tracer_provider = TracerProvider(
    resource=Resource.create({
        "service.name": "my-app",
        "service.instance.id": "my-app-4",
    }),
)
tracer = tracer_provider.get_tracer(__name__)


async def call_redis_command() -> None:
    redis_pool = Redis(host="0.0.0.0", port=6379)
    with tracer.start_as_current_span("redis-command-wrapper"):
        await asyncio.sleep(0.05)
        await redis_pool.execute_command("DEBUG", "SLEEP", "0.1")

async def call_redis_pipeline() -> None:
    redis_pool = Redis(host="0.0.0.0", port=6379)
    with tracer.start_as_current_span("redis-pipeline-wrapper"):
        await asyncio.sleep(0.05)
        p = await redis_pool.pipeline()
        p.execute_command("DEBUG", "SLEEP", "0.1")
        await p.execute()

span_processor = BatchSpanProcessor(span_exporter=exporter)

RedisInstrumentor().instrument(tracer_provider=tracer_provider)

asyncio.run(call_redis_command())
asyncio.run(call_redis_pipeline())
```

Below I have used jaeger to visualise the traces for ease of reading

Expecting to see 100 milliseconds of a redis command, wrapped by the 150 milliseconds parent span
image
(This is a screenshot from when I fix the code)

Actually seeing a really tiny redis command span (< 0.2ms) wrapped by the 150millisecond parent span
image
(This is a screenshot from the results of the example provided script)

Vivanov98:
# Description

Creates a separate wrapper for tracing the async client functions. These ensure that the redis coroutines are executed before ending the spans, which results in correct tracing behaviour for async commands.

(see linked issue for example screenshots of expected behaviour)

Fixes #1829 

## Type of change

# How Has This Been Tested?

The docker integration tests have been updated for redis instrumentation and cover the changed code.

A lightweight unit test was also added in this PR to simplify testing that the async client class is called correctly by the async 

The example reproduce script that is provided in the issue can be used to validate that the traces now fully span the redis calls, here are the results:

Before
image

After
image

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


srikanthccv (approved):
Can you also add some functional tests here https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/tests/opentelemetry-docker-tests/tests/redis/test_redis_functional.py? Otherwise LGTM

Vivanov98:
Thanks for the feedback!

I have added integration tests for each async traceable command that confirm the trace encompasses the full query time, and these tests only work with the fixed code

tests/opentelemetry-docker-tests/tests/redis/test_redis_functional.py
```diff
@@ -367,6 +420,33 @@ async def pipeline_immediate():
...
+                await coro
+                finish_time = time_ns()
+
+        async_call(pipeline_simple())
+
+        spans = self.memory_exporter.get_finished_spans()
+        self.assertEqual(len(spans), 1)
+        span = spans[0]
+        self.assertTrue(span.start_time > coro_created_time)
+        self.assertTrue(span.end_time < finish_time)
```
srikanthccv, line 448: Is this assertion enough? Wouldn't `span.end_time` be less than `finish_time` even if this fix isn't applied?

Vivanov98: Yep this second assertion is kind of redundant in terms of testing the fix, but is useful in general to confirm the span timing isn't doing something weird to extend beyond the expected boundary.

The first assertion that checks the start time is the key one in terms of validating the fix.

To really fully validate that the span fully encompasses the coroutine time would require patching every coroutine to return a timestamp to be captured and comparing that. This would kind of nullify the purpose of these integration tests though as then redis isn't being called. We could instead put these tests into the unit test section?
Additionally, this pattern of time testing doesn't seem to be followed anywhere else in this repository so I am hesitant to add too much testing around this accidental bug

Vivanov98:
@srikanthccv just wanted to check in if this PR is okay to be merged?

srikanthccv:
Please resolve the conflicts. The change seems fine to me, I was wondering if tests could be made better.

Vivanov98:
Thanks @srikanthccv I have resolved the conflicts. As discussed I don't think this is the best context to add a new pattern of timing tests given that this pattern doesn't exist anywhere else in the repository

Vivanov98:
@srikanthccv @shalevr could you please help me get this merged?

shalevr (approved):

PR #1838, opened 2023-06-08
Fix python path to support both windows and linux

liad-miggo:
# Description

The library had a bug regarding the execution of the python path in windows context.
When there trying to execute a python executable containing a space character in the path it actually tried to execute just the first path.
i.e 'C:\Program Files\Python311' there is a space the in program files, tried to execute C:\Program.exe and use the rest as arguments.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Tested against both windows and linux operating systems.

- [ ] No.


shalevr:
Some of the tests failed

liad-miggo:
Yeah I saw that, gonna fix it soon

PR #1850, opened 2023-06-12
Remove use of httpbin

### LINKED ISSUE #1844: Mock httpbin for requests tests

ocelotl:
This is done in order to prevent confusion. We are trying to stop using httpbin.org for our tests. Even when the tests for the requests instrumentation do not actually perform any request to httpbin.org, because the test requests are being mocked with httpretty, having the string httpbin.org in the tests can cause confusion and make the reader think the tests are actually using httpbin.org.

Fixes #1844


ocelotl:
Will merge this PR and others in only one.

PR #1852, opened 2023-06-12
Remove usage of httpbin.org

### LINKED ISSUE #1845: Mock httpbin for urllib tests

ocelotl:
This is done in order to prevent confusion. Even when the urllib tests do not make an external request to httpbin.org, having the httpbin.org URL in our tests is confusing and can give the impression we are actually using httpbin.org.

Fixes #1845


shalevr (approved):

ocelotl:
Will merge this PR with others.

PR #1863, opened 2023-06-15
Unwrap Celery's `ExceptionInfo`

unflxw:
# Description

When a task fails and Celery reports an exception, the exception reported is an `ExceptionInfo` wrapper that Celery provides to the signal handler. The actual exception that was raised is a property of this exception.

This PR changes the instrumentation's behaviour to report the actual exception that is wrapped inside `ExceptionInfo`, ensuring that the resulting exception has a type and traceback that are meaningful and relevant to the application's developer.

Fixes appsignal/appsignal-python#76.

## Type of change

(I consider this a bug fix, in that the useful exception to report is the one that was raised in the task, not its Celery wrapper -- please let me know if you disagree that this is a non-breaking change)

# How Has This Been Tested?

The instrumentation's functional tests have been amended.

- [x] No.

# Checklist:

- [x] Unit tests have been added (and functional tests have been amended)
- [x] ~Documentation has been updated~ N/A

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
@@ -76,8 +76,13 @@ from opentelemetry.propagate import extract, inject
 from opentelemetry.propagators.textmap import Getter
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.trace.status import Status, StatusCode
+from billiard import VERSION
 
-ExceptionWithTraceback = utils.import_exception_with_traceback()
+
+if VERSION >= (4, 0, 1):
+    from billiard.einfo import ExceptionWithTraceback
+else:
+    ExceptionWithTraceback = None
 
 logger = logging.getLogger(__name__)
 
--- a/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
@@ -66,13 +66,6 @@ from typing import Collection, Iterable
 from billiard.einfo import ExceptionInfo
 from celery import signals  # pylint: disable=no-name-in-module
 
-try:
-    from billiard.einfo import (  # pylint: disable=no-name-in-module
-        ExceptionWithTraceback,
-    )
-except ImportError:
-    ExceptionWithTraceback = None
-
 from opentelemetry import trace
 from opentelemetry.instrumentation.celery import utils
 from opentelemetry.instrumentation.celery.package import _instruments
@@ -83,6 +76,8 @@ from opentelemetry.propagate import extract, inject
 from opentelemetry.propagators.textmap import Getter
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.trace.status import Status, StatusCode
+
+ExceptionWithTraceback = utils.import_exception_with_traceback()
 
 logger = logging.getLogger(__name__)
 
--- a/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
@@ -64,6 +64,12 @@ from timeit import default_timer
 from typing import Collection, Iterable
 
 from celery import signals  # pylint: disable=no-name-in-module
+from billiard.einfo import ExceptionInfo
+
+try:
+    from billiard.einfo import ExceptionWithTraceback # pylint: disable=no-name-in-module
+except ImportError:
+    ExceptionWithTraceback = None
 
 from opentelemetry import trace
 from opentelemetry.instrumentation.celery import utils
@@ -271,6 +277,21 @@ class CeleryInstrumentor(BaseInstrumentor):
             return
 
         if ex is not None:
+            # Unwrap the actual exception wrapped by billiard's
+            # `ExceptionInfo` and `ExceptionWithTraceback`.
+            if (
+                isinstance(ex, ExceptionInfo)
+                and ex.exception is not None
+            ):
+                ex = ex.exception
+
+            if (
+                ExceptionWithTraceback is not None
+                and isinstance(ex, ExceptionWithTraceback)
+                and ex.exc is not None
+            ):
+                ex = ex.exc
+
             status_kwargs["description"] = str(ex)
             span.record_exception(ex)
         span.set_status(Status(**status_kwargs))
--- a/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
@@ -63,11 +63,13 @@ import logging
 from timeit import default_timer
 from typing import Collection, Iterable
 
-from celery import signals  # pylint: disable=no-name-in-module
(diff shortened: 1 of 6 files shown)


instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
```diff
@@ -256,6 +257,13 @@ def _trace_failure(*args, **kwargs):
             return
 
         if ex is not None:
+            # Unwrap the actual exception wrapped by
+            # Celery's `ExceptionInfo` and report that.
+            if (
+                isinstance(ex, ExceptionInfo)
```
shalevr, line 263: What version of celery did you test? for celery 5.3 the ExceptionInfo changed and now the ex.exception is ExceptionWithTraceBack, so maybe in this case we need the exception field inside the ExceptionWithTraceBack

unflxw: Right! I'll amend this for newer version. I'll also import these from `billiard.einfo`, since it seems like newer versions of Celery don't expose them from `celery.datastructures`.

instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
```diff
@@ -256,6 +257,19 @@ def _trace_failure(*args, **kwargs):
...
         if ex is not None:
+            # Unwrap the actual exception wrapped by billiard's
+            # `ExceptionInfo` and `ExceptionWithTraceback`.
+            if (
+                isinstance(ex, ExceptionInfo)
+                and ex.exception is not None
+            ):
+                ex = ex.exception
+            elif (
+                isinstance(ex, ExceptionWithTraceback)
```
shalevr, line 268: I think here you meant to check isinstance(ex.exception, ExceptionWithTraceback)
and ex.exception.exc is not None
because the ex is the ExceptionInfo the ex.exception is the ExceptionWithTraceback, right?

unflxw: Is that how it goes? I'm most likely misunderstanding how Celery wraps these exceptions. (It certainly doesn't help that I don't understand _why_ Celery wraps these exceptions)

I thought what you were saying earlier was that `ex` was either an `ExceptionInfo` in older versions (and therefore we unwrap `ex.exception`) or an `ExceptionWithTraceback` in newer versions (and therefore we unwrap `ex.exc`) -- but the way you describe it now, it seems that an `ExceptionInfo` wraps an `ExceptionWithTraceback` which wraps the actual exception? 🤔 

Let me know if I understood it right this time and I'll amend it. To be clear, in that case, would the twice-wrapped exception (`ex.exception.exc`) have a type and traceback that are meaningful to the user? Or should we somehow grab the traceback from `ExceptionWithTraceback` instead?

unflxw: Done! I also made it so that the import of `ExceptionWithTraceback` is allowed to fail, for older versions of Celery or billiards where it doesn't exist.

unflxw:
Is there anything I can do to help get this reviewed and merged?

ocelotl:
> Is there anything I can do to help get this reviewed and merged?

sorry for the late review, taking a look right now...

instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
```diff
@@ -63,8 +63,16 @@ def add(x, y):
 from timeit import default_timer
 from typing import Collection, Iterable
 
+from billiard.einfo import ExceptionInfo
 from celery import signals  # pylint: disable=no-name-in-module
 
+try:
```
ocelotl, line 69: This is a bit convoluted. If I understand correctly, this is done to handle different versions of `billiard`, right? (I mean some version of `billiard` have `billiard.einfo.ExceptionWithTraceback` and others don't, right?)

If that is the case, please check the version of `billiard` and use an `if`/`else` instead of `try`/`except`. For the reader is more clear this way what is being intended here.

instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
```diff
@@ -271,6 +279,18 @@ def _trace_failure(*args, **kwargs):
             return
 
         if ex is not None:
+            # Unwrap the actual exception wrapped by billiard's
```
ocelotl, line 282: We need some testing for these changes.

ocelotl (changes requested):

instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
unflxw: Is it not tested implicitly [here](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1863/files#diff-9d6bf208153ddc0d6bd1c4044cd9244a5282a3b05709e991809f5506ce3a4ae6L282-R282)? The exception is now of type `Exception` (the type that was raised) instead of `ExceptionInfo` (the Billiard wrapper)

unflxw: Added an unit test for a failing task.

instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
unflxw: Done, though I'm not certain if it's much clearer. Moved the `if/else` to `utils.py`, so at least it's out of the way.

unflxw:
@ocelotl Please re-review when you have a moment :)

unflxw:
amended the merge fixing the isort linter issue

unflxw:
Could I get another review @shalevr @ocelotl 🙏 and maybe a merge? 🥺

instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
```diff
@@ -76,6 +77,8 @@ def add(x, y):
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.trace.status import Status, StatusCode
 
+ExceptionWithTraceback = utils.import_exception_with_traceback()
```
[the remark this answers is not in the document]
unflxw: Done.

instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
```diff
@@ -271,6 +274,18 @@ def _trace_failure(*args, **kwargs):
             return
 
         if ex is not None:
+            # Unwrap the actual exception wrapped by billiard's
+            # `ExceptionInfo` and `ExceptionWithTraceback`.
+            if isinstance(ex, ExceptionInfo) and ex.exception is not None:
+                ex = ex.exception
+
+            if (
```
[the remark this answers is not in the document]
unflxw: That would make a lot of sense! Which is why that's not how BIlliard works. 😛 An `ExceptionInfo` wraps an `ExceptionWithTraceback`: https://github.com/celery/billiard/blob/32e51b50aa3e6eaec597f5587c41951376554066/billiard/einfo.py#L157

(In older versions, an `ExceptionInfo` wraps the exception directly)

shalevr (approved):

PR #1892, opened 2023-07-18
Fix pyyaml version

### LINKED ISSUE #1891: CI is broken
Probably related to the new release of Cython: 3.0.0. See [this](https://github.com/yaml/pyyaml/issues/724).

ocelotl:
This is related to the recent release of Cython 3.0.0.

See https://github.com/yaml/pyyaml/issues/724

Fixes #1891

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -493,7 +493,7 @@ commands =
   sphinx-build -E -a -W -b html -T . _build/html
 
 [testenv:spellcheck]
-basepython: python3.10
+basepython: python3.9
 recreate = True
 deps =
   codespell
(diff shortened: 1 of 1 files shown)


srikanthccv (approved):
Do you think this should be changed back in follow-up when things are sorted from upstream deps?

tox.ini
```diff
@@ -502,7 +502,7 @@ commands =
   codespell
 
 [testenv:lint]
-basepython: python3.10
+basepython: python3.9
```
lzchen, line 505: Is this change needed?

lzchen (approved):

tox.ini
ocelotl: Yes, seems related, `lint` fails without this change

ocelotl:
> Do you think this should be changed back in follow-up when things are sorted from upstream deps?

Right, opened #1893

PR #1913, opened 2023-08-15
Fix errors introduced in regression

lzchen:
CI was not able to run successfully for security change: https://github.com/open-telemetry/opentelemetry-python-contrib-ghsa-5rv5-6h4r-h22v/pull/2

1. Fixing typing errors introduced by the above pr for Python 3.7
2. Change fix, name is required to be UNKNOWN instead of NONSTANDARD
3. Fix tests

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
+++ b/instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
@@ -289,7 +289,7 @@ class TestWsgiApplication(WsgiTestBase):
         self.environ["REQUEST_METHOD"]= "NONSTANDARD"
         app = otel_wsgi.OpenTelemetryMiddleware(simple_wsgi)
         response = app(self.environ, self.start_response)
-        self.validate_response(response, span_name="HTTP UNKNOWN", http_method="UNKNOWN")
+        self.validate_response(response, span_name="UNKNOWN /", http_method="UNKNOWN")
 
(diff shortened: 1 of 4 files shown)


instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
```diff
@@ -289,7 +289,7 @@ def test_nonstandard_http_method(self):
         self.environ["REQUEST_METHOD"]= "NONSTANDARD"
         app = otel_wsgi.OpenTelemetryMiddleware(simple_wsgi)
         response = app(self.environ, self.start_response)
-        self.validate_response(response, span_name="HTTP UNKNOWN", http_method="UNKNOWN")
+        self.validate_response(response, span_name="UNKNOWN /", http_method="UNKNOWN")
```
jeremydvoss, line 292: Where does the / come from?

jeremydvoss (approved):

instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
lzchen: I believe it is what `PATH_INFO` in wsgi environ is defaulting to for this response. Similar to https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py#L132

PR #1933, opened 2023-09-05
Relaxed the version range in opentelemetry-instrumentation-starlette

danygielow:
# Description

The Signature of `Starlette.add_middleware()` seems to be quite stable and hasn't changed since version 0.13 (from mid 2020). The current version as of today is 0.31.1. So we can assume that add_middleware Signature won't change in a meaningful way until 1.x or beyond. 

That's why I think relaxing the version range to only set a minimum version would be best for future proofness.

BUGFIXES: 
- I found a bug in a missing leading underscore of `_is_instrumented_by_opentelemetry`.
- Fixed DeprecationWarnings in tests for route decorators

## Type of change

# How Has This Been Tested?

I added checks for the `_is_instrumented_by_opentelemetry` in one test.
I ran the tox test env.

- [X] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/__init__.py
@@ -224,7 +224,7 @@ class StarletteInstrumentor(BaseInstrumentor):
                 tracer_provider=tracer_provider,
                 meter=meter,
             )
-            app.is_instrumented_by_opentelemetry = True
+            app._is_instrumented_by_opentelemetry = True
 
             # adding apps to set for uninstrumenting
             if app not in _InstrumentedStarlette._instrumented_starlette_apps:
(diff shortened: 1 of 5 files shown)


ocelotl:
Hello @danygielow, we need a [CLA](https://docs.linuxfoundation.org/lfx/easycla/v2-current/contributors) signature for any PR in this repo.

instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/__init__.py
```diff
@@ -219,7 +219,7 @@ def instrument_app(
                 tracer_provider=tracer_provider,
                 meter=meter,
             )
-            app.is_instrumented_by_opentelemetry = True
+            app._is_instrumented_by_opentelemetry = True
```
ocelotl, line 222: This is actually a bugfix, right? Can you add a test case? This could be the main change of this PR.

ocelotl (changes requested):

instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/__init__.py
danygielow: I have added some assertions to the `test_metric_for_uninstrument_app_method` test. See below.

PR #1952, opened 2023-09-14
Exclude background task execution from root server span in ASGI middleware 

### LINKED ISSUE #1684: Exclude background tasks from request duration in ASGI middleware
The duration of a request span in FastAPI includes the duration of evaluating background tasks. There is no span whose duration represents the time bewteen receiving a request and sending the response. An example:

```python
import asyncio
from typing import TypedDict

from fastapi import BackgroundTasks, Depends, FastAPI

app = FastAPI()
provider = TracerProvider()
processor = BatchSpanProcessor(exporter)


def get_context(
    background_tasks: BackgroundTasks,
):
    return Context(background_tasks=background_tasks)


async def some_background_task():
    await asyncio.sleep(10)


class Context(TypedDict):
    background_tasks: BackgroundTasks


@app.get("/")
async def main(context=Depends(get_context)):
    print(context.keys())
    context["background_tasks"].add_task(some_background_task)
    return None

```

Either end the server span after we have sent the response (this is how dd-trace solve it in https://github.com/DataDog/dd-trace-py/pull/3799)

*OR* create a new span that stops after the response has been sent.

**Additional context**
This is important to calculation latency metrics using the FastAPI instrumentation

siminn-arnorgj:
# Description

Some ASGI web frameworks, like FastAPI, support scheduling of background tasks that execute after the server has sent a response to the client. Currently, the root server span doesn't end when the response is dispatched but rather when the server has completely processed the request. This can skew aggregate statistics, potentially making it appear as if the server has high latency when, in reality, it's only executing background tasks that don't impact the actual latency.

To address this, I've made changes to end the `server_span` upon sending the final response body. To ensure there are no duplicate attempts at ending the `server_span`, the span is initialized with `end_on_exit=False` and then ended manually if it's still recording.

**However**, one challenge with my solution pertains to error reporting. In the original implementation, exceptions in background tasks get reported on the top-level server span. With the changes in this PR, errors won't be reported unless a tracing context is active within the background task. Perhaps the FastAPI instrumentation should introduce a broader span for background task execution to ensure errors are reported within some span.

The solution is heavily inspired by the Datadog PR for this exact same problem https://github.com/DataDog/dd-trace-py/pull/3799

Before
image

After
image

Fixes #1684

## Type of change

# How Has This Been Tested?

- [X] Ran in production for months (Python 3.11 FastAPI)
- [X] Manually tested with FastApi(0.103.1) in matrix [Normal Request, Request with background execution] x [No Exceptions, Exception before response, Exception in background task*]
- [X] Ran ASGI instrumentation tests
- [X] Ran FastAPI instrumentation tests
- [X] Added additional tests to the ASGI instrumentation

- [X] No.

# Checklist:

- [ ] ~~Documentation has been updated~~

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
@@ -576,7 +576,7 @@ class OpenTelemetryMiddleware:
         if scope["type"] == "http":
             self.active_requests_counter.add(1, active_requests_count_attrs)
         try:
-            with trace.use_span(span, end_on_exit=True) as current_span:
+            with trace.use_span(span, end_on_exit=False) as current_span:
                 if current_span.is_recording():
                     for key, value in attributes.items():
                         current_span.set_attribute(key, value)
@@ -703,5 +703,8 @@ class OpenTelemetryMiddleware:
                         pass
 
                 await send(message)
+            if message["type"] == "http.response.body":
+                if not message.get("more_body", False):
+                    server_span.end()
 
         return otel_send
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
@@ -630,6 +630,8 @@ class OpenTelemetryMiddleware:
                         )
             if token:
                 context.detach(token)
+            if span.is_recording():
+                span.end()
 
     # pylint: enable=too-many-branches
 
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
@@ -655,8 +655,11 @@ class OpenTelemetryMiddleware:
     def _get_otel_send(
         self, server_span, server_span_name, scope, send, duration_attrs
     ):
+        expecting_trailers = False
+
         @wraps(send)
         async def otel_send(message):
+            nonlocal expecting_trailers
             with self.tracer.start_as_current_span(
                 " ".join((server_span_name, scope["type"], "send"))
             ) as send_span:
@@ -670,6 +673,8 @@ class OpenTelemetryMiddleware:
                         ] = status_code
                         set_status_code(server_span, status_code)
                         set_status_code(send_span, status_code)
+
+                        expecting_trailers = message.get("trailers", False)
                     elif message["type"] == "websocket.send":
                         set_status_code(server_span, 200)
                         set_status_code(send_span, 200)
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
```diff
@@ -99,6 +102,63 @@ async def simple_asgi(scope, receive, send):
...
+                ],
+            }
+        )
+        await send(
+            {
+                "type": "http.response.body",
+                "body": b"*",
+            }
+        )
+        time.sleep(_SIMULATED_BACKGROUND_TASK_EXECUTION_TIME_S)
```
aabmass, line 159: Is this a common idiom for background work in ASGI? The user could get the behavior this PR is adding today by using `asyncio.create_task()`.

aabmass (commented):
Thanks for the PR. Do you have any idea on the current state of OpenTelemetry [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.m) around this? I remember some effort to standardize when an HTTP span should be ended, but idk if it was for client or server.

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
aabmass: Thanks for the thorough answer. If this complies with semconv generally, it makes sense to me. The behavior may be slightly surprising to some users but I think it is semantically more meaningful.

Another option would be using span events to indicate when writing the body is complete without ending the span

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -693,5 +695,8 @@ async def otel_send(message):
                         pass
 
                 await send(message)
+            if message["type"] == "http.response.body":
```
aabmass, line 698: Small thing which a coworker pointed out; the request event could indicate that trailers are expected after the request body: https://asgi.readthedocs.io/en/latest/specs/www.html#response-start-send-event:~:text=If%20more_body%20is%20set%20to%20False%2C%20and%20the%20server%20is%20not%20expecting%20Response%20Trailers%20this%20will%20complete%20the%20response. In that case, I think you would want to wait for trailers to be sent before ending the span

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
aabmass: I think you've looked harder at the HTTP semconv than I have so LGTM

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
siminn-arnorgj: In my latest commit(6a2ff51) I added support for HTTP trailers according to the ASGI spec. The drawback of supporting this is of course having to maintain state throughout the request. I currently maintain the state through a nonlocal variable. I'm open to suggestions on more elegant ways to track the state as I don't find the current method very elegant.

However I do think some discussion is warranted around HTTP trailers because AFAIK there is no publicly available ASGI server nor framework which conforms to the spec / supports HTTP trailers. So I'm not exactly sure the instrumentation should support something that nobody seems to use at the cost of increasing complexity.

aabmass: The closure solution looks pretty clean to me, I don't have any issue with it. I agree this may be unimportant in practice, but nice to have for completeness since it's only a few extra lines

aabmass (approved):
Thanks for the PR, LGTM and I think this is the right direction

aabmass:
@open-telemetry/opentelemetry-python-contrib-maintainers can someone take a look and merge if it looks good?

lzchen:
@siminn-arnorgj 

Would you mind rebasing with the latest main?

lzchen (approved):

PR #1959, opened 2023-09-19
urllib/urllib3 - set metric descriptions per semantic conventions

### LINKED ISSUE #1958: `http.client.duration` metric description is incorrect
**Describe your environment** 
This is for the OpenTelemetry demo. This is the loadgenerator service in the demo that uses Locust (2.15.1) to generate a load for the demo application. It uses python 3.11-slim-buster as the base image for Python runtime. In code we leverage the `urllib3` instrumentation package with a call to `URLLib3Instrumentor().instrument()` at the top of the the locustfile.py

The `http.client.duration` metric has a description that does not conform with the OpenTelemetry semantic convention. This creates an issue when used with OpenTelemetry instrumented services that emit the same metric from a different language SDK (ie: Javascript). The OpenTelemetry Collector's, Prometheus exporter will specifically drop a metric if received that has the same name but a different description from a metric previously received.

**Steps to reproduce**
Run the demo and look at the OpenTelemetry collector logs to show an error on the Prometheus export. The error will be for the `http.client.duration` metric, indicating the Help text (description) does not match what was received prior for the same metric name. 

The description for the `http.client.duration` metric should be: `Measures the duration of outbound HTTP requests.`

This is defined in the OpenTelemetry [semantic convention](https://github.com/open-telemetry/semantic-conventions/blob/main/model/metrics/http.yaml#L157) for this metric.

The description for the `http.client.duration` metric is: `measures the duration outbound HTTP requests`

**Additional context**
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/e318c947a23152c8ff1700f0aad44261be0588cd/instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py#L176

puckpuck:
# Description

This sets the description for the `http.client.duration`, `http.client.request.size`, and `http.client.response.size` metrics per what is defined in the [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/model/metrics/http.yaml#L155C44-L179) for the metrics. This is done for both the `urllib` and `urllib3` instrumentation libraries.

When used with other services written in other languages that emit the same metrics, all services must emit the same metric metadata (name, description) when used in conjunction with a Prometheus backend. The OpenTelemetry Collector, Prometheus exporter, will drop metrics that match the name but not the description against already received metrics. We have faced this problem as part of the OpenTelemetry demo project.

Fixes #1958 

## Type of change

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
```diff
@@ -297,17 +297,17 @@ def _create_client_histograms(meter) -> Dict[str, Histogram]:
         MetricInstruments.HTTP_CLIENT_DURATION: meter.create_histogram(
             name=MetricInstruments.HTTP_CLIENT_DURATION,
```
lzchen, line 298: Would a rename of the name be needed as well since metadata is name + description?

puckpuck: The match is made on name, and the check is to ensure the description and data type conform to metrics already received with the same name. So only the description needs to be updated.

instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
```diff
@@ -297,17 +297,17 @@ def _create_client_histograms(meter) -> Dict[str, Histogram]:
...
             unit="ms",
-            description="measures the duration outbound HTTP requests",
+            description="Measures the duration of outbound HTTP requests.",
         ),
         MetricInstruments.HTTP_CLIENT_REQUEST_SIZE: meter.create_histogram(
```
ocelotl, line 302: I think this has been [renamed](https://github.com/open-telemetry/semantic-conventions/commit/9db3c5fdd448ad100c090191fdac0f067f9d1cf4#diff-08cdc1cb5ff57be3c20e0899dae7b45a13c834ea2dad6af35e86218e0e52b8b8R20) to [`request.body.size`](https://github.com/open-telemetry/semantic-conventions/blame/main/model/metrics/http.yaml#L170).

jeremydvoss: That does seem separate from changing the descriptions. Should we do that in another PR?

lzchen: This would be part of the new semantic convention migration and I don't think that should be part of this pr

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- `opentelemetry-instrumentation-urllib` Fix metric descriptions to match semantic conventions
+  ([#1959]((https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1959))
+- `opentelemetry-instrumentation-urllib3` Fix metric descriptions to match semantic conventions
```
lzchen, line 12: Nit: Duplicate?

lzchen (approved):

instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/__init__.py
puckpuck: Would you like me to change the metric names to match the new semconv in this PR, or will that be done in a separate PR?

shalevr (approved):

PR #1967, opened 2023-09-22
fix(auto-instrumentation): Python path no longer required

lperdereau:
# Description

When we implemented the gunicorn.py configuration with a post fork, we wanted to use OTEL auto-instrumentation easily.
But a bug appeared when using `sitecustomize`.

Here is an example of a gunicorn post fork to load auto-instrumentation.
```py
def post_fork(server, worker):
    from opentelemetry.instrumentation.auto_instrumentation import sitecustomize
```

if we run gunicorn like this
```sh
gunicorn my-app.wsgi -c gunicorn.py
```

An error appear :
```
  File "/usr/local/lib/python3.9/site-packages/gunicorn/arbiter.py", line 608, in spawn_worker
    self.cfg.post_fork(self, worker)
  File "gunicorn.py", line 15, in post_fork
    from opentelemetry.instrumentation.auto_instrumentation import sitecustomize
  File "/usr/local/lib/python3.9/site-packages/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py", line 44, in <module>
    initialize()
  File "/usr/local/lib/python3.9/site-packages/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py", line 32, in initialize
    environ["PYTHONPATH"], dirname(abspath(__file__)), pathsep
  File "/usr/local/lib/python3.9/os.py", line 679, in __getitem__
    raise KeyError(key) from None
KeyError: 'PYTHONPATH'
```

**PYTHONPATH should not be required to use this auto-instrument feature.**

### DIFF
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
@@ -28,9 +28,11 @@ logger = getLogger(__name__)
 
 def initialize():
     # prevents auto-instrumentation of subprocesses if code execs another python process
-    environ["PYTHONPATH"] = _python_path_without_directory(
-        environ.get("PYTHONPATH", ""), dirname(abspath(__file__)), pathsep
-    )
+
+    if environ.get("PYTHONPATH") is not None:
+        environ["PYTHONPATH"] = _python_path_without_directory(
+            environ["PYTHONPATH"], dirname(abspath(__file__)), pathsep
+        )
 
     try:
         distro = _load_distro()
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
@@ -29,7 +29,7 @@ logger = getLogger(__name__)
 def initialize():
     # prevents auto-instrumentation of subprocesses if code execs another python process
     environ["PYTHONPATH"] = _python_path_without_directory(
-        environ["PYTHONPATH"], dirname(abspath(__file__)), pathsep
+        environ.get("PYTHONPATH", ""), dirname(abspath(__file__)), pathsep
     )
 
     try:
--- a/opentelemetry-instrumentation/tests/auto_instrumentation/test_sitecustomize.py
+++ b/opentelemetry-instrumentation/tests/auto_instrumentation/test_sitecustomize.py
@@ -24,7 +24,7 @@ class TestSiteCustomize(TestCase):
     def test_unset(self):
         if environ.get("PYTHONPATH"):
             del environ["PYTHONPATH"]  # enforce remove key/value from environ
-        from opentelemetry.instrumentation.auto_instrumentation import (
+        from opentelemetry.instrumentation.auto_instrumentation import (  # noqa
             sitecustomize,
         )
 
@@ -32,7 +32,7 @@ class TestSiteCustomize(TestCase):
 
     @patch.dict("os.environ", {"PYTHONPATH": ""})
     def test_empty(self):
-        from opentelemetry.instrumentation.auto_instrumentation import (
+        from opentelemetry.instrumentation.auto_instrumentation import (  # noqa
             sitecustomize,
         )
 
@@ -40,7 +40,7 @@ class TestSiteCustomize(TestCase):
 
     @patch.dict("os.environ", {"PYTHONPATH": "abc"})
     def test_set(self):
-        from opentelemetry.instrumentation.auto_instrumentation import (
+        from opentelemetry.instrumentation.auto_instrumentation import (  # noqa
             sitecustomize,
         )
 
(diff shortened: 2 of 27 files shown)


lperdereau:
Hello @ocelotl,

Is it possible to have a review on this tiny patch ?
Or is it possible to find out the procedure for obtaining a review?

Regards

ocelotl:
reviewing...

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
```diff
@@ -29,7 +29,7 @@
 def initialize():
     # prevents auto-instrumentation of subprocesses if code execs another python process
     environ["PYTHONPATH"] = _python_path_without_directory(
-        environ["PYTHONPATH"], dirname(abspath(__file__)), pathsep
+        environ.get("PYTHONPATH", ""), dirname(abspath(__file__)), pathsep
```
ocelotl, line 32: Ok, pleaase add an unittest that calls `initialize`.

ocelotl:
This implementation has an issue that is not evident when running the tests because the order of the tests makes them pass. If you run: `tox -e py311-test-opentelemetry-instrumentation -- -k test_unset` this happens:

```
auto_instrumentation/test_run.py::TestRun::test_unset PASSED                                                                                                                             [ 50%] auto_instrumentation/test_sitecustomize.py::TestSiteCustomize::test_unset FAILED                                                                                                         [100%] 
                                                                                                                                                                                                =========================================================================================== FAILURES ===========================================================================================
_________________________________________________________________________________ TestSiteCustomize.test_unset _________________________________________________________________________________
                                                                                                                                                                                                
self = <test_sitecustomize.TestSiteCustomize testMethod=test_unset>                                                                                                                                                                                                                                                                                                                             
    def test_unset(self):                                                                                                                                                                       
        if environ.get("PYTHONPATH"):                                                                                                                                                           
            del environ["PYTHONPATH"]  # enforce remove key/value from environ                                                                                                                  
        from opentelemetry.instrumentation.auto_instrumentation import sitecustomize                                                                                                            
>       self.assertEqual(environ.get("PYTHONPATH"), None)                                                                                                                                       
E       AssertionError: '' != None                                                                                                                                                              
                                                                                                                                                                                                
auto_instrumentation/test_sitecustomize.py:28: AssertionError                                                                                                                                   
```

aabmass (approved):

lzchen (approved):

aabmass (approved):

aabmass:
> because the order of the tests makes them pass.

Is this fixed now?

ocelotl:
> > because the order of the tests makes them pass.
> 
> Is this fixed now?

yes, it is fixed :v:

opentelemetry-instrumentation/tests/auto_instrumentation/test_sitecustomize.py
```diff
@@ -0,0 +1,47 @@
...
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.load_configurators.stop()
+        cls.load_instrumentors.stop()
+
+    def test_unset(self):
+        if environ.get("PYTHONPATH"):
+            del environ["PYTHONPATH"]  # enforce remove key/value from environ
+        from opentelemetry.instrumentation.auto_instrumentation import (  # noqa
```
mariojonke, line 27: i think this will not have any effect if the sitecustomize module was already imported once previously (e.g. by a test from another module or by one of the tests below).
after the first import a module is put into `sys.modules` any subsequent imports will not run the module global code again.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
```diff
@@ -28,9 +28,11 @@
 
 def initialize():
     # prevents auto-instrumentation of subprocesses if code execs another python process
-    environ["PYTHONPATH"] = _python_path_without_directory(
-        environ["PYTHONPATH"], dirname(abspath(__file__)), pathsep
-    )
+
+    if environ.get("PYTHONPATH") is not None:
```
jeremydvoss, line 32: "if "PYTHONPATH" in environ" is the norm

ocelotl (commented):
Left a comment

opentelemetry-instrumentation/tests/auto_instrumentation/test_sitecustomize.py
lperdereau: I think we can try an import only for a test without puting it into `sys.module` or remove the import at the end of the test.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py
lperdereau: In many environment PYTHONPATH isn't set by default, this cause many error to start quickly to use the lib.

opentelemetry-instrumentation/tests/auto_instrumentation/test_sitecustomize.py
ocelotl: @lperdereau ping

Issue #1986, opened 2023-10-03
Prototype HTTP Otel Semantic convention stability opt-in


lzchen:
As part of the effort to [mark http semantic conventions as stable](https://github.com/open-telemetry/opentelemetry-specification/issues/3279) Python has volunteered to provide a http client and server instrumentation prototype. These prototypes will implement the [migration plan](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/http#semantic-conventions-for-http) for http instrumentations to be updated to follow the latest semantic conventions. Since [v.1.21.0](https://github.com/open-telemetry/opentelemetry-python/pull/3251) has been merged in the core repo, we can begin implementing the migration plan with two instrumentations. The work involves:

1. Deriving a mechanism in which we will control the opt-in
2. Prototyping a http client instrumentation
3. Prototyping a http server instrumentation

Part of https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1768

lzchen:
Succeeded by https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1768

lzchen:
Superseded by: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2453

PR #1987, opened 2023-10-03
Implement Otel semantic convention stability opt-in

lzchen:
Part of https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1986

This will cover point number (1) in the linked issue. Please review as I have not added tests yet. Just wanted to see if this a good approach.


opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -152,3 +154,57 @@ def _python_path_without_directory(python_path, directory, path_separator):
         "",
         python_path,
     )
+
+
+_OTEL_SEMCONV_STABILITY_OPT_IN_KEY = "OTEL_SEMCONV_STABILITY_OPT_IN"
```
jeremydvoss, line 159: Is this env var named used prior across languages? If not, we should probably add *_PYTHON_*

lzchen: Good question. I will bring this up with the TC.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -152,3 +154,57 @@ def _python_path_without_directory(python_path, directory, path_separator):
...
+class _OpenTelemetryStabilityMode:
+    # http - emit the new, stable HTTP and networking conventions ONLY
+    HTTP = "http"
+    # http/dup - emit both the old and the stable HTTP and networking conventions
+    HTTP_DUP = "http/dup"
+    # default - continue emitting old experimental HTTP and networking conventions
+    DEFAULT = "default"
+
+
+class _OpenTelemetrySemanticConventionStability:
```
jeremydvoss, line 175: Let's add tests in this package so the burden isn't on each instrumentation that will use this.

lzchen: I didn't add tests yet because I wanted to see what people thought about the approach. If we get a general concensus I will start adding tests.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
ocelotl, line 159:
```suggestion
_OTEL_SEMCONV_STABILITY_OPT_IN_KEY = "OTEL_SEMCONV_STABILITY_OPT_IN_KEY"
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -152,3 +154,57 @@ def _python_path_without_directory(python_path, directory, path_separator):
...
+
+    @classmethod
+    def _initialize(cls):
+        with _OpenTelemetrySemanticConventionStability._lock:
+            if not _OpenTelemetrySemanticConventionStability._initialized:
+                # Users can pass in comma delimited string for opt-in options
+                # Only values for http stability are supported for now
+                opt_in = os.environ.get(_OTEL_SEMCONV_STABILITY_OPT_IN_KEY, "")
+                opt_in_list = []
+                if opt_in:
```
ocelotl, line 188: It seems like we only support one final value for `http_opt_in`. Why support the case where the user can have multiple values in a comma separated string?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -152,3 +154,57 @@ def _python_path_without_directory(python_path, directory, path_separator):
...
+
+_OTEL_SEMCONV_STABILITY_OPT_IN_KEY = "OTEL_SEMCONV_STABILITY_OPT_IN"
+
+
+class _OpenTelemetryStabilitySignalType:
+    HTTP = "http"
+
+
+class _OpenTelemetryStabilityMode:
+    # http - emit the new, stable HTTP and networking conventions ONLY
```
ocelotl, line 167: I think it is better to use semantic convention version numbers here, for example, instead of `http`, `HTTP-1.21`, instead of `default` `HTTP-1.20`.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -152,3 +154,57 @@ def _python_path_without_directory(python_path, directory, path_separator):
...
+
+
+_OTEL_SEMCONV_STABILITY_OPT_IN_KEY = "OTEL_SEMCONV_STABILITY_OPT_IN"
+
+
+class _OpenTelemetryStabilitySignalType:
+    HTTP = "http"
+
+
+class _OpenTelemetryStabilityMode:
```
ocelotl, line 166: This can be an `Enum`.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
lzchen: As defined in the [spec](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/http#semantic-conventions-for-http), the env var is defined to be a comma separated list that could support multiple values. It is designed so that this mechanism can be used in the future for opt-in for other types of semantic conventions as well (db, messaging, etc.)

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
jeremydvoss: This existing env var is in line with the [Migration Plan](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/http#semantic-conventions-for-http)

Is this out of date?

jeremydvoss: Shown in [migration plan](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/http#semantic-conventions-for-http)

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
jeremydvoss: Made enum

jeremydvoss (approved):

Issue #2000, opened 2023-10-12
Implement HTTP request retries and redirects for instrumentations


lzchen:
https://github.com/open-telemetry/semantic-conventions/blob/v1.21.0/docs/http/http-spans.md#http-request-retries-and-redirects

PR #2002, opened 2023-10-13
Requests instrumentation http semantic convention migration

lzchen:
Part https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1986

Does not support redirect/retries. Tracking [issue](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2000)

Opt-in mechanism for certain attributes for new semconvs don't exist yet and will be implemented in the future. (ie. url_scheme)

Tracking issue for updating semconv package to 1.23.X: https://github.com/open-telemetry/opentelemetry-python/issues/3538

Sample below:

```python
import os
import requests
from opentelemetry.instrumentation.requests import RequestsInstrumentor

# Specify "http", "http/dup" or ""
os.environ["OTEL_SEMCONV_STABILITY_OPT_IN"] = "http/dup"

# Enable instrumentation in the requests library.
RequestsInstrumentor().instrument()

trace.get_tracer_provider().add_span_processor(span_processor)

with tracer.start_as_current_span("parent"):
    response = requests.get("https://example.com/", timeout=5)

input()
```

Old semconv with envvar set to "default" or "":
```python
    "attributes": {
        "http.method": "GET",
        "http.url": "https://example.com/",
        "http.status_code": 200
    },
```

New semconv with envvar set to "http":
```python
    "attributes": {
        "http.request.method": "GET",
        "url.full": "https://example.com/",
        "server.address": "example.com",
        "network.peer.address": "example.com",
        "http.response.status_code": 200,
        "net.protocol.version": "1.1"
    },
```

Both semconv with envvar set to "http/dup":
```python
    "attributes": {
        "http.method": "GET",
        "http.request.method": "GET",
        "http.url": "https://example.com/",
        "url.full": "https://example.com/",
        "http.host": "example.com",
        "server.address": "example.com",
        "network.peer.address": "example.com",
        "http.status_code": 200,
        "http.response.status_code": 200,
        "http.flavor": "1.1",
        "net.protocol.version": "1.1"
    },
```


lzchen:
@trask 

Marking as draft for prototyping reasons.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -152,3 +156,123 @@ def _python_path_without_directory(python_path, directory, path_separator):
...
+_OTEL_HTTP_SEMCONV_MIGRATION_MAPPING = {
+    SpanAttributes.HTTP_METHOD: SpanAttributes.HTTP_REQUEST_METHOD,
+    SpanAttributes.HTTP_STATUS_CODE: SpanAttributes.HTTP_RESPONSE_STATUS_CODE,
+    SpanAttributes.HTTP_SCHEME: SpanAttributes.URL_SCHEME,
+    SpanAttributes.HTTP_URL: SpanAttributes.URL_FULL,
+    SpanAttributes.HTTP_REQUEST_CONTENT_LENGTH: SpanAttributes.HTTP_REQUEST_BODY_SIZE,
+    SpanAttributes.HTTP_RESPONSE_CONTENT_LENGTH: SpanAttributes.HTTP_RESPONSE_BODY_SIZE,
+    SpanAttributes.NET_HOST_NAME: SpanAttributes.SERVER_ADDRESS,
+    SpanAttributes.NET_HOST_PORT: SpanAttributes.SERVER_PORT,
+    SpanAttributes.NET_SOCK_HOST_ADDR: SpanAttributes.SERVER_SOCKET_ADDRESS,
```
lmolkova, line 173: will need to change to `SERVER_SOCKET_ADDRESS|PORT` to `network.peer.address|port`, bit it's minor

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -152,3 +156,123 @@ def _python_path_without_directory(python_path, directory, path_separator):
...
+    SpanAttributes.NET_SOCK_HOST_PORT: SpanAttributes.SERVER_SOCKET_PORT,
+    SpanAttributes.NET_TRANSPORT: SpanAttributes.NETWORK_TRANSPORT,
+    SpanAttributes.NET_PROTOCOL_NAME: SpanAttributes.NETWORK_PROTOCOL_NAME,
+    SpanAttributes.NET_PROTOCOL_VERSION: SpanAttributes.NETWORK_PROTOCOL_VERSION,
+    SpanAttributes.NET_SOCK_FAMILY: SpanAttributes.NETWORK_TYPE,
+    SpanAttributes.NET_PEER_IP: SpanAttributes.CLIENT_SOCKET_ADDRESS,
```
lmolkova, line 179: this mapping (if necessary) should depend on the span kind - it maps to `client` on server and to `server` on client.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -152,3 +156,123 @@ def _python_path_without_directory(python_path, directory, path_separator):
...
+    SpanAttributes.NET_HOST_IP: SpanAttributes.SERVER_SOCKET_ADDRESS,
```
lmolkova, line 180: should map to `network.local.address` (not in the released semconv yet)

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -287,8 +315,11 @@ def _instrument(self, **kwargs):
         )
         duration_histogram = meter.create_histogram(
```
lmolkova, line 316: the metric has been renamed and should also be emitted depending on the feature flag (the old one is `http.client.duration` with `ms` unit and the new one is  `http.client.request.duration` with `s` unit)

lmolkova:
Thanks a lot @lzchen!

A few follow-up questions:
- no retry tracking: how hard do you think this would be? Any blockers you envision?
- can we add prototype for `error.type` as well? 
- span names were changed

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
```diff
@@ -196,9 +196,10 @@ def sanitize_method(method: Optional[str]) -> Optional[str]:
     method = method.upper()
     if (environ.get(OTEL_PYTHON_INSTRUMENTATION_HTTP_CAPTURE_ALL_METHODS) or
         # Based on https://www.rfc-editor.org/rfc/rfc7231#section-4.1 and https://www.rfc-editor.org/rfc/rfc5789#section-2.
+        # Defined list: https://github.com/open-telemetry/semantic-conventions/blob/v1.21.0/docs/http/http-spans.md#common-attributes
         method in ["GET", "HEAD", "POST", "PUT", "DELETE", "CONNECT", "OPTIONS", "TRACE", "PATCH"]):
         return method
-    return "UNKNOWN"
+    return "_OTHER"
```
lmolkova, line 202: we also need to populate `http.request.method_original` but only on spans

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
jeremydvoss: Got it. Currently working on figuring out how this can work with the HTTP_DUP case.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
jeremydvoss: Doesn't this conflict with this commit? It seems net.* was changed to server.* and client.* Was it changed back to network.peer and network.local?
https://github.com/open-telemetry/semantic-conventions/commit/5b8861e8c54bc42593952faca3b9b8332cfb2d22

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
jeremydvoss: Similar to above comment. It seems these have changed multiple times. I would like to figure out where the source of confusion was before changing leighton's code

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
lmolkova: the schema transformation does not capture all the changes (some of them cannot be expressed in transformation), so it's best to use [changelog](https://github.com/open-telemetry/semantic-conventions/blob/main/CHANGELOG.md) or [the spec](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md) as a source of truth.

In any case, this change seems trivial.

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
jeremydvoss: This will require some refactoring. Is this new to the HTTP sem conv, or should we do this regardless of the stability mode?

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
```diff
@@ -196,9 +196,10 @@ def sanitize_method(method: Optional[str]) -> Optional[str]:
     method = method.upper()
     if (environ.get(OTEL_PYTHON_INSTRUMENTATION_HTTP_CAPTURE_ALL_METHODS) or
```
jeremydvoss, line 197: May want to allow people to override the list.

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
jeremydvoss: Confirmed this is only for new stable HTTP

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
jeremydvoss: Got it. It seems this was just misesd because it was a newer change. The [instances of server.socket.address ](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md?plain=1#L429)in the current sem conv are just mistakes. Will make this change

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
jeremydvoss: Same as above. Will impliment.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
jeremydvoss: Will do.

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
jeremydvoss: Done in latest push

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
jeremydvoss: Done in latest push

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
jeremydvoss: Done in latest push

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
jeremydvoss: Done in latest push

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -183,15 +212,28 @@ def get_or_create_headers():
                 result = getattr(exc, "response", None)
             finally:
                 elapsed_time = max(
-                    round((default_timer() - start_time) * 1000), 0
+                    round((default_timer() - start_time)), 0
```
lmolkova, line 215: nit: not sure why old duration needed to be rounded, but the new one (in seconds) should not be for sure.

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -183,15 +212,28 @@ def get_or_create_headers():
...
+                    }
+                    additional_span_attributes = _get_attributes_based_on_stability_mode(
+                        sem_conv_opt_in_mode,
+                        additional_span_attributes,
                     )
+                    version = ""
+                    if result.raw is not None:
+                        version = getattr(result.raw, "version", None)
+                        if version:
+                            version = "1.1" if version == 11 else "1.0"
```
lmolkova, line 232: nit: no HTTP/2 or /3 support?

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -200,17 +242,23 @@ def get_or_create_headers():
...
                 ] = result.status_code
 
-                if result.raw is not None:
-                    version = getattr(result.raw, "version", None)
-                    if version:
-                        metric_labels[SpanAttributes.HTTP_FLAVOR] = (
-                            "1.1" if version == 11 else "1.0"
-                        )
+                metric_labels[
+                    SpanAttributes.NETWORK_PROTOCOL_NAME
```
lmolkova, line 246: it's opt-in on metrics

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
jeremydvoss: Got it. This will require some refactoring of the way I made the duration histogram array

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
jeremydvoss: NETWORK_PROTOCOL_VERSION is already a part of the mapping and so, should be set as part of _get_attributes_based_on_stability_mode. My guess is that this is covering some edge case. Will explore and if needed talk to Leighton about what the intent was when he returns.

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
jeremydvoss: Got it. So, we should set the NETWORK_PROTOCOL_NAME network label only for http and http_dup?

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
lzchen: Fixed.

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
lzchen: Requests library only supports HTTP/1 https://stackoverflow.com/questions/34570991/http-2-request-in-python-2-7#:~:text=As%20others%20mentioned%20in%20the%20comments%20to%20the,of%20is%20hyper%2C%20which%20quoting%20from%20the%20docs%3A

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
lzchen: Removing and leaving opt-in as to-do

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -0,0 +1,209 @@
...
+
+_client_duration_attrs_new = [
+    _SPAN_ATTRIBUTES_ERROR_TYPE,
+    SpanAttributes.HTTP_REQUEST_METHOD_ORIGINAL,
+    SpanAttributes.HTTP_REQUEST_METHOD,
+    SpanAttributes.HTTP_RESPONSE_STATUS_CODE,
+    SpanAttributes.NET_PROTOCOL_VERSION,
+    SpanAttributes.SERVER_ADDRESS,
+    SpanAttributes.SERVER_PORT,
+    SpanAttributes.URL_SCHEME,
```
lmolkova, line 44: it's now opt-in

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -0,0 +1,209 @@
...
+        set_string_attribute(result, SpanAttributes.NET_PEER_NAME, peer_name)
+    if _report_new(sem_conv_opt_in_mode):
+        set_string_attribute(result, SpanAttributes.SERVER_ADDRESS, peer_name)
+
+
+def _set_http_port(result, port, sem_conv_opt_in_mode):
+    if _report_old(sem_conv_opt_in_mode):
+        set_string_attribute(result, SpanAttributes.NET_PEER_PORT, port)
+    if _report_new(sem_conv_opt_in_mode):
+        set_string_attribute(result, SpanAttributes.SERVER_PORT, port)
```
lmolkova, line 125:
```suggestion
        set_int_attribute(result, SpanAttributes.SERVER_PORT, port)
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -0,0 +1,209 @@
...
+            result, SpanAttributes.HTTP_RESPONSE_STATUS_CODE, code
+        )
+
+
+def _set_http_network_protocol_version(result, version, sem_conv_opt_in_mode):
+    if _report_old(sem_conv_opt_in_mode):
+        set_string_attribute(result, SpanAttributes.HTTP_FLAVOR, version)
+    if _report_new(sem_conv_opt_in_mode):
+        set_string_attribute(
+            result, SpanAttributes.NET_PROTOCOL_VERSION, version
```
lmolkova, line 142:
```suggestion
            result, SpanAttributes.NETWORK_PROTOCOL_VERSION, version
```

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -182,35 +230,68 @@ def get_or_create_headers():
...
                     response_hook(span, request, result)
 
-            duration_histogram.record(elapsed_time, attributes=metric_labels)
+            if exception is not None and _report_new(sem_conv_opt_in_mode):
+                span.set_attribute(
+                    _SPAN_ATTRIBUTES_ERROR_TYPE, type(exception).__name__
+                )
+                metric_labels[_SPAN_ATTRIBUTES_ERROR_TYPE] = type(
+                    exception
+                ).__name__
```
lmolkova, line 278:
```suggestion
                ).__qualname__
```

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -182,35 +230,68 @@ def get_or_create_headers():
...
+                for key, val in span_attributes.items():
+                    span.set_attribute(key, val)
 
                 if callable(response_hook):
                     response_hook(span, request, result)
 
-            duration_histogram.record(elapsed_time, attributes=metric_labels)
+            if exception is not None and _report_new(sem_conv_opt_in_mode):
+                span.set_attribute(
+                    _SPAN_ATTRIBUTES_ERROR_TYPE, type(exception).__name__
```
lmolkova, line 274:
```suggestion
                    _SPAN_ATTRIBUTES_ERROR_TYPE, type(exception).__qualname__
```

instrumentation/opentelemetry-instrumentation-requests/tests/test_requests_integration.py
```diff
@@ -133,6 +152,63 @@ def test_basic(self):
...
+        self.assertEqual(
+            span.attributes,
+            {
+                SpanAttributes.HTTP_METHOD: "GET",
+                SpanAttributes.HTTP_REQUEST_METHOD: "GET",
+                SpanAttributes.HTTP_URL: self.URL,
+                SpanAttributes.URL_FULL: self.URL,
+                SpanAttributes.HTTP_HOST: self.HOST,
+                SpanAttributes.SERVER_ADDRESS: self.HOST,
+                _SPAN_ATTRIBUTES_NETWORK_PEER_ADDRESS: "mock",
```
lmolkova, line 198: `server.port` is now required even if 80/443

instrumentation/opentelemetry-instrumentation-requests/tests/test_requests_integration.py
```diff
@@ -214,6 +290,51 @@ def test_not_foundbasic(self):
...
+            url_404,
+            status=404,
+        )
+        result = self.perform_request(url_404)
+        self.assertEqual(result.status_code, 404)
+
+        span = self.assert_span()
+
+        self.assertEqual(
+            span.attributes.get(SpanAttributes.HTTP_RESPONSE_STATUS_CODE), 404
```
lmolkova, line 306: also `error.type` should be equal to `"404"`

instrumentation/opentelemetry-instrumentation-requests/tests/test_requests_integration.py
```diff
@@ -214,6 +290,51 @@ def test_not_foundbasic(self):
...
+        result = self.perform_request(url_404)
+        self.assertEqual(result.status_code, 404)
+
+        span = self.assert_span()
+
+        self.assertEqual(
+            span.attributes.get(SpanAttributes.HTTP_STATUS_CODE), 404
+        )
+        self.assertEqual(
+            span.attributes.get(SpanAttributes.HTTP_RESPONSE_STATUS_CODE), 404
```
lmolkova, line 330: please also check `error.type`

instrumentation/opentelemetry-instrumentation-requests/tests/test_requests_integration.py
```diff
@@ -368,6 +489,27 @@ def test_requests_exception_without_response(self, *_, **__):
...
+
+        span = self.assert_span()
+        self.assertEqual(
+            span.attributes,
+            {
+                SpanAttributes.HTTP_REQUEST_METHOD: "GET",
+                SpanAttributes.URL_FULL: self.URL,
+                SpanAttributes.SERVER_ADDRESS: self.HOST,
+                _SPAN_ATTRIBUTES_NETWORK_PEER_ADDRESS: "mock",
+                _SPAN_ATTRIBUTES_ERROR_TYPE: "RequestException",
```
lmolkova, line 508: exception type should be fully qualified

instrumentation/opentelemetry-instrumentation-requests/tests/test_requests_integration.py
```diff
@@ -368,6 +489,27 @@ def test_requests_exception_without_response(self, *_, **__):
...
+        with self.assertRaises(requests.RequestException):
+            self.perform_request(self.URL)
+
+        span = self.assert_span()
+        self.assertEqual(
+            span.attributes,
+            {
+                SpanAttributes.HTTP_REQUEST_METHOD: "GET",
+                SpanAttributes.URL_FULL: self.URL,
+                SpanAttributes.SERVER_ADDRESS: self.HOST,
```
lmolkova, line 506: let's also check `server.port`

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -0,0 +1,209 @@
...
+    SpanAttributes.NET_PEER_PORT,
+    SpanAttributes.NET_PEER_NAME,
+    SpanAttributes.HTTP_METHOD,
+    SpanAttributes.HTTP_FLAVOR,
+    SpanAttributes.HTTP_SCHEME,
+]
+
+_client_duration_attrs_new = [
+    _SPAN_ATTRIBUTES_ERROR_TYPE,
+    SpanAttributes.HTTP_REQUEST_METHOD_ORIGINAL,
```
lmolkova, line 38: this attributes should be set on spans only

instrumentation/opentelemetry-instrumentation-requests/tests/test_requests_integration.py
lzchen: Changed to `__qualname__` in code.

lzchen:
> Thanks a lot @lzchen!
> 
> A few follow-up questions:
> 
> * no retry tracking: how hard do you think this would be? Any blockers you envision?
> * can we add prototype for `error.type` as well?
> * span names were changed

@lmolkova 

1. Are there other cases in which we expect retries to happen automatically with the underlying instrumented library besides redirects? The requests library handles [redirects](https://docs.python-requests.org/en/latest/user/quickstart/#redirection-and-history) automatically by default.
     - What do we expect to happen here? The count being stamped on one span or multiple spans created for the initial call and then the redirect?
     - For cases like 503 where the library does not handle this automatically, are multiple spans expected to be created by the instrumentation library or by the user?
    If we are just stamping the one span, I don't think it will be that difficult to implement. Just needs clarification on behavior.
2. Done in recent commits
3. Are you referring to the change of "HTTP {METHOD NAME} to just {METHOD NAME}" ?

lmolkova (approved):
LGTM!

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -182,35 +232,78 @@ def get_or_create_headers():
...
+                        metric_labels, result.status_code, sem_conv_opt_in_mode
                     )
-
-                metric_labels[
-                    SpanAttributes.HTTP_STATUS_CODE
-                ] = result.status_code
+                    status_code = http_status_to_status_code(
+                        result.status_code
+                    )
+                    span.set_status(status_code)
```
jeremydvoss, line 252: Status code is separate from attributes?

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -182,35 +232,78 @@ def get_or_create_headers():
                 exception = exc
                 result = getattr(exc, "response", None)
             finally:
-                elapsed_time = max(
-                    round((default_timer() - start_time) * 1000), 0
-                )
+                elapsed_time = max(default_timer() - start_time, 0)
```
jeremydvoss, line 235: So, elapsed time is always seconds but then multiplied for the old histrogram. I like this approach

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -0,0 +1,217 @@
...
+
+_client_duration_attrs_new = [
+    _SPAN_ATTRIBUTES_ERROR_TYPE,
+    SpanAttributes.HTTP_REQUEST_METHOD,
+    SpanAttributes.HTTP_RESPONSE_STATUS_CODE,
+    SpanAttributes.NETWORK_PROTOCOL_VERSION,
+    SpanAttributes.SERVER_ADDRESS,
+    SpanAttributes.SERVER_PORT,
+    # TODO: Support opt-in for scheme in new semconv
+    # SpanAttributes.URL_SCHEME,
```
jeremydvoss, line 45: Shouldn't we just create our own constant in the meantime?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -0,0 +1,217 @@
...
+    if _report_new(sem_conv_opt_in_mode):
+        set_string_attribute(result, SpanAttributes.URL_FULL, url)
+
+
+def _set_http_scheme(result, scheme, sem_conv_opt_in_mode):
+    if _report_old(sem_conv_opt_in_mode):
+        set_string_attribute(result, SpanAttributes.HTTP_SCHEME, scheme)
+    # TODO: Support opt-in for scheme in new semconv
+    # if _report_new(sem_conv_opt_in_mode):
+    #     set_string_attribute(result, SpanAttributes.URL_SCHEME, scheme)
```
jeremydvoss, line 105: Similar to above comment. Shouldn't we set this even if that constant doesn't exist so as to follow the new semconv?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
lzchen: The TODO refers to the feature for opting into enabling these attributes, not that the attributes don't exist yet.

jeremydvoss (approved):

PR #2010, opened 2023-10-20
issue 1918: Ensure string values for span status and attribute values

### LINKED ISSUE #1918: pymongo Invalid type in attribute sequence when using sessions
#1555 introduced a bug when using pymongo sessions, the instrumentor now outputs the message

```
Invalid type dict in attribute value sequence. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or None
```

while exiting the session.
After some debugging, I realized that this id triggered by the "endSessions" command, that triggers the CommandStartedEvent with a SON command like this:

```
SON([('endSessions', [{'id': Binary(b'U\xf4\x19x\xdd\xd8FK\x80\x13\x97\xcc_U\xd5L', 4)}]), ('$db', 'admin')])
```

and the [`collection` variable](https://github.com/avzis/opentelemetry-python-contrib/blob/b513d1f9102557d94f8b920baf50ec9a647e6ed9/instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py#L132) becomes a list of dictionaries instead of the expected string.

**Describe your environment**

- python 3.10
- windows 10, but also on linux/docker
- pymongo 4.4.1
- opentelemetry-instrumentation-pymongo 0.40b0
- opentelemetry 1.19.0

**Steps to reproduce**

```python
from opentelemetry.instrumentation.pymongo import PymongoInstrumentor
from pymongo import MongoClient

provider = TracerProvider()

PymongoInstrumentor().instrument()
client = MongoClient()
session = client.start_session()
RECORD = {"test": "123"}
db = client["MongoDB_Database"]
collection = db["MongoDB_Collection"]
collection.find_one(RECORD)
session.end_session()
client.close()
```

[`span.set_attribute(SpanAttributes.DB_MONGODB_COLLECTION)`](https://github.com/avzis/opentelemetry-python-contrib/blob/b513d1f9102557d94f8b920baf50ec9a647e6ed9/instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py#L142-L145) should not be called if the `collection` variable is not a string

The `set_attribute`  method is called and the message above is output in the console.

**Additional context**
Another solution would be to check for the pymongo command names that doesn't have/need a collection name

cmpilato:
# Description

This aims to fix two cases where data types were deemed invalid, resulting in telemetry not being transmitted to our collector.  Specifically:
* don't set the `SpanAttributes.DB_MONGODB_COLLECTION` attribute unless the collection is already a string value.
* `str()`-ify the event failure details

Fixes #1918

## Type of change

# How Has This Been Tested?

So far, only adhoc testing has been done.  Without these changes, we fail to receive telemetry data and see these logged warnings:

```
opentelemetry.trace.status.py, line 54 "Invalid status description type, expected str"
opentelemetry.attributes._init_.py, line 70 "Invalid type dict in attribute value sequence. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or None"
```

With the changes, the warnings are gone and telemetry data flows as expected.  We are using pymongo 4.5.0.

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
@@ -139,7 +139,7 @@ class CommandTracer(monitoring.CommandListener):
                 )
                 span.set_attribute(SpanAttributes.DB_NAME, event.database_name)
                 span.set_attribute(SpanAttributes.DB_STATEMENT, statement)
-                if collection:
+                if collection and isinstance(collection, str):
                     span.set_attribute(
                         SpanAttributes.DB_MONGODB_COLLECTION, collection
                     )
@@ -193,7 +193,7 @@ class CommandTracer(monitoring.CommandListener):
         if span is None:
             return
         if span.is_recording():
-            span.set_status(Status(StatusCode.ERROR, event.failure))
+            span.set_status(Status(StatusCode.ERROR, str(event.failure)))
             try:
                 self.failed_hook(span, event)
             except (


xrmx:
@cmpilato could you please rebase this?

cmpilato:
@xrmx Rebased as requested.

xrmx (commented):
LGTM, would be nice to add tests for checking that the code now works where previously it failed. Also would be nice to have a changelog entry.

cmpilato:
@xrmx I've added a ChangeLog entry, but unfortunately I lack the overall OTEL domain knowledge (and time) to consider a regression test.  Maybe somebody could use the issue #1918 reproduction recipe to whip up such a contribution for at least the "collection isn't a string" problem?

PR #2020, opened 2023-10-26
Remove URL credentials (httpx integration)

ods:
# Description

This PR fixes issue #367 for `httpx` instrumentation. This would ensure that URLs passed in the form of `https://username:password@example.com/` would result in a http span attribute value of `https://example.com/`.

Based on https://github.com/open-telemetry/opentelemetry-python-contrib/pull/538

## Type of change

# How Has This Been Tested?

Additional tests are provided:

- [x] `test_httpx_integration.py::TestSyncIntegration::test_credential_removal`
- [x] `test_httpx_integration.py::TestAsyncIntegration::test_credential_removal`

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
@@ -206,6 +206,7 @@ from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.trace import SpanKind, TracerProvider, get_tracer
 from opentelemetry.trace.span import Span
 from opentelemetry.trace.status import Status
+from opentelemetry.util.http import remove_url_credentials
 
 _logger = logging.getLogger(__name__)
 
@@ -269,7 +270,7 @@ def _extract_parameters(args, kwargs):
         # In httpx >= 0.20.0, handle_request receives a Request object
         request: httpx.Request = args[0]
         method = request.method.encode()
-        url = request.url
+        url = remove_url_credentials(str(request.url))
         headers = request.headers
         stream = request.stream
         extensions = request.extensions
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `opentelemetry-instrumentation-httpx` Remove URL credentials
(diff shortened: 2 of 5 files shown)


ods:
@ocelotl @lzchen Could you please take a look / assign reviewers?

ocelotl (approved):

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -269,7 +270,7 @@ def _extract_parameters(args, kwargs):
         # In httpx >= 0.20.0, handle_request receives a Request object
         request: httpx.Request = args[0]
         method = request.method.encode()
-        url = request.url
+        url = remove_url_credentials(str(request.url))
```
anuraaga, line 273: Hello - when updating to latest instrumentation, I noticed that `url` stopped being a `httpx.URL` and is a `str` now because of the conversion here. Was it intended? If so, the type is still the old one, though I think `httpx.URL` would be more useful if possible

https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2020/files#diff-eed55d62a8d5f01b055c1bb2d63934c48cfd767416c03826f2c19d4442a76e7bR227

ods: No, it wasn't intentional. Moreover, it violates [current type annotation](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py#L254). But I don't understand how it may break something if the result of this function always goes through `_prepare_attributes`  function which [converts it to `str`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py#L256) anyway.

samypr100: @ocelotl This is definitely an breaking issue preventing upgrades on my end to v44. We rely on request.url in my request hooks being httpx.URL type

`def request_hook(span: Span, request: Request) -> None` request.url is now a string and it used to be a httpx.URL < v44.

The solution is to wrap it back to httpx.URL:

`httpx.URL(remove_url_credentials(str(request.url)))`

Issue #2022, opened 2023-10-26
celery instrumentation: Attribute 'celery.timelimit' mixes types str and int in attribute value sequence


maxkomarychev:
**Describe your environment** Describe any aspect of your environment relevant to the problem, including your Python version, [platform](https://docs.python.org/3/library/platform.html), version numbers of installed dependencies, information about your cloud hosting provider, etc. If you're reporting a problem with a specific version of a library in this repo, please check whether the problem has been fixed on main.

- python 3.9.9
- celery 5.3.4
- celery 4.4.2

**Steps to reproduce**
Describe exactly how to reproduce the error. Include a code sample if applicable.

WIP

**What is the expected behavior?**
What did you expect to see?

To not see a warning

**What is the actual behavior?**
What did you see instead?

I am observing a warning `Attribute 'celery.timelimit' mixes types str and int in attribute value sequence` from OTEL

**Additional context**

I have debugged and logged some values from https://github.com/open-telemetry/opentelemetry-python-contrib/blob/6f3aead436911d84c59fc822cbe2fcae32002316/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/utils.py#L65-L69 I see that `values` is actually `[None, 3600]` (None and a number) which makes this code https://github.com/open-telemetry/opentelemetry-python-contrib/blob/6f3aead436911d84c59fc822cbe2fcae32002316/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/utils.py#L68-L69 to produce `[str, int]` and trigger a warning from https://github.com/open-telemetry/opentelemetry-python/blob/34d11d59e9b37052eb3aa6706288641ecb3056e7/opentelemetry-api/src/opentelemetry/attributes/__init__.py#L86-L93

maxkomarychev:
I would like to solve this but I'm not sure what's the best way to identify "None"? Set it to "-1" ?

ocelotl:
This implementation fails its test cases:

```
    def test_set_attributes_partial_timelimit_soft_limit_numeric(self):                                                                                                                         
        # it should extract only relevant keys                                                                                                                                                  
        context = {
            "correlation_id": "44b7f305",
            "delivery_info": {"eager": True},
            "eta": "soon",
            "expires": "later",
            "hostname": "localhost",
            "id": "44b7f305",
            "reply_to": "44b7f305",
            "retries": 4,
[...]

PR #2025, opened 2023-10-27
fix(celery): prevent a warning when only one of time limits is set

maxkomarychev:
I am not 100% sure what is the correct way to fix this issue and looking for feedback from maintainers

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2022

# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

- [] Followed the style guidelines of this project


srikanthccv:
`None` is allowed. I would personally prefer (None, 42) over ("", "42") because the original type is preserved and you don't have to worry about the type when data reaches the backend.

maxkomarychev:
> None is allowed. [...] original type is preserved and you don't have to worry about the type when data reaches the backend.

I do not disagree with this. 

At the same time I believe this fix is conforming to the [spec](https://opentelemetry.io/docs/specs/otel/common/) which says that arrays MUST be homogenous and `null` values are not allowed in arrays:

> An array of primitive type values. The array MUST be homogeneous, i.e., it MUST NOT contain values of different types.

> Attribute values of null are not valid and attempting to set a null value is undefined behavior.
> null values SHOULD NOT be allowed in arrays. 


Maybe a better fix would be to split a single array into two independent fields and track them separately or change the spec :)

What do you think?

srikanthccv:
The spec also says

> However, if it is impossible to make sure that no null values are accepted (e.g. in languages that do not have appropriate compile-time type checking), null values within arrays MUST be preserved as-is (i.e., passed on to span processors / exporters as null). If exporters do not support exporting null values, they MAY replace those values by 0, false, or empty strings. This is required for map/dictionary structures represented as two arrays with indices that are kept in sync (e.g., two attributes header_keys and header_values, both containing an array of strings to represent a mapping header_keys[i] -> header_values[i]).

If there are `None` values in the array, the spec explicitly calls out they should be preserved as-is in languages that don't have compile-time checks. When the data reaches exporter, depending on the protocol support, it might be replaced with zero values of each data type.

I don't have expertise in celery but I think a number(s) property of instrumentation should be a number(s) attribute, not a string, especially not to fix the warning message.

PR #2026, opened 2023-10-30
feat: configure header extraction for ASGI middleware via constructor params

### LINKED ISSUE #2009: Allow configuring header extraction via runtime parameters
Currently, web framework instrumentations only allow configuring header extraction via environment variables. I want to add the option to configure them via constructor parameters. I think the code should be quite straightforward but I want to check if this would be an acceptable change before making a  PR. Some background on why it wasn't implemented in the first place might also be helpful if it's something other than 'no one asked for it yet'.

adriangb:
This adds constructor parameters to the ASGI OpenTelemetryMiddleware that allow configuration extraction and sanitization of header values via constructor parameters. Previously this was only configurable via environment variables.

Fixes #2009


instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -249,7 +254,18 @@ def process_request(self, request):
...
+                            get_custom_headers(
+                                OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST
+                            ),
+                            SanitizeValue(
+                                get_custom_headers(
+                                    OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SANITIZE_FIELDS
+                                )
+                            ),
+                            normalise_request_header_name,
+                        )
```
adriangb, line 268: I think a future PR should make this implementation mirror the asig middleware implementation but I'm trying to minimize the sprawl for this PR

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -530,7 +503,20 @@ def __init__(
...
+        self.http_capture_headers_server_response = http_capture_headers_server_response or (
+            get_custom_headers(OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_RESPONSE)
+        ) or None
+        self.http_capture_headers_sanitize_fields = SanitizeValue(
+            http_capture_headers_sanitize_fields or (
+                get_custom_headers(OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SANITIZE_FIELDS)
+            ) or []
+        )
+
+    async def __call__(self, scope: dict[str, Any], receive: Callable[[], Awaitable[dict[str, Any]]], send: Callable[[dict[str, Any]], Awaitable[None]]) -> None:
```
adriangb, line 519: I gave this module some typing love at the points my changes were touching but I feel like it overall needs to be improved.

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
```diff
@@ -84,9 +84,9 @@ def sanitize_header_value(self, header: str, value: str) -> str:
         )
 
     def sanitize_header_values(
-        self, headers: dict, header_regexes: list, normalize_function: callable
-    ) -> dict:
-        values = {}
+        self, headers: dict[str, str], header_regexes: list[str], normalize_function: Callable[[str], str]
+    ) -> dict[str, str]:
+        values: dict[str, str] = {}
```
adriangb, line 89: Happy to revert these, I mainly added them to help myself understand the logic in these functions

adriangb:
@srikanthccv can you review? There may be some test failures but I'm pretty confident they can be resolved.

I don't seem to be able to run any tox commends either locally or in a blank CodeSpaces:

```
py310-test-instrumentation-asgi: commands_pre[1] /workspaces/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-asgi/tests> pip install '"opentelemetry-api[test] @ "git+https://github.com/open-telemetry/opentelemetry-python.git@main"'
ERROR: Invalid requirement: '"opentelemetry-api[test] @ "git+https://github.com/open-telemetry/opentelemetry-python.git@main"'
Hint: It looks like a path. File '"opentelemetry-api[test] @ "git+https://github.com/open-telemetry/opentelemetry-python.git@main"' does not exist.
py310-test-instrumentation-asgi: exit 1 (0.45 seconds) /workspaces/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-asgi/tests> pip install '"opentelemetry-api[test] @ "git+https://github.com/open-telemetry/opentelemetry-python.git@main"'
```

adriangb:
Ah good catch I think I was using v4, let me try that

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
aabmass: Thanks for adding annotations, we should definitely keep them. but iirc you need the imports from typing before python 3.9 https://docs.python.org/3/library/typing.html#deprecation-timeline-of-major-features

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
```diff
@@ -801,18 +801,15 @@ class TestAsgiApplicationRaisingError(AsgiTestBase):
     def tearDown(self):
         pass
 
-    @mock.patch(
-        "opentelemetry.instrumentation.asgi.collect_custom_request_headers_attributes",
-        side_effect=ValueError("whatever"),
-    )
-    def test_asgi_issue_1883(
-        self, mock_collect_custom_request_headers_attributes
-    ):
```
adriangb, line 810: IMO this was a really bad way to test this. The alternative I'm proposing here avoids monkey patching an unrelated function, is easier to understand and is even less LOC.

instrumentation/opentelemetry-instrumentation-starlette/tests/test_starlette_instrumentation.py
```diff
@@ -466,16 +466,18 @@ async def _(websocket: WebSocket) -> None:
...
+    @patch.dict(
+        "os.environ",
+        {
+            OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SANITIZE_FIELDS: ".*my-secret.*",
+            OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST: "Custom-Test-Header-1,Custom-Test-Header-2,Custom-Test-Header-3,Regex-Test-Header-.*,Regex-Invalid-Test-Header-.*,.*my-secret.*",
+            OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_RESPONSE: "Custom-Test-Header-1,Custom-Test-Header-2,Custom-Test-Header-3,my-custom-regex-header-.*,invalid-regex-header-.*,.*my-secret.*",
+        },
+    )
+    def setUp(self) -> None:
+        super().setUp()
```
adriangb, line 479: Previously the middleware checked the env variables for every request. Now it only checks it once when the middleware is created. I think this is a positive change because it will reduce the work done for each request and I don't think that changing env variables between requests is something people do.

The reason then I had to move the `patch` call to `setUp()` is that `patch.dict` only patches `test_*` methods by default (see [docs](https://docs.python.org/3/library/unittest.mock.html#patch-dict)).

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_custom_headers.py
```diff
@@ -72,21 +73,20 @@ async def websocket_app_with_custom_headers(scope, receive, send):
...
-        OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_RESPONSE: "Custom-Test-Header-1,Custom-Test-Header-2,Custom-Test-Header-3,my-custom-regex-header-.*,invalid-regex-header-.*,.*my-secret.*",
-    },
-)
 class TestCustomHeaders(AsgiTestBase, TestBase):
+    constructor_params = {}
+    __test__ = False
+
+    def __init_subclass__(cls) -> None:
+        if cls is not TestCustomHeaders:
+            cls.__test__ = True
```
adriangb, line 82: The idea here is to share all of the tests between the env var and constructor configurations without duplicating code.

adriangb:
@srikanthccv @aabmass tests are passing, I've left some comments on the implementation, could you review please?

srikanthccv (approved):
Thanks for the clean up!

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
```diff
@@ -215,15 +217,15 @@ def sanitize_method(method: Optional[str]) -> Optional[str]:
         return method
     return "UNKNOWN"
 
-
-def get_custom_headers(env_var: str) -> List[str]:
-    custom_headers = environ.get(env_var, [])
+def get_custom_headers(env_var: str) -> list[str]:
+    custom_headers = environ.get(env_var, None)
```
shalevr, line 221: Why is this change necessary?

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
```diff
@@ -215,15 +217,15 @@ def sanitize_method(method: Optional[str]) -> Optional[str]:
...
     if custom_headers:
```
shalevr, line 222: if you keep the change with None as default, it's better to change the condition: 
```suggestion
    if custom_headers is not None:
```

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
adriangb: It avoids allocating a list

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
adriangb: Sure but if it’s `””` there’s no point in taking that branch anyway

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
shalevr: If I understand this correctly, the list is allocated anyway(in line 228)

adriangb: Ah true, it’s just the pattern I’m used to to avoid the common gotcha.

In any case I believe what was there could return `""` which is why I changed it in the first place.

adriangb:
Is there anything missing to get this merged? Thanks

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -332,55 +334,29 @@ def collect_request_attributes(scope):
...
-        normalise_request_header_name,
-    )
-
-
-def collect_custom_response_headers_attributes(message):
-    """returns custom HTTP response headers to be added into SERVER span as span attributes
-    Refer specification https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers
+    Refer specifications:
+     - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers
+     - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers
```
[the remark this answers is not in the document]
adriangb:
```suggestion
```

PR #2061, opened 2023-11-14
Clean up use of suppress_instrumentation in context and fix httpx bug

adriangb:
Currently, the exporters set the `_SUPPRESS_INSTRUMENTATION_KEY` key but httpx only checks for `"suppress_instrumentation"`.

This fixes that bug and does some cleanup to prevent similar bugs in the future.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
@@ -86,7 +86,6 @@ from botocore.endpoint import Endpoint
 from botocore.exceptions import ClientError
 from wrapt import wrap_function_wrapper
 
-from opentelemetry import context as context_api
 from opentelemetry.instrumentation.botocore.extensions import _find_extension
 from opentelemetry.instrumentation.botocore.extensions.types import (
     _AwsSdkCallContext,
--- a/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
@@ -87,9 +87,6 @@ from botocore.exceptions import ClientError
 from wrapt import wrap_function_wrapper
 
 from opentelemetry import context as context_api
-
-# FIXME: fix the importing of this private attribute when the location of the _SUPPRESS_HTTP_INSTRUMENTATION_KEY is defined.
-from opentelemetry.context import _SUPPRESS_HTTP_INSTRUMENTATION_KEY
 from opentelemetry.instrumentation.botocore.extensions import _find_extension
 from opentelemetry.instrumentation.botocore.extensions.types import (
     _AwsSdkCallContext,
@@ -98,7 +95,8 @@ from opentelemetry.instrumentation.botocore.package import _instruments
 from opentelemetry.instrumentation.botocore.version import __version__
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
 from opentelemetry.instrumentation.utils import (
-    _SUPPRESS_INSTRUMENTATION_KEY,
+    is_instrumentation_enabled,
+    suppress_http_instrumentation,
     unwrap,
 )
 from opentelemetry.propagators.aws.aws_xray_propagator import AwsXRayPropagator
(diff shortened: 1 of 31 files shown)


opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -152,3 +152,21 @@ def _python_path_without_directory(python_path, directory, path_separator):
...
+        or 
+        context.get_value(context._SUPPRESS_INSTRUMENTATION_KEY)  # type: ignore
+    ):
+        return False
+    return True
+
+
+def is_http_instrumentation_enabled() -> bool:
+    return (
+        is_http_instrumentation_enabled()
```
srikanthccv, line 169: did you mean `is_instrumentation_enabled`?

adriangb: Probably! Still need to run linting etc on this

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
```diff
@@ -197,25 +195,22 @@ def _patched_api_call(self, original_func, instance, args, kwargs):
...
-            except ClientError as error:
-                result = getattr(error, "response", None)
-                _apply_response_attributes(span, result)
-                _safe_invoke(extension.on_error, span, error)
-                raise
-            else:
-                _apply_response_attributes(span, result)
-                _safe_invoke(extension.on_success, span, result)
+                with suppress_http_instrumentation():
+                    result = None
```
adriangb, line 200: I tried to preserve the logic with the exception catching and finally here but a second set of eyes would be good, especially given that I don't know if calling `_safe_invoke(extension.after_service_call)` after de-suppressing instrumentation is on purpose or it just happened to be like that

instrumentation/opentelemetry-instrumentation-cassandra/src/opentelemetry/instrumentation/cassandra/__init__.py
```diff
@@ -65,7 +65,10 @@ def _traced_execute_async(func, instance, args, kwargs):
             if span.is_recording():
                 span.set_attribute(SpanAttributes.DB_NAME, instance.keyspace)
                 span.set_attribute(SpanAttributes.DB_SYSTEM, "cassandra")
-                span.set_attribute(SpanAttributes.NET_PEER_NAME, instance.cluster.contact_points)
+                span.set_attribute(
+                    SpanAttributes.NET_PEER_NAME,
+                    instance.cluster.contact_points,
+                )
```
adriangb, line 71: I ran `.tox/lint/bin/black .` after `tox -e lint` (which failed with `Could not find a version that satisfies the requirement opentelemetry-semantic-conventions==0.42b0.dev` for `opentelemetry-instrumentation-wsgi[test]`; I'm not sure how to point tox to my local version; I'm using `tox==3.27.1`) and it did some reformatting. I can try to revert this if I can get some pointers on the dependency issue.

adriangb: I manually reverted most of these for now

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -212,3 +217,44 @@ def _get_opentelemetry_stability_opt_in(
...
+        yield
+    finally:
+        context.detach(token)
+
+
+@contextmanager
+def suppress_instrumentation() -> Iterable[None]:
+    """Suppress instrumentation within the context."""
+    with _suppress_instrumentation(_SUPPRESS_INSTRUMENTATION_KEY):
+        yield
```
adriangb, line 253: We might want to also set the `"suppress_instrumentation"` (with no uuid) key here for any users that were relying on it.

adriangb:
@srikanthccv would you mind reviewing again? Thanks!

srikanthccv:
Thank you @adriangb. The changes are a little bigger than usual. I will take some time to review this one.

adriangb:
No problem. I could make the diff smaller by e.g. adding the new functions and only replacing the old ones in one spot as a proof of concept then doing a followup PR that is just replacing other uses of the context.

adriangb:
@srikanthccv let me know what you'd like to do with this PR so we don't get more conflicts to keep resolving

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
srikanthccv: This looks fine

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
srikanthccv: We kept these suppress keys internal so far so it should be fine to not set that.

srikanthccv (approved):

PR #2101, opened 2023-12-21
Refactor bootstrap generation

### LINKED ISSUE #2053: ERROR: No matching distribution found for opentelemetry-instrumentation-aiohttp-server==0.42b0
Python Version - 3.8

**Steps to reproduce**
Using 'opentelemetry-bootstrap -a install' Command to Install the Dependencies for Auto Instrumentation.

package : opentelemetry-instrumentation-aiohttp-server==0.42b0 should be downloaded from pypi

ERROR: No matching distribution found for opentelemetry-instrumentation-aiohttp-server==0.42b0

ocelotl:
This makes the bootstrap script get the package version directly from pypi instead of from our lists of packages. This makes sure that the packages are actually available for the end user to install.

### DIFF
--- a/scripts/otel_packaging.py
+++ b/scripts/otel_packaging.py
@@ -12,23 +12,36 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
-
-import tomli
+from tomli import load
+from os import path, listdir
+from subprocess import check_output, CalledProcessError
 from requests import get
 
-scripts_path = os.path.dirname(os.path.abspath(__file__))
-root_path = os.path.dirname(scripts_path)
-instrumentations_path = os.path.join(root_path, "instrumentation")
+scripts_path = path.dirname(path.abspath(__file__))
+root_path = path.dirname(scripts_path)
+instrumentations_path = path.join(root_path, "instrumentation")
 
 
 def get_instrumentation_packages():
-    for pkg in sorted(os.listdir(instrumentations_path)):
-        pkg_path = os.path.join(instrumentations_path, pkg)
-        if not os.path.isdir(pkg_path):
+    for pkg in sorted(listdir(instrumentations_path)):
+        pkg_path = path.join(instrumentations_path, pkg)
+        if not path.isdir(pkg_path):
             continue
 
         error = f"Could not get version for package {pkg}"
+
+        try:
+            hatch_version = check_output(
+                "hatch version",
+                shell=True,
+                cwd=pkg_path,
+                universal_newlines=True
+            )
+
+        except CalledProcessError as exc:
+            print(f"Could not get hatch version from path {pkg_path}")
+            print(exc.output)
+
         try:
             response = get(f"https://pypi.org/pypi/{pkg}/json", timeout=10)
 
@@ -40,16 +53,14 @@ def get_instrumentation_packages():
             print(error)
             continue
 
-        version = response.json()["info"]["version"]
-
-        pyproject_toml_path = os.path.join(pkg_path, "pyproject.toml")
+        pyproject_toml_path = path.join(pkg_path, "pyproject.toml")
 
         with open(pyproject_toml_path, "rb") as file:
-            pyproject_toml = tomli.load(file)
+            pyproject_toml = load(file)
 
         instrumentation = {
             "name": pyproject_toml["project"]["name"],
-            "version": version.strip(),
+            "version": hatch_version.strip(),
             "instruments": pyproject_toml["project"]["optional-dependencies"][
                 "instruments"
             ],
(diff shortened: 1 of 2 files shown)


lzchen:
Was there a [reason](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/514) why we initially had the source versions being generated instead of the latest from pypi? Were we installing the source versions when running `opentelemetry-bootstrap`?

lzchen:
I think we allow users to specify [version](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py#L134) in bootstrap so probably have to still reflect this in the install.

scripts/otel_packaging.py
```diff
@@ -29,17 +28,19 @@ def get_instrumentation_packages():
...
-                universal_newlines=True,
-            )
-        except CalledProcessError as exc:
-            print(f"Could not get hatch version from path {pkg_path}")
-            print(exc.output)
-            raise exc
+            response = get(f"https://pypi.org/pypi/{pkg}/json", timeout=10)
+
+        except Exception:
+            print(error)
```
lzchen, line 36: Shouldn't we log instead of print?

lzchen:
Perhaps we want to add a changelog since this would cause users to install pypi versions instead of source versions?

ocelotl:
> Just to clarify, in our release process, we change each package to the release [version](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2092/files#diff-a9435a736abdea2d4f8a1fb9bde32809fe522eb98f43bf7025f5592626b62da4) and then we [regenerate](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/scripts/prepare_release.sh?rgh-link-date=2023-12-27T17%3A33%3A59Z#L23) `bootstrap_gen.py`. I'm curious as to how you generated the `0.43b0` versions of `bootstrap_gen.py`, did you run the new script or did you replace these manually? I think we just need to add a try catch for erroneous installs, there's no need to change what is generated in our dev branch right?

Right, reverted to `44b0.dev`.

ocelotl:
> > Just to clarify, in our release process, we change each package to the release [version](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2092/files#diff-a9435a736abdea2d4f8a1fb9bde32809fe522eb98f43bf7025f5592626b62da4) and then we [regenerate](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/scripts/prepare_release.sh?rgh-link-date=2024-01-02T23%3A32%3A32Z#L23) `bootstrap_gen.py`. I'm curious as to how you generated the `0.43b0` versions of `bootstrap_gen.py`, did you run the new script or did you replace these manually? I think we just need to add a try catch for erroneous installs, there's no need to change what is generated in our dev branch right?
> 
> Right, reverted to `44b0.dev`.

Which fails `tox -e generate`...

scripts/otel_packaging.py
ocelotl: I'm fine with printing this is just a script. Logging is more susceptible to logging configuration that may produce no output.

ocelotl:
> I'm not sure if this is exactly a workaround for #2053 since users will still see an error message if the package doesn't exist in Pypi right? I think there isn't really a workaround since there's no way to provide the actual package (the most we can do is not display the error message and just explain on the issues that the package is not available yet).

This is a workaround because no exception is being raised now, only an error message is now produced for the particular package.

aabmass:
> Fixes #2053

I think this fixes the bootstrap script issue, but doesn't actually solve the PyPI error you would when trying to install the non-existent release

tammy-baylis-swi:
> I think this fixes the bootstrap script issue, but doesn't actually solve the PyPI error you would when trying to install the non-existent release

Seconding this, and was also mentioned at today's sig meeting. In this PR description I've included an example error if a user with `aiohttp` installed tries to bootstrap with the latest release: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2299. It would be great to have this part fixed.

lzchen (approved):
LGTM. @ocelotl please remove the "fixes" line in the pr description as this pr doesn't actually fully resolve the original issue.

npuichigo:
any update on this?

scripts/otel_packaging.py
```diff
@@ -29,17 +28,19 @@ def get_instrumentation_packages():
...
+            continue
+
+        if response.status_code != 200:
+            print(error)
+            continue
+
+        version = response.json()["info"]["version"]
```
tammy-baylis-swi, line 43: Hi @ocelotl , the `generate` check is failing because of PyPI verses local dev versions. Please could this be fixed?

tammy-baylis-swi: Hi again, please could this be fixed?

PR #2127, opened 2024-01-17
Remove [test] package from flask instrumentation

### LINKED ISSUE #2126: Remove [test] package from flask instrumentation
See #2182

ocelotl:
Fixes #2126

~I am not a fan of using additional executable `bash` scripts but this exemplifies what I am trying to do (remove the `[test]` packages and separate the testing requirements for every package into a file in each package).~

If we agree on this strategy I will add the changes for the rest of the packages in this PR. For the time being I am adding only one since I want to make it easy for others to review and understand this strategy.

For the record, I would prefer not to use `tox` but [`nox`](https://nox.thea.codes/en/stable/) which has the advantage of using a Python file instead of an `ini` file. I think that by using `nox` we can find a pure-Python way of handing our test dependencies. In fact, I would prefer if we had one noxfile for every package and a Python script that just calls `nox` in every package directory.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -83,9 +83,14 @@ envlist =
     pypy3-test-instrumentation-fastapi
 
     ; opentelemetry-instrumentation-flask
-    py3{8,9,10,11}-test-instrumentation-flask-{213,220}
-    py3{8,9,10,11}-test-instrumentation-flask-{300}
-    pypy3-test-instrumentation-flask-{213,220}
+    ; The numbers at the end of the environment names
+    ; below mean these dependencies are being used:
+    ; 0: Flask ==2.1.3 Werkzeug <3.0.0
+    ; 1: Flask ==2.2.0 Werkzeug <3.0.0
+    ; 2: Flask >=3.0.0 Werkzeug >=3.0.0
+    py3{8,9,10,11}-test-instrumentation-flask-{0,1}
+    py3{8,9,10,11}-test-instrumentation-flask-{2}
+    pypy3-test-instrumentation-flask-{0,1}
 
     ; opentelemetry-instrumentation-urllib
     py3{8,9,10,11}-test-instrumentation-urllib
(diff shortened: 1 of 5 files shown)


ocelotl:
BTW, the reason I am using a `bash` executable file here instead of a `test-requirements.txt` file is because we need to install packages from PyPi (`flask`, `werkzeug`) and also _local_ packages from this repo.

lzchen:
> For the record, I would prefer not to use tox but [nox](https://nox.thea.codes/en/stable/) which has the advantage of using a Python file instead of an ini file.

I'd prefer not to use `nox` since we had precedence in `opencensus` for it being a bit complex to use in terms of setting up configuration and matrices in exactly the way we want it.

ocelotl:
@lzchen ok, using `nox` (or not) can be discussed later, are you ok with the removal of `[test]` packages?

lzchen:
@ocelotl 

>  ok, using nox (or not) can be discussed later, are you ok with the removal of [test] packages?

Yup, these seems fine with me. Thanks for working on this!

shalevr (approved):

ocelotl:
I need to update the requirements file, marking as draft to prevent accidental merging.

aabmass (approved):

tox.ini
```diff
@@ -83,9 +83,14 @@ envlist =
...
 
     ; opentelemetry-instrumentation-flask
-    py3{8,9,10,11}-test-instrumentation-flask-{213,220}
-    py3{8,9,10,11}-test-instrumentation-flask-{300}
-    pypy3-test-instrumentation-flask-{213,220}
+    ; The numbers at the end of the environment names
+    ; below mean these dependencies are being used:
+    ; 0: Flask ==2.1.3 Werkzeug <3.0.0
+    ; 1: Flask ==2.2.0 Werkzeug <3.0.0
+    ; 2: Flask >=3.0.0 Werkzeug >=3.0.0
```
aabmass, line 90: Would still prefer to also name them flask-21 flask-22 flask-33 but approved

ocelotl: Thanks, we can check this approach again later :v:

lzchen (approved):

ocelotl:
Converting to draft to prevent accidental merging.

lzchen (approved):

PR #2150, opened 2024-02-02
fix: elasticsearch `bulk` API produces lists

silverjam:
# Description

Fixes an issue seen in production where the elasticsearch bulk API would end up sending a list through this code path and not a dictionary.

For example:

```
File \"/code/platform_be/core/logic/ingest.py\", line 378, in update_source_doc_version_history
        bulk(es_client, bulk_es_updates)
      File \"/opt/python/lib/python3.10/site-packages/elasticsearch/helpers/actions.py\", line 521, in bulk
        for ok, item in streaming_bulk(
      File \"/opt/python/lib/python3.10/site-packages/elasticsearch/helpers/actions.py\", line 436, in streaming_bulk
        for data, (ok, info) in zip(
      File \"/opt/python/lib/python3.10/site-packages/elasticsearch/helpers/actions.py\", line 339, in _process_bulk_chunk
        resp = client.bulk(*args, operations=bulk_actions, **kwargs)  # type: ignore[arg-type]
      File \"/opt/python/lib/python3.10/site-packages/elasticsearch/_sync/client/utils.py\", line 414, in wrapped
        return api(*args, **kwargs)
      File \"/opt/python/lib/python3.10/site-packages/elasticsearch/_sync/client/__init__.py\", line 704, in bulk
        return self.perform_request(  # type: ignore[return-value]
      File \"/opt/python/lib/python3.10/site-packages/elasticsearch/_sync/client/_base.py\", line 285, in perform_request
        meta, resp_body = self.transport.perform_request(
      File \"/opt/python/lib/python3.10/site-packages/opentelemetry/instrumentation/elasticsearch/__init__.py\", line 242, in wrapper
        attributes[SpanAttributes.DB_STATEMENT] = sanitize_body(
      File \"/opt/python/lib/python3.10/site-packages/opentelemetry/instrumentation/elasticsearch/utils.py\", line 58, in sanitize_body
        flatten_body = _flatten_dict(body)
      File \"/opt/python/lib/python3.10/site-packages/opentelemetry/instrumentation/elasticsearch/utils.py\", line 31, in _flatten_dict
        for k, v in d.items():
    AttributeError: 'list' object has no attribute 'items'"
```

## Type of change

# How Has This Been Tested?

- [x] Unit tests

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
@@ -246,7 +246,7 @@ def _wrap_perform_request(
                     attributes["elasticsearch.method"] = method
                 if body:
                     # Don't set db.statement for bulk requests, as it can be very large
-                    if not isinstance(body, list):
+                    if isinstance(body, dict):
                         attributes[SpanAttributes.DB_STATEMENT] = sanitize_body(
                             body
                         )
--- a/instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
@@ -245,9 +245,11 @@ def _wrap_perform_request(
                 if method:
                     attributes["elasticsearch.method"] = method
                 if body:
-                    attributes[SpanAttributes.DB_STATEMENT] = sanitize_body(
-                        body
-                    )
+                    # Don't set db.statement for bulk requests, as it can be very large
+                    if not isinstance(body, list):
+                        attributes[SpanAttributes.DB_STATEMENT] = sanitize_body(
+                            body
+                        )
                 if params:
                     attributes["elasticsearch.params"] = str(params)
                 if doc_id:


silverjam:
Hi @lzchen / @ocelotl / @shalevr -- Is there anything I can do to help move this forward?  This is preventing us from using ES instrumentation in our service.  Thanks!

pquentin:
Hello! I maintain the elasticsearch and elastic-transport packages at Elastic (and am actually working on native instrumentation - see https://github.com/elastic/elasticsearch-py/issues/2435). That said, I'm new to OpenTelemetry and I've never contributed to opentelemtry-python-contrib.

While most Elasticsearch APIs accept dictionaries, indeed the [bulk API](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html) accepts a list of dictionaries, as you found out. However, the [Semantic Conventions for Elasticsearch](https://opentelemetry.io/docs/specs/semconv/database/elasticsearch/) define `db.statement` as follows:

> The request body for a [search-type query](https://www.elastic.co/guide/en/elasticsearch/reference/current/search.html), as a json string.

This makes sense. The bulk API can be used to send arbitrarily large amount of data, and you would not want that to be part of your span attributes. I believe the correct fix here is to only set `db.statement` when one of those APIs is used: Search, Async search, Multi search, EQL search, ES|QL query, Terms enum, Search template, Multi search template or Render search template. This isn't very easy to do in opentelemetry-python-contrib because the client does not pass the name of the endpoint to the transport (yet, I'm working on it). Which means it has to be guessed from the URL.

I guess an easy way out here could also be to *not* set `db.statement` for bulk requests, which would solve your immediate issue, but would still not be compliant.

silverjam:
> I guess an easy way out here could also be to _not_ set `db.statement` for bulk requests, which would solve your immediate issue, but would still not be compliant.

Thanks for the context @pquentin -- I assume that the path forward for us would be to either wait for your native instrumentation or to do a stopgap fix here.  From what you described it sounds like the stopgap is to just not generate any data if `db.statement` happens to be a list.  This feels a bit like a heuristic, but if you think it's OK stopgap I'm happy to implement this.

pquentin:
Native instrumentation will only be available with a future version of elasticsearch-py, so I think you want to keep the existing instrumentation for older versions? (Though it would be nice to automatically disable for newer versions maybe? But it would be a breaking change since the native implementation follows the semantic convention which probably did not exist when this repo was started.)

Regarding the fix, the proper fix is definitely inspecting the URL to figure out if the request is a search request, but I also offered the stopgap as way to fix the most pressing issue with less effort. It's really up to you to decide what you implement! (But if you do ping me in the pull request, I'll be happy to take a look.)

silverjam:
Made this change much simpler and ran tests with `CORE_REPO_SHA=a1253585f66d63e7c05a19f070f3bfe0ab6460c1 tox3 -e py311-test-instrumentation-elasticsearch-6` -- @pquentin or @xrmx if you have ideas on how to test this I'm all ears, or is this good enough?

xrmx (commented):
It improves current status and document the issue a bit so it's fine for me. It could use a test to avoid regressions though.

silverjam:
Ok, I'll look into adding a simple test

instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
```diff
@@ -245,9 +245,11 @@ def wrapper(wrapped, _, args, kwargs):
                 if method:
                     attributes["elasticsearch.method"] = method
                 if body:
-                    attributes[SpanAttributes.DB_STATEMENT] = sanitize_body(
-                        body
-                    )
+                    # Don't set db.statement for bulk requests, as it can be very large
+                    if not isinstance(body, list):
```
pquentin, line 249: I would suggest to only accept dicts instead:

```suggestion
                    # Don't set db.statement for bulk requests, as it can be very large
                    if isinstance(body, dict):
```

since `body` can also be a string or bytes.

Not sure about the test, sorry.

xrmx:
@silverjam I can write the tests but I need from you an example of the elastic client call that send a bulk request please.

silverjam:
> @silverjam I can write the tests but I need from you an example of the elastic client call that send a bulk request please.
> 
> UPDATE: ah sending more than one item in the bulk iteratable was enough :)

Thanks, if it helps anyway, our calls look something like this:

```python
from elasticsearch.helpers import bulk
# ...
    bulk_es_updates = []
    updated_previous_source_docs = []
    for previous_source_doc in previous_source_docs:
        if cast(bool, previous_source_doc.is_latest_version):
            previous_source_doc.is_latest_version = False
            updated_previous_source_docs.append(previous_source_doc)
            bulk_es_updates.append(
                {
                    "_op_type": "update",
                    "_index": ES_SOURCE_DOC_ALIAS,
                    "_id": str(previous_source_doc.id),
                    "doc": {"is_latest_version": False},
                }
            )
# ...
    es_client = get_es_client()
    if bulk_es_updates:
        bulk(es_client, bulk_es_updates)
```

silverjam:
Closing in favor of #2355

Issue #2180, opened 2024-02-14
Remove pkg_resource and replace for importlib.metadata


Rodrigo-Novas:
**Describe your environment**

Docker environment
Image so: ubuntu:20.04
Python version: 3.9

Otel dependencies:

opentelemetry-api = "^1.14.0"
opentelemetry-semantic-conventions = "0.41b0"  # At the moment there are no releases without beta
opentelemetry-instrumentation-urllib3 = "0.41b0"  # At the moment there are no releases without beta
opentelemetry-instrumentation-httpx = "0.41b0" # At the moment there are no releases without beta

**Steps to reproduce**

Instrument any app and run it, it will throws a warning deprecation message about pkg_resource.

I created a CM of spans with something like this:

``` python
from opentelemetry import trace
@contextlib.contextmanager
def create_span(tracer: trace.Tracer) -> Generator:
    "Start the span as current span."
    name = "name"
    _kind = "kind"
    with tracer.start_as_current_span(
        name,
        kind=_kind,
        record_exception=False,
    ) as current_span:
        yield current_span
```

**What is the expected behavior?**

Otel doesnt throws a deprecationWarning

**What is the actual behavior?**

The output is a deprecation warning:

`DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html`

**Additional context**

[Use of pkg_resources is deprecated](https://setuptools.pypa.io/en/latest/pkg_resources.html)

Rodrigo-Novas:
I created a fix about this on https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2181

theunkn0wn1:
FYI `pkg_resources` is deprecated AND REMOVED in modern python versions.
This package now depends on an undeclared dependency and is thus broken.

```
backend        |     from opentelemetry.instrumentation.sqlalchemy import SQLAlchemyInstrumentor
backend        |   File "/home/nonroot/.local/lib/python3.12/site-packages/opentelemetry/instrumentation/sqlalchemy/__init__.py", line 105, in <module>
backend        |     from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
backend        |   File "/home/nonroot/.local/lib/python3.12/site-packages/opentelemetry/instrumentation/instrumentor.py", line 27, in <module>
backend        |     from opentelemetry.instrumentation.dependencies import (
[...]

Issue #1970, opened 2023-09-26
Opentelemetry instrumentation uses deprecated pkg_resources.


jenshnielsen:
**Describe your environment** 
Python 3.11
Setuptools >= 67.5.0
Windows 11

**Steps to reproduce**

Any operation that imports opentelemetry\instrumentation\dependencies.py

**What is the expected behavior?**

No deprecation warnings

**What is the actual behavior?**
What did you see instead?

`C:\Users\jenielse\Miniconda3\envs\myenv\Lib\site-packages\opentelemetry\instrumentation\dependencies.py:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html`

**Additional context**

This was removed from opentelemetry sdk/api in https://github.com/open-telemetry/opentelemetry-python/pull/3047

In addition to being deprecated pkg_resources is known to be problematic since it indexes all packages on start up adding significant overhead to the package import. Normally this should be replaced by importlib.resources or for older python version support importlib_resources  

As a bonus this will also remove the runtime dependency on setuptools

PR #2181, opened 2024-02-14
Remove pkg_resource and replace for importlib.metadata in autoinstrumentation

Rodrigo-Novas:
# Description

Pkgresource dependency is [deprecated](https://setuptools.pypa.io/en/latest/pkg_resources.html#package-discovery-and-resource-access-using-pkg-resources), on this pr i replaced pkg_resource for importlib.metadata.

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2180

## Type of change

# How Has This Been Tested?

- [x] Run with tox the unit tests

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Breaking changes
 
 - Deprecation of pkg_resource in favor of importlib.metadata
-  ([#2180](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2180))
+  ([#2181](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2181))
 
 ## Version 1.27.0/0.48b0 ()
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Breaking changes
+
+- Deprecation of pkg_resource in favor of importlib.metadata
+  ([#2180](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2180))
+
 ## Version 1.27.0/0.48b0 ()
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Deprecation of pkg_resource in favor of importlib.metadata
-  ([#2180](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2180))
 - `opentelemetry-instrumentation-kafka-python` Instrument temporary fork, kafka-python-ng inside kafka-python's instrumentation
   ([#2537](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2537))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `opentelemetry-instrumentation-kafka-python` Instrument temporary fork, kafka-python-ng
-  inside kafka-python's instrumentation
+- Deprecation of pkg_resource in favor of importlib.metadata
+  ([#2180](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2180))
+- `opentelemetry-instrumentation-kafka-python` Instrument temporary fork, kafka-python-ng inside kafka-python's instrumentation
   ([#2537](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2537))
 
 ### Breaking changes
--- a/opentelemetry-instrumentation/pyproject.toml
+++ b/opentelemetry-instrumentation/pyproject.toml
@@ -26,7 +26,6 @@ classifiers = [
 ]
 dependencies = [
   "opentelemetry-api ~= 1.4",
-  "setuptools >= 16.0",
   "wrapt >= 1.0.0, < 2.0.0",
 ]
 
--- a/opentelemetry-instrumentation/pyproject.toml
+++ b/opentelemetry-instrumentation/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 dependencies = [
   "opentelemetry-api ~= 1.4",
   "wrapt >= 1.0.0, < 2.0.0",
-  "importlib-metadata >= 6.0, < 7.0",
+  "importlib-metadata >= 6.0, < 8.0",
   "packaging >= 18.0",
 ]
 
--- a/opentelemetry-instrumentation/pyproject.toml
+++ b/opentelemetry-instrumentation/pyproject.toml
@@ -27,6 +27,8 @@ classifiers = [
 dependencies = [
   "opentelemetry-api ~= 1.4",
   "wrapt >= 1.0.0, < 2.0.0",
+  "importlib-metadata >= 6.0, < 7.0",
+  "packaging >= 18.0",
 ]
 
 [project.scripts]
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
@@ -48,9 +48,7 @@ def run() -> None:
 
     argument_otel_environment_variable = {}
 
-    for entry_point in entry_points().get(
-        "opentelemetry_environment_variables", []
-    ):
+    for entry_point in entry_points(group="opentelemetry_environment_variables"):
         environment_variable_module = entry_point.load()
 
         for attribute in dir(environment_variable_module):
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
@@ -13,13 +13,12 @@
 # limitations under the License.
 
 from argparse import REMAINDER, ArgumentParser
+from importlib.metadata import entry_points
 from logging import getLogger
 from os import environ, execl, getcwd
 from os.path import abspath, dirname, pathsep
 from re import sub
 from shutil import which
-
-from importlib.metadata import entry_points
 
 from opentelemetry.instrumentation.version import __version__
 
@@ -48,7 +47,9 @@ def run() -> None:
 
     argument_otel_environment_variable = {}
 
-    for entry_point in entry_points(group="opentelemetry_environment_variables"):
+    for entry_point in entry_points(
+        group="opentelemetry_environment_variables"
+    ):
         environment_variable_module = entry_point.load()
 
         for attribute in dir(environment_variable_module):
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
@@ -19,7 +19,7 @@ from os.path import abspath, dirname, pathsep
 from re import sub
 from shutil import which
 
-from pkg_resources import iter_entry_points
+from importlib.metadata import entry_points
 
 from opentelemetry.instrumentation.version import __version__
 
@@ -48,8 +48,8 @@ def run() -> None:
 
     argument_otel_environment_variable = {}
 
-    for entry_point in iter_entry_points(
-        "opentelemetry_environment_variables"
+    for entry_point in entry_points().get(
+        "opentelemetry_environment_variables", []
     ):
         environment_variable_module = entry_point.load()
 
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
@@ -56,7 +56,7 @@ class _EntryPointDistFinder:
 
 def _load_distro() -> BaseDistro:
     distro_name = environ.get(OTEL_PYTHON_DISTRO, None)
-    for entry_point in entry_points().get("opentelemetry_distro", []):
+    for entry_point in entry_points(group="opentelemetry_distro"):
         try:
             # If no distro is specified, use first to come up.
             if distro_name is None or distro_name == entry_point.name:
@@ -87,10 +87,10 @@ def _load_instrumentors(distro):
         # to handle users entering "requests , flask" or "requests, flask" with spaces
         package_to_exclude = [x.strip() for x in package_to_exclude]
 
-    for entry_point in entry_points().get("opentelemetry_pre_instrument", []):
+    for entry_point in entry_points(group="opentelemetry_pre_instrument"):
         entry_point.load()()
 
-    for entry_point in entry_points().get("opentelemetry_instrumentor", []):
+    for entry_point in entry_points(group="opentelemetry_instrumentor"):
         if entry_point.name in package_to_exclude:
             _logger.debug(
                 "Instrumentation skipped for library %s", entry_point.name
@@ -115,14 +115,14 @@ def _load_instrumentors(distro):
             _logger.exception("Instrumenting of %s failed", entry_point.name)
             raise exc
 
-    for entry_point in entry_points().get("opentelemetry_post_instrument", []):
+    for entry_point in entry_points(group="opentelemetry_post_instrument"):
         entry_point.load()()
 
 
 def _load_configurators():
     configurator_name = environ.get(OTEL_PYTHON_CONFIGURATOR, None)
     configured = None
-    for entry_point in entry_points().get("opentelemetry_configurator", []):
+    for entry_point in entry_points(group="opentelemetry_configurator"):
         if configured is not None:
             _logger.warning(
                 "Configuration of %s not loaded, %s already loaded",
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
@@ -15,7 +15,7 @@
 from logging import getLogger
 from os import environ
 
-from pkg_resources import iter_entry_points
+from importlib.metadata import entry_points
 
 from opentelemetry.instrumentation.dependencies import (
     get_dist_dependency_conflicts,
@@ -33,7 +33,7 @@ _logger = getLogger(__name__)
 
 def _load_distro() -> BaseDistro:
     distro_name = environ.get(OTEL_PYTHON_DISTRO, None)
-    for entry_point in iter_entry_points("opentelemetry_distro"):
+    for entry_point in entry_points().get("opentelemetry_distro", []):
         try:
             # If no distro is specified, use first to come up.
             if distro_name is None or distro_name == entry_point.name:
@@ -63,10 +63,10 @@ def _load_instrumentors(distro):
         # to handle users entering "requests , flask" or "requests, flask" with spaces
         package_to_exclude = [x.strip() for x in package_to_exclude]
 
-    for entry_point in iter_entry_points("opentelemetry_pre_instrument"):
+    for entry_point in entry_points().get("opentelemetry_pre_instrument", []):
         entry_point.load()()
 
-    for entry_point in iter_entry_points("opentelemetry_instrumentor"):
+    for entry_point in entry_points().get("opentelemetry_instrumentor", []):
         if entry_point.name in package_to_exclude:
             _logger.debug(
                 "Instrumentation skipped for library %s", entry_point.name
@@ -90,14 +90,14 @@ def _load_instrumentors(distro):
             _logger.exception("Instrumenting of %s failed", entry_point.name)
             raise exc
 
-    for entry_point in iter_entry_points("opentelemetry_post_instrument"):
+    for entry_point in entry_points().get("opentelemetry_post_instrument", []):
         entry_point.load()()
 
 
 def _load_configurators():
     configurator_name = environ.get(OTEL_PYTHON_CONFIGURATOR, None)
     configured = None
-    for entry_point in iter_entry_points("opentelemetry_configurator"):
+    for entry_point in entry_points().get("opentelemetry_configurator", []):
         if configured is not None:
             _logger.warning(
                 "Configuration of %s not loaded, %s already loaded",
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
@@ -12,10 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from importlib.metadata import EntryPoint, distributions, entry_points
 from logging import getLogger
 from os import environ
-
-from importlib.metadata import entry_points, EntryPoint, distributions
 
 from opentelemetry.instrumentation.dependencies import (
     get_dist_dependency_conflicts,
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
@@ -15,7 +15,7 @@
 from logging import getLogger
 from os import environ
 
-from importlib.metadata import entry_points
+from importlib.metadata import entry_points, EntryPoint, distributions
 
 from opentelemetry.instrumentation.dependencies import (
     get_dist_dependency_conflicts,
@@ -29,6 +29,29 @@ from opentelemetry.instrumentation.environment_variables import (
 from opentelemetry.instrumentation.version import __version__
 
 _logger = getLogger(__name__)
+
+
+class _EntryPointDistFinder:
+    def __int__(self):
+        self._mapping = None
+
+    def dist_for(self, entry_point: EntryPoint):
+        dist = getattr(entry_point, "dist", None)
+        if dist:
+            return dist
+
+        if self._mapping is None:
+            self._mapping = {
+                self._key_for(ep): dist
+                for ep in dist.entry_points
+                for dist in distributions()
+            }
+
+        return self._mapping.get(self._key_for(entry_point))
+
+    @staticmethod
+    def _key_for(entry_point: EntryPoint):
+        return f"{entry_point.group}:{entry_point.name}:{entry_point.value}"
 
 
 def _load_distro() -> BaseDistro:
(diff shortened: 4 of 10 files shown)


ocelotl:
@Rodrigo-Novas please sign the CLA :v:

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
```diff
@@ -19,7 +19,7 @@
 from re import sub
 from shutil import which
 
-from pkg_resources import iter_entry_points
```
mariojonke, line 22: you might want to remove the `setuptools` dependency in the [pyproject.toml](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/pyproject.toml#L28) file since it shouldn't be required anymore.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py
```diff
@@ -84,17 +84,10 @@ def _pip_check():
...
-        pkg_resources.get_distribution(req)
-    except pkg_resources.DistributionNotFound:
-        return False
-    except pkg_resources.VersionConflict as exc:
-        logger.warning(
-            "instrumentation for package %s is available but version %s is installed. Skipping.",
-            exc.req,
-            exc.dist.as_requirement(),  # pylint: disable=no-member
-        )
+        distribution(req)
```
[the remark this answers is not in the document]
Rodrigo-Novas: It would seem that this snnipet is correct, I don't see any improvements.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
```diff
@@ -48,8 +48,8 @@ def run() -> None:
 
     argument_otel_environment_variable = {}
 
-    for entry_point in iter_entry_points(
-        "opentelemetry_environment_variables"
+    for entry_point in entry_points().get(
```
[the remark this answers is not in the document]
Rodrigo-Novas: Okay, maybe we can leave this for the future, there is also the possibility of using
`entry_points(group="opentelemetry_distro") ` but in older versions like 3.9 and 3.8 the group parameter does not exist, so I don't think it is a valid option

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -63,10 +63,10 @@ def _load_instrumentors(distro):
         # to handle users entering "requests , flask" or "requests, flask" with spaces
         package_to_exclude = [x.strip() for x in package_to_exclude]
 
-    for entry_point in iter_entry_points("opentelemetry_pre_instrument"):
+    for entry_point in entry_points().get("opentelemetry_pre_instrument", []):
         entry_point.load()()
 
-    for entry_point in iter_entry_points("opentelemetry_instrumentor"):
+    for entry_point in entry_points().get("opentelemetry_instrumentor", []):
```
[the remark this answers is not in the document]
Rodrigo-Novas: that snipet seems fine to me

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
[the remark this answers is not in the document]
mariojonke: the `opentelmetry-api` is also making use of `entry_points` and there it was solved by taking a dependency on the [importlib_metdata](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/pyproject.toml#L30) (imports are happening [here|https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/util/_importlib_metadata.py]). This might have some side effects though, see: https://github.com/open-telemetry/opentelemetry-python/issues/3234.

Another option would be to implement/use a unified version of the `entry_points` API that works with the python builtin `importlib.metadata` module. E.g. the code below should work with all the current python versions:
```python
from sys import version_info

if version_info.minor in (8, 9):
    # in python 3.8 and 3.9 'entry_points' returns a dict. For a unified API
    #   the 'EntryPoints' class is (partly) copied from a later release
    #   https://github.com/python/importlib_metadata/blob/v6.1.0/importlib_metadata/__init__.py#L277
    #

    import itertools
    from importlib.metadata import EntryPoint
    from importlib.metadata import entry_points as _importlib_entry_points
    from importlib.metadata import version

    def _matches_ep(entry_point: EntryPoint, **params) -> bool:
        return (
            params.get("name", entry_point.name) == entry_point.name
            and params.get("group", entry_point.group) == entry_point.group
            and params.get("value", entry_point.value) == entry_point.value
        )

    class EntryPoints(tuple):
        __slots__ = ()

        def __getitem__(self, name):  # -> EntryPoint:
            try:
                return next(iter(self.select(name=name)))
            except StopIteration:
                raise KeyError(name)

        def select(self, **params):
            return EntryPoints(
                ept for ept in self if _matches_ep(ept, **params)
            )

        @property
        def names(self):
            return {ep.name for ep in self}

        @property
        def groups(self):
            return {ep.group for ep in self}

    def entry_points(**params) -> EntryPoints:
        eps = itertools.chain.from_iterable(_importlib_entry_points().values())
        return EntryPoints(eps).select(**params)

elif version_info.minor in (10, 11):
    # in 3.10 and 3.11 when called without parameters 'entry_points' returns
    #   a dict-like (SelectableGroups) object. Called with parameters a list-like
    #   (EntryPoints) object is returned. For a unified API the SelectableGroups
    #   return value is avoided and always the 'EntryPoints' object is returned.

    from importlib.metadata import version  # pylint:disable=no-name-in-module
    from importlib.metadata import (  # pylint:disable=no-name-in-module
        EntryPoint,
        EntryPoints,
        SelectableGroups,
    )
    from importlib.metadata import (
        entry_points as _importlib_entry_points,  # pylint:disable=no-name-in-module
    )

    def entry_points(**params) -> EntryPoints:
        eps = _importlib_entry_points(**params)
        if isinstance(eps, SelectableGroups):
            return getattr(eps, "_all")
        return eps

else:
    from importlib.metadata import (  # pylint:disable=no-name-in-module
        EntryPoint,
        EntryPoints,
        entry_points,
        version,
    )

__all__ = ["entry_points", "version", "EntryPoint", "EntryPoints"]
```

Which of the two approaches is the better, i don't know.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py
```diff
@@ -17,7 +17,8 @@
 import subprocess
 import sys
 
-import pkg_resources
+from packaging.requirements import Requirement
```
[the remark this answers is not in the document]
Rodrigo-Novas: mmm i think maybe the 18.0, because that version solved the https://github.com/pypa/packaging/issues/94 issue related with the requirements.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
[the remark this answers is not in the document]
Rodrigo-Novas: Using a constraint on a version of importlib-metadata would ensure that we could use groups and thus avoid having to use .get on Entry_Points and that's a good thing. On the other hand, I see that we would be restricting a version which could cause compatibility problems.
There is an interesting [post](https://iscinumpy.dev/post/bound-version-constraints/#when-is-it-okay-to-set-an-upper-limit) that report this.

The second approach is fine but I think it would make it difficult for us to maintain the code in the future, every time Python releases a new version we would have to update it and perhaps have to implement a new method if the imp changes. I think it would add an unnecessary workload to us in the future. Additionally, we would be repeating already made code.


If I have to take sides with something, it is to use .get() for the moment.
**But in case I cannot use the .get() approach I would use the version constraint solution**

opentelemetry-instrumentation/src/opentelemetry/instrumentation/dependencies.py
```diff
@@ -27,9 +22,9 @@ def __str__(self):
 def get_dist_dependency_conflicts(
     dist: Distribution,
 ) -> Optional[DependencyConflict]:
-    main_deps = dist.requires()
+    main_deps = dist.requires
     instrumentation_deps = []
-    for dep in dist.requires(("instruments",)):
+    for dep in requires(("instruments",)):
```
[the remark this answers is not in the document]
Rodrigo-Novas: I agree it is a good approach

opentelemetry-instrumentation/pyproject.toml
```diff
@@ -25,8 +25,9 @@ classifiers = [
 ]
 dependencies = [
   "opentelemetry-api ~= 1.4",
-  "setuptools >= 16.0",
   "wrapt >= 1.0.0, < 2.0.0",
+  "importlib-metadata >= 6.0, < 7.0",
```
mariojonke, line 29: the upper bound probably should be `< 8.0`, see https://github.com/open-telemetry/opentelemetry-python/issues/3570 (or maybe no boundary at all, but not sure about that)

mariojonke (approved):

mariojonke (approved):

opentelemetry-instrumentation/pyproject.toml
Rodrigo-Novas: Although I don't like version constraints, in this case I would leave it because of cpython's reputation for changing its interface 🤣

WillDaSilva:
I believe this would also fix:
- https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1970

psi29a:
Hope this gets picked back up! Great work so far!

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
```diff
@@ -19,7 +19,7 @@
 from re import sub
 from shutil import which
 
-from pkg_resources import iter_entry_points
+from importlib.metadata import entry_points
```
mlasch, line 22: You are adding `importlib-metadata` to the dependencies, but this import is from the Python standard library. There is a discussion here: https://github.com/open-telemetry/opentelemetry-python/issues/3234 about using  `importlib-metadata` and its implications. Note my comment at the end.

Rodrigo-Novas: Hello, sorry for the delay. 
If I understand correctly, you mean having a standard library with its own constraint within Opentelemetry would be wrong. I also understand that the change of importlib-metadata version could be made later if an error is reported.

https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2181#discussion_r1504854087

pablo-sumup:
Hey there! Could you prioritize the code review for this PR? It's for migrating to Python 3.12. Thanks a bunch! :D

theunkn0wn1:
Any ETA on this?
the package is broken in modern python versions due to undeclared `pkg_resources` usage.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
theunkn0wn1: IMHO its fine to depend on `importlib-metadata`, as its the python-version independent backport of the stdlib feature.
it will be easier to develop against.

lzchen:
@Rodrigo-Novas 

Are you still working on this? If so we can prioritize getting this reviewed and merged. It is difficult to know what the current status is so please close the conversations that you think have been resolved and note down the open topics that still need to be discussed. Feel free to also join the Python SIG if you want to bring this issue up to the community specifically.

michaelgmiller1:
We are also getting hit with this on every unit test run. It would be great to get a fix in.

Rodrigo-Novas:
> @Rodrigo-Novas
> 
> Are you still working on this? If so we can prioritize getting this reviewed and merged. It is difficult to know what the current status is so please close the conversations that you think have been resolved and note down the open topics that still need to be discussed. Feel free to also join the Python SIG if you want to bring this issue up to the community specifically.

@lzchen 

Its ok for me, you can review it!

Rodrigo-Novas:
> We are also getting hit with this on every unit test run. It would be great to get a fix in.

i will try to fix it

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `opentelemetry-bootstrap` Remove `opentelemetry-instrumentation-aws-lambda` from the defaults instrumentations
   ([#2786](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2786))
+- Deprecation of pkg_resource in favor of importlib.metadata
+  ([#2180](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2180))
```
lzchen, line 21:
```suggestion
  ([#2181](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2181))
```

opentelemetry-instrumentation/pyproject.toml
```diff
@@ -26,8 +26,9 @@ classifiers = [
 ]
 dependencies = [
   "opentelemetry-api ~= 1.4",
-  "setuptools >= 16.0",
   "wrapt >= 1.0.0, < 2.0.0",
+  "importlib-metadata >= 6.0, < 8.0",
```
Arnatious, line 30: We already vendor importlib-metadata in opentelemetry-api (https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/util/_importlib_metadata.py) - should we reuse this rather than re-introduce the dependency?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
```diff
@@ -48,8 +47,8 @@ def run() -> None:
 
     argument_otel_environment_variable = {}
 
-    for entry_point in iter_entry_points(
-        "opentelemetry_environment_variables"
+    for entry_point in entry_points(
```
lzchen, line 50: I don't think this method of selection is supported for Python 3.8 and 3.9 -> https://docs.python.org/3/library/importlib.metadata.html#entry-points

Arnatious: Related to my earlier comment https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2181#discussion_r1737100608 - if we use the importlib_metadata already taken as a dependency in opentelemetry-api, it's version >=6.4.0 which supports this syntax.

Arnatious:
```
>>> for entry_point in importlib_metadata.entry_points(group="console_scripts"):
...   print(entry_point)
... 
EntryPoint(name='pip', value='pip._internal.cli.main:main', group='console_scripts')
EntryPoint(name='pip3', value='pip._internal.cli.main:main', group='console_scripts')
EntryPoint(name='pip3.9', value='pip._internal.cli.main:main', group='console_scripts')
>>> importlib_metadata.version('importlib_metadata')
'6.4.0'
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -31,9 +30,32 @@
...
+
+    def dist_for(self, entry_point: EntryPoint):
+        dist = getattr(entry_point, "dist", None)
+        if dist:
+            return dist
+
+        if self._mapping is None:
+            self._mapping = {
+                self._key_for(ep): dist
+                for ep in dist.entry_points
```
lzchen, line 45: Is there a more elegant way to obtain the `dist` for a specific `entrypoint` rather than generating and storing info on all installed `dist` s?

Arnatious: I don't believe so - importlib doesn't do any sort of caching like pkg_resources did, so this is the most performant way to do this lookup.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -31,9 +30,32 @@
 _logger = getLogger(__name__)
 
 
+class _EntryPointDistFinder:
+    def __int__(self):
+        self._mapping = None
```
Arnatious, line 35: Consider making this a `functools.cached_property`?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
Rodrigo-Novas: Maybe I'm misunderstanding but we discuss about that here https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2181#discussion_r1494303135.

And here https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2181#discussion_r1504854087 about the version constraint.

> Using a constraint on a version of importlib-metadata would ensure that we could use groups and thus avoid having to use .get() on Entry_Points and that's a good thing. On the other hand, I suppose that we would be restricting a version which could cause compatibility problems.
> There is an interesting [post](https://iscinumpy.dev/post/bound-version-constraints/#when-is-it-okay-to-set-an-upper-limit) that report this.
> 
> The second approach is fine but I think it would make it difficult for us to maintain the code in the future, every time Python releases a new version we would have to update it and perhaps have to implement a new method if the imp changes. I think it would add an unnecessary workload to us in the future. Additionally, we would be repeating already made code.
> 
> If I have to take sides with something, it is to use .get() at the moment.
> But in case we cannot use the .get() approach I would use the version constraint solution


@xrmx @Arnatious

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
Rodrigo-Novas: Something like this?:

```
    @cached_property
    def dist_for(self, entry_point: EntryPoint):
        dist = getattr(entry_point, "dist", None)
        if dist:
            return dist

        if self._mapping is None:
            self._mapping = {
                self._key_for(ep): dist
                for ep in dist.entry_points
                for dist in distributions()
            }

        return self._mapping.get(self._key_for(entry_point))
```



This would always return the same dist, right?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -58,23 +80,25 @@ def _load_distro() -> BaseDistro:
...
         if entry_point.name in package_to_exclude:
             _logger.debug(
                 "Instrumentation skipped for library %s", entry_point.name
             )
             continue
 
         try:
-            conflict = get_dist_dependency_conflicts(entry_point.dist)
+            entry_point_dist = entry_point_finder.dist_for(entry_point)
+            conflict = get_dist_dependency_conflicts(entry_point_dist)
```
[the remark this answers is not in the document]
Rodrigo-Novas: entry_point.dist used in [line 77](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2181/files#diff-6ecd6cbdf97812bf55619474db901385d54a6a6931b91b6b2fc5b308b1fb8ab3R77) doesn't seem to exist in [older python versions](https://github.com/python/cpython/blob/3.8/Lib/importlib/metadata.py#L41) and is an optional attribute in the [latest version](https://github.com/python/cpython/blob/3.12/Lib/importlib/metadata/__init__.py#L194).

opentelemetry-instrumentation/pyproject.toml
Rodrigo-Novas: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2181#discussion_r1739214916

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
Arnatious: I'm suggesting replacing all uses of `importlib.metadata` or `importlib_metadata` with `opentelemetry.util._importlib_metadata` and taking it as a transitive dependency from opentelemetry-api itself.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
Arnatious: I meant the _mapping attribute within _EntrypointDistFinder - it'll do exactly what you're already doing and only compute _mapping once

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -31,9 +30,32 @@
 _logger = getLogger(__name__)
 
 
+class _EntryPointDistFinder:
+    def __int__(self):
```
Arnatious, line 34: Typo? `__int__` vs `__init__`

(Can probably just be deleted if https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2181/files#r1739126880 is accepted)

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
Rodrigo-Novas: Okay, my bad. It seems reasonable to use the alias.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
Rodrigo-Novas: Okay yes, I agree

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
Rodrigo-Novas: Yes it is a typo

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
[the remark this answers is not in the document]
xrmx: If we are going to use importlib_metadata (via opentelemetry api util) we just need to be assured that it's there since 6+ no? Looking at some old pkg_resources code, dist was default None there too.

serkan-ozal:
Switching from `pkg_resource` to `importlib.metadata` also helps us to reduce OTEL Python layer Lambda coldstart.

bhaskarbanerjee:
When is this PR expected to be approved and merged?

xrmx:
> When is this PR expected to be approved and merged?

When it's ready

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
ocelotl: Yes, please use `opentelemetry.util._importlib_metadata`. We defnitely *don't* want to use neither `importlib.metadata` nor `imporlib_metadata` anywhere else outside of `opentelemetry.util._importlib_metadata`. The implementation of `importlib.metadata` and `importlib_metadata` is plagued with API changes between Python versions and we want to contain all the issues this causes into one single place.

BTW, since there is a lot of interest in moving this forward, I added some commits on top of your PR in #2854. I cannot continue debugging the failing test cases right now but I encourage you to use the code in #2854, which removes usage of `importlib.metadata` in test requirements files and uses `opentelemetry.util._importlib_metadata` in source code.

Also, you'll notice that the `CORE_REPO_SHA` environment variable in #2854 is set to a particular commit, that commit should always be the latest commit of [this](https://github.com/open-telemetry/opentelemetry-python/pull/4177). Once that PR gets merged, the value of `CORE_REPO_SHA` should be set back to `main` in the `tox.ini` file and the `.github/workflows/*.yml` files.

ocelotl:
> When is this PR expected to be approved and merged?

We are trying to get more eyes on this PR, we realize it is important for many people. The issue at hand here is not particularly easy to solve (kudos to @Rodrigo-Novas for the work so far!) since it deals with replacing an old library with a new one where there is not necessarily a 1:1 relationship between old library and new library components. To make matters even more complicated, the new library has been implemented very inconsistently, so its behavior for Python 3.x is not the same as for Python 3.y.

We are working on this and we'll merge as soon as we can :slightly_smiling_face:

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
xrmx: @Arnatious isn't that missing the dist parameter though?

xrmx: I've tried this (with the two for lines swapped) but it looks like the code is not used with our current tests, see 
https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2871/files#r1759148502

So the question is: do we really need this cache?

Arnatious:
For anyone following - this is currently blocked by (at minimum) https://github.com/open-telemetry/opentelemetry-python/pull/4177, as well as a new release needed for https://github.com/open-telemetry/opentelemetry-python/pull/4181 .  Some of our assumptions in the previous discussions mentioned need to be re-addressed if we swap to relying on the stdlib in python <= 3.11, which will delay this a bit further.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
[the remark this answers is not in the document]
Arnatious: We're not going to be guaranteed 6+ if https://github.com/open-telemetry/opentelemetry-python/pull/4177 goes through - we'll only have 6 on python 3.12 and higher.

xrmx:
> For anyone following - this is currently blocked by (at minimum) [open-telemetry/opentelemetry-python#4177](https://github.com/open-telemetry/opentelemetry-python/pull/4177), as well as a new release needed for [open-telemetry/opentelemetry-python#4181](https://github.com/open-telemetry/opentelemetry-python/pull/4181) . Some of our assumptions in the previous discussions mentioned need to be re-addressed if we swap to relying on the stdlib in python <= 3.11, which will delay this a bit further.

We're keeping importlib_metadata as dependency

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
Arnatious: Looks like according to coverage everything we test has a dist attribute, and therefore exits before the call to ._mapping

Arnatious: Granted, not sure that's necessarily something _this_ pr needs to fix.

multani:
@xrmx I think this should have been closed by #2871, right?

xrmx:
Closing, since this has been included in #2871. Thanks everyone!

PR #2260, opened 2024-02-23
Fix HTTP header normalization

samuelcolvin:
# Description

HTTP request and response headers (attributes `http.request.header.*` and `http.response.header.*`) should be normalized by just being converted to lowercase, not replacing underscore with dash.

This is made clear in the docs [here](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/):

image
image

The current incorrect behavior of `replace('_', '-')` is extremely problematic since you can't know what the original header was - e.g. if you see a key `http.request.header.foo_bar` what was the original value? It could have been `foo-bar` or `foo_bar`.

## Type of change

# How Has This Been Tested?

- [ ] replaced all the tests expecting the incorrect behavior

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
@@ -132,7 +132,7 @@ is the normalized HTTP header name (lowercase, with ``-`` replaced by ``_``). Th
 single item list containing all the header values.
 
 For example:
-``http.request.header.custom_request_header = ["<value1>,<value2>"]``
+``http.request.header.custom-request-header = ["<value1>,<value2>"]``
 
 Response headers
 ****************
@@ -166,7 +166,7 @@ is the normalized HTTP header name (lowercase, with ``-`` replaced by ``_``). Th
 single item list containing all the header values.
 
 For example:
-``http.response.header.custom_response_header = ["<value1>,<value2>"]``
+``http.response.header.custom-response-header = ["<value1>,<value2>"]``
 
 Sanitizing headers
 ******************
--- a/instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_custom_headers.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_custom_headers.py
@@ -105,17 +105,17 @@ class TestCustomHeaders(AsgiTestBase, TestBase):
         self.get_all_output()
         span_list = self.exporter.get_finished_spans()
         expected = {
-            "http.request.header.custom_test_header_1": (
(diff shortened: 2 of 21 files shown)


instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -132,7 +132,7 @@ def client_response_hook(span: Span, message: dict):
 single item list containing all the header values.
```
federicobond, line 132: Note in the line above this one that it still says it replaces `-` with `_`. We should leave "(lowercase)" only.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -166,7 +166,7 @@ def client_response_hook(span: Span, message: dict):
...
```
federicobond, line 166: Same here.

federicobond (approved):

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
samuelcolvin: good point, fixed.

samuelcolvin:
FWIW those failures are on an install task, and i'm seeing them locally too while working on #2266.

lzchen:
@aabmass 

This might be a bigger discussion on how we should be handling the new http semantic convention migrations. For now, please hold off on this PR. Will bring this up in the weekly SIG.

samuelcolvin:
@lzchen can we at least run tests, so I know it's passing.

Also, if this won't get merged yet, could we merge #2266 instead first please? There's no "http semantic convention migrations" issue there, just a bug.

aabmass:
> can we at least run tests, so I know it's passing.

Running now

> Also, if this won't get merged yet, could we merge #2266 instead first please? There's no "http semantic convention migrations" issue there, just a bug.

That SGTM as it's just a bug 👍

lzchen:
From discussion of 07/11/2024 SIG, just an update on what is blocking this PR. Some of the instrumentations that this pr touches still has not been migrated to new semantic conventions as of yet, most notably django, falcon, pyramid, starlette and tornado. Since `django` is the only one remaining for the [stable http semantic migration plan](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2453), once that instrumentation is migrated, we can go ahead with this change as we have agreed that it is fine to make breaking/semantic convention changes outright to the other instrumentations since they are all in beta.

PR #2266, opened 2024-02-23
avoid losing repeated HTTP headers

samuelcolvin:
# Description

Currently with some http middleware (ASGI and WSGI), if you send multiple headers with the same key, only the last value is kept. This is obviously wrong.

**Subtlety:** [the docs](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#:~:text=is%20not%20recommended.-,The%20attribute%20value%20MUST%20consist%20of%20either%20multiple%20header%20values%20as%20an%20array%20of%20strings%20or%20a%20single%2Ditem%20array%20containing%20a%20possibly%20comma%2Dconcatenated%20string%2C%20depending%20on%20the%20way%20the%20HTTP%20library%20provides%20access%20to%20headers.,-%5B2%5D%3A%20The) allow you to use either a list of header values where the keys are repeated or a single item list with the values comma separated.

Obviously a list of each value would be preferable (no information is lost, easier to reconstruct), however:
* this would means changing the signature of numerous public methods
* and, changing the docstring which perversely (presumably according to some old spec) specificly say you'll get a single item list with values comma seprated

I've therefor chosen to do the non-breaking but less correct thing so far, happy to change if that's preferred - if so, please let me know how it should be changed.

I also assume doing it this way avoids issues around `OTEL_SEMCONV_STABILITY_OPT_IN`.

## Type of change

# How Has This Been Tested?

- [x] test are updated

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -347,11 +347,17 @@ def collect_custom_headers_attributes(
      - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers
     """
     # Decode headers before processing.
-    headers: dict[str, str] = {
-        _key.decode("utf8"): _value.decode("utf8")
-        for (_key, _value) in scope_or_response_message.get("headers")
-        or cast("list[tuple[bytes, bytes]]", [])
-    }
+    headers: dict[str, str] = {}
```
samuelcolvin, line 350: I can use `defaultdict` if you prefer.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -347,11 +347,17 @@ def collect_custom_headers_attributes(
...
+    raw_headers = scope_or_response_message.get("headers")
+    if raw_headers:
+        for _key, _value in raw_headers:
+            key = _key.decode().lower()
```
samuelcolvin, line 354: `lower()` now to avoid loosing repeats if you had

```http
Foo: v1
foo: v2
```

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_custom_headers.py
```diff
@@ -121,6 +139,25 @@ def test_http_custom_request_headers_in_span_attributes(self):
...
+        self.seed_app(self.app)
+        self.send_default_request()
+        self.get_all_output()
+        span_list = self.exporter.get_finished_spans()
+        expected = {
+            "http.request.header.custom_test_header_1": (
+                "test-header-value-1,test-header-value-2",
+            ),
+        }
+        span = next(span for span in span_list if span.kind == SpanKind.SERVER)
```
samuelcolvin, line 158: other tests do a slightly dangerous thing where they'd fail if no `span` had `span.kind == SpanKind.SERVER` but I've left them unchanged to avoid increasing the diff.

samuelcolvin:
cc @adriangb as we were discussing this.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
samuelcolvin: wrong, ignore me, wouldn't work unless we had `dict[str, list[str]`.

samuelcolvin:
@aabmass you mentioned in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2260#issuecomment-1961830196 this this sounds good to you, please could this be reviewed (or at least have tests started).

xrmx (approved):

samuelcolvin:
I have resolved conflicts again, is there anything stopping this from being merged?

ocelotl (approved):

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -358,12 +358,17 @@ def collect_custom_request_headers_attributes(environ):
...
-    headers = {
-        key[_CARRIER_KEY_PREFIX_LEN:].replace("_", "-"): val
-        for key, val in environ.items()
-        if key.startswith(_CARRIER_KEY_PREFIX)
-    }
+    for key, val in environ.items():
+        if key.startswith(_CARRIER_KEY_PREFIX):
+            header_key = (
+                key[_CARRIER_KEY_PREFIX_LEN:].replace("_", "-").lower()
+            )
```
aabmass, line 367: I don't think any changes are needed to the WSGI instrumentation. The [WSGI/CGI specs](https://datatracker.ietf.org/doc/html/draft-coar-cgi-v11-03#:~:text=Meta%2Dvariables%20with%20names,onto%20a%20single%20line.) says that HTTP headers should already be canonicalized into upper case with underscores and since it's a dictionary, multiple header values should be merged with `,`

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
```diff
@@ -671,6 +671,22 @@ def test_custom_request_header_added_in_server_span(self):
...
+        headers = [
+            ("Custom-Test-Header-1", "Test Value 1"),
+            ("Custom-Test-Header-1", "Test Value 2"),
+        ]
+        resp = self.client.get("/hello/123", headers=headers)
+        self.assertEqual(200, resp.status_code)
+        span = self.memory_exporter.get_finished_spans()[0]
+        expected = {
+            "http.request.header.custom_test_header_1": (
+                "Test Value 1, Test Value 2",
```
aabmass, line 684: Is the space after the comma expected?

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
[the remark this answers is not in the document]
samuelcolvin: The file has `from __future__ import annotations`

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
samuelcolvin: For request headers I think you're right, but presumably response headers do need to be changed.

aabmass (commented):
> **Subtlety:** [the docs](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#:~:text=is%20not%20recommended.-,The%20attribute%20value%20MUST%20consist%20of%20either%20multiple%20header%20values%20as%20an%20array%20of%20strings%20or%20a%20single%2Ditem%20array%20containing%20a%20possibly%20comma%2Dconcatenated%20string%2C%20depending%20on%20the%20way%20the%20HTTP%20library%20provides%20access%20to%20headers.,-%5B2%5D%3A%20The) allow you to use either a list of header values where the keys are repeated or a single item list with the values comma separated.

In the case of WSGI request headers which come pre-joined, I think it makes sense to leave them comma separated. For WSGI response headers and ASGI overall, where we have access to repeated headers, it might make sense to use the list for the reasons you mentioned. I think that is the sprit of the spec when it says "depending on the way the HTTP library provides access to headers".

> this would means changing the signature of numerous public methods
> and, changing the docstring which perversely (presumably according to some old spec) specificly say you'll get a single item list with values comma seprated

@ocelotl @xrmx any thoughts on this?

I was actually confused reading the code, because It looks like this signature for [`SanitizeValue.sanitize_header_values()`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/e923938646341c2ec54d7ba55cb486b1a2b6e7b7/util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py#L88-L93) is already wrong (it's returning a `dict[str, list[str]]` 😓 Were there other signatures?

ocelotl:
@samuelcolvin falcon test cases are failing, please take a look.

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
samuelcolvin: yes, it's added by WSGI when it concatenates the values while setting environment variables.

samuelcolvin:
> @samuelcolvin falcon test cases are failing, please take a look.

Thanks for reviewing @ocelotl, tests are hopefully now passing. Everything I've run has passed.

xrmx:
> > this would means changing the signature of numerous public methods
> > and, changing the docstring which perversely (presumably according to some old spec) specificly say you'll get a single item list with values comma seprated
> 
> @ocelotl @xrmx any thoughts on this?

No particular thoughts except that maybe we can track this on a new issue 
 
> I was actually confused reading the code, because It looks like this signature for [`SanitizeValue.sanitize_header_values()`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/e923938646341c2ec54d7ba55cb486b1a2b6e7b7/util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py#L88-L93) is already wrong (it's returning a `dict[str, list[str]]` 😓 Were there other signatures?

Fixing typing is considered breaking compatibility?

aabmass:
> Fixing typing is considered breaking compatibility?

I would say this is clearly a bug. I think that was the only signature to change but I could be wrong. I also doubt any "user" is using these private methods.

samuelcolvin:
Well the methods aren't private, which makes them public IMHO.

In Pydantic we've been bitten a lot by people using non-private methods even when they're not documented.

The current situation is a definite bug, and this is strictly better than what's on `main`, so I would suggest merging this and discussion further in a new issue?

samuelcolvin:
> Fixing typing is considered breaking compatibility?

@xrmx To be clear, I'm not talking about just type hints, to fix this "properly" we need to change some methods to return `dict[str, str]` to `dict[str, list[str]`.

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
aabmass: Right but I would have expected them to be comma joined by the http spec.

Looks specific to the [werkzeug test client's WSGI impl](https://github.com/pallets/werkzeug/blob/32e69512134c2f8183c6438b2b2e13fd24e9d19f/src/werkzeug/test.py#L747-L748). Just wanted to check that our instrumentation isn't doing something weird

aabmass (approved):
Thanks for the PR!

aabmass:
> @xrmx To be clear, I'm not talking about just type hints, to fix this "properly" we need to change some methods to return `dict[str, str]` to `dict[str, list[str]`.

Which methods do you mean? I think the existing functions like `collect_custom_headers_attributes()` and `collect_custom_response_headers_attributes` are already returning `dict[str, list[str]]` because the signature of `sanitize_header_values()` is wrong

aabmass:
I guess you mean this parameter would need to changed https://github.com/open-telemetry/opentelemetry-python-contrib/blob/113d346b5b301d1f198f153a3c5dcdcaba81f5ce/util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py#L90. I think we could add an overload for a `dict[str, list[str]]` input

samuelcolvin:
Okay, I'll update those functions to work with header values as lists.

samuelcolvin:
> Okay, I'll update those functions to work with header values as lists.

Okay, I've re-read. I think this is ready and follow can happen on #2360 and a new PR.

ocelotl:
> > Okay, I'll update those functions to work with header values as lists.
> 
> Okay, I've re-read. I think this is ready and follow can happen on #2360 and a new PR.

@samuelcolvin tests are failing, please take a look. @aabmass are we good to merge once these tests are fixed?

ocelotl:
> > > Okay, I'll update those functions to work with header values as lists.
> > 
> > 
> > Okay, I've re-read. I think this is ready and follow can happen on #2360 and a new PR.
> 
> @samuelcolvin tests are failing, please take a look. @aabmass are we good to merge once these tests are fixed?

Disregard, seems like the test case failures are unrelated to this PR. :v:

aabmass:
@ocelotl good to merge?

ocelotl:
> @ocelotl good to merge?

good to merge :sunglasses:

PR #2311, opened 2024-02-28
Remove [test] package for django instrumentation

### LINKED ISSUE #2198: Remove [test] package from django instrumentation
See #2182

ocelotl:
Fixes #2198

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -52,10 +52,18 @@ envlist =
     ; Only officially supported Python versions are tested for each Django
     ; major release. Updated list can be found at:
     ; https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django
+    ; The numbers at the end of the environment names
+    ; below mean these dependencies are being used:
+    ; 0: django~=2.0
+    ; 1: django~=3.0
+    ; 2: django>=4.0b1,<5.0 backports.zoneinfo==0.2.1
+    ; 3: django>=4.0b1,<5.0
+    py3{8,9}-test-instrumentation-django-0
+    py3{8,9}-test-instrumentation-django-1
     py3{8,9}-test-instrumentation-django-2
-    py3{8,9,10,11}-test-instrumentation-django-3
-    py3{8,9,10,11}-test-instrumentation-django-4
-    pypy3-test-instrumentation-django-{2,3}
+    py3{10,11}-test-instrumentation-django-1
+    py3{10,11}-test-instrumentation-django-3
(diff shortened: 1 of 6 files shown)


tox.ini
```diff
@@ -539,7 +550,7 @@ commands_pre =
   python -m pip install -e {toxinidir}/instrumentation/opentelemetry-instrumentation-asgi[test]
   python -m pip install -e {toxinidir}/instrumentation/opentelemetry-instrumentation-botocore[test]
   python -m pip install -e {toxinidir}/instrumentation/opentelemetry-instrumentation-boto3sqs[test]
-  python -m pip install -e {toxinidir}/instrumentation/opentelemetry-instrumentation-django[test]
+  pip install -r {toxinidir}/instrumentation/opentelemetry-instrumentation-django/test-requirements-3.txt
```
xrmx, line 553: Keep `python -m pip` as other packages?

tox.ini
```diff
@@ -52,10 +52,18 @@ envlist =
     ; Only officially supported Python versions are tested for each Django
     ; major release. Updated list can be found at:
     ; https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django
+    ; The numbers at the end of the environment names
+    ; below mean these dependencies are being used:
+    ; 0: django~=2.0
+    ; 1: django~=3.0
+    ; 2: django>=4.0b1,<5.0 backports.zoneinfo==0.2.1
+    ; 3: django>=4.0b1,<5.0
+    py3{8,9}-test-instrumentation-django-0
```
xrmx, line 61: Do you think it would be possible to use the major version of django instead of using an enumeration for the requirements filenames?

lzchen: I think @aabmass pointed this out as well. Seems like we do have a number people who feels that is a better way to indicate different tests. @ocelotl

ocelotl: Yeah, we agreed on this approach already.  We can still change it later in another PR. This is only needed for some instrumentarions.

tox.ini
ocelotl: No, we are installing from a requirements file here and we don't need to call Python but pip directly

tox.ini
xrmx: I'm reviewing all your PR and am approving the ones that does not have this issue. Since there will be conflicts anyway I think it would be better to update current PRs instead of opening new ones later so you can reuse the current numbering convention and will be easier to review

lzchen (approved):

PR #2361, opened 2024-03-20
Repeated headers list for ASGI frameworks

### LINKED ISSUE #2358: `SanitizeValue.sanitize_header_values()` type signature is incorrect
> **Subtlety:** [the docs](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#:~:text=is%20not%20recommended.-,The%20attribute%20value%20MUST%20consist%20of%20either%20multiple%20header%20values%20as%20an%20array%20of%20strings%20or%20a%20single%2Ditem%20array%20containing%20a%20possibly%20comma%2Dconcatenated%20string%2C%20depending%20on%20the%20way%20the%20HTTP%20library%20provides%20access%20to%20headers.,-%5B2%5D%3A%20The) allow you to use either a list of header values where the keys are repeated or a single item list with the values comma separated.

In the case of WSGI request headers which come pre-joined, I think it makes sense to leave them comma separated. For WSGI response headers and ASGI overall, where we have access to repeated headers, it might make sense to use the list for the reasons you mentioned. I think that is the sprit of the spec when it says "depending on the way the HTTP library provides access to headers".

> this would means changing the signature of numerous public methods
> and, changing the docstring which perversely (presumably according to some old spec) specificly say you'll get a single item list with values comma seprated

@ocelotl @xrmx any thoughts on this?

I was actually confused reading the code, because It looks like this signature for [`SanitizeValue.sanitize_header_values()`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/e923938646341c2ec54d7ba55cb486b1a2b6e7b7/util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py#L88-L93) is already wrong (it's returning a `dict[str, list[str]]` 😓 Were there other signatures?

_Originally posted by @aabmass in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2266#pullrequestreview-1938012361_

### LINKED ISSUE #2360: Preserve multiple HTTP header values as a list instead of a joined string
Right now, HTTP headers are stored as list type span attribute with one item. If there are repeated headers, they are joined with a comma as (being implemented in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2266). We could instead preserve each header value as a separate string.

Current behavior:
```
{
  "http.request.header.custom_test_header_1": ("test-header-value-1,test-header-value-2",),
}
```
Proposed behavior
```
{
  "http.request.header.custom_test_header_1": ("test-header-value-1", "test-header-value-2"),
}
```

This may require changing some public signatures or adding new overloads to keep backward compatibility.

samuelcolvin:
# Description

Use a list, rather than a comma separate string to represent repeated HTTP request and response headers.

fixes #2360
fixes #2358
followup to #2266

## Type of change

# How Has This Been Tested?

tests updated

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

- [x] Unit tests have been added/updated


instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -339,24 +340,19 @@ def collect_custom_headers_attributes(
...
     if raw_headers:
-        for _key, _value in raw_headers:
-            key = _key.decode().lower()
-            value = _value.decode()
-            if key in headers:
-                headers[key] += f",{value}"
-            else:
-                headers[key] = value
+        for key, value in raw_headers:
+            headers[key.decode()].append(value.decode())
```
xrmx, line 355: Is the removal of the .lower() intentional?

xrmx:
@samuelcolvin please rebase

samuelcolvin:
Okay, _think_ this is good.

I got so confused about all the wsgi based web frameworks, so use enviroment variables to pass request headers (and repsonse headers? 🤷 ) some don't, sometimes. It's not helped by the fact that lots of the tests combine "multiple" headers into one key-value pair with comma separation.

I therefore reverted all the changes in wsgi and flask, and concentrated on getting the ASGI based frameworks right.

srikanthccv (approved):

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -11,9 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 import unittest
 from timeit import default_timer
+from typing import Mapping, Tuple
```
xrmx, line 16: Mapping should be imported from `collections.abc` as the alias is deprecated

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
```diff
@@ -18,7 +18,7 @@
 from re import IGNORECASE as RE_IGNORECASE
 from re import compile as re_compile
 from re import search
-from typing import Callable, Iterable, Optional
+from typing import Callable, Iterable, Mapping, Optional
```
xrmx, line 21: Here is fine because subscription has been added in 3.9 AFAIU

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
```diff
@@ -14,11 +14,12 @@
 
 from __future__ import annotations
 
+from collections.abc import Mapping
```
xrmx, line 17: This should fail with old python because it's subscribed

ocelotl: What do you mean by subscribed?

xrmx: `Mapping[..]` like in line 91 https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2361/files#diff-9f93b3514e50dc2c60507d1deb1ad7bd80b8dcca02a5d1cb4bcec412b5cdcbaeR91

samuelcolvin:
Sorry, I've been busy and dropped the ball, I'll try to fix soon.

Samuel

util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
ocelotl: It didn't fail any test :shrug:

ocelotl (approved):

PR #2363, opened 2024-03-20
fix(grpc): aio grpc client interceptor incorrect metadata handling

legau:
# Description

This fixes a bug when using grpcio>=1.60.0 and the gRPC telemetry. Introduced by https://github.com/grpc/grpc/pull/34347 a tuple metadata is always converted to a `Metadata` object which follows the Python grpc docs and typing. In the current opentelemetry implementation there is a conversion to an `OrderedDict` which causes :
```python
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = Metadata((('key', 'value'),)), key = ('key', 'value')

    def __getitem__(self, key: MetadataKey) -> MetadataValue:
        """When calling <metadata>[<key>], the first element of all those
        mapped for <key> is returned.
        """
        try:
>           return self._metadata[key][0]
E           KeyError: ('key', 'value')
```

This PR makes use of the `Metadata` object directly and propagates it as such. 
This could cause regressions for some following interceptors handling metadata as tuples (same way this regression happened), I'm not sure if this would qualify as a breaking change though.

## Type of change

# How Has This Been Tested?

Added metadata in grpc tests

- [x] No.

# Checklist:


xrmx (approved):

aabmass:
Thanks for the PR @legau 

> This fixes a bug when using grpcio>=1.60.0 and the gRPC telemetry. Introduced by [grpc/grpc#34347](https://github.com/grpc/grpc/pull/34347)

Do you know if this change is backward compatible to older gRPC versions before v1.60.0?

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_aio_client.py
```diff
@@ -55,20 +54,19 @@ def callback(call):
...
     @staticmethod
-    def propagate_trace_in_details(client_call_details):
+    def propagate_trace_in_details(client_call_details: ClientCallDetails):
         metadata = client_call_details.metadata
         if not metadata:
-            mutable_metadata = OrderedDict()
+            mutable_metadata = Metadata()
         else:
-            mutable_metadata = OrderedDict(metadata)
+            mutable_metadata = Metadata(*tuple(metadata))
```
aabmass, line 62: It looks like this was based on the previous assumption that `metadata` was an immutable tuple. `Metadata` looks to be mutable and retains previously set values (as a multimap)–can we just pass the metadata directly?

aabmass (approved):

legau:
This transforms the tuple of <1.60 into a Metadata object so yes it is backwards compatible

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_aio_client.py
aabmass: I guess the current approach would work with tuple already. I'm OK to just leave it as is.

aabmass (approved):

legau:
I added the changelog entry

xrmx (approved):

PR #2374, opened 2024-03-27
fix: do not convert grpc.aio.Metadata to OrderedDict

### LINKED ISSUE #2373: `aio` interceptor fails because `grpc.aio.Metadata` doesn't correctly implement `abc.Mapping`
This is related to a known `grpc` bug: https://github.com/grpc/grpc/issues/26498

### Environment

I believe this applies to any environment using the current latest Python `grpc` and `opentelemetry-instrumentation-grpc` packages.

### Steps to reproduce

Perform a gRPC call using `aio`, with non-empty metadata and a configured interceptor (e.g. using autoinstrumentation).

To see a simple demonstration that the code in `propagate_trace_in_details` doesn't work, just build an example `Metadata` object in a Python terminal and attempt to construct an `OrderedDict` from it:

```
>>> from collections import OrderedDict
>>> from grpc.aio import Metadata
>>> m = grpc.aio.Metadata({"foo", "bar"})
>>> OrderedDict(m)
  File "<stdin>", line 1, in <module>
  File "<path>/.venv/lib/python3.9/site-packages/grpc/aio/_metadata.py", line 64, in __getitem__
    return self._metadata[key][0]
KeyError: ('bar', 'foo')
```

### What is the expected behavior?

The call should succeed, and a corresponding trace span should be created.

There is a key error indicating that lookup of a metadata `(key, value)` pair failed.

BatmanAoD:
# Description

gRPC metadata has traditionally been represented as nested tuples, so when injecting new metadata, interceptors needed to convert to a mutable type and then back into a nested tuple. However, for async gRPC, the `aio.Metadata` class breaks this conversion, and the conversion is unnecessary anyway since the data is already mutable.

Fixes #2373 

## Type of change

# How Has This Been Tested?

TODO

- [ ] Test A

- [x] No.

# Checklist:


BatmanAoD:
I'm not sure why I'm not covered under the CLA. When I click the "not covered" button and go through the process, it says I'm ready to go (as a corporate contributor under Rigetti), but the status of the EasyCLA check doesn't seem to change.

For comparison, in this PR, it marked me as covered: https://github.com/grpc/grpc-go/pull/7038

Is it perhaps because one of my commits here is unverified? (I'm not sure why that is, but I can try to fix it if that's the problem.)

thpierce:
I think this issue was resolved with https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2363, note that the issue persisted (for me) even after https://github.com/grpc/grpc/issues/26498 was completed.

BatmanAoD:
I haven't confirmed whether or not this is resolved, but this PR isn't mergeable anyway.

Issue #2296, opened 2024-02-25
Request hooks are not passed into middleware when using DjangoInstrumentor with gunicorn


halkony:
**Describe your environment** Running Gunicorn on Django inside of kubernetes, Python 3.10

**Steps to reproduce**
Define the following in `gunicorn.conf.py`
```
from opentelemetry.instrumentation.django import DjangoInstrumentor
import opentelemetry

def post_fork(server, worker):
    from opentelemetry.instrumentation.auto_instrumentation import sitecustomize
    server.log.info("Worker spawned (pid: %s)", worker.pid)
    
    def request_hook(span, request):
        server.log.info("Running request hook...")
        span.update_name(f"{request.method} {request.path}")
 
# passing the hooks in this intializer does NOT work
    DjangoInstrumentor().instrument(request_hook=request_hook)
    # applying the hook directly to the middleware works as intended. copied from the implementation
opentelemetry.instrumentation.django.middleware.otel_middleware._DjangoMiddleware._otel_request_hook = request_hook
```

**What is the expected behavior?**
I expected to have the spans renamed and a log print out in my console.

**What is the actual behavior?**
The function is never called. I confirmed that it's not even in the middleware by shelling into my pod and evaluating `opentelemetry.instrumentation.django.middleware.otel_middleware._DjangoMiddleware._otel_request_hook`. It evaluates to None.

**Additional context**
Not sure if this is a gunicorn specific issue or not. My solution is incredibly hacky.

Issue #2043, opened 2023-11-07
Django Instrumentation does not work with gunicorn and wsgi


jeremydvoss:
**Describe your environment** 
Using the [Azure Django Sample](https://github.com/Azure-Samples/msdocs-python-django-webapp-quickstart) as my example app. I have tested the differences in between running `python manage.py runserver` and `gunicorn quickstartproject.wsgi`. In both cases, the app works. However, the instrumentation only worked with `python manage.py runserver`. Gunicorn or wsgi seem to result in the OTel middleware not being triggered when a requests is made.

**Steps to reproduce**

1. Clone the repo
2. Add django instrumentations to views.py. I add custom print statements to the instrumentation and middleware
3. Run app with `python manage.py runserver`
4. Confirm django instrumentation code is hit
5. Run app with `gunicorn quickstartproject.wsgi`
6. Confirm django instrumentation code is not hit

**What is the expected behavior?**
Django instrumentation code should work regardless of how a django app is started. It should also work regardless of whether the instrumentation is manually added to the django app or through auto-instrumentation.

**What is the actual behavior?**
Django instrumentation only works with manage.py

**Additional context**
My guess is that the issue is that wsgi sets the settings (including middleware) before views or other parts of the django app are triggered, and that the middleware settings are not being updated thereafter.

jeremydvoss:
The issue is that when running through wsgi.py, the middleware is set up BEFORE the instrumentation middleware is prepended in veiws.py

jeremydvoss:
Adding instrument to wsgi or using auto instrumentation to make sure instrument runs before anything else fixes the issue. So, the location of the instrument function should depend on the django app's triggering mechanism

PR #2408, opened 2024-04-10
Add log to _DjangoMiddleware (opentelemetry-instrumentation-django)

nenoNaninu:
# Description

When I added the `opentelemetry-instrumentation-django` package and started the server with `python manage.py runserver` calling `DjangoInstrumentor().instrument()` in manage.py, HTTP request spans were correctly created

However, when using **Django with Gunicorn**, there was a problem: HTTP request spans were not created.
To solve this problem, I spent a lot of time reading the `opentelemetry-instrumentation-django` code and found that adding `opentelemetry-instrumentation-asgi` would solve this problem. I have arrived at the solution.

The current implementation silently turns off instrumentation in certain situations, such as using Gunicorn. [otel_middleware.py#L199-L200](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/v0.45b0/instrumentation/ opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py#L199-L200)

Instrumentation library users will expect the instrumentation to work correctly if the call to `DjangoInstrumentor().instrument()` succeeds and there are no warnings in the logs. Also, the documentation does not mention the need to add `opentelemetry-instrumentation-asgi`.

- https://opentelemetry-python.readthedocs.io/en/latest/examples/django/README.html
- https://opentelemetry-python.readthedocs.io/en/latest/examples/fork-process-model/README.html?highlight=gunicorn#gunicorn-post- fork-hook

This wastes a lot of time for many developers. Therefore, I have added a log that shows how to deal with this problem instead of silently disabling instrumentation when it occurs.

Related: #2043 #2296

## Type of change

# How Has This Been Tested?

- No test, because only adding log.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
+++ b/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
@@ -197,6 +197,9 @@ class _DjangoMiddleware(MiddlewareMixin):
 
         is_asgi_request = _is_asgi_request(request)
         if not _is_asgi_supported and is_asgi_request:
+            _logger.warning(
+                "ASGI request detected but ASGI instrumentation is not available. Skipping instrumentation. You may need to install the `opentelemetry-instrumentation-asgi` package."
+            )
             return
 
         # pylint:disable=W0212


xrmx:
Thanks for the  PR, care to add a test too?

nenoNaninu:
Is it essential to add test code to add a mere line of logging?

xrmx:
> Is it essential to add test code to add a mere line of logging?

It's not essential of course, but it'll help :)

srikanthccv:
I don't think we should log such a warning more than once. Please use warnings.warn instead.

nenoNaninu:
If developers are using `manage.py runserver` in a local dev environment, it will probably output plain text logs, so `warnings.warn` will not be a problem. However, this problem only occurs when Django is started using like gunicorn. And it's not the local development environment that uses gunicorn etc.; it's the stg environment. The stg environment uses structured logs in most cases. So, it is easier to find the problem if the developer has a structured warning log using `logger.warning` rather than `warnings.warn`. I don't think it's worth throwing away the structured logs to warn only once.

Moreover, if multiple workers run using gunicorn, the warnings will be logged more than once even if we use `warnings.warn`.

xrmx:
> Moreover, if multiple workers run using gunicorn, the warnings will be logged more than once even if we use `warnings.warn`.

This will warn at each request not for each worker

nenoNaninu:
> This will warn at each request not for each worker

Of course, I know.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -197,6 +197,9 @@ def process_request(self, request):
 
         is_asgi_request = _is_asgi_request(request)
         if not _is_asgi_supported and is_asgi_request:
```
ocelotl, line 199: Maybe something like this?

```suggestion

        self._unsupported_asgi_warned = False
        ...
        if not _is_asgi_supported and is_asgi_request and not self._unsupported_asgi_warned:
            ...
            self._unsupported_asgi_warned
```

Could be a good idea, or a terrible one :shrug:

ocelotl (commented):
Added a comment

nenoNaninu:
@srikanthccv @ocelotl I fixed to log a warning only once.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -197,6 +209,7 @@ def process_request(self, request):
...
+            _OnceLogger.log_warning_asgi_instrumentation()
```
ocelotl, line 212: Please add a test case, see [this](https://github.com/open-telemetry/opentelemetry-python/pull/3360/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R74-R84).

nenoNaninu: Is there already a test for a pattern where the request type is `ASGIRequest` and `opentelemetry-instrumentation-asgi` is not installed?

PR #2420, opened 2024-04-15
elasticsearch: tests against elasticsearch 8

xrmx:
# Description

Test against recent elasticsearch versions

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
@@ -94,7 +94,7 @@ from opentelemetry.instrumentation.elasticsearch.version import __version__
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
 from opentelemetry.instrumentation.utils import unwrap
 from opentelemetry.semconv.trace import SpanAttributes
-from opentelemetry.trace import SpanKind, get_tracer
+from opentelemetry.trace import SpanKind, Status, StatusCode, get_tracer
 
 from .utils import sanitize_body
 
@@ -103,6 +103,7 @@ from .utils import sanitize_body
 es_transport_split = elasticsearch.VERSION[0] > 7
 if es_transport_split:
     import elastic_transport
+    from elastic_transport._models import DefaultType
 
 logger = getLogger(__name__)
 
@@ -173,7 +174,12 @@ class ElasticsearchInstrumentor(BaseInstrumentor):
 
     def _uninstrument(self, **kwargs):
         # pylint: disable=no-member
-        unwrap(elasticsearch.Transport, "perform_request")
+        transport_class = (
+            elastic_transport.Transport
+            if es_transport_split
+            else elasticsearch.Transport
+        )
+        unwrap(transport_class, "perform_request")
 
 
 _regex_doc_url = re.compile(r"/_doc/([^/]+)")
@@ -182,6 +188,7 @@ _regex_doc_url = re.compile(r"/_doc/([^/]+)")
 _regex_search_url = re.compile(r"/([^/]+)/_search[/]?")
 
 
+# pylint: disable=too-many-statements
 def _wrap_perform_request(
     tracer,
     span_name_prefix,
@@ -234,7 +241,22 @@ def _wrap_perform_request(
             kind=SpanKind.CLIENT,
         ) as span:
             if callable(request_hook):
-                request_hook(span, method, url, kwargs)
+                # elasticsearch 8 changed the parameters quite a bit
+                if es_transport_split:
+
+                    def normalize_kwargs(k, v):
+                        if isinstance(v, DefaultType):
+                            v = str(v)
+                        elif isinstance(v, elastic_transport.HttpHeaders):
+                            v = dict(v)
+                        return (k, v)
+
+                    hook_kwargs = dict(
+                        normalize_kwargs(k, v) for k, v in kwargs.items()
+                    )
+                else:
+                    hook_kwargs = kwargs
+                request_hook(span, method, url, hook_kwargs)
 
             if span.is_recording():
                 attributes = {
@@ -260,16 +282,41 @@ def _wrap_perform_request(
                     span.set_attribute(key, value)
 
             rv = wrapped(*args, **kwargs)
-            if isinstance(rv, dict) and span.is_recording():
+
+            body = rv.body if es_transport_split else rv
+            if isinstance(body, dict) and span.is_recording():
                 for member in _ATTRIBUTES_FROM_RESULT:
-                    if member in rv:
+                    if member in body:
                         span.set_attribute(
                             f"elasticsearch.{member}",
-                            str(rv[member]),
+                            str(body[member]),
                         )
 
+            # since the transport split the raising of exceptions that set the error status
+            # are called after this code so need to set error status manually
+            if es_transport_split and span.is_recording():
+                if not (method == "HEAD" and rv.meta.status == 404) and (
+                    not 200 <= rv.meta.status < 299
+                ):
+                    exception = elasticsearch.exceptions.HTTP_EXCEPTIONS.get(
+                        rv.meta.status, elasticsearch.exceptions.ApiError
+                    )
+                    message = str(body)
+                    if isinstance(body, dict):
+                        error = body.get("error", message)
+                        if isinstance(error, dict) and "type" in error:
+                            error = error["type"]
+                        message = error
+
+                    span.set_status(
+                        Status(
+                            status_code=StatusCode.ERROR,
+                            description=f"{exception.__name__}: {message}",
+                        )
+                    )
+
             if callable(response_hook):
-                response_hook(span, rv)
+                response_hook(span, body)
             return rv
 
     return wrapper
--- a/instrumentation/opentelemetry-instrumentation-elasticsearch/tests/test_elasticsearch.py
+++ b/instrumentation/opentelemetry-instrumentation-elasticsearch/tests/test_elasticsearch.py
@@ -51,25 +51,25 @@ Article = helpers.Article
 
 
 def normalize_arguments(doc_type, body=None):
-    if major_version == 7:
-        return {"document": body} if body else {}
-    return (
-        {"body": body, "doc_type": doc_type}
-        if body
-        else {"doc_type": doc_type}
-    )
+    if major_version < 7:
+        return (
+            {"body": body, "doc_type": doc_type}
+            if body
+            else {"doc_type": doc_type}
+        )
+    return {"document": body} if body else {}
 
 
 def get_elasticsearch_client(*args, **kwargs):
     client = Elasticsearch(*args, **kwargs)
-    if major_version == 7:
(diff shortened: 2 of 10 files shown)


instrumentation/opentelemetry-instrumentation-elasticsearch/test-requirements-8.txt
```diff
@@ -0,0 +1,23 @@
+asgiref==3.7.2
+attrs==23.2.0
+Deprecated==1.2.14
+elasticsearch==8.12.1
+elasticsearch-dsl==8.12.0
+elastic-transport==8.12.0
```
pquentin, line 6: Is the idea that for 8 we support a recent release (8.12)) but not the latest (8.13)?

instrumentation/opentelemetry-instrumentation-elasticsearch/tests/test_elasticsearch.py
```diff
@@ -54,9 +58,37 @@
...
+
+def ElasticsearchFactory(*args, **kwargs):
+    client = Elasticsearch(*args, **kwargs)
+    if major_version == 8:
+        client._verified_elasticsearch = True
+    elif major_version == 7:
+        client.transport._verified_elasticsearch = True
+    return client
+
+class PerformRequestMock:
```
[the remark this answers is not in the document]
pquentin: Yeah it would be nice to use helpers and avoid all the conditionals here. I don't know what the support policy is, but the EOL dates are:

| Version  | EOL date |
| ------------ | ------------- |
| 1.x  | January 2017  |
| 2.x  | February 2018 |
| 5.x  | March 2019 |
| 6.x | February 2022 |

Source: https://web.archive.org/web/20220703152519/https://www.elastic.co/support/eol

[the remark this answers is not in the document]
pquentin: Is an alternative to have a generic `helpers_es.py` that applies to 2/5/6/7? My understanding is that we would also avoid the conditionals that way?

instrumentation/opentelemetry-instrumentation-elasticsearch/tests/test_elasticsearch.py
```diff
@@ -38,7 +38,11 @@
 
 major_version = elasticsearch.VERSION[0]
 
+NodeAPIResponse, ApiResponseMeta, HttpHeaders = None, None, None
+
 if major_version == 8:
+    from elastic_transport._node import NodeApiResponse
```
pquentin, line 44: Let's use that instead. I made a mistake when reviewing a pull request that exposed `NodeApiResponse`, it should be `TransportApiResponse` instead. That way in the future you will avoid the private import.

```suggestion
    from elastic_transport._transport import TransportApiResponse
```

instrumentation/opentelemetry-instrumentation-elasticsearch/test-requirements-8.txt
xrmx: That is for handling the native instrumentation in following PRs, more tests were failing because of double spans :)

instrumentation/opentelemetry-instrumentation-elasticsearch/tests/test_elasticsearch.py
```diff
@@ -54,9 +58,37 @@
...
+        return {"document": body} if body else {}
+
+
+def ElasticsearchFactory(*args, **kwargs):
+    client = Elasticsearch(*args, **kwargs)
+    if major_version == 8:
+        client._verified_elasticsearch = True
+    elif major_version == 7:
+        client.transport._verified_elasticsearch = True
+    return client
```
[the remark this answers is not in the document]
xrmx: No, it's the client that checks that and fails otherwise, an alternative would be to pass the headers to the mocked responses

instrumentation/opentelemetry-instrumentation-elasticsearch/tests/test_elasticsearch.py
xrmx: With that means i should also patch at the Transport.perform_request level instead of the Urllib3 backend right?

pquentin: Yes, which will cover more backends.

instrumentation/opentelemetry-instrumentation-elasticsearch/tests/test_elasticsearch.py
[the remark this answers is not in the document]
xrmx: Moved to helpers, after dropping every version < 6 we just have 2 versions older than 8.

instrumentation/opentelemetry-instrumentation-elasticsearch/tests/test_elasticsearch.py
xrmx: Ok, will revise once tests pass with 8.12

instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
```diff
@@ -234,7 +239,8 @@ def wrapper(wrapped, _, args, kwargs):
             kind=SpanKind.CLIENT,
         ) as span:
             if callable(request_hook):
-                request_hook(span, method, url, kwargs)
+                hook_kwargs = dict(kwargs["headers"]) if es_transport_split else kwargs
```
xrmx, line 242: not sure this is the right thing to do

instrumentation/opentelemetry-instrumentation-elasticsearch/tests/test_elasticsearch.py
```diff
@@ -439,12 +447,19 @@ def request_hook(span, method, url, kwargs):
...
+        if major_version == 8:
+            expected_url += "?realtime=true&refresh=true"
         self.assertEqual(
             expected_url,
             spans[0].attributes[request_hook_url_attribute],
         )
 
-        if major_version == 7:
+        if major_version == 8:
+            # FIXME: kwargs passed to request_hook on 8 are completely different
```
xrmx, line 458: need to revise this

instrumentation/opentelemetry-instrumentation-elasticsearch/tests/test_elasticsearch.py
```diff
@@ -220,14 +222,16 @@ def _test_trace_error(self, code, exc):
         spans = self.get_finished_spans()
         self.assertEqual(1, len(spans))
         span = spans[0]
-        self.assertFalse(span.status.is_ok)
-        self.assertEqual(span.status.status_code, code)
-        self.assertEqual(
-            span.status.description, f"{type(exc).__name__}: {exc}"
-        )
+        if major_version < 8:
```
xrmx, line 225: this needs to be resolved but I'm clueless, as the FIXME says span status is not set properly with es8

xrmx: It looks like now the exception raised by the transport is catched somewhere and before it escaped up to the tracer where the span status was updated accordingly.

instrumentation/opentelemetry-instrumentation-elasticsearch/tests/test_elasticsearch.py
xrmx: Transport.perform_request is the method we are wrapping for our instrumentation so mocking one level below in tests seems right to avoid testing our own mocks

instrumentation/opentelemetry-instrumentation-elasticsearch/tests/test_elasticsearch.py
xrmx: I think I understood this at last! At least for elasticsearch 8 we need to handle failure manually in the instrumentation because we are mocking at the transport layer but exceptions are raised at the elasticsearch layer so after we created our span. Need to check if it's working fine for older versions :sweat_smile:

instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
```diff
@@ -260,16 +266,17 @@ def wrapper(wrapped, _, args, kwargs):
                     span.set_attribute(key, value)
 
             rv = wrapped(*args, **kwargs)
```
xrmx, line 268: We need to set the error on the span in the following cases at least if `es_transport_split` because for elasticsearch 8 exceptions are raised later in the client:

```
        if not (method == "HEAD" and meta.status == 404) and (
            not 200 <= meta.status < 299
            and (
                self._ignore_status is DEFAULT
                or self._ignore_status is None
                or meta.status not in self._ignore_status
            )
        ):
```

xrmx: Added

instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
xrmx: Added normalization for all the kwargs

instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
```diff
@@ -103,6 +103,7 @@ def response_hook(span, response):
 es_transport_split = elasticsearch.VERSION[0] > 7
```
tammy-baylis-swi, line 103: I know `es_transport_split` is not introduced here but we are adding more checks that use it in this PR. Could we have an added comment that briefly summarizes how Otel instrumentation is affected by changes in elasticsearch 8?

xrmx: I've added 3 more users of es_transport_split:
- 1 in uninstrument to match what was already done in instrument
- 1 in the request hook to normalize args that are not json serializable
- 1 is for taking into account the wrapped `perform_request` have a different return value
- 1 is for settings the status error in case of failure since it has been moved to upper layers, I already added a comment here

From a user point of view I think the request hook parameters are the most impacting change, added a comment.

xrmx: BTW this is testing against 8.12 that has not the native OTel instrumentation. With an 8.13 client with the native instrumentation active the span created would be different. Hopefully we'll converge when updating ours to more recent semantic conventions.

tammy-baylis-swi (approved):

ocelotl (approved):

pquentin (commented):
Thanks! LGTM. Thanks for fixing the mock to be correct for urllib3 nodes.

PR #2424, opened 2024-04-15
Remove unnecessary package installations

### LINKED ISSUE #2423: Do not install unnecessary packages
We install unnecessary packages [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/e84bfff6cf8e57169aa18560a90e4d1d52c3f8d4/tox.ini#L326-L3300). Fix that.

ocelotl:
Fixes #2423

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -239,6 +239,10 @@ envlist =
     pypy3-test-instrumentation-tortoiseorm
 
     ; opentelemetry-instrumentation-httpx
+    ; The numbers at the end of the environment names
+    ; below mean these dependencies are being used:
+    ; 0: httpx>=0.18.0,<0.19.0 respx~=0.17.0
+    ; 1: httpx>=0.19.0 respx~=0.20.1
     py3{8,9,10,11}-test-instrumentation-httpx-{0,1}
     pypy3-test-instrumentation-httpx-{0,1}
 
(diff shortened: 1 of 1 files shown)


ocelotl:
This PR adds more code to `tox.ini` but it now allows us to know exactly what test dependencies each package has. I want to move in a direction where the testing information for each package lives in each package and this is a step in that direction.

tammy-baylis-swi (approved):
Thanks, I think this is good as a step in reorganization.

tox.ini
```diff
@@ -319,68 +323,143 @@ setenv =
   CORE_REPO=git+https://github.com/open-telemetry/opentelemetry-python.git@{env:CORE_REPO_SHA}
 
 commands_pre =
-; Install without -e to test the actual installation
-  py3{8,9,10,11}: python -m pip install -U pip setuptools wheel
-; Install common packages for all the tests. These are not needed in all the
-; cases but it saves a lot of boilerplate in this file.
-  test: pip install opentelemetry-api@{env:CORE_REPO}\#egg=opentelemetry-api&subdirectory=opentelemetry-api
```
lzchen, line 326: Didn't we have this here to remove the exact boiler plate code below? I know it is more explicit for what dependencies are installed per environment but is that better than repeated the same install lines? Open to discussing it during this weeks sig.

ocelotl: Yes, my plan is to fragment the `tox.ini` file in one per package so that we don't have this huge file, this PR is a step in that direction. That hopefully also would make `lint` faster by taking advantage of paralellization in CI and would allow the user to only have to run `lint` for the packages that the user changes.

PR #2425, opened 2024-04-15
HTTP transition for wsgi 

lzchen:
Http migration plan for wsgi.

For changes NOT related to span attributes, we simply snap to the new semantic conventions like for [span name](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#name).

Credits to @lmolkova for the original [pr](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2024/files).


opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -41,24 +46,147 @@
...
+    HTTP = "http"
+
+
+class _OpenTelemetryStabilityMode(Enum):
+    # http - emit the new, stable HTTP and networking conventions ONLY
+    HTTP = "http"
+    # http/dup - emit both the old and the stable HTTP and networking conventions
+    HTTP_DUP = "http/dup"
+    # default - continue emitting old experimental HTTP and networking conventions
+    DEFAULT = "default"
```
ocelotl, line 100: These modes have priorities, that idea should be represented in the `Enum` itself by using numerical values:

```suggestion
    # http/dup - emit both the old and the stable HTTP and networking conventions
    HTTP_DUP = 0
    # http - emit the new, stable HTTP and networking conventions ONLY
    HTTP = 1
    # default - continue emitting old experimental HTTP and networking conventions
    DEFAULT = 2
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -41,24 +46,147 @@
...
+    _OTEL_SEMCONV_STABILITY_SIGNAL_MAPPING = {}
+
+    @classmethod
+    def _initialize(cls):
+        with _OpenTelemetrySemanticConventionStability._lock:
+            if not _OpenTelemetrySemanticConventionStability._initialized:
+                # Users can pass in comma delimited string for opt-in options
+                # Only values for http stability are supported for now
+                opt_in = os.environ.get(_OTEL_SEMCONV_STABILITY_OPT_IN_KEY, "")
+                opt_in_list = []
```
ocelotl, line 123: Do not implement like this, use the ordering provided by a numerical `Enum` to get the stability mode:

```python
opentelemetry_stability_mode = sorted(
    [
        getattr(OptIn, element) for element in set(
            environ.get(
                "OPENTELEMETRY_SEMCONV_STABILITY_OPT_IN",
                OptIn.default.value
            ).split(",")
        )
        if element in OptIn.__members__
    ],
    key=lambda x: x.value
)[0]

```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -41,24 +46,147 @@
...
+
+_OTEL_SEMCONV_STABILITY_OPT_IN_KEY = "OTEL_SEMCONV_STABILITY_OPT_IN"
+
 
-def _filter_duration_attrs(attrs, sem_conv_opt_in_mode):
+class _OpenTelemetryStabilitySignalType:
+    HTTP = "http"
+
+
+class _OpenTelemetryStabilityMode(Enum):
```
ocelotl, line 94:
```suggestion
class _HTTPStabilityMode(Enum):
```

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -584,20 +636,31 @@ def __call__(self, environ, start_response):
...
+                    self._sem_conv_opt_in_mode,
                 )
                 iterable = self.wsgi(environ, start_response)
                 return _end_span_after_iterating(iterable, span, token)
         except Exception as ex:
             if span.is_recording():
-                span.set_status(Status(StatusCode.ERROR, str(ex)))
+                if _report_new(self._sem_conv_opt_in_mode):
+                    span.set_attribute(_SPAN_ATTRIBUTES_ERROR_TYPE, type(ex).__qualname__ )
+            span.set_status(Status(StatusCode.ERROR, str(ex)))
```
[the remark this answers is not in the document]
lzchen: Not intentional. Fixed.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -41,24 +46,147 @@
...
+    SpanAttributes.NET_HOST_NAME,
+    SpanAttributes.NET_HOST_PORT,
+]
+
+_server_active_requests_count_attrs_new = [
+    SpanAttributes.HTTP_REQUEST_METHOD,
+    SpanAttributes.URL_SCHEME,
+]
+
+_OTEL_SEMCONV_STABILITY_OPT_IN_KEY = "OTEL_SEMCONV_STABILITY_OPT_IN"
```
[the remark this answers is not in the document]
lzchen: I am not sure why this is showing up in the diffs but we already have this name commited. I am open to changing it though https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py#L147

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -41,24 +46,147 @@
...
+
 
-def _filter_duration_attrs(attrs, sem_conv_opt_in_mode):
+class _OpenTelemetryStabilitySignalType:
```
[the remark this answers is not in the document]
lzchen: There will multiple supported types in the future.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
lzchen: The enum values must be the string values that we read from the env vars, not numbers.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -41,24 +46,147 @@
...
+class _OpenTelemetryStabilityMode(Enum):
+    # http - emit the new, stable HTTP and networking conventions ONLY
+    HTTP = "http"
+    # http/dup - emit both the old and the stable HTTP and networking conventions
+    HTTP_DUP = "http/dup"
+    # default - continue emitting old experimental HTTP and networking conventions
+    DEFAULT = "default"
+
+
+def _report_new(mode):
```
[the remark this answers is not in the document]
lzchen: Similar to above, not sure how this diff is here. We already have been using this https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py#L163.

Also, it's not http, http/dup or /default. Sometimes you report BOTH old and new, or just old or just new. It is much simpler this way so we don't have to keep checking what mode it is but rather what mode it is NOT.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -41,24 +46,147 @@
...
+    return mode.name != _OpenTelemetryStabilityMode.DEFAULT.name
+
+
+def _report_old(mode):
```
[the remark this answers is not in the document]
lzchen: Yes, with this way to do http/dup i can make a call to `report_old` and `report_new` instead of having to check for a third "if" check.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -41,24 +46,147 @@
...
+    _initialized = False
+    _lock = threading.Lock()
+    _OTEL_SEMCONV_STABILITY_SIGNAL_MAPPING = {}
+
+    @classmethod
+    def _initialize(cls):
+        with _OpenTelemetrySemanticConventionStability._lock:
+            if not _OpenTelemetrySemanticConventionStability._initialized:
+                # Users can pass in comma delimited string for opt-in options
+                # Only values for http stability are supported for now
```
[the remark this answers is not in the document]
lzchen: The specs specifically state this [env var](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/migration-guide.md#http-semantic-convention-stability-migration-guide) to be named this way. With this, non-http stability will be a comma separated list in the future which is why we have to handle it this way.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
ocelotl: You already have those string names in the `enum` names:

```python
from enum import Enum
from ipdb import set_trace
from os import environ


class OptIn(Enum):
    dup = 0
    http = 1
    default = 2


opt_in = sorted(
    [
        getattr(OptIn, element) for element in set(
            environ.get(
                "OPENTELEMETRY_SEMCONV_STABILITY_OPT_IN",
                OptIn.default.name
            ).split(",")
        )
        if element in OptIn.__members__
    ],
    key=lambda x: x.value
)[0]

set_trace()
```

```bash
tigre@hilleman:~/sandbox$ python3 opt_in_example.py 
--Return--
None
> /home/tigre/sandbox/opt_in_example.py(25)<module>()
      6 class OptIn(Enum):
      7     dup = 0
      8     http = 1
      9     default = 2
     10 
     11 
     12 opt_in = sorted(
     13     [
     14         getattr(OptIn, element) for element in set(
     15             environ.get(
     16                 "OPENTELEMETRY_SEMCONV_STABILITY_OPT_IN",
     17                 OptIn.default.name
     18             ).split(",")
     19         )
     20         if element in OptIn.__members__
     21     ],
     22     key=lambda x: x.value
     23 )[0]
     24 
---> 25 set_trace()

ipdb> opt_in
<OptIn.default: 2>
ipdb> opt_in.value
2
ipdb> opt_in.name
'default'

```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
lzchen: We cannot have a numerical Enum since we need the value to represent the explicit string value in the env var. In your example, `HTTP_DUP.value` does not equal "http/dup" that will be read from envvar. Theoretically I guess we can implement a named tuple but I don't think the ordering provides much value.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
lzchen: OptIn.Dup.name does not map to "http/dup"

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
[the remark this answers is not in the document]
ocelotl: Yes, let's change it. We have been using the same string for environment variable names and their values.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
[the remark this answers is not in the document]
ocelotl: But are these types something else than just strings? It makes little sense to have an enum where the enum "key" is `HTTP` and the "value" is `http`.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
ocelotl: > We cannot have a numerical Enum since we need the value to represent the explicit string value in the env var. In your example, `HTTP_DUP.value` does not equal "http/dup" that will be read from envvar. Theoretically I guess we can implement a named tuple but I don't think the ordering provides much value.

That's just because `/` cannot be a character in an `Enum` attribute name, there is a workaround:

```python
from enum import Enum
from ipdb import set_trace
from os import environ


class OptIn(Enum):
    http_dup = 0
    http = 1
    default = 2


opt_in = sorted(
    [
        getattr(OptIn, element.replace("/", "_")) for element in set(
            environ.get(
                "OPENTELEMETRY_SEMCONV_STABILITY_OPT_IN",
                # OptIn.default.name
                "http/dup"
            ).split(",")
        )
        if element.replace("/", "_") in OptIn.__members__
    ],
    key=lambda x: x.value
)[0]

assert opt_in.name == "http_dup"
assert opt_in.value == 0

set_trace()
```
I find important that the objects that represent the stability type can represent their priority themselves.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
ocelotl: > OptIn.Dup.name does not map to "http/dup"

See [this](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2425/files#r1569487644) :v:

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -41,24 +46,147 @@
...
+    return mode.name != _HTTPStabilityMode.DEFAULT.name
+
+
+def _report_old(mode):
+    return mode.name != _HTTPStabilityMode.HTTP.name
+
 
-def _filter_duration_attrs(attrs, sem_conv_opt_in_mode):
+class _OpenTelemetrySemanticConventionStability:
+    _initialized = False
```
ocelotl, line 112: Not used. We want to get the stability mode only once, that is a job for [`Once`](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/util/_once.py#L19).

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -41,24 +46,147 @@
...
+
+    @classmethod
+    # Get OpenTelemetry opt-in mode based off of signal type (http, messaging, etc.)
+    def _get_opentelemetry_stability_opt_in_mode(
+        cls,
+        signal_type: _OpenTelemetryStabilitySignalType,
+    ) -> _HTTPStabilityMode:
+        return _OpenTelemetrySemanticConventionStability._OTEL_SEMCONV_STABILITY_SIGNAL_MAPPING.get(
+            signal_type, _HTTPStabilityMode.DEFAULT
+        )
```
ocelotl, line 150: I don't think we need this class, we can just have a module level `_http_stability_mode = Once(_get_http_stability_mode)` where `_get_http_stability_mode` is a function that gets the stability mode from environment variables.

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -495,42 +533,56 @@ def __init__(
         tracer_provider=None,
         meter_provider=None,
     ):
+        # initialize semantic conventions opt-in if needed
+        _OpenTelemetrySemanticConventionStability._initialize()
```
ocelotl, line 537: This class-based approach requires the instrumentations to worry about calling for the initialization themselves, that should not be the case, `from opentelemetry.x.y.z.semconv import http_stability_mode` should take care of getting the HTTP stability mode by itself.

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -284,16 +286,22 @@ def get_or_create_headers():
                 ).__qualname__
 
             if duration_histogram_old is not None:
-                duration_attrs_old = _filter_duration_attrs(
-                    metric_labels, _OpenTelemetryStabilityMode.DEFAULT
+                duration_attrs_old = _filter_semconv_duration_attrs(
+                    metric_labels,
+                    _client_duration_attrs_old,
+                    _client_duration_attrs_new,
+                    _HTTPStabilityMode.DEFAULT
```
ocelotl, line 293: Seems hardocded, shouldn't this value depend on that environment variable value?

ocelotl (commented):
Adding more comments

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -584,20 +636,31 @@ def __call__(self, environ, start_response):
...
         finally:
-            duration = max(round((default_timer() - start) * 1000), 0)
-            self.duration_histogram.record(duration, duration_attrs)
+            duration_s = default_timer() - start
+            if self.duration_histogram_old:
+                duration_attrs_old = _parse_duration_attrs(req_attrs, _HTTPStabilityMode.DEFAULT)
+                self.duration_histogram_old.record(max(round(duration_s * 1000), 0), duration_attrs_old)
+            if self.duration_histogram_new:
+                duration_attrs_new = _parse_duration_attrs(req_attrs, _HTTPStabilityMode.HTTP)
+                self.duration_histogram_new.record(max(round(duration_s * 1000), 0), duration_attrs_new)
```
lmolkova, line 663: the new one should be in seconds
```suggestion
                self.duration_histogram_new.record(max(round(duration_s), 0), duration_attrs_new)
```

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -584,20 +636,31 @@ def __call__(self, environ, start_response):
...
+                    response_hook,
+                    req_attrs,
+                    self._sem_conv_opt_in_mode,
                 )
                 iterable = self.wsgi(environ, start_response)
                 return _end_span_after_iterating(iterable, span, token)
         except Exception as ex:
             if span.is_recording():
+                if _report_new(self._sem_conv_opt_in_mode):
+                    span.set_attribute(_SPAN_ATTRIBUTES_ERROR_TYPE, type(ex).__qualname__ )
```
lmolkova, line 650: I think we should also add it to the duration_attrs_new (regardless of `span.is_recording`)

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -41,24 +46,147 @@
     SpanAttributes.NETWORK_PROTOCOL_VERSION,
     SpanAttributes.SERVER_ADDRESS,
     SpanAttributes.SERVER_PORT,
-    # TODO: Support opt-in for scheme in new semconv
-    # SpanAttributes.URL_SCHEME,
+    SpanAttributes.URL_SCHEME,
```
lmolkova, line 49: is it intentional?

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
lzchen: No, we have two histograms representing the old and new semantic conventions.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
lzchen: Nope, accident. Readded.

lzchen:
As discussed with @ocelotl , we will be separating out the architectural changes into a different issue. This pr will only pertain to changes related to adding migration/new semconv to wsgi to keep it smaller and simple.

ocelotl (approved):

Issue #2351, opened 2024-03-14
Vote on which instrumentations to implement the migration plan!


lzchen:
Based off of the recent [discussion](https://github.com/open-telemetry/semantic-conventions/issues/791#issuecomment-1988772078) on alleviating some of the implementation burden of implementing the migration plan for instrumentations moving to new semantic conventions, we are now allowed to choose a subset of instrumentations to implement this for. Please comment here on the preferences for "important instrumentations" that we should include as part of the migration plan. Keep in mind non-http instrumentations are included because this affects their eventual migration. A couple of things to also consider:

1. We can iterate on this list, where we can INCLUDE instrumentations in this list as we see fit but we most likely do not want to REMOVE ones that we already decide to be important.
2. Once we decide on a preliminary set of instrumentations, we can begin implementing migration plans for them INDIVIDUALLY.
3. For instrumentations NOT included in the set, we can begin to discuss whether we want to allow for contributions that update PARTIALLY to new semantic conventions.

For starters, I vote:

`opentelemetry-instrumentation-aiohttp-client`
`opentelemetry-instrumentation-asgi`
`opentelemetry-instrumentation-django`
`opentelemetry-instrumentation-fastapi`
`opentelemetry-instrumentation-httplib`
`opentelemetry-instrumentation-httpx`
`opentelemetry-instrumentation-flask`
`opentelemetry-instrumentation-psycopg2`
`opentelemetry-instrumentation-requests`
`opentelemetry-instrumentation-sqlalchemy`
`opentelemetry-instrumentation-requests`
`opentelemetry-instrumentation-urllib`
`opentelemetry-instrumentation-urllib3`
`opentelemetry-instrumentation-wsgi`

xrmx:
Maybe `opentelemetry-instrumentation-dbapi`?

aabmass:
> `opentelemetry-instrumentation-sqlalchemy`
> Maybe `opentelemetry-instrumentation-dbapi`?

Wouldn't this only affect HTTP libraries right now?

samuelcolvin:
In case it helps, here are the download counts for most of the libraries for February.

I would suggest these download counts should be used as the primary indicator of which projects are most popular (obviously that won't work for instrumentation of internal standard lib packages like asyncio or sqlite3).

|  project   | downloads |
|------------|-----------|
| urllib3    | 471792853 |
| requests   | 392426582 |
| jinja2     | 151665370 |
[...]

PR #2433, opened 2024-04-16
Add section for semantic convention status for instrumentations

lzchen:
As a continuation of https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2351 , it's a good idea to have a consolidated place where the semantic convention status of each instrumentation exists.

### DIFF
--- a/instrumentation/README.md
+++ b/instrumentation/README.md
@@ -1,51 +1,51 @@
 
-| Instrumentation | Supported Packages | Metrics support |
-| --------------- | ------------------ | --------------- |
-| [opentelemetry-instrumentation-aio-pika](./opentelemetry-instrumentation-aio-pika) | aio_pika >= 7.2.0, < 10.0.0 | No
-| [opentelemetry-instrumentation-aiohttp-client](./opentelemetry-instrumentation-aiohttp-client) | aiohttp ~= 3.0 | No
-| [opentelemetry-instrumentation-aiohttp-server](./opentelemetry-instrumentation-aiohttp-server) | aiohttp ~= 3.0 | No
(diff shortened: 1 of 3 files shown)


instrumentation/README.md
```diff
@@ -1,51 +1,51 @@
...
-| [opentelemetry-instrumentation-starlette](./opentelemetry-instrumentation-starlette) | starlette ~= 0.13.0 | Yes
-| [opentelemetry-instrumentation-system-metrics](./opentelemetry-instrumentation-system-metrics) | psutil >= 5 | No
-| [opentelemetry-instrumentation-threading](./opentelemetry-instrumentation-threading) | threading | No
-| [opentelemetry-instrumentation-tornado](./opentelemetry-instrumentation-tornado) | tornado >= 5.1.1 | Yes
-| [opentelemetry-instrumentation-tortoiseorm](./opentelemetry-instrumentation-tortoiseorm) | tortoise-orm >= 0.17.0 | No
-| [opentelemetry-instrumentation-urllib](./opentelemetry-instrumentation-urllib) | urllib | Yes
-| [opentelemetry-instrumentation-urllib3](./opentelemetry-instrumentation-urllib3) | urllib3 >= 1.0.0, < 3.0.0 | Yes
-| [opentelemetry-instrumentation-wsgi](./opentelemetry-instrumentation-wsgi) | wsgi | Yes
\ No newline at end of file
+| Instrumentation | Supported Packages | Metrics support | Semconv status |
```
ocelotl, line 2: We should add an explanation on what does this new column means

ocelotl (changes requested):

instrumentation/README.md
lzchen: @ocelotl 

The explanation is in the README.

ocelotl: Right, missed it, sorry :v:

ocelotl (approved):

PR #2439, opened 2024-04-18
DO NOT MERGE: Auto instrumentation paramaters proposal

jeremydvoss:
# Description

Currently, the only way for custom distros to configure is through setting or defaulting environment variables. While it makes sense for customers to configure distros with env vars. There is no reason distros should need be able to directly configure start up (exporters, samplers...etc). Exposing distro configuration of otel via env var defaults is invasive, confusing, and easy for users to mess up. For instrance, even a blank env var or a left over setting from previous run could interfere with a distro starting up.

This is a draft proposal for what changing this approach could look like. In this approach, distros can output kwargs that our sitecustomize setup can then pass to configurators, and by extension the _initialize_components method which currently _only_ considers env vars.

main repo pr: https://github.com/open-telemetry/opentelemetry-python/pull/3864
contrib pr: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2439

### DIFF
--- a/opentelemetry-distro/src/opentelemetry/distro/__init__.py
+++ b/opentelemetry-distro/src/opentelemetry/distro/__init__.py
@@ -41,14 +41,5 @@ class OpenTelemetryDistro(BaseDistro):
         # Since the distro sets these env vars, these params are not necesary.
         # However they could replace these env var defaults.
         # Otherwise, they just serve as an example.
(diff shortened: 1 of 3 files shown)


jeremydvoss:
Closing because custom configurators can handle kwargs instead. No need for the distros to return anything.

PR #2392, opened 2024-04-04
Fix while calling asyncio `exception()`

janario:
# Description

While updating our setup from operator 0.94.0 we started to have errors in python applications that were fine before.

It looks like asyncio was added on version 0.95.0 https://github.com/open-telemetry/opentelemetry-operator/pull/2677/files#diff-e8feb6bbad63df255d68837a72bc1c440ad0cac92b0406281d64879682e41c52R19

```
[2024-04-04 10:06:41,746] [7] [ERROR] Exception in callback AsyncioInstrumentor.trace_future.<locals>.callback(<Future cancelled>) at /otel-auto-instrumentation-python/opentelemetry/instrumentation/asyncio/__init__.py:309
handle: <Handle AsyncioInstrumentor.trace_future.<locals>.callback(<Future cancelled>) at /otel-auto-instrumentation-python/opentelemetry/instrumentation/asyncio/__init__.py:309>
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/asyncio/events.py", line 84, in _run
    self._context.run(self._callback, *self._args)
  File "/otel-auto-instrumentation-python/opentelemetry/instrumentation/asyncio/__init__.py", line 310, in callback
    exception = f.exception()
                ^^^^^^^^^^^^^
asyncio.exceptions.CancelledError 
```

Our scenario is an application using faust to consume kafka messages. https://faust.readthedocs.io/en/latest/
Quite a big legacy application with many consumers, so it was a bit hard to extract the exact scenario.

The problem is when a callback invokes the `exception()` function, which is expected to throw exception in some conditions.

In this PR we catch them instead of letting them be propagated.

## Type of change

Please delete options that are not relevant.

- [ x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

I've re-build it and the errors logs were gone

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [ ] Followed the style guidelines of this project
- [ ] Changelogs have been updated
- [ ] Unit tests have been added
- [ ] Documentation has been updated

PR #2461, opened 2024-04-26
Bugfix/check future cancelled

### LINKED ISSUE #2460: Asyncio Instrumentation future cancelledError

**Steps to reproduce**
Using python auto instrumentation version 0.44b0  the real trace source is being deprecated and we see only this:
`{"line":44,"module":"runners","timestamp":"2024-04-18 15:02:21.526124+00:00","message":"[asyncio] Exception in callback <function AsyncioInstrumentor.trace_future.<locals>.callback at 0x7fa8d2f47c70>\nhandle: <Handle AsyncioInstrumentor.trace_future.<locals>.callback>","commit":"-","service_name":"******","hostname":"***********","level":"ERROR","exception_type":"CancelledError","traceback":"File \"uvloop/cbhandles.pyx\", line 63, in uvloop.loop.Handle._run\n\nFile \"/otel-auto-instrumentation-python/opentelemetry/instrumentation/asyncio/__init__.py\", line 310, in callback\n    309  def callback(f):\n--> 310      exception = f.exception()\n    311      attr = {\n    ..................................................\n     f = <Future cancelled>\n    ..................................................\n\nCancelledError"}`

What did you expect to see?
SPAN collected with no errors

What did you see instead?
trace source does not run with an error log

**Additional context**
Calling the exception() method when future is in the cancelled state is causing a CancelledError, so I think we should check the cancelled state first and call f.exception() only if it's not cancelled.

bourbonkk:
# Description

Using python auto instrumentation version 0.44b0  the real trace source is being deprecated.
Calling the exception() method when future is in the cancelled state is causing a CancelledError

Fixes #2460 

## Type of change

# How Has This Been Tested?

- [x] Execute Test Code
- [x] gitflow validation

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


xrmx:
Could you please add a test?

Also #2392 looks a different solution of the same issue, like they choose to always pass an exception

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_future_cancellation.py
```diff
@@ -0,0 +1,26 @@
+import asyncio
+
+import pytest
+from opentelemetry.test.test_base import TestBase
+
+
+class TestTraceFuture(TestBase):
+
+    @pytest.mark.asyncio
+    def test_trace_future_cancelled(self):
```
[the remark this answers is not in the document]
bourbonkk: Fixed the test code.

bourbonkk:
> Could you please add a test?
> 
> Also #2392 looks a different solution of the same issue, like they choose to always pass an exception

i've looked into the issue.
It's good code, but I have a slightly different opinion.
I don't report canceledError as an error in the span because there is a subject that did the cancel, and the coroutine follows the same strategy.

```
        # CancelledError is raised when a coroutine is cancelled
        # before it has a chance to run. We don't want to record
        # this as an error.
```

instrumentation/opentelemetry-instrumentation-asyncio/src/opentelemetry/instrumentation/asyncio/__init__.py
```diff
@@ -307,13 +296,17 @@ def trace_future(self, future):
         )
 
         def callback(f):
-            exception = f.exception()
             attr = {
                 "type": "future",
+                "state": (
+                    "cancelled"
+                    if f.cancelled()
+                    else determine_state(f.exception())
```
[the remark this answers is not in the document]
bourbonkk: I'm using determine_state() with a coroutine, would it be better to separate them?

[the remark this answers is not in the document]
aabmass: Understood. No worries then

instrumentation/opentelemetry-instrumentation-asyncio/src/opentelemetry/instrumentation/asyncio/__init__.py
```diff
@@ -116,17 +116,6 @@ class AsyncioInstrumentor(BaseInstrumentor):
...
-    def __init__(self):
-        super().__init__()
-        self.process_duration_histogram = None
-        self.process_created_counter = None
-
-        self._tracer = None
-        self._meter = None
-        self._coros_name_to_trace: set = set()
-        self._to_thread_name_to_trace: set = set()
-        self._future_active_enabled: bool = False
```
aabmass, line 128: Is the related to the issue or just some old dead code?

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_future_cancellation.py
```diff
@@ -0,0 +1,59 @@
...
+            .scope_metrics[0]
+            .metrics
+        ):
+            if metric.name == "asyncio.process.duration":
+                for point in metric.data.data_points:
+                    self.assertEqual(point.attributes["type"], "future")
+            if metric.name == "asyncio.process.created":
+                for point in metric.data.data_points:
+                    self.assertEqual(point.attributes["type"], "future")
+                    self.assertEqual(point.attributes["state"], "cancelled")
```
aabmass, line 59: I think these tests will still pass if these metrics are not present

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_future_cancellation.py
```diff
@@ -0,0 +1,59 @@
...
+    @patch.dict(
+        "os.environ", {OTEL_PYTHON_ASYNCIO_FUTURE_TRACE_ENABLED: "true"}
+    )
+    def setUp(self):
+        super().setUp()
+        self._tracer = get_tracer(
+            __name__,
+        )
+        self.loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(self.loop)
```
aabmass, line 24: Is this needed? I think we can just use `asnycio.run()` in place and the framework will handle loop lifecycles

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_future_cancellation.py
```diff
@@ -0,0 +1,59 @@
...
+        self.instrumentor = AsyncioInstrumentor()
+        self.instrumentor.instrument()
+
+    def tearDown(self):
+        super().tearDown()
+        self.instrumentor.uninstrument()
+        self.loop.close()
+
+    @pytest.mark.asyncio
```
aabmass, line 33: Do we have pytest-asyncio installed?

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_future_cancellation.py
bourbonkk: When you run it with asyncio.run(), it converts the future to a coroutine, so you're running the trace_coroutine() logic.

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_future_cancellation.py
bourbonkk: Removed the if statement.

instrumentation/opentelemetry-instrumentation-asyncio/src/opentelemetry/instrumentation/asyncio/__init__.py
bourbonkk: I removed it because it's not related to the issue and I don't think it's needed, and I thought it was right to remove it because sometimes when I run the test code, the process_duration_histogram and process_created_counter are None.

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_future_cancellation.py
bourbonkk: Yes, i've seen it being used in a lot of places.

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_future_cancellation.py
aabmass: See comment below about `async def`. Can you just await the future then? If not I'll take your work for it

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_future_cancellation.py
bourbonkk: I understand, I've made changes based on your feedback.
Thank you.

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_future_cancellation.py
bourbonkk: I've modified the conditions in this part to be a bit simpler.

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_future_cancellation.py
aabmass: I left a suggestion below

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_future_cancellation.py
```diff
@@ -0,0 +1,63 @@
...
+        async def future_cancelled():
+            with self._tracer.start_as_current_span("root"):
+                future = asyncio.Future()
+                future = self.instrumentor.trace_future(future)
+                future.cancel()
+
+        try:
+            asyncio.run(future_cancelled())
+        except asyncio.CancelledError as exc:
+            self.assertEqual(isinstance(exc, asyncio.CancelledError), True)
```
aabmass, line 41: With `pytest.mark.asyncio` I was thinking you could do this 
```suggestion
    async def test_trace_future_cancelled(self):
        with self._tracer.start_as_current_span("root"):
            future = asyncio.Future()
            future = self.instrumentor.trace_future(future)
            future.cancel()

        try:
            await future
        except asyncio.CancelledError as exc:
            self.assertEqual(isinstance(exc, asyncio.CancelledError), True)
```

Does that work?

aabmass (approved):
Thanks for the PR! Apologies for all the comments

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_future_cancellation.py
bourbonkk: That code seems to ignore the assert.
If you modify it with this code, and replace the 2 below with a 1, it will pass.
```
        self.assertEqual(len(metrics), 2)
```

Does the test work?

aabmass: It sounds like pytest-asyncio is not installed then. You can just leave teh code as is but please remove the `@pytest.mark.asyncio` as it's a bit misleading

bourbonkk: i removed it. Thank you for your feedback.

xrmx (approved):

ocelotl (approved):

PR #2467, opened 2024-04-26
Separate lint into several jobs

### LINKED ISSUE #2419: Lint is very slow
[This](https://github.com/open-telemetry/opentelemetry-python/pull/3848) cannot be applied to this repo because there are many packages that share the same dependency but require different versions of it.

We should instead separate linting in as many actions as packages we have. This will speed up linting in CI significantly thanks to the fact that many jobs can run in parallel.

ocelotl:
Fixes #2419


xrmx (commented):
FYI black, isort and flake8 can be run once per repository instead of per package, that would avoid duplicating a ton of lines in tox (and also work)

ocelotl:
> FYI black, isort and flake8 can be run once per repository instead of per package, that would avoid duplicating a ton of lines in tox (and also work)

Sure, but I want to produce a scenario as similar as possible to a scenario where there are multiple `tox.ini` files, one per package.

tox.ini
```diff
@@ -206,37 +247,46 @@ envlist =
...
     ; instrumentation-remoulade intentionally excluded from pypy3
+    lint-instrumentation-remoulade
 
     ; opentelemetry-instrumentation-celery
     py3{8,9,10,11}-test-instrumentation-celery
     pypy3-test-instrumentation-celery
+    lint-instrumentation-celery
 
     ; opentelemetry-instrumentation-sklearn
     py3{8}-test-instrumentation-sklearn
```
ocelotl, line 268: Probably lint fails for sklearn because we are running lint with python 3.11.

ocelotl: Created a new section for sklearn in the workflow file.

ocelotl:
Lint tests take ~14 minutes right now: https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/8839570393/job/24273211949

This PR speeds up this process to ~2 minutes: https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/8862990640

xrmx:
> > FYI black, isort and flake8 can be run once per repository instead of per package, that would avoid duplicating a ton of lines in tox (and also work)
> 
> Sure, but I want to produce a scenario as similar as possible to a scenario where there are multiple `tox.ini` files, one per package.

I see that, I would had hoped to make lint faster by using faster tools before splitting stuff though.

ocelotl:
> > > FYI black, isort and flake8 can be run once per repository instead of per package, that would avoid duplicating a ton of lines in tox (and also work)
> > 
> > 
> > Sure, but I want to produce a scenario as similar as possible to a scenario where there are multiple `tox.ini` files, one per package.
> 
> I see that, I would had hoped to make lint faster by using faster tools before splitting stuff though.

Sure, we can move to another tool later, this does not introduce anyting new that needs further CI testing and already makes a huge difference.

ocelotl:
> Would be nice to sort packages in the matrix though

Right, I just sorted them :v:

PR #2471, opened 2024-04-27
Allow broad exceptions in pylint

### LINKED ISSUE #2470: Fix warning when running lint
Similar to [https://github.com/open-telemetry/opentelemetry-python/issues/3671](this issue).

ocelotl:
Fixes #2470

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -81,7 +81,8 @@ disable=missing-docstring,
       missing-module-docstring, # temp-pylint-upgrade
       import-error, # needed as a workaround as reported here: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/290
       cyclic-import,
-      not-context-manager
+      not-context-manager,
+      broad-exception-raised,
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
@@ -486,10 +487,3 @@ max-statements=50
 
 # Minimum number of public methods for a class (see R0903).
 min-public-methods=2
-
-
-[EXCEPTIONS]
-
-# Exceptions that will emit a warning when being caught. Defaults to
-# "Exception".
-overgeneral-exceptions=Exception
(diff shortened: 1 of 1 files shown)


xrmx (commented):
No strong opinion, but why not keeping the check and just use `builtins.Exception` in the config?

ocelotl:
> No strong opinion, but why not keeping the check and just use `builtins.Exception` in the config?

So, I tried this approach because it was the one that worked for the [core repo](https://github.com/open-telemetry/opentelemetry-python/pull/3685).

.pylintrc
```diff
@@ -486,10 +487,3 @@ max-statements=50
 
 # Minimum number of public methods for a class (see R0903).
 min-public-methods=2
-
-
-[EXCEPTIONS]
-
-# Exceptions that will emit a warning when being caught. Defaults to
-# "Exception".
-overgeneral-exceptions=Exception
```
xrmx, line 495: Before giving up and remove the warning I would try with builtins.Exception, I think this warning is useful

xrmx (changes requested):

.pylintrc
ocelotl: All right, changed it.

.pylintrc
```diff
@@ -81,7 +81,8 @@ disable=missing-docstring,
       missing-module-docstring, # temp-pylint-upgrade
       import-error, # needed as a workaround as reported here: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/290
       cyclic-import,
-      not-context-manager
+      not-context-manager,
+      broad-exception-raised
```
xrmx, line 85: this should not be added then

ocelotl: Did that and `lint` [failed](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2471/commits/4532a7f6ceb1d03927771fbdceb0a8bab157b158).

xrmx: @ocelotl I've opened a draft PR to fix them, failing because I got some the pylint exception wrong and missed to update tests when used different Exception types, will update it tomorrow

ocelotl:
Closing in favor of #2536

PR #2481, opened 2024-04-30
Fix dependency notes in tox.ini

ocelotl:
Fixes #24800

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -64,10 +64,10 @@ envlist =
     ; https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django
     ; The numbers at the end of the environment names
     ; below mean these dependencies are being used:
-    ; 0: django~=2.0
-    ; 1: django~=3.0
-    ; 2: django>=4.0b1,<5.0 backports.zoneinfo==0.2.1
-    ; 3: django>=4.0b1,<5.0
+    ; 0: django==2.2.28
+    ; 1: django==3.2.25
+    ; 2: django==4.2.11 backports.zoneinfo==0.2.1
+    ; 3: django==4.2.11
     py3{8,9}-test-instrumentation-django-0
     py3{8,9}-test-instrumentation-django-1
     py3{8,9}-test-instrumentation-django-2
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -100,9 +100,9 @@ envlist =
     ; py310 does not work with falcon 1
     ; The numbers at the end of the environment names
     ; below mean these dependencies are being used:
-    ; 0: falcon ==1.4.1
-    ; 1: falcon >=2.0.0,<3.0.0
-    ; 2: falcon >=3.0.0,<4.0.0
```
aabmass, line 105: I thought the goal of these comments was to preserve the actual range so when it comes time to upgrade the pinned version, we can make sure it meets this requirement?

tox.ini
```diff
@@ -116,9 +116,9 @@ envlist =
     ; opentelemetry-instrumentation-flask
     ; The numbers at the end of the environment names
     ; below mean these dependencies are being used:
-    ; 0: Flask ==2.1.3 Werkzeug <3.0.0
-    ; 1: Flask ==2.2.0 Werkzeug <3.0.0
-    ; 2: Flask >=3.0.0 Werkzeug >=3.0.0
+    ; 0: Flask==2.1.3 Werkzeug==2.3.8
+    ; 1: Flask==2.2.0 Werkzeug==2.3.8
+    ; 2: Flask==3.0.2 Werkzeug==3.0.1
```
aabmass, line 121: I'm a bit confused what's going on here. This version constraint is impossible to fulfill

How did we get here? 😅

tox.ini
```diff
@@ -116,9 +116,9 @@ envlist =
     ; opentelemetry-instrumentation-flask
     ; The numbers at the end of the environment names
     ; below mean these dependencies are being used:
-    ; 0: Flask ==2.1.3 Werkzeug <3.0.0
-    ; 1: Flask ==2.2.0 Werkzeug <3.0.0
-    ; 2: Flask >=3.0.0 Werkzeug >=3.0.0
```
aabmass, line 121: These are the same number, something went wrong?

aabmass (changes requested):

ocelotl:
I won't be workin on this PR for the foreseeable future, closing.

PR #2493, opened 2024-05-03
Fix net peer attribute for unix socket connection

shadchin:
# Description

After switch on `.get(...)`, `KeyError` exception is never raised

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

```
import redis
from opentelemetry.instrumentation.redis import RedisInstrumentor

RedisInstrumentor().instrument()

client = redis.Redis(unix_socket_path="/foobar")  # or redis.Redis.from_url("unix:///foobar")
# Span attribute `net.peer.name` must be equal to `/foobar`
```

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [x] Followed the style guidelines of this project
- [ ] Changelogs have been updated
- [ ] Unit tests have been added
- [ ] Documentation has been updated

PR #2501, opened 2024-05-05
Support cursor based queries

bartv:
# Description

This PR add support for tracing on queries using a cursor instead of directly on the connection.

## Type of change

# How Has This Been Tested?

- [ ] Test against a code base that uses cursor extensively 

- [ ] No.

# Checklist:


xrmx (commented):
Thanks for the PR but you need to add tests

bartv:
> Thanks for the PR but you need to add tests

What do you have in mind? The existing code is also just testing if the function gets wrapped. Do the same for the Cursor class?

bartv:
Do you have examples on how to do this? The current code does not have any of this at the moment. This also means I will need a postgres database to actually connect to.

xrmx:
> Do you have examples on how to do this? The current code does not have any of this at the moment. This also means I will need a postgres database to actually connect to.

This is redis mocking the wrapped function:
https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2493/files#diff-83dd33b6baa2337f393ad1da857dbc683f48d0f52b460c5900c0c63139f2c279R235

Can something like this work?

xrmx:
> It would test if we can instrument a mock and that it generates spans. Not sure what the value is of that?

The value is to check that `_do_cursor_execute` does what it we expected to do :sweat_smile:

xrmx:
> This project also seems to use pytest however it is not clear to me how it can execute async test cases. Do you have pointer on this?

I would take a look that the instrumentation current tests in `instrumentation/opentelemetry-instrumentation-psycopg/tests/test_psycopg_integration.py`

bartv:
I looked at them. The problem is that there is something different there that makes pytest execute those tests correct and not the ones that I added.

Both when invoking pytest directly and using the predefined tox environments.

```
instrumentation/opentelemetry-instrumentation-asyncpg/tests/test_asyncpg_wrapper.py::TestAsyncPGInstrumentation::test_cursor_span_creation
  /usr/lib64/python3.11/unittest/case.py:579: RuntimeWarning: coroutine 'TestAsyncPGInstrumentation.test_cursor_span_creation' was never awaited
    if method() is not None:
  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.

instrumentation/opentelemetry-instrumentation-asyncpg/tests/test_asyncpg_wrapper.py::TestAsyncPGInstrumentation::test_cursor_span_creation
  /usr/lib64/python3.11/unittest/case.py:678: DeprecationWarning: It is deprecated to return a value that is not None from a test case (<bound method TestAsyncPGInstrumentation.test_cursor_span_creation of <tests.test_asyncpg_wrapper.TestAsyncPGInstrumentation testMethod=test_cursor_span_creation>>)
    return self.run(*args, **kwds)
```

It does not detect them as async. I will push my change, maybe you can spot the issue. This test should fail but it doesn't because it never gets awaited.

instrumentation/opentelemetry-instrumentation-asyncpg/tests/test_asyncpg_wrapper.py
```diff
@@ -34,3 +37,32 @@ def test_duplicated_uninstrumentation(self):
...
+        assert_wrapped(self.assertFalse)
+
+    async def test_cursor_span_creation(self):
+        def mock_fn(*args, **kwargs):
+            pass
+
+        cursor_mock = mock.Mock()
+
+        apg = AsyncPGInstrumentor()
+        await apg._do_cursor_execute(mock_fn, cursor_mock, "SELECT * FROM test", {})
```
xrmx, line 65: If you look at https://github.com/open-telemetry/opentelemetry-python-contrib/blob/d0500c2f8a2160ef21ae4ef3fa3f522f3eea7f94/instrumentation/opentelemetry-instrumentation-psycopg/tests/test_psycopg_integration.py#L224
You'll see:
- they are not awaiting the async function but running the async function in asyncio.run
- they are not calling the internal helpers directly but testing the instrumented function

bartv: I saw that they use asyncio.run however they marked their function as aysnc and pytest executes it as such. This is not the case for asyncpg tests and I really do not know why. But I can use asyncio.run it just makes the code at lot less clear to read.

In our code base we have 1000'ends of async tests but there we use plain pytest style and not the unittest base classes. Something is off here and I do not know what.

About the testing itself, the only way to test it is by mocking out connection and cursor and then instrument it. It looks like I am testing the instrumented function in that case, however it is exactly the same as what I am doing. Here it is at least explicit. I will add a comment to explain that reasoning.

xrmx: > About the testing itself, the only way to test it is by mocking out connection and cursor and then instrument it. It looks like I am testing the instrumented function in that case, however it is exactly the same as what I am doing. Here it is at least explicit. I will add a comment to explain that reasoning.

If you call the instrumented function we may change the instrumentation implementation without need to touch tests

xrmx:
@bartv regarding async tests https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2540

instrumentation/opentelemetry-instrumentation-asyncpg/tests/test_asyncpg_wrapper.py
```diff
@@ -34,3 +39,52 @@ def test_duplicated_uninstrumentation(self):
...
+        async def mock_fn_stop(*args, **kwargs):
+            raise StopAsyncIteration()
+
+        cursor_mock = mock.Mock()
+        cursor_mock._query = "SELECT * FROM test"
+
+        apg = AsyncPGInstrumentor()
+        apg.instrument(tracer_provider=self.tracer_provider)
+
+        # Call the wrapper function directly. They only way to be able to do this on the real classes is to mock all of the
```
xrmx, line 75: If you update this to use the wrapped function, as most of the other tests in this repository I guess, I'll approve and eventually move things forward

bartv: I now test the functions directly. This requires to mock out all connection related handling in asyncpy.cursor

xrmx:
@bartv next time please configure pre-commit as documented in CONTRIBUTING

bartv:
> @bartv next time please configure pre-commit as documented in CONTRIBUTING

ok. done. Do you squash merge or do I need to squash the commit history myself?

bartv:
How do I handle this error?

```
lint-instrumentation-asyncpg: commands[4]> pylint /home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-asyncpg/tests
************* Module tests.test_asyncpg_wrapper
instrumentation/opentelemetry-instrumentation-asyncpg/tests/test_asyncpg_wrapper.py:103:24: C2801: Unnecessarily calls dunder method __anext__. Use anext built-in function. (unnecessary-dunder-call)
```

anext only exists in python3.10 and later. I switched to calling __anext__ directly to make the tests pass on 3.8

xrmx:
> How do I handle this error?
> 
> ```
> lint-instrumentation-asyncpg: commands[4]> pylint /home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-asyncpg/tests
> ************* Module tests.test_asyncpg_wrapper
> instrumentation/opentelemetry-instrumentation-asyncpg/tests/test_asyncpg_wrapper.py:103:24: C2801: Unnecessarily calls dunder method __anext__. Use anext built-in function. (unnecessary-dunder-call)
> ```
> 
> anext only exists in python3.10 and later. I switched to calling **anext** directly to make the tests pass on 3.8

Does something like this work? Otherwise just silence the warning with a pylint comment and open an issue with the exact line to fix properly

```
diff --git a/.pylintrc b/.pylintrc
index 114dadef..dc3a0525 100644
--- a/.pylintrc
+++ b/.pylintrc
@@ -45,6 +45,8 @@ suggestion-mode=yes
 # active Python interpreter and may run arbitrary code.
 unsafe-load-any-extension=no
 
+# Run python dependant checks considering the baseline version
+py-version=3.8
 
 [MESSAGES CONTROL]
 
```

xrmx (approved):

ocelotl:
> @xrmx What do I need to do to get this merged?
> 
> image

I'll merge this PR :v:

PR #2508, opened 2024-05-08
WIP

### LINKED ISSUE #2478: Separate `tox.ini` into multiple files
Right now, `tox.ini` is extremely long and it is easy to overlook important testing details. Separate `tox.ini` into multiple files, one per package.

A "global" `tox.ini` file may still remain after this change if it is convenient.

ocelotl:
Fixes #2478


tox.ini
```diff
@@ -1169,8 +1126,6 @@ commands =
   lint-exporter-prometheus-remote-write: pylint {toxinidir}/exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry
   lint-exporter-prometheus-remote-write: pylint {toxinidir}/exporter/opentelemetry-exporter-prometheus-remote-write/tests
 
-  coverage: {toxinidir}/scripts/coverage.sh
```
ocelotl, line 1172: Right now, we should not run coverage nor benchmark tests in CI. As they are configured right now, the coverage tests will not fail if a decrease in coverage is detected, the benchmark test will not fail if a decrease in performance is detected. For this reason, they cannot fail CI in a useful manner and it makes no sense to spend resources running them.

For these tests to be usefully executed in CI we need to configure them so that they can fail CI if a decrease in performance (for the benchmark tests) or a decrease in coverage (for the coverage tests) is detected.

tox.ini
```diff
@@ -1120,20 +1091,6 @@ commands =
 
   test-util-http: pytest {toxinidir}/util/opentelemetry-util-http/tests {posargs}
 
-  test-sdk-extension-aws: pytest {toxinidir}/sdk-extension/opentelemetry-sdk-extension-aws/tests {posargs}
```
ocelotl, line 1123: The size of this `tox.ini` file will be greatly reduced when the rest of the packages are refactored.

lzchen:
I'm not really super enthusiastic about having so many `tox` files. Is the largeness of the singular tox file so much that it warrants separating these out? 

From the issue:

> Right now, tox.ini is extremely long and it is easy to overlook important testing details.

What "important testing details" are you referring to specifically? Maybe if we call them out, there are smaller, less intrusive changes we can make.

ocelotl:
> I'm not really super enthusiastic about having so many `tox` files. Is the largeness of the singular tox file so much that it warrants separating these out?
> 
> From the issue:
> 
> > Right now, tox.ini is extremely long and it is easy to overlook important testing details.
> 
> What "important testing details" are you referring to specifically? Maybe if we call them out, there are smaller, less intrusive changes we can make.

When I was introducing the changes that split `lint` into multiple tests, I noticed that we were missing several missing testing for a few of our components. It was hard to notice among the hundreds of lines in our `tox.ini`.

Nevertheless, splitting the `tox.ini` file is just another enhancement that this PR will introduce, please take a look at the advantages listed in this PR description.

aabmass:
My main concern with this approach is that the common parts of the config that all tox.ini share may drift over time and become inconsistent or difficult to maintain. For example, adding/removing a supported Python version will require updating every tox file. More subtle things like changing the flags to a command or adding a new tool would be tricky.

IIRC tox doesn't really have any mechanism for re-use like include/extend?

ocelotl:
> My main concern with this approach is that the common parts of the config that all tox.ini share may drift over time and become inconsistent or difficult to maintain. For example, adding/removing a supported Python version will require updating every tox file.

That's right, but we have the very same problem now. We can forget to add a new python version in some package in the `tox.ini` file as it is right now too. Having one single `tox.ini` file is no guarantee of consistency either.

> More subtle things like changing the flags to a command or adding a new tool would be tricky.

Yes, but again, we have the same problem right now, we can make a mistake and forget a command argument for a particular package even if we only have one `tox.ini` file.

tox.ini
aabmass: 👍 makes sense. We should a file a bug regardless of this PR to set up coverage in CI

tox.ini
```diff
@@ -371,10 +361,8 @@ deps =
   lint: -r dev-requirements.txt
   test: pytest
   test: pytest-benchmark
-  coverage: pytest
-  coverage: pytest-cov
+  grpc: pytest-asyncio
```
xrmx, line 364: You don't need this anymore

.github/workflows/instrumentations_0.yml
```diff
@@ -120,4 +120,4 @@ jobs:
             ~/.cache/pip
           key: v7-build-tox-cache-${{ env.RUN_MATRIX_COMBINATION }}-${{ hashFiles('tox.ini', 'gen-requirements.txt', 'dev-requirements.txt') }}
       - name: run tox
-        run: tox -f ${{ matrix.python-version }}-${{ matrix.package }} -- -ra --benchmark-json=${{ env.RUN_MATRIX_COMBINATION }}-benchmark.json
```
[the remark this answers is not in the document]
ocelotl: Fixed in #2660

ocelotl:
#2666, we have typos in our `tox.ini` file.

ocelotl:
Another problem we have with a big `tox.ini` file is that we can't have consistent `tox` environment names.

We need `py3{8,9,10,11,12}-test-opentelemetry-instrumentation` and `py3{8,9,10,11,12}-test-instrumentation-aiohttp-client` instead of `py3{8,9,10,11,12}-test-opentelemetry-instrumentation-aiohttp-client` because we have commands like this:

`test-opentelemetry-instrumentation: pytest {toxinidir}/opentelemetry-instrumentation/tests {posargs}`

That command above would incorrectly match `py3{8,9,10,11,12}-test-opentelemetry-instrumentation-aiohttp-client` too.

We can save ourselves the trouble of handling these collisions by having one `tox.ini` file per package.

PR #2517, opened 2024-05-10
Add error handling to opentelemetry-bootstrap -a

### LINKED ISSUE #2516: Add error handling for uninstallable packages to `opentelemetry-bootstrap -a install`
Right now we are doing this error handling [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/46d2ce6acea9a1a6cb1a4d4c863077002f5f7d21/scripts/otel_packaging.py#L46). It should be [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/46d2ce6acea9a1a6cb1a4d4c863077002f5f7d21/opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py#L51).

The situation we are right now is that if we add a new instrumentation that does not have a third-party package as a dependency (like `threading`), then it won't be added to the [`default_instrumentations`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/46d2ce6acea9a1a6cb1a4d4c863077002f5f7d21/opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap_gen.py#L184) as it should. This happens because the [script that gets all instrumentations](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/46d2ce6acea9a1a6cb1a4d4c863077002f5f7d21/scripts/otel_packaging.py#L46) won't get this instrumentation because it first checks if it is in Pypi (it is obviously not because we are just adding it to the repo).

ocelotl:
Fixes #2516

### DIFF
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py
@@ -14,8 +14,14 @@
 
 import argparse
 import logging
-import subprocess
 import sys
+from subprocess import (
+    PIPE,
+    CalledProcessError,
+    Popen,
+    SubprocessError,
+    check_call,
+)
 
 import pkg_resources
 
@@ -34,7 +40,7 @@ def _syscall(func):
             if package:
                 return func(package)
             return func()
-        except subprocess.SubprocessError as exp:
+        except SubprocessError as exp:
             cmd = getattr(exp, "cmd", None)
             if cmd:
                 msg = f'Error calling system command "{" ".join(cmd)}"'
@@ -48,18 +54,21 @@ def _syscall(func):
 @_syscall
 def _sys_pip_install(package):
     # explicit upgrade strategy to override potential pip config
-    subprocess.check_call(
-        [
-            sys.executable,
-            "-m",
-            "pip",
-            "install",
-            "-U",
-            "--upgrade-strategy",
-            "only-if-needed",
-            package,
-        ]
-    )
+    try:
+        check_call(
+            [
+                sys.executable,
+                "-m",
+                "pip",
+                "install",
+                "-U",
+                "--upgrade-strategy",
+                "only-if-needed",
+                package,
+            ]
+        )
+    except CalledProcessError as error:
+        print(error)
 
 
 def _pip_check():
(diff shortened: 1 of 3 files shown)


xrmx:
Isn't this reintroducing #2053 though?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py
```diff
@@ -48,18 +54,21 @@ def wrapper(package=None):
...
+                "-m",
+                "pip",
+                "install",
+                "-U",
+                "--upgrade-strategy",
+                "only-if-needed",
+                package,
+            ]
+        )
+    except CalledProcessError as error:
```
lzchen, line 70: To clarify the experience, `bootstrap_gen.py` will now contain the generated instrumentations from the local directories (which may include unreleased instrumentations). When users use `opentelemetry-bootstrap`, the logic will run this code, and the logic iterates through the generated file but instead of throwing an exception, we catch it and print the error instead correct?

lzchen (approved):
LGTM with one clarifying question.

lzchen:
> Isn't this reintroducing #2053 though?

@xrmx 

I think it is reintroducing the error however we are catching the exception, printing the error and then continuing instead of failing the operation.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py
tammy-baylis-swi: That handling is what I'm seeing with `opentelemetry-bootstrap` with this change (version changed to `0.45b0`). The logs include these so the messages are still visible:

```
ERROR: Could not find a version that satisfies the requirement opentelemetry-instrumentation-threading==0.45b0 (from versions: none)
ERROR: No matching distribution found for opentelemetry-instrumentation-threading==0.45b0
Command '['/usr/local/bin/python', '-m', 'pip', 'install', '-U', '--upgrade-strategy', 'only-if-needed', 'opentelemetry-instrumentation-threading==0.45b0']' returned non-zero exit status 1.

ERROR: Could not find a version that satisfies the requirement opentelemetry-instrumentation-aiohttp-server==0.45b0 (from versions: none)
ERROR: No matching distribution found for opentelemetry-instrumentation-aiohttp-server==0.45b0
Command '['/usr/local/bin/python', '-m', 'pip', 'install', '-U', '--upgrade-strategy', 'only-if-needed', 'opentelemetry-instrumentation-aiohttp-server==0.45b0']' returned non-zero exit status 1.
```

Bootstrapping does not crash; no more traceback with uncaught `subprocess.CalledProcessError`. 👍 👍  I can then `opentelemetry-instrument` my test service and the other, successfully-installed instrumentors generate spans.

tammy-baylis-swi (approved):
I gave this a little smoke test ([comment](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2517#discussion_r1598875092)) and lgtm, thank you again @ocelotl

xrmx (approved):

opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py
ocelotl: > To clarify the experience, `bootstrap_gen.py` will now contain the generated instrumentations from the local directories (which may include unreleased instrumentations). When users use `opentelemetry-bootstrap`, the logic will run this code, and the logic iterates through the generated file but instead of throwing an exception, we catch it and print the error instead correct?

That's right :v:

PR #2524, opened 2024-05-15
elasticsearch: don't produce spans if native elasticsearch support is enabled

### LINKED ISSUE #2393: Handle elasticsearch client native instrumentation
**Describe your environment**

Elasticsearch client added native opentelemetry instrumentation in [8.13.0](https://www.elastic.co/guide/en/elasticsearch/client/python-api/master/release-notes.html#rn-8-13-0) and so currently without manual intervention it is possible to have our instrumentation and the client one create spans concurrently.

**Steps to reproduce**

Add opentelemetry instrumentation using elasticsearch client 8.13.0

Only the elasticsearch client creates spans.

Both client and opentelemetry instrumentation creates their own spans.

**Additional context**

Java opentelemetry instrumentation disables its instrumentation if a new enough elasticsearch client is found https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/9337/

xrmx:
# Description

If elasticsearch is found with native OTel support enabled just call the wrapped function without creating our own spans.

Fixes #2393

## Type of change

# How Has This Been Tested?

- [x] tox -e py310-test-instrumentation-elasticsearch-2

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


xrmx:
cc @pquentin

tox.ini
```diff
@@ -92,7 +92,7 @@ envlist =
     ; below mean these dependencies are being used:
     ; 0: elasticsearch-dsl==6.4.0 elasticsearch==6.8.2
     ; 1: elasticsearch-dsl==7.4.1 elasticsearch==7.17.9
-    ; 2: elasticsearch-dsl>=8.0,<8.13 elasticsearch>=8.0,<8.13
+    ; 2: elasticsearch-dsl==8.13.1 elasticsearch==8.13.1
```
pquentin, line 95: Are we still testing versions 8.0 to 8.12?

pquentin (commented):
Thanks for this Riccardo! I wonder, though, should we add a warning in that case?

The native instrumentation is more correct, following the semantic conventions, but also very different. If you install this package and elasticsearch-py 8.13+, you're going to be very surprised. So a warning saying to uninstall opentelemetry-instrumentation-elasticsearch because it does not cover elasticsearch-py 8.13+ would be useful IMO.

xrmx:
> Thanks for this Riccardo! I wonder, though, should we add a warning in that case?

I can add that only if we have a rate limited logger because we are still checking the span at each perform request call

> The native instrumentation is more correct, following the semantic conventions, but also very different. If you install this package and elasticsearch-py 8.13+, you're going to be very surprised. So a warning saying to uninstall opentelemetry-instrumentation-elasticsearch because it does not cover elasticsearch-py 8.13+ would be useful IMO.

Will add something to the documentation

tox.ini
xrmx: Nope, i tested it manually. Do you think it would be different enough we need to run separate tests? Thinking about it the test should fail with elasticsearch < 8.13

pquentin: As long as we're testing a version 8 client, it's fine. I was worried that by only using 8.13 we would not exercise the code path used in 8.12, for example.

pquentin:
> I can add that only if we have a rate limited logger because we are still checking the span at each perform request call

Warnings are deduplicated already, you can run this example to see for yourself:

```python
import warnings

def deprecated_api(message):
    warnings.warn(message, Warning, stacklevel=2)

for i in range(10):
    deprecated_api("test")
```

tox.ini
xrmx: We are running all tests with native otel disabled and just one with it enabled

xrmx:
Updated PR: fixed typo in env variables, added warning to the docs and code, limited tests with native Otel to elasticsearch > 8.13

instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
```diff
@@ -197,6 +206,14 @@ def _wrap_perform_request(
 ):
     # pylint: disable=R0912,R0914
     def wrapper(wrapped, _, args, kwargs):
+        # if wrapped elasticsearch has native OTel instrumentation just call the wrapped function
+        otel_span = kwargs.get("otel_span")
+        if otel_span and otel_span.otel_span:
+            warnings.warn(
+                "Using native elasticsearch OTel instrumentation", Warning
```
pquentin, line 213: This isn't enough, we should:

 * say that the instrumentation for this package is disabled in favor of the Elasticsearch package
 * ideally link to a page that contains the information 
 * ideally figure out the correct stacklevel so that the warning appears in user code

Here's a good example: https://github.com/urllib3/urllib3/blob/b8589ec9f8c4da91511e601b632ac06af7e7c10e/src/urllib3/poolmanager.py#L422-L430

instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
```diff
@@ -16,6 +16,14 @@
 This library allows tracing HTTP elasticsearch made by the
 `elasticsearch <https://elasticsearch-py.readthedocs.io/en/master/>`_ library.
 
+.. warning::
+    The elasticsearch package got native OpenTelemetry support since version
+    `8.13 <https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/release-notes.html#rn-8-13-0>`_.
+    To avoid duplicated tracing this instrumentation disables itself if it finds an elasticsearch client
+    that has OpenTelemetry support enabled.
+
+    Please be aware that the two libraries may use a different semantic convention.
```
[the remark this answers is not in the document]
xrmx: Added link to elasticsearch documentation

instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
xrmx: Added 1) and 2). For 3) I prefer not poking with stacktraces since this is a _service_ issue not local to some callers.

instrumentation/opentelemetry-instrumentation-elasticsearch/src/opentelemetry/instrumentation/elasticsearch/__init__.py
```diff
@@ -197,6 +207,16 @@ def _wrap_perform_request(
 ):
     # pylint: disable=R0912,R0914
     def wrapper(wrapped, _, args, kwargs):
+        # if wrapped elasticsearch has native OTel instrumentation just call the wrapped function
+        otel_span = kwargs.get("otel_span")
+        if otel_span and otel_span.otel_span:
+            warnings.warn(
+                "Instrumentation disabled, relying on elasticsearch native OTel support, see "
+                "https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/elasticsearch/elasticsearch.html ",
```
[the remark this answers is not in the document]
xrmx: thanks

xrmx, line 215:
```suggestion
                "https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/elasticsearch/elasticsearch.html",
```

pquentin (approved):

tammy-baylis-swi (approved):
Thanks for comments as well

PR #2532, opened 2024-05-20
kafka-python-ng: Instrument temporary fork kafka-python-ng by copying kafka-python's instrumentation

rjduffner:
# Description

This PR creates a new kafka-python-ng instrumentation to instrument the new (and temporary) fork of kafka-python called kafka-python-ng (https://github.com/wbarnha/kafka-python-ng). It copies all of the instrumentation from the kafka-python instrumentation and places it in the new folder.

The old maintainer of kafka-python is no longer able to support the project and has passed the project off to a new maintainer. However, the new maintainer does not yet have access to the kafka-python namespace in pypi so cannot release a new version. See https://github.com/dpkp/kafka-python/issues/2431

Since this fork looks to be semi temporary, I think its a little strange to create a whole new instrumentation for it but the issue linked above gives no timeline on when we might expect the reunification.

I originally tried to use the old instrumentation and just point it at either kafka-python or kafka-python-ng but it didn't seemed to treat comma delineated packages as ANDs instead of ORs. If there is a way to say, "this package or that package" I think that might be a better solution but I did not see any examples of it. 

## Type of change

# How Has This Been Tested?

Tested locally. Same tests as the previous kafka-python instrumentation

- [ ] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#2397](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2397)))
 - `opentelemetry-processor-baggage` Initial release
   ([#2436](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2436))
+- `opentelemetry-instrumentation-kafka-python` Instrument temporary fork kafka-python-ng alongside kafka-python 
+  ([#2532](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2532))
 
 ### Fixed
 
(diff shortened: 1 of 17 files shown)


xrmx:
Yeah, I think it would be better to fix the ability to wrap the same code for more than one package instead of publishing duplicated or even temporary packages.

rjduffner:
@xrmx, ya i think the same thing. And it may not be too hard to do if the packages installed in seperate namespaces/import paths.

As they sit now both packages (kafka-python and kafka-python-ng) both use "import kafka". So in the __init__.py where I would presumably hook in here, https://github.com/rjduffner/opentelemetry-python-contrib/blob/kafka-python-ng-add-rd/instrumentation/opentelemetry-instrumentation-kafka-python-ng/src/opentelemetry/instrumentation/kafka/__init__.py#L91-L122 I can't differentiate (as far as I know) between the two packages.

While the case where one or the other is installed is fine, if the user has them both installed by accident, this may lead to strange results.

rjduffner:
Second stab at the original idea of instrument either kafka-python or kafka-python-ng here https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2537

xrmx:
Closing in favor of #2537

PR #2533, opened 2024-05-20
Update functions detector readme

jeremydvoss:
# Description

Update Functions Resource Detector Readme.


xrmx:
Do you consider them experimental because something is missing or just for the semantic convention status?

jeremydvoss:
> Do you consider them experimental because something is missing or just for the semantic convention status?

The spec for this detector has not been finalized and marked as stable in OTel. It is currently an internal spec and may change slgihtly in the future as it goes stable.

lzchen (approved):

PR #2537, opened 2024-05-21
kafka-python Instrument temporary fork, kafka-python-ng inside kafka-python's instrumentation

rjduffner:
# Description

This PR adds kafka-python-ng, a temporary fork of kafka-python due to pypi access issues, to the kafka-python instrumentation.

## Type of change

# How Has This Been Tested?

- [ ] Test 1:  Setup the code locally and then made sure that a sample pyramid app using kafka-python-ng instrumented without any errors.

- [ ] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/package.py
+++ b/instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/package.py
@@ -13,4 +13,7 @@
 # limitations under the License.
 
 
-_instruments = ("kafka-python >= 2.0",)
+_instruments_kafka_python = "kafka-python >= 2.0, < 3.0"
+_instruments_kafka_python_ng = "kafka-python-ng >= 2.0, < 3.0"
+
+_instruments = (_instruments_kafka_python, _instruments_kafka_python_ng)
--- a/tox.ini
+++ b/tox.ini
@@ -946,6 +946,9 @@ commands =
   lint-instrumentation-kafka-python: flake8 --config {toxinidir}/.flake8 {toxinidir}/instrumentation/opentelemetry-instrumentation-kafka-python
   lint-instrumentation-kafka-python: sh -c "cd instrumentation && pylint --rcfile ../.pylintrc opentelemetry-instrumentation-kafka-python"
 
+  ; Test only for kafka-pythonng instrumentation as the only difference between kafka-python and kafka-pythonng is the version of kafka-python
+  test-instrumentation-kafka-pythonng: pytest {toxinidir}/instrumentation/opentelemetry-instrumentation-kafka-python/tests {posargs}
+
   test-instrumentation-confluent-kafka: pytest {toxinidir}/instrumentation/opentelemetry-instrumentation-confluent-kafka/tests {posargs}
   lint-instrumentation-confluent-kafka: black --diff --check --config {toxinidir}/pyproject.toml {toxinidir}/instrumentation/opentelemetry-instrumentation-confluent-kafka
   lint-instrumentation-confluent-kafka: isort --diff --check-only --settings-path {toxinidir}/.isort.cfg {toxinidir}/instrumentation/opentelemetry-instrumentation-confluent-kafka
--- a/tox.ini
+++ b/tox.ini
@@ -350,7 +350,9 @@ envlist =
 
     ; opentelemetry-instrumentation-kafka-python
     py3{8,9,10,11}-test-instrumentation-kafka-python
+    py3{8,9,10,11,12}-test-instrumentation-kafka-pythonng
     pypy3-test-instrumentation-kafka-python
+    pypy3-test-instrumentation-kafka-pythonng
     lint-instrumentation-kafka-python
 
     ; opentelemetry-instrumentation-confluent-kafka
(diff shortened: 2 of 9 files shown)


rjduffner:
@xrmx, take a look at this one. I still need to do some doc updates but I think this is a better option vs #2537

xrmx:
@rjduffner My preferred approach would be something like this but unfortunately we are using a field with the semantics already defined (must be strings) https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2409

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/package.py
```diff
@@ -13,4 +13,5 @@
 # limitations under the License.
 
 
-_instruments = ("kafka-python >= 2.0",)
+_instruments = ("kafka-python >= 2.0, < 3.0",)
```
xrmx, line 16: Could you please keep _instruments unique and handle it in something like this?
https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2702/files#diff-616b2ef8de6c8fe9cc6278d0adaadb35a4a81667400d476ff472373049b6b20aR16

emdneto (commented):
I think would also be nice to have `kafka-python-ng` in a test-requirements file to check we support it on tests as well

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/package.py
rjduffner: Updated

rjduffner:
@emdneto, it looks like fastapi and fastapi-slim do something similar

```  fastapi: pip install opentelemetry-api@{env:CORE_REPO}\#egg=opentelemetry-api&subdirectory=opentelemetry-api
  fastapi: pip install opentelemetry-semantic-conventions@{env:CORE_REPO}\#egg=opentelemetry-semantic-conventions&subdirectory=opentelemetry-semantic-conventions
  fastapi: pip install opentelemetry-sdk@{env:CORE_REPO}\#egg=opentelemetry-sdk&subdirectory=opentelemetry-sdk
  fastapi: pip install opentelemetry-test-utils@{env:CORE_REPO}\#egg=opentelemetry-test-utils&subdirectory=tests/opentelemetry-test-utils
  fastapi: pip install -r {toxinidir}/instrumentation/opentelemetry-instrumentation-fastapi/test-requirements.txt
  fastapi-slim: pip install opentelemetry-api@{env:CORE_REPO}\#egg=opentelemetry-api&subdirectory=opentelemetry-api
  fastapi-slim: pip install opentelemetry-semantic-conventions@{env:CORE_REPO}\#egg=opentelemetry-semantic-conventions&subdirectory=opentelemetry-semantic-conventions
  fastapi-slim: pip install opentelemetry-sdk@{env:CORE_REPO}\#egg=opentelemetry-sdk&subdirectory=opentelemetry-sdk
  fastapi-slim: pip install opentelemetry-test-utils@{env:CORE_REPO}\#egg=opentelemetry-test-utils&subdirectory=tests/opentelemetry-test-utils
  fastapi-slim: pip install -r {toxinidir}/instrumentation/opentelemetry-instrumentation-fastapi/test-requirements-slim.txt
```

Ive already copied what they are doing for the instruments part, ill copy them for the tests too. Thanks for the suggestion.

emdneto (commented):
@rjduffner It seems we have a problem with the way fastapi-slim support was implemented, so hopefully we can use the same here. See https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2756

rjduffner:
@emdneto

Ha, I just saw that and switched my code back to doing it the way the fix for fastAPI is doing it. Great timing

.github/workflows/instrumentations_0.yml
```diff
@@ -76,8 +76,6 @@ jobs:
         exclude:
           - python-version: py312
             package: "boto"
-          - python-version: py312
-            package: "kafka-python"
```
rjduffner, line 80: So while kafka-python does not currently support py312, kafka-python-ng does and I removed this.

rjduffner:
@xrmx and @emdneto

Thanks for the feedback. I think it is all integrated at this time.

tox.ini
```diff
@@ -443,6 +445,10 @@ commands_pre =
   kafka-python: pip install opentelemetry-semantic-conventions@{env:CORE_REPO}\#egg=opentelemetry-semantic-conventions&subdirectory=opentelemetry-semantic-conventions
   kafka-python: pip install opentelemetry-sdk@{env:CORE_REPO}\#egg=opentelemetry-sdk&subdirectory=opentelemetry-sdk
   kafka-python: pip install -r {toxinidir}/instrumentation/opentelemetry-instrumentation-kafka-python/test-requirements.txt
+  kafka-python-ng: pip install opentelemetry-api@{env:CORE_REPO}\#egg=opentelemetry-api&subdirectory=opentelemetry-api
```
xrmx, line 448: Do you the same problem we have with fastapi-slim so that `kafka-python` pre-commands would run for `kafka-python-ng` too? We resolved by using `fastapislim` without the dash to avoid that.

.github/workflows/instrumentations_0.yml
xrmx: A bit confused on how this could possibly work, I think we should rename the rule and and a specific entry here in the workflow for the `-ng` package, see https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2537/files#r1699836631

emdneto: In that case, I'd vote to keep as-is and just add `kafka-pythonng` to CI

tox.ini
rjduffner: Switched to kafka-pythonng and followed fastapislim's pr here https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2756/files

.github/workflows/instrumentations_0.yml
rjduffner: Since we are separating out kafka-python-ng and kafka-python, i readded this and followed fastapislim's example in instrumentations_1.yaml

rjduffner:
@xrmx, I updated the tox to switch everything to kafka-pythonng from kafka-python-ng.

However, now the action only installs the kafka-python-ng instrumentation and doesn't actually test with it because I didn't update these to test kafka-python-ng side

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/tox.ini#L937-L941

I saw that fastapislim didn't update those but that seems like an oversight. Im guessing we want a section here for the kafka-python-ng tests that would be identical to the kafka-python ones.

emdneto:
@rjduffner, you're right; the job isn't running the test. I will submit a fix.

> @xrmx, I updated the tox to switch everything to kafka-pythonng from kafka-python-ng.
> 
> However, now the action only installs the kafka-python-ng instrumentation and doesn't actually test with it because I didn't update these to test kafka-python-ng side
> 
> https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/tox.ini#L937-L941
> 
> I saw that fastapislim didn't update those but that seems like an oversight. Im guessing we want a section here for the kafka-python-ng tests that would be identical to the kafka-python ones.

rjduffner:
@emdneto cool, ill do the same here. Thanks for the confirmation.

xrmx (approved):

ocelotl (approved):

PR #2538, opened 2024-05-21
Ensure httpx non-client methods are instrumented

rbagd:
# Description

This is an update of [previous MR](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2011) for which the original author unfortunately didn't submit the updates required for merging into `main`. We actively use this instrumentation and this instrumentation gap is quite annoying to deal with.

Quote from MR

The `httpx.get`, `httpx.post`, etc. methods ultimately use `httpx._api.request`, which makes instantiates `httpx._api.Client`. But `httpx.Client` is modified during instrumentation without updating the value imported in the `httpx._api` module.

Fixing this was as easy as also patching the `Client` value in `httpx._api`.

All credit for the fix goes to @dmontagu.

Fixes #1742 

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Unit test which tests instrumentation of various ways to make a request in `httpx` without `httpx.Client`. Uninstrumentation is also tested.

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [x] Followed the style guidelines of this project
- [x] Changelogs have been updated
- [x] Unit tests have been added
- [ ] Documentation has been updated

Issue #1742, opened 2023-04-03
httpx instrumentation doesn't work for `httpx.get`, `httpx.post`, etc.


macieyng:
**Describe your environment** Describe any aspect of your environment relevant to the problem, including your Python version, [platform](https://docs.python.org/3/library/platform.html), version numbers of installed dependencies, information about your cloud hosting provider, etc. If you're reporting a problem with a specific version of a library in this repo, please check whether the problem has been fixed on main.
```
#Python
Python 3.8.16

# OTEL
opentelemetry-api==1.15.0 ; python_version >= '3.7'
opentelemetry-instrumentation==0.36b0 ; python_version >= '3.7'
opentelemetry-instrumentation-httpx==0.36b0
opentelemetry-instrumentation-requests==0.36b0
opentelemetry-sdk==1.15.0 ; python_version >= '3.7'

# 3rd Party
httpx==0.23.3
requests==2.28.2 ; 
python_version >= '3.6'
```

Azure Function run locally.

**Steps to reproduce**
Describe exactly how to reproduce the error. Include a code sample if applicable.
1. Install `requests` and `httpx` libs and instrumentations
2. Initialize instrumentors
3. Make a `get/post/etc` request to some api without initializing a client.
4. Make a `get/post/etc` request to some api with initializing a client.

```python
import httpx 
import requests

from opentelemetry.instrumentation.httpx import HTTPXClientInstrumentor
from opentelemetry.instrumentation.requests import RequestsInstrumentor

HTTPXClientInstrumentor().instrument()
RequestsInstrumentor().instrument()


requests.get("www.google.com")
httpx.get("www.google.com")


with httpx.Client() as client:
    response = client.get("https://www.google.com")
with requests.Session() as session:
    response = session.get("https://www.google.com")

```

**What is the expected behavior?**
What did you expect to see?

Whenever I use `httpx.get` (or similar) a new dependency/span should be created and exported - in my case to Azure Application Insights.

**What is the actual behavior?**
What did you see instead?

I don't see expected dependencies/spans/traces exported.

<img width="628" alt="Zrzut ekranu 2023-04-3 o 12 53 01" src="https://user-images.githubusercontent.com/82878433/229490157-edb23b78-a14e-4b21-80b8-f82171414449.png">

**Additional context**
Full example of the function. 

```python
import httpx
import logging
import requests

import azure.functions as func
from opentelemetry.instrumentation.httpx import HTTPXClientInstrumentor
[...]

PR #2546, opened 2024-05-23
feat(asgi,fastapi,starlette)!: provide both send and receive hooks with `scope` and `message`

### LINKED ISSUE #2560: ASGI, FastAPI, Starlette: provide both send and receive hooks with `scope` and `message`

Currently, a «receive» hook is only provided with `scope`, but not `message`. On contrary, a «send» hook is only provided with `message`, but not `scope`. In particular, it makes it impossible for a receive hook to inspect events.

Unify both hooks so that they both would accept the both parameters.

**Additional context**

The feature request is already implemented in this PR: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2546

This is going to be a breaking change, since the hooks will accept additional parameters.

eigenein:
# Description

Currently, a «receive» hook is only provided with `scope`, but not `message`. On contrary, a «send» hook is only provided with `message`, but not `scope`.

This change unifies the both hooks and provides them with additional request info.

This is a breaking change:

- The hooks will now be provided with one more positional argument
- `client_request_hook` will be called _after_ `receive()` instead of _before_ `receive()`

Resolves #2560.

## Type of change

# How Has This Been Tested?

Existing unit tests have been updated.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#2425](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2425))
 - `opentelemetry-instrumentation-flask` Add `http.method` to `span.name`
   ([#2454](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2454))
+- ASGI, FastAPI, Starlette: provide both send and receive hooks with `scope` and `message` for internal spans ([#2546](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2546))
 
 ### Added
 
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
@@ -81,15 +81,15 @@ For example,
 
 .. code-block:: python
 
-    def server_request_hook(span: Span, scope: dict):
+    def server_request_hook(span: Span, scope: dict[str, Any]):
         if span and span.is_recording():
             span.set_attribute("custom_user_attribute_from_request_hook", "some-value")
 
-    def client_request_hook(span: Span, scope: dict):
+    def client_request_hook(span: Span, scope: dict[str, Any], message: dict[str, Any]):
         if span and span.is_recording():
             span.set_attribute("custom_user_attribute_from_client_request_hook", "some-value")
 
-    def client_response_hook(span: Span, message: dict):
+    def client_response_hook(span: Span, scope: dict[str, Any], message: dict[str, Any]):
         if span and span.is_recording():
             span.set_attribute("custom_user_attribute_from_response_hook", "some-value")
 
@@ -200,6 +200,11 @@ from typing import Any, Awaitable, Callable, Tuple
 from asgiref.compatibility import guarantee_single_callable
 
 from opentelemetry import context, trace
+from opentelemetry.instrumentation.asgi.types import (
+    ClientRequestHook,
+    ClientResponseHook,
+    ServerRequestHook,
+)
 from opentelemetry.instrumentation.asgi.version import __version__  # noqa
 from opentelemetry.instrumentation.propagators import (
     get_global_response_propagator,
@@ -212,7 +217,7 @@ from opentelemetry.metrics import get_meter
 from opentelemetry.propagators.textmap import Getter, Setter
 from opentelemetry.semconv.metrics import MetricInstruments
 from opentelemetry.semconv.trace import SpanAttributes
-from opentelemetry.trace import Span, set_span_in_context
+from opentelemetry.trace import set_span_in_context
 from opentelemetry.trace.status import Status, StatusCode
 from opentelemetry.util.http import (
     OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SANITIZE_FIELDS,
@@ -226,10 +231,6 @@ from opentelemetry.util.http import (
     normalise_response_header_name,
     remove_url_credentials,
 )
-
-_ServerRequestHookT = typing.Optional[typing.Callable[[Span, dict], None]]
-_ClientRequestHookT = typing.Optional[typing.Callable[[Span, dict], None]]
-_ClientResponseHookT = typing.Optional[typing.Callable[[Span, dict], None]]
 
 
 class ASGIGetter(Getter[dict]):
@@ -454,10 +455,10 @@ class OpenTelemetryMiddleware:
                       Optional: Defaults to get_default_span_details.
         server_request_hook: Optional callback which is called with the server span and ASGI
                       scope object for every incoming request.
-        client_request_hook: Optional callback which is called with the internal span and an ASGI
-                      scope which is sent as a dictionary for when the method receive is called.
-        client_response_hook: Optional callback which is called with the internal span and an ASGI
-                      event which is sent as a dictionary for when the method send is called.
+        client_request_hook: Optional callback which is called with the internal span, and ASGI
+                      scope and event which are sent as dictionaries for when the method receive is called.
+        client_response_hook: Optional callback which is called with the internal span, and ASGI
+                      scope and event which are sent as dictionaries for when the method send is called.
         tracer_provider: The optional tracer provider to use. If omitted
             the current globally configured one is used.
     """
@@ -468,9 +469,9 @@ class OpenTelemetryMiddleware:
         app,
         excluded_urls=None,
         default_span_details=None,
-        server_request_hook: _ServerRequestHookT = None,
-        client_request_hook: _ClientRequestHookT = None,
-        client_response_hook: _ClientResponseHookT = None,
+        server_request_hook: ServerRequestHook = None,
+        client_request_hook: ClientRequestHook = None,
+        client_response_hook: ClientResponseHook = None,
         tracer_provider=None,
         meter_provider=None,
         meter=None,
@@ -666,9 +667,9 @@ class OpenTelemetryMiddleware:
             with self.tracer.start_as_current_span(
                 " ".join((server_span_name, scope["type"], "receive"))
             ) as receive_span:
-                if callable(self.client_request_hook):
-                    self.client_request_hook(receive_span, scope)
                 message = await receive()
+                if callable(self.client_request_hook):
+                    self.client_request_hook(receive_span, scope, message)
                 if receive_span.is_recording():
                     if message["type"] == "websocket.receive":
                         set_status_code(receive_span, 200)
@@ -691,7 +692,7 @@ class OpenTelemetryMiddleware:
                 " ".join((server_span_name, scope["type"], "send"))
             ) as send_span:
                 if callable(self.client_response_hook):
-                    self.client_response_hook(send_span, message)
+                    self.client_response_hook(send_span, scope, message)
                 if send_span.is_recording():
                     if message["type"] == "http.response.start":
                         status_code = message["status"]
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/types.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/types.py
@@ -1,0 +1,49 @@
+# Copyright The OpenTelemetry Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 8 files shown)


eigenein:
Fixed the linter errors, let's try it again

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -227,10 +232,6 @@ def client_response_hook(span: Span, message: dict):
     remove_url_credentials,
 )
 
-_ServerRequestHookT = typing.Optional[typing.Callable[[Span, dict], None]]
-_ClientRequestHookT = typing.Optional[typing.Callable[[Span, dict], None]]
-_ClientResponseHookT = typing.Optional[typing.Callable[[Span, dict], None]]
```
eigenein, line 232: Extracted these into `types.py` and made public, so that they could be reused in the FastAPI and Starlette instrumentations

eigenein:
Ready for review

tammy-baylis-swi:
Thank you for submitting this! A couple of asks:

> client_request_hook will be called after receive() instead of before receive()

(1) How has this been tested? Does the hook still work the same way to e.g. set span name or attributes?

(2) Please could you also create and link a github Issue describing the feature, like how [this PR](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2538) links [this issue](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1742), especially because this would be a breaking change.

eigenein:
Sure!

1. There are existing tests for hooks, that verify the hooks keep working the same way (they test exactly setting span name & attributes and would've got red should I break anything):
    - [ASGI](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py#L682-L713)
    - [FastAPI](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py#L353-L383)
    - [Starlette](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-starlette/tests/test_starlette_instrumentation.py#L270-L304)

    Of course, they went red because I added the new parameters, so I just updated the tests.

    The [call order change here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L670-L672) basically ensures that the hook could access the ASGI event. No functional changes, otherwise. I don't think the call order was part of the public interface, but pointed it out – just in case.

2. Would you like me to just copy-paste the description, or would need some additional info in there? This pull request is pretty self-contained, there was no bug or feature request in the first place.

tammy-baylis-swi:
>    The [call order change here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L670-L672) basically ensures that the hook could access the ASGI event. No functional changes, otherwise. I don't think the call order was part of the public interface, but pointed it out – just in case.

Great thanks!

> 2. Would you like me to just copy-paste the description, or would you need some additional info in there? This pull request is pretty self-contained, there was no bug or feature request in the first place.

Yes please a copy-paste works 🙂

tammy-baylis-swi (approved):
Also Lgtm, nice to see those types in shared ASGI module. We'll get eyes from the maintainers as well.

eigenein:
Done!

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -666,9 +667,9 @@ async def otel_receive():
             with self.tracer.start_as_current_span(
                 " ".join((server_span_name, scope["type"], "receive"))
             ) as receive_span:
-                if callable(self.client_request_hook):
-                    self.client_request_hook(receive_span, scope)
                 message = await receive()
+                if callable(self.client_request_hook):
+                    self.client_request_hook(receive_span, scope, message)
```
lzchen, line 672: Im confused as to why we are changing the order of this? Conceptually this was supposed to represent a hook that users can add that will be called BEFORE the receive call is made and right AFTER a span is created, such as in [requests](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py#L216) and [flask](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py#L437). It also follows the precedence of the [server request hook](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L617). This way users can implement logic like timing how long a request has taken.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#2425](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2425))
 - `opentelemetry-instrumentation-flask` Add `http.method` to `span.name`
   ([#2454](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2454))
+- ASGI, FastAPI, Starlette: provide both send and receive hooks with `scope` and `message` ([#2546](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2546))
```
lzchen, line 22:
```suggestion
- ASGI, FastAPI, Starlette: provide both send and receive hooks with `scope` and `message` for internal spans ([#2546](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2546))
```

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -81,15 +81,15 @@ async def hello():
 
 .. code-block:: python
 
-    def server_request_hook(span: Span, scope: dict):
+    def server_request_hook(span: Span, scope: dict[str, Any]):
         if span and span.is_recording():
             span.set_attribute("custom_user_attribute_from_request_hook", "some-value")
 
-    def client_request_hook(span: Span, scope: dict):
+    def client_request_hook(span: Span, scope: dict[str, Any], message: dict[str, Any]):
```
lzchen, line 88: > The [call order change here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L670-L672) basically ensures that the hook could access the ASGI event.

If request hooks were meant to be called right after span creation and before requests are made, should the hook even be allowed to accept the `message` field?

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
eigenein: > Conceptually this was supposed to represent a hook that users can add that will be called BEFORE the receive call is made and right AFTER a span is created, such as in [requests](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py#L216) and [flask](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py#L437).

I believe that my change actually makes it more aligned with those:

- In Flask, request is _already read_ and available in [`flask.request`](https://flask.palletsprojects.com/en/3.0.x/api/#flask.request) by the moment the hook is called. Similarly, `client_request_hook` would get called right after the event has been read, but still _before it gets processed_ in the [`self.app(...)`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L632) call down the chain
- In `requests`, the `request` is _already prepared_ by the moment the request hook is called
- In Django, the request hook is called [at the very end](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py#L289), when `request` is fully prepared

> It also follows the precedence of the [server request hook](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L617). This way users can implement logic like timing how long a request has taken.

Server request hook gets called before [`self.app(...)`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eaacb51b22fc88ae426ac5b8c665363e969613de/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L632) and hence, before `otel_receive` even gets called and before the hook/receive execution order gets to play its role. So, my change does not affect it at all, and I'm not sure if I understand your point well. Could you elaborate on that?

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
eigenein: I don't think there is even concensus on that, but the existing instrumentations seem to more agree with me: in Django, Requests, `urllib`, `urllib3`, and HTTPX instrumentations, the request hook is able to access the request instance. It's either this or there seems to be no concensus at all. E.g. Flask is a little weird as it receives only the `environ` part of the request instance even though the latter is available.

Also, I could argue that this:

> If request hooks were meant to be called right after span creation and before requests are made

depends heavily on definition of «before request is made». In ASGI instrumentation, we have `server_request_hook` that gets called before any reading and processing happens.

lzchen: Trying to consolidate what was said in [this](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2546#discussion_r1621092953) thread and the this one. I agree we should have a concensus on what these hooks should be doing and what parameters would add the most value/functionality to customers. Currently taken from the docs: 

> This instrumentation supports request and response hooks. These are functions that get called
right after a span is created for a request and right before the span is finished for the response.

This definition doesn't seem too helpful in terms of implementation as well as functionality. If we were to define how hooks would be useful, it probably would be "a place where information pertaining to the current request/response is most present so that I can use that information to run custom logic".

If we agree on this definition, then I see the value of having as much information available for the hooks as possible pertaining to those specific steps. However, I do have one question:

Also, apologies for my inexperience with ASGI, but if we were to make both client* hooks have the `message` and `scope` arguments and we change the ordering of the `client_request_hook` AFTER the actual recieve, what would be a use case in which the `client_request_hook` must be used but the `client_response_hook` can't? What would be the purpose of having both in this case?

lzchen (approved):

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
lzchen: Great! Thanks for the explanation!

PR #2580, opened 2024-06-05
Do not use `asgi` name and version for tracer/meter for instrumentations using Asgi Middleware

lzchen:
From [comment](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1199#discussion_r940917893)

For instrumentations that use asgi `OtelMiddleware` as an implementation detail, we want to use the `__name__` and `__version__` of the originating instrumentation instead of the inner `opentelemetry-instrumentation-asgi` for the `tracer` and `meter` components used.


xrmx (commented):
LGTM, maybe add a test in the asgi implementation to assert that?

lzchen:
@xrmx 

Done

xrmx (approved):

PR #2585, opened 2024-06-07
Add support for ALB events in opentelemetry-instrumentation-aws-lambda

jbfenton:
# Description

The current `opentelemetry-instrumentation-aws-lambda` module only supports conventional API Gateway events which typically contain a `headers` field that the trace context can be extracted from. 

ALB events will only contain a `headers` field when `Multi value headers` is disabled for the ALB in question. Otherwise the necessary headers will be found under the 'multiValueHeaders` field:
[ELB: Lambda functions as targets](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#multi-value-headers)

As a part of implementing this change, types + interfaces of the instrumentation have been hardened so that we avoid passing `Optional` parameters down the call chain wherever it is practical.

Handlers for processing types of events have also been placed in *wrappers* to make it a little easier to identify where we are pulling the trace context from, and where/why a given set of span attributes is being extracted + applied.

The hope is that in the future if someone wanted to add an additional span attribute for an API Gateway Proxy they could simply add it to `APIGatewayProxyWrapper.set_pre_execution_span_attributes` without risk of breaking any of the other event handlers.

## Type of change

    - Current documentation doesn't cover `event_context_extractor` that has changed.

# How Has This Been Tested?

- Sample events have been added for all types of events that the instrumentation supports.
- Unit tests have also been aded for new/modified utility functions.

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-aws-lambda/tests/test_aws_lambda_instrumentation_manual.py
```diff
@@ -264,6 +279,7 @@ def custom_event_context_extractor(lambda_event):
             ),
         ]
         for test in tests:
+            print(f"Executing test: {test.name}")
```
xrmx, line 282: please use self.subTest https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

xrmx:
@jbfenton need a changelog entry

instrumentation/opentelemetry-instrumentation-aws-lambda/tests/test_aws_lambda_instrumentation_manual.py
jbfenton: @xrmx good call, the `print` statement was originally added because it was difficult to nail down exactly which sub-test was failing. I've incorporated your feedback and now each sub-test is wrapped in a `subTest` context manager.

jbfenton:
> @jbfenton need a changelog entry

Changelog updated in 393dcf2d8bcebbdeb7c1ec27349fdb5fbf1bd923

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -431,6 +844,89 @@ def _instrumented_lambda_handler_call(  # noqa pylint: disable=too-many-branches
...
+    Determine the context extractor to use for the Lambda instrumentation.
+    The context extract should accept an EventWrapper and return an OTel Context.
+
+    Args:
+        context_extractor_override: Optional context extractor to use.
+
+    Returns: A context extractor function.
+    """
+
+    if context_extractor_override and isinstance(
```
Charlie-lizhihan, line 922: Suggested Change for Improved Readability and Simplicity
```
if callable(context_extractor_override):
    return context_extractor_override
```

jbfenton: Good call, change pushed in 3ed3b1cabab25d3c3c3e5865290c842a8bf9b0f6

xrmx (commented):
Took a look at this today, it's great to have the tests with proper mock responses but the amount of code that has been added was intimidating :sweat_smile: 

Would it possible to first add the tests and add whatever minimal change is eventually needed to have them pass and do the refactoring later?

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -110,223 +110,694 @@ def custom_event_context_extractor(lambda_event):
...
+        tracer_provider = tracer_provider_override
 
-        if xray_env_var:
-            parent_context = AwsXRayPropagator().extract(
-                {TRACE_HEADER_KEY: xray_env_var}
-            )
+    if not hasattr(tracer_provider, "force_flush"):
+        tracer_provider.force_flush = provider_warning(
+            provider_type="TracerProvider"
+        )
```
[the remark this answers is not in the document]
jbfenton: Functionally the Flushable Providers don't serve any purpose other than fixing up some type-hinting issues. 
I'll have a look back through and see if it can be optimised.

jbfenton:
> Take a look at this today, it's great to have the tests with proper mock responses but the amount of code that has been added was intimidating 😅
> 
> Would it possible to first add the tests and add whatever minimal change is eventually needed to have them pass and do the refactoring later?

Thanks for taking a look!

I'm happy to split this across a couple of PRs, I'll transition this PR back to draft, and spin up a new PR to introduce tests first.

PR #2590, opened 2024-06-08
consistently use of suppress_instrumentation utils

### LINKED ISSUE #2543: `suppress_instrumentation` utils are not used consistently
Searching for `suppress_instrumentation` (https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-python-contrib%20suppress_instrumentation&type=code) reveals places using the old plain string `"suppress_instrumentation"` context key, e.g:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/f4f3042f85fa68565c4da3f96d437b15d78e1063/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py#L350

Since this is not how instrumentation is actually suppressed any more, this can lead to bugs.

There are also places checking/setting the key 'manually' instead of via the utility functions, which also seems error-prone.

Additionally, if it's allowed for a user to have different versions of `opentelemetry-instrumentation` and `opentelemetry-instrumentation-httpx` (or another specific library instrumentation package) installed, then the util functions `suppress_instrumentation` and `is_instrumentation_enabled` should also check the plain string key  `"suppress_instrumentation"` for backward compatibility.

emdneto:
# Description

According to this [search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-python-contrib%20suppress_instrumentation&type=code), some instrumentations aren't using the `suppression_instrumentation` util properly. Found these ones:
* opentelemetry-instrumentation-httpx
* opentelemetry-instrumentation-aiohttp-serve

Please let me know if I missed something. 

Fixes #2543

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- `opentelemetry-instrumentation-aiohttp-server`, `opentelemetry-instrumentation-httpx`, `opentelemetry-resource-detector-azure` Ensure consistently use of suppress_instrumentation utils; adapt `opentelemetry.instrumentation.utils` util functions `suppress_instrumentation` and `is_instrumentation_enabled` to also check the old plain string key "suppress_instrumentation" for backward compatibility
```
xrmx, line 10: azure detector has its own changelog, you should update that too and just list instrumentation bumps there

emdneto: Missed this. Thanks

opentelemetry-instrumentation/tests/test_utils.py
```diff
@@ -186,3 +196,47 @@ def test_add_sql_comments_without_comments(self):
...
+    def test_suppress_http_instrumentation(self):
+        with suppress_http_instrumentation():
+            self.assertFalse(is_http_instrumentation_enabled())
+            self.assertTrue(is_instrumentation_enabled())
+
+        self.assertTrue(is_instrumentation_enabled())
+        self.assertTrue(is_http_instrumentation_enabled())
+
+    def test_suppress_instrumentation_key(self):
+        self.assertIsNone(get_value(_SUPPRESS_INSTRUMENTATION_KEY))
```
emdneto, line 221: maybe mock it to have a static value?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -161,9 +165,10 @@ def _python_path_without_directory(python_path, directory, path_separator):
 
 
 def is_instrumentation_enabled() -> bool:
```
emdneto, line 167: Add comment with the reason why checking both keys

xrmx: That could also be a question: should we support the old key?

emdneto: I think we shouldn't. Does anybody have thoughts about this?

lzchen: This [comment](https://github.com/open-telemetry/opentelemetry-python/issues/1737#issuecomment-815557962) explains the history of why we use a unique object instead of just the string. In terms of functionality, it would be nice to support plain text as well since it allows for easy use of this functionality from outside of the sdk (libraries won't have to explicitly reference `_SUPPRESS_HTTP_INSTRUMENTATION_KEY`) but it would defeat the purpose of having this context variable represented as a key. Might be a good discussion topic to bring up in the SIG

emdneto: @lzchen It's good to know the motivation behind that. I'll bring up this topic in the next SIG call.

emdneto: Discussed at SIG on 7/13 and we agree that we'll move forward supporting both old and new context keys.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -37,6 +37,10 @@
 
 propagator = TraceContextTextMapPropagator()
 
+_SUPPRESS_INSTRUMENTATION_KEY_PLAIN = (
```
aabmass, line 40: I wonder if this can be cleaned up now? https://github.com/open-telemetry/opentelemetry-python-contrib/blob/7249cc07301309748de26a4b1b941fad7f8d1551/opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py

emdneto: I'm not sure if I got it. do you mean remove L40?

aabmass: Sorry I put the wrong link! I meant here https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/context/__init__.py#L159-L165

emdneto: I found this issue https://github.com/open-telemetry/opentelemetry-python/issues/2184, which explains the motivation behind that. I think we don't have a clear decision on where this should be moved,  but I'm ok with keeping things as they are

lzchen: I think we still want to support the actual key object in case others have already taken a dependency

lzchen (approved):

PR #2597, opened 2024-06-11
add OTEL_ENABLE_SERVER_ATTRIBUTES_FOR_REQUEST_DURATION flag 

zeitlinger:
# Description

add `OTEL_ENABLE_SERVER_ATTRIBUTES_FOR_REQUEST_DURATION` flag to add `server.address` and `server.port` for `http.server.request.duration` - which are [opt-in](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration) 

This env var does not exist for any other language so far.

Also fixes the problem that `http.route` was not added to `http.server.request.duration` for flask - can be extracted to separate PR if required.

## Type of change

# How Has This Been Tested?

tests included in this PR

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
+++ b/instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
@@ -24,6 +24,7 @@ from opentelemetry.instrumentation._semconv import (
     OTEL_SEMCONV_STABILITY_OPT_IN,
     _OpenTelemetrySemanticConventionStability,
     _server_active_requests_count_attrs_new,
+    _server_duration_attrs_new_with_server_attributes,
     _server_active_requests_count_attrs_old,
     _server_duration_attrs_new,
     _server_duration_attrs_old,
@@ -99,7 +100,7 @@ _recommended_metrics_attrs_old = {
 }
 _recommended_metrics_attrs_new = {
     "http.server.active_requests": _server_active_requests_count_attrs_new,
-    "http.server.request.duration": _server_duration_attrs_new,
+    "http.server.request.duration": _server_duration_attrs_new_with_server_attributes,
 }
 _server_active_requests_count_attrs_both = (
     _server_active_requests_count_attrs_old
@@ -132,6 +133,7 @@ class TestProgrammatic(InstrumentationTest, WsgiTestBase):
             "os.environ",
             {
                 "OTEL_PYTHON_FLASK_EXCLUDED_URLS": "http://localhost/env_excluded_arg/123,env_excluded_noarg",
+                "OTEL_ENABLE_SERVER_ATTRIBUTES_FOR_REQUEST_DURATION": "true",
                 OTEL_SEMCONV_STABILITY_OPT_IN: sem_conv_mode,
             },
         )
@@ -519,6 +521,9 @@ class TestProgrammatic(InstrumentationTest, WsgiTestBase):
                             histogram_data_point_seen = True
                         if isinstance(point, NumberDataPoint):
                             number_data_point_seen = True
+                        for attr in _recommended_metrics_attrs_new[metric.name]:
+                            if attr != _SPAN_ATTRIBUTES_ERROR_TYPE:
+                                self.assertIn(attr, point.attributes)
                         for attr in point.attributes:
                             self.assertIn(
                                 attr,
@@ -596,9 +601,12 @@ class TestProgrammatic(InstrumentationTest, WsgiTestBase):
         self.client.get("/hello/756")
         expected_duration_attributes = {
             "http.request.method": "GET",
+            "http.route": "/hello/<int:helloid>",
             "url.scheme": "http",
             "network.protocol.version": "1.1",
             "http.response.status_code": 200,
+            "server.address": "localhost",
+            "server.port": 80,
         }
         expected_requests_count_attributes = {
             "http.request.method": "GET",
@@ -642,6 +650,8 @@ class TestProgrammatic(InstrumentationTest, WsgiTestBase):
             "url.scheme": "http",
             "network.protocol.version": "1.1",
             "http.response.status_code": 405,
+            "server.address": "localhost",
+            "server.port": 80,
         }
         expected_requests_count_attributes = {
             "http.request.method": "_OTHER",
@@ -667,6 +677,8 @@ class TestProgrammatic(InstrumentationTest, WsgiTestBase):
             "url.scheme": "http",
             "network.protocol.version": "1.1",
             "http.response.status_code": 405,
+            "server.address": "localhost",
+            "server.port": 80,
         }
         expected_requests_count_attributes = {
             "http.request.method": "NONSTANDARD",
--- a/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
@@ -208,6 +208,7 @@ API
 """
 
 import functools
+import os
 import typing
 import wsgiref.util as wsgiref_util
 from timeit import default_timer
@@ -225,6 +226,7 @@ from opentelemetry.instrumentation._semconv import (
     _report_new,
     _report_old,
     _server_active_requests_count_attrs_new,
+    _server_duration_attrs_new_with_server_attributes,
     _server_active_requests_count_attrs_old,
     _server_duration_attrs_new,
     _server_duration_attrs_old,
@@ -463,7 +465,9 @@ def _parse_duration_attrs(
     return _filter_semconv_duration_attrs(
         req_attrs,
         _server_duration_attrs_old,
-        _server_duration_attrs_new,
+        _server_duration_attrs_new_with_server_attributes
+        if os.environ.get("OTEL_ENABLE_SERVER_ATTRIBUTES_FOR_REQUEST_DURATION")
+        else _server_duration_attrs_new,
         sem_conv_opt_in_mode,
     )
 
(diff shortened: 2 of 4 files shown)


instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -463,7 +465,9 @@ def _parse_duration_attrs(
     return _filter_semconv_duration_attrs(
         req_attrs,
         _server_duration_attrs_old,
-        _server_duration_attrs_new,
+        _server_duration_attrs_new_with_server_attributes
+        if os.environ.get("OTEL_ENABLE_SERVER_ATTRIBUTES_FOR_REQUEST_DURATION")
```
xrmx, line 469: souldn't we use a python specific env var if it's not specified?

zeitlinger: good point - changed

zeitlinger:
@xrmx can you check again?

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -424,8 +422,8 @@ def process_response(self, request, response):
         if request_start_time is not None:
             duration_s = default_timer() - request_start_time
             if self._duration_histogram_old:
-                duration_attrs_old = _parse_duration_attrs(
-                    duration_attrs, _HTTPStabilityMode.DEFAULT
+                duration_attrs_old = _filter_semconv_server_duration_attrs(
+                    duration_attrs,
```
emdneto, line 426: I would like to explicitly have the `_HTTPStabilityMode.DEFAULT`. The migration to a new semconv can cause misunderstandings sometimes and this helps to better follow the code while reading

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
```diff
@@ -104,8 +106,8 @@ def expected_attributes_new(override_attributes):
     "http.server.duration": _server_duration_attrs_old_copy,
 }
 _recommended_metrics_attrs_new = {
-    "http.server.active_requests": _server_active_requests_count_attrs_new,
-    "http.server.request.duration": _server_duration_attrs_new_copy,
+    "http.server.active_requests": _server_active_requests_count_attrs_new_with_server_attributes,
+    "http.server.request.duration": _server_duration_attrs_new_with_server_attributes,
```
emdneto, line 110: Are we missing reflecting this change for the tests needed for `both_semconv`?

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
```diff
@@ -138,6 +140,8 @@ def setUp(self):
             "os.environ",
             {
                 "OTEL_PYTHON_FLASK_EXCLUDED_URLS": "http://localhost/env_excluded_arg/123,env_excluded_noarg",
+                "OTEL_PYTHON_HTTP_SERVER_REQUEST_DURATION_SERVER_ATTRIBUTES_ENABLED": "true",
```
emdneto, line 143: I think it's better to create a new bunch of tests for that, wdyt? For default, new_semconv and both_semconv

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -90,6 +90,17 @@
     URL_SCHEME,
 ]
 
+_server_duration_attrs_new_with_server_attributes = [
```
emdneto, line 93: can we just extend this to support the opt-in attributes?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -154,6 +171,23 @@ def _initialize(cls):
                 _OpenTelemetrySemanticConventionStability._OTEL_SEMCONV_STABILITY_SIGNAL_MAPPING[
                     _OpenTelemetryStabilitySignalType.HTTP
                 ] = http_opt_in
+
+                _OpenTelemetrySemanticConventionStability.server_duration_attrs_new_effective = (
+                    _server_duration_attrs_new_with_server_attributes
+                    if os.environ.get(
```
emdneto, line 177: We don't need to run this for all instrumentation migrated to new semconv, right? because not every instrumentation has metrics implemented. Wdyt we have a way to explicitly initialize this per instrumentation we want to support the opt-in attributes?

emdneto (commented):
Thanks for working on this one. Left some comments

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
zeitlinger: I don't understand

zeitlinger: ok, I reverted this change after applying the suggestion below

zeitlinger:
@emdneto please have another look :smile:

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-instrumentation-kafka-python` Instrument temporary fork, kafka-python-ng
   inside kafka-python's instrumentation
   ([#2537](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2537))
+- Add `OTEL_PYTHON_HTTP_SERVER_REQUEST_DURATION_SERVER_ATTRIBUTES_ENABLED` flag to add
```
lzchen, line 17: Please use a similar format as the other CHANGELOG entries (e.g. `([#2537](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2537))`)

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -101,7 +109,13 @@
 _server_active_requests_count_attrs_new = [
     HTTP_REQUEST_METHOD,
     URL_SCHEME,
-    # TODO: Support SERVER_ADDRESS AND SERVER_PORT
+]
+
+_server_active_requests_count_attrs_new_with_server_attributes = [
+    HTTP_REQUEST_METHOD,
```
lzchen, line 115: Nit: `_server_active_requests_count_attrs_new ` and append the two new attributes.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
lzchen: +1 to supporting this individually per instrumentation. `_OpenTelemetrySemanticConventionStability` should only be used for semantic convention migration. We do not want to mix this functionality with semantic conventions.

zeitlinger: is this going in the right direction?

https://github.com/zeitlinger/opentelemetry-python-contrib/compare/http-metrics-server-arributes...zeitlinger:opentelemetry-python-contrib:http-metrics-server-arributes-asgi-trial?expand=1

lzchen: @zeitlinger 

Yeah you mostly like don't need to store the config as part of the global `_OpenTelemetrySemanticConventionStability `. I think you can store the config information in each instrumentation instance themselves when creating them. You can then have a static set of attributes with the one without the flag enabled and one with the flag enabled.

emdneto: I'm sympathetic to the @lzchen  approach. With that we can be more selective in which instrumentation we want to support to opt-in attributes

zeitlinger: ok - changed now :smile:

zeitlinger:
feedback from SIG meeting

- Should not add new environment variables - opens up too many possibilities (and file config will take care of that)
- But it should be possible to do programmatically (or a PR to allow this would be welcome)

xrmx:
Maybe I am late but would adding a single environment variable that takes a list of additional attributes to export would be more acceptable? The caveat would be to requires specifying the accepted literals with types so at least it's clear what the code supports.
I understand the concern for avoiding too many environment variables but I'm sympathetic to the auto-instrumentation cause :sweat_smile:

emdneto:
@xrmx That's the suggestion I made at SIG yesterday. I don't know if my explanation was very clear, but we discussed and agreed to do that per instrumentation using the kwargs at the instrument level.

lzchen:
@xrmx 

I'm open to having a separate discussion about what efforts we should make for auto instrumentation to not block this PR. Seems like we have enough interest for autoinstrumentation that we should definitely do an audit of it.

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
```diff
@@ -615,6 +620,35 @@ def test_basic_metric_success_new_semconv(self):
...
         )
 
+    @patch.dict(
+        "os.environ",
+        {
+            OTEL_PYTHON_HTTP_SERVER_REQUEST_DURATION_SERVER_ATTRIBUTES_ENABLED: "1",
+            OTEL_PYTHON_HTTP_SERVER_ACTIVE_REQUESTS_COUNT_SERVER_ATTRIBUTES_ENABLED: "1",
+        },
+    )
+    def test_basic_metric_success_new_semconv_server_address(self):
```
lzchen, line 630: Would you be able to add tests for django, asgi and wsgi as well?

lzchen (approved):

lzchen (changes requested):
Blocking until we remove env var mechanism.

PR #2611, opened 2024-06-14
Use generated symbols from semantic conventions package

lzchen:
Also will use stable attributes when possible.


instrumentation/opentelemetry-instrumentation-flask/tests/base_test.py
```diff
@@ -85,7 +85,6 @@ def _custom_response_headers():
         resp.headers["My-Custom-Regex-Header-2"] = (
             "my-custom-regex-value-3,my-custom-regex-value-4"
         )
-        resp.headers["my-secret-header"] = "my-secret-value"
```
xrmx, line 88: this removal makes test fail

lzchen: Oops good catch.

xrmx (approved):

PR #2106, opened 2024-01-02
Update wsgi __init__.py to include HTTP_TARGET.  

philipcwhite:
Add HTTP_TARGET to server duration and active requests

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # Currently metrics exclude the HTTP_TARGET.  This does not allow use to get specific resource counts or durations for visited endpoints.  By adding HTTP_TARGET, we can now get detailed stats for individual page access.  This will create additional metrics and increase cardinality.  I feel like the granularity is needed to make the metric useful.  It also brings this metric in-line with the spanmetrics connector. 

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ x ] Test A

Changed package and ran locally using auto-instrumentation->local collector with Prometheus exporter.  Adding variables allowed prometheus to create correct metrics with targets.

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [ ] Followed the style guidelines of this project
- [ ] Changelogs have been updated
- [ ] Unit tests have been added
- [ ] Documentation has been updated

Issue #1457, opened 2022-11-18
Include http.target in flask instrumentation metrics attributes


ElementalWarrior:
As part of the semantic conventions, `http.target` should be included in the metrics attributes. However it is only present for span attributes.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#parameterized-attributes

Currently it isn't part of the `duration_attrs`. Which it probably shouldn't until there is a solution to set the value to the flask url_rule instead of the raw input request:
- caller: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py#L295
- callee: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py#L236

martinrw:
+1 We have the same requirement but Falcon rather than Flask

sakthiraam:
+1 We have the same requirement for both Falcon and Flask.

Hi @ElementalWarrior , Good Day! I see you mentioned "Which it probably shouldn't until there is a solution to set the value to the flask url_rule instead of the raw input request".

Currently we are using the latest version of contrib opentelemetry-distro==0.41b0 in which we see the traces are already having the http.route and http.target labels

For Example, In traces it auto replaces like below

Static Resources

http.route -- /script/js/min.js to /script/<path:filename>

http.target -- /script/js/min.js to /script/js/min.js

Server Side resources

http.route -/about/xyz

http.target - /about/xyz.

I suppose the http.route will auto truncate any query params and keep only the route. With that is it ok to add http.route to the duration attributes ?

alexmojaki:
https://github.com/open-telemetry/opentelemetry-specification/pull/2818 changed the metric attribute name from `http.target` to `http.route`. The FastAPI instrumentation uses `http.target`. Questions:

alexmojaki:
https://github.com/open-telemetry/opentelemetry-specification/pull/2818 changed the metric attribute name from `http.target` to `http.route`. The FastAPI instrumentation uses `http.target`. What's the recommended path forward?

1. Use `http.target` everywhere
2. Use `http.route` everywhere, except if `http.target` is already being used
[...]

PR #2621, opened 2024-06-19
add metric attribute

GonzaloGuasch:
# Description

Taking up [this closed PR](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2106), when sending metrics in the Flask instrumentation, the http.route attribute is added to them.
This attribute is super important in the use of metrics and will greatly improve the experience of using Flask. This change was tested locally.

## Type of change

# How Has This Been Tested?
 running: tox -e  test-instrumentation-flask

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `opentelemetry-instrumentation-flask` Add http route to metric attributes
+  ([#2506](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2506))
 - `opentelemetry-sdk-extension-aws` Register AWS resource detectors under the
   `opentelemetry_resource_detector` entry point
   ([#2382](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2382))
--- a/instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
@@ -251,7 +251,6 @@ from packaging import version as package_version
 import opentelemetry.instrumentation.wsgi as otel_wsgi
 from opentelemetry import context, trace
 from opentelemetry.instrumentation._semconv import (
-    _METRIC_ATTRIBUTES_SERVER_DURATION_NAME,
     _get_schema_url,
     _HTTPStabilityMode,
     _OpenTelemetrySemanticConventionStability,
@@ -268,6 +267,9 @@ from opentelemetry.instrumentation.propagators import (
 from opentelemetry.instrumentation.utils import _start_internal_or_server_span
 from opentelemetry.metrics import get_meter
 from opentelemetry.semconv.metrics import MetricInstruments
+from opentelemetry.semconv.metrics.http_metrics import (
+    HTTP_SERVER_REQUEST_DURATION,
+)
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.util.http import (
     get_excluded_urls,
@@ -282,6 +284,7 @@ _ENVIRON_SPAN_KEY = "opentelemetry-flask.span_key"
 _ENVIRON_ACTIVATION_KEY = "opentelemetry-flask.activation_key"
 _ENVIRON_REQCTX_REF_KEY = "opentelemetry-flask.reqctx_ref_key"
 _ENVIRON_TOKEN = "opentelemetry-flask.token"
+_ENVIRON_REQUEST_ROUTE_KEY = "request-route_key"
 
 _excluded_urls_from_env = get_excluded_urls("FLASK")
 
@@ -344,6 +347,11 @@ def _rewrapped_app(
                 excluded_urls is None
                 or not excluded_urls.url_disabled(flask.request.url)
             ):
+                if flask.request.url_rule:
+                    wrapped_app_environ[_ENVIRON_REQUEST_ROUTE_KEY] = str(
+                        flask.request.url_rule
+                    )
+
                 span = flask.request.environ.get(_ENVIRON_SPAN_KEY)
 
                 propagator = get_global_response_propagator()
@@ -386,6 +394,12 @@ def _rewrapped_app(
             duration_attrs_old = otel_wsgi._parse_duration_attrs(
                 attributes, _HTTPStabilityMode.DEFAULT
             )
+
+            if wrapped_app_environ.get(_ENVIRON_REQUEST_ROUTE_KEY, None):
+                duration_attrs_old[SpanAttributes.HTTP_ROUTE] = (
+                    wrapped_app_environ.get(_ENVIRON_REQUEST_ROUTE_KEY)
+                )
+
             duration_histogram_old.record(
                 max(round(duration_s * 1000), 0), duration_attrs_old
             )
@@ -393,6 +407,12 @@ def _rewrapped_app(
             duration_attrs_new = otel_wsgi._parse_duration_attrs(
                 attributes, _HTTPStabilityMode.HTTP
             )
+
+            if wrapped_app_environ.get(_ENVIRON_REQUEST_ROUTE_KEY, None):
+                duration_attrs_new[SpanAttributes.HTTP_ROUTE] = (
+                    wrapped_app_environ.get(_ENVIRON_REQUEST_ROUTE_KEY)
+                )
+
             duration_histogram_new.record(
                 max(duration_s, 0), duration_attrs_new
             )
@@ -553,7 +573,7 @@ class _InstrumentedFlask(flask.Flask):
         duration_histogram_new = None
         if _report_new(_InstrumentedFlask._sem_conv_opt_in_mode):
             duration_histogram_new = meter.create_histogram(
-                name=_METRIC_ATTRIBUTES_SERVER_DURATION_NAME,
+                name=HTTP_SERVER_REQUEST_DURATION,
                 unit="s",
                 description="measures the duration of the inbound HTTP request",
             )
@@ -684,7 +704,7 @@ class FlaskInstrumentor(BaseInstrumentor):
             duration_histogram_new = None
             if _report_new(sem_conv_opt_in_mode):
                 duration_histogram_new = meter.create_histogram(
-                    name=_METRIC_ATTRIBUTES_SERVER_DURATION_NAME,
+                    name=HTTP_SERVER_REQUEST_DURATION,
                     unit="s",
                     description="measures the duration of the inbound HTTP request",
                 )
--- a/instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
+++ b/instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
@@ -20,7 +20,6 @@ from flask import Flask, request
 
 from opentelemetry import trace
 from opentelemetry.instrumentation._semconv import (
-    _SPAN_ATTRIBUTES_ERROR_TYPE,
     OTEL_SEMCONV_STABILITY_OPT_IN,
     _OpenTelemetrySemanticConventionStability,
     _server_active_requests_count_attrs_new,
@@ -40,6 +39,7 @@ from opentelemetry.sdk.metrics.export import (
     NumberDataPoint,
 )
 from opentelemetry.sdk.resources import Resource
+from opentelemetry.semconv.attributes.error_attributes import ERROR_TYPE
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.test.wsgitestutil import WsgiTestBase
 from opentelemetry.util.http import (
@@ -379,7 +379,7 @@ class TestProgrammatic(InstrumentationTest, WsgiTestBase):
                 SpanAttributes.URL_PATH: "/hello/500",
                 SpanAttributes.HTTP_ROUTE: "/hello/<int:helloid>",
                 SpanAttributes.HTTP_RESPONSE_STATUS_CODE: 500,
-                _SPAN_ATTRIBUTES_ERROR_TYPE: "500",
+                ERROR_TYPE: "500",
                 SpanAttributes.URL_SCHEME: "http",
             }
         )
@@ -405,7 +405,7 @@ class TestProgrammatic(InstrumentationTest, WsgiTestBase):
                 {
                     SpanAttributes.URL_PATH: "/hello/500",
                     SpanAttributes.HTTP_RESPONSE_STATUS_CODE: 500,
-                    _SPAN_ATTRIBUTES_ERROR_TYPE: "500",
+                    ERROR_TYPE: "500",
                     SpanAttributes.URL_SCHEME: "http",
                 }
             )
@@ -459,6 +459,7 @@ class TestProgrammatic(InstrumentationTest, WsgiTestBase):
         self.assertEqual(len(span_list), 1)
 
     def test_flask_metrics(self):
+        _server_duration_attrs_old.append("http.route")
         start = default_timer()
         self.client.get("/hello/123")
         self.client.get("/hello/321")
@@ -570,6 +571,7 @@ class TestProgrammatic(InstrumentationTest, WsgiTestBase):
         self.client.get("/hello/756")
         expected_duration_attributes = {
             "http.method": "GET",
+            "http.route": "/hello/<int:helloid>",
             "http.host": "localhost",
             "http.scheme": "http",
             "http.flavor": "1.1",
@@ -597,6 +599,7 @@ class TestProgrammatic(InstrumentationTest, WsgiTestBase):
         expected_duration_attributes = {
             "http.request.method": "GET",
             "url.scheme": "http",
+            "http.route": "/hello/<int:helloid>",
             "network.protocol.version": "1.1",
             "http.response.status_code": 200,
         }


instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
```diff
@@ -459,6 +459,7 @@ def test_exclude_lists_from_explicit(self):
         self.assertEqual(len(span_list), 1)
 
     def test_flask_metrics(self):
+        _server_duration_attrs_old.append("http.route")
```
alexmojaki, line 462: IMO `_server_duration_attrs_old` itself should be updated at the source. If that's not an option, I suggest that `_recommended_metrics_attrs_old` and `_recommended_metrics_attrs_both` use some copy of `_server_duration_attrs_old` that includes `http.route`. Modifying this global state here could easily break other tests in confusing ways.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -284,6 +284,7 @@ def response_hook(span: Span, status: str, response_headers: List):
 _ENVIRON_ACTIVATION_KEY = "opentelemetry-flask.activation_key"
 _ENVIRON_REQCTX_REF_KEY = "opentelemetry-flask.reqctx_ref_key"
 _ENVIRON_TOKEN = "opentelemetry-flask.token"
+_ENVIRON_REQUEST_ROUTE_KEY = "request-route_key"
```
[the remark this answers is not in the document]
alexmojaki: Actually, is there any reason to use `wrapped_app_environ` rather than just a plain variable which is set in `_start_response` using `nonlocal`? Or maybe even just modifying `attributes` directly?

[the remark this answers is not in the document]
GonzaloGuasch: I use `wrapped_app_environ` becouse is used above in the method. Do you see this as a problem? In mi point of view it is not a problem or bad decision to used it!

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
GonzaloGuasch: I agree with you on changing the _server_duration_attrs_old at the source, but they are imported from the semconv package, and I cannot modify them. I committed the changes and made a copy for the test. Please let me know if this is what you intended

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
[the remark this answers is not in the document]
alexmojaki: I think the other cases need to use `wrapped_app_environ` because the value is accessed somewhere else. Here a local variable is an option and would be more efficient and probably simpler. But ultimately it's a small difference.

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
alexmojaki: > I agree with you on changing the _server_duration_attrs_old at the source, but they are imported from the semconv package, and I cannot modify them.

It's in this same repo, so I *think* you could change them in this PR. I've done something similar in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2624. But maybe that won't be accepted.

Your `_server_duration_attrs_old_copy` is an OK solution for the sake of this test, but the line being commented on here (`_server_duration_attrs_old.append("http.route")`) still needs to be removed.

GonzaloGuasch: Sorry, I may have misunderstood, but if I remove the line `_server_duration_attrs_old.append("http.route")`, the test where I assert to have the `http.route` as a metric attribute will fail (`test_basic_metric_success`, for example). I can also remove the `http.route` from the assertion list, and this will result in a successful run for all tests. Is that the solution you're looking for?

alexmojaki: `test_basic_metric_success` should pass if you run it by itself, it shouldn't depend on another test. Sounds like there's a bug in the implementation.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -388,13 +394,25 @@ def _start_response(status, response_headers, *args, **kwargs):
...
             duration_histogram_old.record(
                 max(round(duration_s * 1000), 0), duration_attrs_old
             )
         if duration_histogram_new:
             duration_attrs_new = otel_wsgi._parse_duration_attrs(
                 attributes, _HTTPStabilityMode.HTTP
             )
+
+            if wrapped_app_environ.get(_ENVIRON_REQUEST_ROUTE_KEY, None):
+                duration_attrs_new[SpanAttributes.HTTP_ROUTE] = (
```
emdneto, line 412: I think would be nice to use the new semconv symbol
from opentelemetry.semconv.attributes.http_attributes import HTTP_ROUTE
```suggestion
                duration_attrs_new[HTTP_ROUTE] = (
```

GonzaloGuasch: Done!

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
GonzaloGuasch: My mistake!! i was having troubles with the `tox` in local. Now its done :)

CHANGELOG.md
```diff
@@ -61,6 +61,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `opentelemetry-instrumentation-flask` Add `http.route` to metric attributes
```
xrmx, line 64: You should move this to the next version

GonzaloGuasch: Done! moved to unreleased section

xrmx: Please move it to the Added section of the unreleased chapter

GonzaloGuasch:
@ocelotl do you think this PR is ready to merge?

CHANGELOG.md
```diff
@@ -6,9 +6,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+  ([#2153](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2153))
+- `opentelemetry-instrumentation-asyncio` instrumented `asyncio.wait_for` properly raises `asyncio.TimeoutError` as expected
+  ([#2637](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2637))
+- `opentelemetry-instrumentation-aws-lambda` Bugfix: AWS Lambda event source key incorrect for SNS in instrumentation library. 
+  ([#2612](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2612))
+- `opentelemetry-instrumentation-system-metrics` Permit to use psutil 6.0+.
+  ([#2630](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2630))
 
 ### Added
-
```
emdneto, line 11: Keep the empty line here

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -388,13 +394,25 @@ def _start_response(status, response_headers, *args, **kwargs):
             duration_attrs_old = otel_wsgi._parse_duration_attrs(
                 attributes, _HTTPStabilityMode.DEFAULT
             )
+
+            if wrapped_app_environ.get(_ENVIRON_REQUEST_ROUTE_KEY, None):
```
lzchen, line 398: `HTTP_ROUTE` is not a defined attribute in the [old semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.11.0/specification/metrics/semantic_conventions/http-metrics.md#attributes) so this would just be a feature for the new sem conv.

alexmojaki: @lzchen sounds like you can answer https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1457#issuecomment-2179059109

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -388,13 +394,25 @@ def _start_response(status, response_headers, *args, **kwargs):
...
+                duration_attrs_old[HTTP_ROUTE] = wrapped_app_environ.get(
```
alexmojaki, line 399:
```suggestion
                duration_attrs_old[HTTP_TARGET] = wrapped_app_environ.get(
```

to resolve the comment above. Of course this will also require updating imports.

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
```diff
@@ -570,6 +573,7 @@ def test_basic_metric_success(self):
         self.client.get("/hello/756")
         expected_duration_attributes = {
             "http.method": "GET",
+            "http.route": "/hello/<int:helloid>",
```
emdneto, line 576: You can drop this too, since the test if for old semconv.

alexmojaki: I think rather than dropping it should be updated:

```suggestion
            "http.target": "/hello/<int:helloid>",
```

emdneto: I'm also okay with that; I would rename the PR then and the changelog entry to make it explicit that we are adding those two attributes instead of only `http.route`

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
emdneto: I would use the `spanAttributes.HTTP_TARGET` instead of the incubating symbol

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
GonzaloGuasch: Just to understand, the idea is to add `http.target` and `http.route` in this PR?

alexmojaki: Yes. It's the same attribute but with different names depending on the schema version. See https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1457#issuecomment-2229266571 and after.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -388,13 +397,31 @@ def _start_response(status, response_headers, *args, **kwargs):
...
                 attributes, _HTTPStabilityMode.HTTP
             )
+
+            if wrapped_app_environ.get(_ENVIRON_REQUEST_ROUTE_KEY, None):
+                duration_attrs_new[HTTP_ROUTE] = wrapped_app_environ.get(
+                    _ENVIRON_REQUEST_ROUTE_KEY
+                )
+                duration_attrs_new[HTTP_TARGET] = wrapped_app_environ.get(
+                    _ENVIRON_REQUEST_ROUTE_KEY
+                )
```
alexmojaki, line 423:
```suggestion
            if wrapped_app_environ.get(_ENVIRON_REQUEST_ROUTE_KEY, None):
                duration_attrs_old[HTTP_TARGET] = wrapped_app_environ.get(
                    _ENVIRON_REQUEST_ROUTE_KEY
                )

            duration_histogram_old.record(
                max(round(duration_s * 1000), 0), duration_attrs_old
            )
        if duration_histogram_new:
            duration_attrs_new = otel_wsgi._parse_duration_attrs(
                attributes, _HTTPStabilityMode.HTTP
            )

            if wrapped_app_environ.get(_ENVIRON_REQUEST_ROUTE_KEY, None):
                duration_attrs_new[HTTP_ROUTE] = wrapped_app_environ.get(
                    _ENVIRON_REQUEST_ROUTE_KEY
                )
```

`target` for old, `route` for new, not both for both.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -388,13 +397,25 @@ def _start_response(status, response_headers, *args, **kwargs):
             duration_attrs_old = otel_wsgi._parse_duration_attrs(
                 attributes, _HTTPStabilityMode.DEFAULT
             )
+
+            if wrapped_app_environ.get(_ENVIRON_REQUEST_ROUTE_KEY, None):
+                duration_attrs_old[HTTP_TARGET] = wrapped_app_environ.get(
```
emdneto, line 402:
```suggestion
                duration_attrs_old[SpanAttributes.HTTP_TARGET] = wrapped_app_environ.get(
```

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
GonzaloGuasch: Done!

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
GonzaloGuasch: Done!

alexmojaki (approved):

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -425,7 +443,7 @@ def _before_request():
         if flask.request.url_rule:
             # For 404 that result from no route found, etc, we
             # don't have a url_rule.
-            attributes[SpanAttributes.HTTP_ROUTE] = flask.request.url_rule.rule
+            attributes[HTTP_ROUTE] = flask.request.url_rule.rule
```
emdneto, line 446:
```suggestion
            attributes[SpanAttributes.HTTP_ROUTE] = flask.request.url_rule.rule
```

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
emdneto: @GonzaloGuasch since we already did the semconv transition for flask, here I would keep 
`from opentelemetry.semconv.attributes.http_attributes import HTTP_ROUTE` symbol

GonzaloGuasch: Done!

GonzaloGuasch: with this changes do you see this PR ready to be merge? :)

emdneto (approved):

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -346,6 +348,11 @@ def _start_response(status, response_headers, *args, **kwargs):
                 excluded_urls is None
                 or not excluded_urls.url_disabled(flask.request.url)
             ):
+                if flask.request.url_rule:
+                    wrapped_app_environ[_ENVIRON_REQUEST_ROUTE_KEY] = str(
```
lzchen, line 352: Why do we need to store this value in the environ? It looks like it's only being used in `_start_response`?

GonzaloGuasch: I set the value in the `_start_response` where I can access the variable `flask.request.url_rule` to populate the` http.route` or `http.target`. Otherwise, if I try using this value directly without a variable, I get the error:
`RuntimeError: Working outside of request context.`
This was the motivation for using the variable. For more context, we already discussed this in the first comment of this PR.

alexmojaki: But you can use a normal Python variable with `nonlocal`, you don't need the environ. That's what I said in my first comment.

GonzaloGuasch: I change to nonlocal variable in mi last commit. Hope its okey :)! @alexmojaki @lzchen

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -340,12 +341,17 @@ def _wrapped_app(wrapped_app_environ, start_response):
...
+        request_route = None
 
         def _start_response(status, response_headers, *args, **kwargs):
             if flask.request and (
                 excluded_urls is None
                 or not excluded_urls.url_disabled(flask.request.url)
             ):
+                if flask.request.url_rule:
+                    nonlocal request_route
+                    request_route = str(flask.request.url_rule)
```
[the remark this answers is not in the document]
alexmojaki: The point was to get rid of the `if flask.request.url_rule:`

alexmojaki (approved):

xrmx (approved):

lzchen (approved):

PR #2624, opened 2024-06-19
Add `http.target` to Django duration metric attributes

alexmojaki:
# Description

This is similar to https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2621 which is for https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1457, but this PR is for Django rather than Flask.

It adds the `http.target` attribute to the metric named `http.server.duration`.

## Type of change

"New feature" feels right and I've chosen it to match https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2621 and its changelog update. Arguably this is a bug fix because the attribute is required by the spec, and technically it could be considered a breaking change because this will increase the number of metric data points emitted, but the attribute has low cardinality.

# How Has This Been Tested?

The existing test failed until I updated it to also expect the `http.target` attribute.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -290,26 +293,6 @@ def process_request(self, request):
...
-            and self._environ_span_key in request.META.keys()
-        ):
-            span = request.META[self._environ_span_key]
-
-            if span.is_recording():
-                match = getattr(request, "resolver_match", None)
-                if match:
-                    route = getattr(match, "route", None)
-                    if route:
-                        span.set_attribute(SpanAttributes.HTTP_ROUTE, route)
```
alexmojaki, line 311: This is the only thing this method does, so I deleted it since the route attribute is now set when the span is started.

I should note that getting the route here has a couple of advantages. It uses `request.resolver_match` which is already set, whereas `_get_span_name_and_route` (previously `_get_span_name`) uses `resolve(request.path)` (it tries `request.resolver_match` first, but AFAICT it's never set at that point). Using `request.resolver_match` is:

1. Faster, whereas `resolve(request.path)` duplicates a nontrivial computation that tests a bunch of regexes.
2. Possibly more correct, since `request.resolver_match` is set by Django itself. In fact it seems to be the result of `resolve(request.path_info)`, and `request.path_info` is not always the same as `request.path`. This could be fixed, but there may be other differences now or in the future.

The problem is that the span name needs to include the route. This could be solved by starting the span with a simpler name and using `Span.update_name` after getting the route here from `request.resolver_match`, but I saw a comment in the flask instrumentation saying `that API is "highly discouraged" so we better avoid it`. Is that still the case? It seems unfortunate.

xrmx: AFAICS the API or the SDK does not mention any deprecation or discouraging of using `update_name`.

Wondering how good the tests coverage of these discrepancies between resolve and resolver_match is :sweat:

alexmojaki: I searched and found this: https://github.com/open-telemetry/opentelemetry-rust/issues/725

It links to https://github.com/open-telemetry/opentelemetry-rust/blob/561426e4d5e04fff10f896c681c7501adbe95929/opentelemetry/src/trace/span.rs#L156 which says:

```
    /// Updates the `Span`'s name. After this update, any sampling behavior based on the
    /// name will depend on the implementation.
    ///
    /// It is highly discouraged to update the name of a `Span` after its creation.
    /// `Span` name is often used to group, filter and identify the logical groups of
    /// spans. Often, filtering logic will be implemented before the `Span` creation
    /// for performance reasons, and the name update may interfere with this logic.
    ///
    /// The method name is called `update_name` to differentiate this method from the
    /// regular property. It emphasizes that this operation signifies a
    /// major change for a `Span` and may lead to re-calculation of sampling or
    /// filtering decisions made previously depending on the implementation.
```

But a comment in that issue says:

> We should probably update the doc to be in line with the current version of specs.

And now the latest commit (https://github.com/open-telemetry/opentelemetry-rust/blob/e0fb7fe3b7ab343cdf2f702b04d04976dc5542ca/opentelemetry/src/trace/span.rs) only says:

```
    /// Updates the span's name.
    ///
    /// After this update, any sampling behavior based on the name will depend on
    /// the implementation.
```

So maybe it's less discouraged now?

If there are samplers extracting the route from the span name and using that to make decisions, removing the route from the initial span name will be a breaking change for them. But it should usually still be possible to make the necessary sampling decision by looking at the `http.url` attribute. You could even extract the path from the URL and then `resolve` the route from the path. The caveat is that this is similar to `resolve(request.path)` - I don't think `request.path_info` can be determined from `http.url`. If Django set `http.target` (it currently doesn't at all) to `request.path_info` then that would be solved. I don't know if that would be following the OTEL spec.

I don't know what 'filtering logic' the original comment is talking about that could apply here outside of sampling.

alexmojaki:
> The way I've done this may also affect other instrumentations that use `_duration_attrs` in `util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py` if those instrumentations set `http.route` in the right place.

I've just found that the FastAPI tests fail because of this: the metrics now have both `http.route` and `http.target` with the same value. What should be done about this? I listed some options in https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1457#issuecomment-2179059109

alexmojaki:
`http.route` is also appearing in the starlette metrics, which is causing those tests to fail, but that just means that this fix applies to both starlette and django.

alexmojaki:
> > The way I've done this may also affect other instrumentations that use `_duration_attrs` in `util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py` if those instrumentations set `http.route` in the right place.
> 
> I've just found that the FastAPI tests fail because of this: the metrics now have both `http.route` and `http.target` with the same value. What should be done about this? I listed some options in [#1457 (comment)](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1457#issuecomment-2179059109)

Would it be helpful for me to just pick an option and push that, so that (1) the tests pass and (2) the effect is easily visible in the git diff?

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -290,26 +293,6 @@ def process_request(self, request):
                 span, request
             )
 
-    # pylint: disable=unused-argument
-    def process_view(self, request, view_func, *args, **kwargs):
```
aabmass, line 294: Just to be clear, this runs after `process_request()`, right?

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
aabmass: I think it would be pretty nice for samplers to have the route available. [The spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#updatename) is helpful here

> Alternatives for the name update may be late Span creation, when Span is started with the explicit timestamp from the past at the moment where the final Span name is known

It looks like we are already recording start time in (several?) variables in `process_request()`. Would this backdating approach work?

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
xrmx: Yes

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
alexmojaki: That would mean that the span only gets started after all the `process_request` of all other middleware has already been called. Since other middleware is quite likely to start spans (e.g. for an instrumented DB request) that should presumably be children of the Django span, I don't think backdating will work.

alexmojaki: In https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions I see:

> `http.route` MUST be provided at span creation time if and only if it's already available. If it becomes available after span starts, instrumentation MUST populate it anytime before span ends.

But I'm not sure if that route counts as 'available' at this stage.

alexmojaki: > If Django set http.target (it currently doesn't at all) to request.path_info then that would be solved. I don't know if that would be following the OTEL spec.

It looks like this wouldn't be following the spec. On `http.route` at least is a note saying "SHOULD include the [application root](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/http.md#http-server-definitions) if there is one." This actually implies that the current `http.route` in Django is incomplete since I think it's missing the application root if there is one. And `http.target` is described as "The full request target as passed in a HTTP request line or equivalent" which also sounds like it includes the route.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -236,6 +236,7 @@ def test_basic_metric_success(self):
             "net.host.port": 443,
             "http.status_code": 200,
             "http.target": "/foobar",
+            "http.route": "/foobar",
```
alexmojaki, line 239: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2624#issuecomment-2200646062

Note that here `http.target` and `http.route` are equal.

alexmojaki:
This should be mergeable now. Only thing I'm unsure about is the changelog entry. This doesn't need a documentation update, right?

emdneto (approved):

lzchen (approved):
Thanks for the change @alexmojaki ! I know the semantic convention stuff was confusing haha.

Issue #2453, opened 2024-04-23
Implement migration plan for selected instrumentations


lzchen:
A followup of https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2351

Tracking issue to track progress of semantic convention stability migration for all instrumentations

Finished:

opentelemetry-instrumentation-requests https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2002
opentelemetry-instrumentation-wsgi https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2425

Remaining: 

opentelemetry-instrumentation-aiohttp-client
opentelemetry-instrumentation-asgi
opentelemetry-instrumentation-django
opentelemetry-instrumentation-fastapi
opentelemetry-instrumentation-httplib
opentelemetry-instrumentation-httpx
opentelemetry-instrumentation-flask
opentelemetry-instrumentation-psycopg2
opentelemetry-instrumentation-sqlalchemy
opentelemetry-instrumentation-requests
opentelemetry-instrumentation-urllib
opentelemetry-instrumentation-urllib3

Charlie-lizhihan:
Hey, @lzchen

I'm a beginner contributor, just fixed some bugs and would love to contribute more to the community. After discussing with @emdneto, we believe maybe I could try to contribute this one.

I would like to contribute to the opentelemetry-instrumentation-httplib. I read through the previous issues #2351 and #791, and read through your previous PRs: #2002, #2425, and #2454. They are really complex for me, so I think the steps to update httplib are like:

1. `__init__.py` under util/opentelemetry-util-http/src/opentelemetry/util/http:
  Introduce Environment Variable: Add logic to check OTEL_SEMCONV_STABILITY_OPT_IN.
  Update Attributes: Replace old attributes with new semantic conventions.
  Conditional Logic: Implement conditional logic to emit old and new attributes based on the environment variable.

2. Instrumentation File (httplib.py):
  Introduce Environment Variable: Similar to __init__.py.
  Update Span Attributes: Use new semantic conventions.
  Conditional Logic: Add logic to conditionally emit attributes.

3. Test Files:
  Update Tests: Add tests to ensure both old and new conventions are correctly handled based on the environment variable.


I think my next step is to completely understand the httplib instrumentation, as I'm not really familiar with this one yet. May I ask if my steps are correct? Any other resources that could be helpful during my development?

Thank you!

lzchen:
[...]

PR #2631, opened 2024-06-24
HTTP semantic convention stability migration for httpx

emdneto:
# Description
Part of this https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2453 and based on https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2002

## Type of change

# How Has This Been Tested?
Sample below
```python
import os
import httpx
import asyncio
import logging

from opentelemetry.sdk.trace.export import (
    BatchSpanProcessor,
    ConsoleSpanExporter,
)
# Specify "http", "http/dup" or ""
os.environ["OTEL_SEMCONV_STABILITY_OPT_IN"] = "http/dup"
logging.basicConfig(level=logging.DEBUG)

provider = TracerProvider()
tracer = trace.get_tracer("my.tracer.name")

def sync_request():
        response = client.get("https://example.com", timeout=1)
        logging.info(f"Sync response status: {response.status_code}")

async def async_request():
    async with httpx.AsyncClient() as client:
        response = await client.get("https://example.com", timeout=1)
        logging.info(f"Async response status: {response.status_code}")

async def main():
    await async_request()

sync_request()
asyncio.run(main())

HTTPXClientInstrumentor().uninstrument()

        HTTPXClientInstrumentor.instrument_client(client)
        response = client.get("https://example.com")
```
Old semconv with envvar set to "default" or "":
```json
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "http.method": "GET",
        "http.url": "https://example.com/",
        "http.status_code": 200
    },
```
New semconv with envvar set to "http" exception, error, and ok:
```json
    "status": {
        "status_code": "ERROR",
        "description": "ConnectTimeout: timed out"
    },
    "attributes": {
        "http.request.method": "GET",
        "url.full": "https://example.com:8443/",
        "server.address": "example.com",
        "network.peer.address": "example.com",
        "server.port": 8443,
        "network.peer.port": 8443,
        "error.type": "ConnectTimeout"
    },
```
```json
    "status": {
        "status_code": "ERROR"
    },
    "attributes": {
        "http.request.method": "GET",
        "url.full": "https://example.com/ok",
        "server.address": "example.com",
        "network.peer.address": "example.com",
        "http.response.status_code": 404,
        "error.type": "404",
        "network.protocol.version": "1.1"
    },
```
```json
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "http.request.method": "GET",
        "url.full": "https://example.com/",
        "server.address": "example.com",
        "network.peer.address": "example.com",
        "http.response.status_code": 200,
        "network.protocol.version": "1.1"
    },
```

Both semconv with envvar set to "http/dup":
```json
    "name": "GET",
    "kind": "SpanKind.CLIENT",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "http.method": "GET",
        "http.request.method": "GET",
        "http.url": "https://example.com/",
        "url.full": "https://example.com/",
        "http.host": "example.com",
        "server.address": "example.com",
        "network.peer.address": "example.com",
        "http.status_code": 200,
        "http.response.status_code": 200,
        "http.flavor": "1.1",
        "network.protocol.version": "1.1"
    },
```

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-httpx/tests/test_httpx_integration.py
```diff
@@ -115,12 +134,35 @@ class BaseTest(TestBase, metaclass=abc.ABCMeta):
...
+                sem_conv_mode = "http"
+            elif "both_semconv" in test_name:
+                sem_conv_mode = "http/dup"
+            self.env_patch = mock.patch.dict(
+                "os.environ",
+                {
+                    OTEL_SEMCONV_STABILITY_OPT_IN: sem_conv_mode,
+                },
+            )
+            self.env_patch.start()
```
xrmx, line 151: Wondering if instead of making this under the hood we can make it more explicit and add the three decorators instead, e.g `@default_http_semconv`, `@new_http_semconv`, `@both_http_semconv`

Maybe we can avoid the one for the default one if the others can override the patch started here on setUp.

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -279,7 +281,7 @@ def _extract_parameters(args, kwargs):
     else:
         # In httpx < 0.20.0, handle_request receives the parameters separately
         method = args[0]
-        url = args[1]
+        url = httpx.URL(args[1])
```
lzchen, line 284: Was this a bug previously?

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -318,12 +391,17 @@ def __init__(
         request_hook: typing.Optional[RequestHook] = None,
         response_hook: typing.Optional[ResponseHook] = None,
     ):
+        _OpenTelemetrySemanticConventionStability._initialize()
+        self.semconv = _OpenTelemetrySemanticConventionStability._get_opentelemetry_stability_opt_in_mode(
```
lzchen, line 395:
```suggestion
        self._semconv = _OpenTelemetrySemanticConventionStability._get_opentelemetry_stability_opt_in_mode(
```

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -355,39 +434,55 @@ def handle_request(
         method, url, headers, stream, extensions = _extract_parameters(
             args, kwargs
         )
-        span_attributes = _prepare_attributes(method, url)
+        method_original = method.decode()
```
lzchen, line 437: What's the reason for the `decode()` addition?

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
emdneto: I don't think so. The expected type is `typing. Tuple[bytes, bytes, typing.Optional[int], bytes]`. The else statement is just to please httpx older versions regarding parameter extraction, but in those versions httpx.URL also [supports this signature](https://github.com/encode/httpx/blob/0.18.2/httpx/_models.py#L122). What I'm doing here is also returning a httpx.URL for those versions as well. I don't think it's a bug, just did that to avoid doing `httpx.URL` again in `_apply_request_client_attributes_to_span`.

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
emdneto: Actually we are already using `decode()` in L258 of the current implementation `_method = method.decode().upper()`. What I'm doing here is to get it as str to pass to the `_get_default_span_name` and later to the `sanitize_method`.

instrumentation/opentelemetry-instrumentation-httpx/tests/test_httpx_integration.py
emdneto: I'm happy to do that in a follow-up PR if you're fine with that. My concern is about these changes being compatible with what we already have in other transitions; after this gets merged, I can work on a better approach. 
@lzchen let me know if you are you ok with that as well.

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
emdneto: I can keep it as `url = args[1]` and do the parser later just to get the URL attributes like host and etc.

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -299,6 +301,77 @@ def _inject_propagation_headers(headers, args, kwargs):
...
+
+
+def _apply_request_client_attributes_to_span(
+    span_attributes, url, method_original, span_name, semconv
+):
+    # http semconv transition: http.method -> http.request.method
+    _set_http_method(span_attributes, method_original, span_name, semconv)
+    # http semconv transition: http.url -> url.full
+    _set_http_url(span_attributes, str(url), semconv)
+    try:
```
lzchen, line 331: Is the try except there for in the case `url` does not have `host` or `port` fields?

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -355,39 +434,55 @@ def handle_request(
...
+                    )
+
+                if callable(self._response_hook):
+                    self._response_hook(
+                        span,
+                        request_info,
+                        ResponseInfo(status_code, headers, stream, extensions),
+                    )
+
+            if exception and _report_new(self.semconv):
```
lzchen, line 480: Think this should be within the `is_recording()` check.

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -299,6 +301,77 @@ def _inject_propagation_headers(headers, args, kwargs):
...
+    span, status_code, http_version, semconv
+):
+    span_attributes = {}
+    # http semconv transition: http.status_code -> http.response.status_code
+    _set_http_status_code(
+        span_attributes,
+        status_code,
+        semconv,
+    )
+    http_status_code = http_status_to_status_code(status_code)
```
lzchen, line 356: Would you be able to use [set_status](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py#L344) for this part?

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
emdneto: Yes

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
lzchen: Edit: Can probably handle setting `error.type` to exception name if you use [set_status](_apply_response_client_attributes_to_span) and pass in `status_code_str`

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
emdneto: I'll use `_sem_conv_opt_in_mode` like the other ones

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
emdneto: Yes. I think it's possible, I remember I did it this way because I saw you were working on some changes of this function in asgi transition and I was unsure about the parameters order

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
emdneto: Do you mean something like that:

```
if span.is_recording() and exception is not None:
  _set_status(span, {}, type(exception).__qualname__, -1, self._sem_conv_opt_in_mode)
```

emdneto: I think the `if exception and _report_new(self._sem_conv_opt_in_mode):` makes it more explicit that will apply only for new semconv no?

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
emdneto: In that case, `host` and `port` are properties of httpx.URL and they have defaults. Think I can remove this.

instrumentation/opentelemetry-instrumentation-httpx/tests/test_httpx_integration.py
xrmx: Sure

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
emdneto: # TODO: use _set_status when it's stable for http clients

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
emdneto: Opted to keep as-is and do httpx.URL again when generating attributes.

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
lzchen: I don't believe we have exception handling in most places for expected fields from using instrumented libraries. At best I think we can check if the fields are there with `hasattr` and if they do not exist, we simply exclude the span attribute.

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
lzchen: We don't set attributes if the span is not recording.

emdneto: @lzchen, I agree. I was rreferring to _set_status. I made an update to check if it is recording and created a test to validate this.

lzchen (approved):

PR #2635, opened 2024-06-26
Add additional attributes for redis.search methods create_index, search

mpozniak95:
# Description

Add additional attributes for redisearch methods: redis.commands.search, create_index:

For create_index method: https://redis-py.readthedocs.io/en/stable/redismodules.html#redis.commands.search.commands.SearchCommands.create_index
Attributes:
- fields - to show from which fields index was created

For search method: https://redis-py.readthedocs.io/en/stable/redismodules.html#redis.commands.search.commands.SearchCommands.search
Attributes:
- query - shows query string
- total - shows number of documents returned
- xdoc{index} - shows returned documents

## Type of change

# How Has This Been Tested?

- [ ] Added functional tests for these methods

- [ ] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py
```diff
@@ -68,3 +68,25 @@ def _format_command_args(args):
         out_str = ""
 
     return out_str
+
+
+def _check_skip(name):
+    skip_list = ["FT.SEARCH", "FT.CREATE", "FT.AGGREGATE"]
```
lzchen, line 74: Why are we not tracing these commands in particular?

lzchen:
@mpozniak95 

I am not too familiar with the inner workings of Redis. Could you update the description of the PR to be more descriptive of why are we making these changes? Is the plan to add INTERNAL spans or should these calls be treated as DATABASE spans?

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py
mpozniak95: They are traced but with new methods (_traced_search(), etc.). I added that to avoid having repeated spans (one from _traced_execute_command and one from new method). I am not sure if that is the best way to handle it. Maybe we should just keep two spans for those functions or edit current implementation to add those additional attributes to the span?

mpozniak95:
Idea is to integrate it with: https://github.com/traceloop/openllmetry. Since those three methods are probably the most used Redis methods in LLM applications we would like to have additional attributes in spans of that methods.

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py
lzchen: Where is `_traced_search()` defined?

lzchen:
@mpozniak95 

Please update the description of the pr to reflect what you are trying to do. I am not familiar with the inner workings of https://github.com/traceloop/openllmetry so please summarize exactly what you are trying to do and what new methods you are trying to trace (linking to redis docs would be most helpful as well).

> Is the plan to add INTERNAL spans or should these calls be treated as DATABASE spans?

You are currently also creating `INTERNAL` spans for the current implementation. Is this correct or should we treating the spans as `CLIENT` with type database which follows database  [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/redis.md)

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py
```diff
@@ -68,3 +68,25 @@ def _format_command_args(args):
...
+    for method in skip_list:
+        if method in name:
+            return True
+    return False
+
+
+def _set_span_attribute(span, name, value):
+    if value is not None:
```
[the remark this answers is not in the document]
xrmx: `if value:` then?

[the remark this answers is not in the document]
lzchen: Probably still want to keep value if it is `False`.

mpozniak95:
You are right the spans should be `CLIENT` type, I changed it.

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py
lzchen: Ahh I see, this is an ignore list because these are the commands you are adding more sophisticated tracing for. 

> edit current implementation to add those additional attributes to the span?

In that case I believe this would be the best way to handle this. Utilize the already existing `_traced_execute_command` and create helper methods to determine the span name as well as what attributes should be populated on the span depending on what command it is.

mpozniak95:
I changed the implementation to catch those functions in `_traced_execute_command` and edit their span_name and attributes. Also for now I deleted the `aggregate` from the PR, I will add it in separate PR once we merge this.

I feel that it got a little complicated right now, it was easy to extract the response from redis.commands.search function directly (because it was already parsed by `_parse_response `function in redis package) but `execute_command` returns unparsed response. I added some comments to the code (I hope they make any sense, maybe it will be easier to review)

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
```diff
@@ -230,6 +238,64 @@ def _traced_execute_pipeline(func, instance, args, kwargs):
...
+        # According to: https://github.com/redis/redis-py/blob/master/redis/commands/search/commands.py#L155 schema is last argument for execute command
+        try:
+            schema_index = args.index("SCHEMA")
+        except ValueError:
+            return
+        schema = args[schema_index:]
+        field_attribute = ""
+        # Schema in format:
+        # [first_field_name, first_field_type, first_field_some_attribute1, first_field_some_attribute2, second_field_name, ...]
+        field_types = ["NUMERIC", "TEXT", "GEO", "TAG", "VECTOR"]
```
emdneto, line 254: Can we define a constant for this? And also use list comprehension here?
Something like this:

```
field_attribute = "".join(
    f"Field(name: {schema[index - 1]}, type: {schema[index]});"
    for index in range(1, len(schema))
    if schema[index] in _FIELD_TYPES
)
```

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
```diff
@@ -230,6 +238,64 @@ def _traced_execute_pipeline(func, instance, args, kwargs):
...
+        # https://redis.io/docs/latest/commands/ft.search/
+        # Response in format:
+        # [number_of_returned_documents, index_of_first_returned_doc, first_doc(as a list), index_of_second_returned_doc, second_doc(as a list) ...]
+        # Returned documents in array format:
+        # [first_field_name, first_field_value, second_field_name, second_field_value ...]
+        number_of_returned_documents = _value_or_none(response, 0)
+        _set_span_attribute(
+            span, "redis.search.total", number_of_returned_documents
+        )
+        if "NOCONTENT" in args:
```
emdneto, line 283: do we also need to early return if not number_of_returned_documents?

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py
```diff
@@ -68,3 +68,16 @@ def _format_command_args(args):
         out_str = ""
 
     return out_str
+
+
+def _set_span_attribute(span, name, value):
```
xrmx, line 73: Maybe call this `_set_span_attribute_if_set` to hint that there is a filter?

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
mpozniak95: We can, then we don't need another if in line 285.

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/util.py
mpozniak95: I changed it to `_set_span_attribute_if_value`

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
mpozniak95: Done

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
```diff
@@ -192,6 +200,10 @@ def _traced_execute_command(func, instance, args, kwargs):
             if callable(request_hook):
                 request_hook(span, instance, args, kwargs)
             response = func(*args, **kwargs)
+            if span.name == "redis.search":
```
lzchen, line 203: I think these should be within the `span.is_recording()` check.

mpozniak95: Changed

lzchen (approved):
LGTM. @mpozniak95  Please add a CHANGELOG entry and we can get this merged!

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
```diff
@@ -181,17 +190,21 @@ def _instrument(
...
             if span.is_recording():
                 span.set_attribute(SpanAttributes.DB_STATEMENT, query)
                 _set_connection_attributes(span, instance)
                 span.set_attribute("db.redis.args_length", len(args))
+                if span.name == "redis.create_index":
+                    _add_create_attributes(span, args)
             if callable(request_hook):
                 request_hook(span, instance, args, kwargs)
             response = func(*args, **kwargs)
+            if span.is_recording():
```
lzchen, line 205: Nit: Can't we include this as part of the above "if"?

mpozniak95: I think no, because one of the additional attributes is based on the response of search method so we need this to be placed after:
`response = func(*args, **kwargs)`
which as I understand should happen even if span is not recording

mpozniak95:
So, can we merge it :)?

PR #2645, opened 2024-06-28
Fix wrong unit in Flask new semconv test

### LINKED ISSUE #2596: flask instrumentation flaky test
[instrumentations-0 (pypy3, flask, ubuntu-20.04)](https://github.com/emdneto/opentelemetry-python-contrib/actions/runs/9465152589/job/26074132278#logs)
```shell

_______________ TestProgrammatic.test_flask_metrics_new_semconv ________________

self = <tests.test_programmatic.TestProgrammatic testMethod=test_flask_metrics_new_semconv>

    def test_flask_metrics_new_semconv(self):
        start = default_timer()
        self.client.get("/hello/123")
        self.client.get("/hello/321")
        self.client.get("/hello/756")
        duration = max(round((default_timer() - start) * 1000), 0)
        metrics_list = self.memory_metrics_reader.get_metrics_data()
        number_data_point_seen = False
        histogram_data_point_seen = False
        self.assertTrue(len(metrics_list.resource_metrics) != 0)
        for resource_metric in metrics_list.resource_metrics:
            self.assertTrue(len(resource_metric.scope_metrics) != 0)
            for scope_metric in resource_metric.scope_metrics:
                self.assertTrue(len(scope_metric.metrics) != 0)
                for metric in scope_metric.metrics:
                    self.assertIn(metric.name, _expected_metric_names_new)
                    data_points = list(metric.data.data_points)
                    self.assertEqual(len(data_points), 1)
                    for point in data_points:
                        if isinstance(point, HistogramDataPoint):
                            self.assertEqual(point.count, 3)
>                           self.assertAlmostEqual(
                                duration, point.sum, delta=10
                            )
E                           AssertionError: 17 != 0.011322596000013618 within 10 delta (16.988677403999986 difference)
```

**Steps to reproduce**
`Run tox -f pypy3-flask`

test pass everytime

**What did you see instead**?
test is flaky

**Additional context**

emdneto:
# Description
Changes duration to be measured in seconds during the test since in [stable semconv](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/migration-guide.md#http-client-duration-metric), the unit changed from ms to s and we are recording in [seconds](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py#L398)

Probably Fixes #2596

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
+++ b/instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
@@ -514,7 +514,7 @@ class TestProgrammatic(InstrumentationTest, WsgiTestBase):
                         if isinstance(point, HistogramDataPoint):
                             self.assertEqual(point.count, 3)
                             self.assertAlmostEqual(
-                                duration_s, point.sum, delta=10
+                                duration_s, point.sum, delta=2
                             )
                             histogram_data_point_seen = True
                         if isinstance(point, NumberDataPoint):
--- a/instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
+++ b/instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
@@ -514,7 +514,7 @@ class TestProgrammatic(InstrumentationTest, WsgiTestBase):
                         if isinstance(point, HistogramDataPoint):
                             self.assertEqual(point.count, 3)
                             self.assertAlmostEqual(
-                                duration_s, point.sum, delta=2
+                                duration_s, point.sum, places=2
                             )
                             histogram_data_point_seen = True
                         if isinstance(point, NumberDataPoint):
--- a/instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
+++ b/instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
@@ -497,8 +497,7 @@ class TestProgrammatic(InstrumentationTest, WsgiTestBase):
         self.client.get("/hello/123")
         self.client.get("/hello/321")
         self.client.get("/hello/756")
-        # new semconv Unit: ms -> s
-        duration = max(default_timer() - start, 0)
+        duration_s = max(default_timer() - start, 0)
         metrics_list = self.memory_metrics_reader.get_metrics_data()
         number_data_point_seen = False
         histogram_data_point_seen = False
@@ -515,7 +514,7 @@ class TestProgrammatic(InstrumentationTest, WsgiTestBase):
                         if isinstance(point, HistogramDataPoint):
                             self.assertEqual(point.count, 3)
                             self.assertAlmostEqual(
-                                duration, point.sum, delta=10
+                                duration_s, point.sum, delta=10
                             )
                             histogram_data_point_seen = True
                         if isinstance(point, NumberDataPoint):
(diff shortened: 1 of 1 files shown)


instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
```diff
@@ -497,7 +497,8 @@ def test_flask_metrics_new_semconv(self):
         self.client.get("/hello/123")
         self.client.get("/hello/321")
         self.client.get("/hello/756")
-        duration = max(round((default_timer() - start) * 1000), 0)
+        # new semconv Unit: ms -> s
```
xrmx, line 500: Instead of the comment I would rename the variable to `duration_s`

xrmx (commented):
I don't think this is the only test that has the same duration calculation

xrmx (commented):
Given this is now in seconds, isn't 10 seconds a bit too much for delta?

emdneto:
> I don't think this is the only test that has the same duration calculation

Do you mean in other instrumentations? If so, I think yes, that's right. For Flask, I think this is the only one that uses new_semconv. I think the other ones are correct because they are using the default opt-in (old semconv), and the histogram unit is in `ms`

xrmx:
> > I don't think this is the only test that has the same duration calculation
> 
> Do you mean in other instrumentations? If so, I think yes, that's right. For Flask, I think this is the only one that uses new_semconv. I think the other ones are correct because they are using the default opt-in (old semconv), and the histogram unit is in `ms`



> > I don't think this is the only test that has the same duration calculation
> 
> Do you mean in other instrumentations? If so, I think yes, that's right it might have other tests. For Flask, I believe this is the only one that uses new_semconv. I think the other tests in Flask are correct because they are using the default opt-in (old semconv), and the histogram unit is in `ms`

emdneto:
> Given this is now in seconds, isn't 10 seconds a bit too much for delta?
Makes sense

smoke (approved):

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
```diff
@@ -514,7 +514,7 @@ def test_flask_metrics_new_semconv(self):
                         if isinstance(point, HistogramDataPoint):
                             self.assertEqual(point.count, 3)
                             self.assertAlmostEqual(
-                                duration, point.sum, delta=10
+                                duration_s, point.sum, delta=10
```
xrmx:
```suggestion
                                duration_s, point.sum, places=2
```

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
```diff
@@ -514,7 +514,7 @@ def test_flask_metrics_new_semconv(self):
                         if isinstance(point, HistogramDataPoint):
                             self.assertEqual(point.count, 3)
                             self.assertAlmostEqual(
-                                duration, point.sum, delta=10
+                                duration_s, point.sum, delta=2
```
xrmx, line 517:
```suggestion
                                duration_s, point.sum, places=2
```

xrmx (approved):
Thanks!

lzchen (approved):

PR #2650, opened 2024-07-02
fix(instrumentation-asgi): remove high cardinal path from span name

srikanthccv:
# Description

The `path` value of the ASGI `scope` is a non-generalized path. This is not an issue in other frameworks such as fastapi or falsk where the templatized path is accessible. This PR updates the span name to not contain the high cardinal value for ASGI instrumentation span names. The highly cardinal span names will become a problem when used with the span metrics connection from the collector.

The spec guidelines say https://opentelemetry.io/docs/specs/otel/trace/api/#span

>The span name concisely identifies the work represented by the Span, for example, an RPC method name, a function name, or the name of a subtask or stage within a larger computation. The span name SHOULD be the most general string that identifies a (statistically) interesting class of Spans, rather than individual Span instances while still being human-readable. That is, “get_user” is a reasonable name, while “get_user/314159”, where “314159” is a user ID, is not a good name due to its high cardinality. Generality SHOULD be prioritized over human-readability.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
@@ -403,13 +403,9 @@ def get_default_span_details(scope: dict) -> Tuple[str, dict]:
     Returns:
         a tuple of the span name, and any attributes to attach to the span.
     """
-    path = scope.get("path", "").strip()
-    method = scope.get("method", "").strip()
-    if method and path:  # http
-        return f"{method} {path}", {}
-    if path:  # websocket
-        return path, {}
-    return method, {}  # http with no path
+    if scope.get("type") == "http":
+        return scope.get("method", ""), {}
+    return scope.get("type", ""), {}
 
 
 def _collect_target_attribute(
--- a/instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
@@ -227,6 +227,53 @@ async def error_asgi(scope, receive, send):
         await send({"type": "http.response.body", "body": b"*"})
 
 
+# New ASGI app for user update
+async def user_update_app(scope, receive, send):
+    assert scope["type"] == "http"
+    assert scope["method"] == "PUT"
+    assert scope["path"].startswith("/api/v3/io/users/")
+
+    user_id = scope["path"].split("/")[-1]
+
+    message = await receive()
+    if message["type"] == "http.request":
+        await send(
+            {
+                "type": "http.response.start",
+                "status": 200,
+                "headers": [
+                    [b"Content-Type", b"application/json"],
+                ],
+            }
+        )
+        await send(
+            {
+                "type": "http.response.body",
+                "body": f'{{"status": "updated", "user_id": "{user_id}"}}'.encode(),
+            }
+        )
+
+
+# New ASGI app for WebSocket
+async def websocket_session_app(scope, receive, send):
+    assert scope["type"] == "websocket"
+    assert scope["path"].startswith("/ws/")
+
+    session_id = scope["path"].split("/")[-1]
+
+    await send({"type": "websocket.accept"})
+
+    while True:
+        event = await receive()
+        if event["type"] == "websocket.disconnect":
+            break
+        if event["type"] == "websocket.receive":
+            if event.get("text") == "ping":
+                await send(
+                    {"type": "websocket.send", "text": f"pong:{session_id}"}
+                )
+
+
 # pylint: disable=too-many-public-methods
 class TestAsgiApplication(AsgiTestBase):
     def validate_outputs(self, outputs, error=None, modifiers=None):
@@ -266,12 +313,12 @@ class TestAsgiApplication(AsgiTestBase):
         span_list = self.memory_exporter.get_finished_spans()
         expected = [
             {
-                "name": "GET / http receive",
+                "name": "GET http receive",
                 "kind": trace_api.SpanKind.INTERNAL,
                 "attributes": {"asgi.event.type": "http.request"},
             },
             {
-                "name": "GET / http send",
+                "name": "GET http send",
                 "kind": trace_api.SpanKind.INTERNAL,
                 "attributes": {
                     SpanAttributes.HTTP_STATUS_CODE: 200,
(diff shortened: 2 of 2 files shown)


shalevr (approved):

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
emdneto, line 500: Can we update the comment and also reference the one you mentioned in the PR description?

CHANGELOG.md
```diff
@@ -42,6 +42,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3882](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3882))
 - `opentelemetry-instrumentation-aiohttp-server`: delay initialization of tracer, meter and excluded urls to instrumentation for testability
   ([#3836](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3836))
+- `opentelemetry-instrumentation-asgi` remove high cardinal path from span name
```
emdneto, line 45: plus remove the protocol type from the span name if http?

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
aabmass, line 1: Good to see you @srikanthccv 👋 

Is the span name being modified here supposed to follow the OTel [HTTP server semconv](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name) or is it a more internal span? This change would be a breakage for anyone relying on this particular name, but if it's not the main `{method} {target}` span it may be OK.

srikanthccv: >Is the span name being modified here supposed to follow the OTel [HTTP server semconv](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name) or is it a more internal span?

Correct, specifically this part of the spec: _If there is no (low-cardinality) {target} available, HTTP span names SHOULD be {method}._ The `path` value of the ASGI scope for both HTTP and WebSocket contains the non-templatized path. The change in this PR removes the `path` as the `target` and just sets the method as recommended.


<details>
  <summary>Some examples</summary>
image

</details>


> This change would be a breakage for anyone relying on this particular name

Agreed, but at the same time, it doesn't follow the semantic convention and spec guidelines for span names. I would say it's a necessary breaking change. Even from the end-user perspective, I believe this is a useful change; otherwise, IDs (from messaging-adjacent apps, in my experience) become part of the name, leading to no meaningful aggregation for server spans. In many deployments, the spanmetrics connector is used to derive metrics from these spans, which creates a problem.

aabmass: > Agreed, but at the same time, it doesn't follow the semantic convention and spec guidelines for span names. I would say it's a necessary breaking change. Even from the end-user perspective, I believe this is a useful change

Ya I agree, maybe we can just gate this behind the new semconv stability opt in like we do here?https://github.com/open-telemetry/opentelemetry-python-contrib/blob/bd3c1f2aa2e58490e45ba9c95e140112e9016c56/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L741-L743

How hard would that be to do?

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
```diff
@@ -357,25 +404,25 @@ def validate_outputs(
         span_list = self.memory_exporter.get_finished_spans()
         expected_old = [
             {
-                "name": "GET / http receive",
+                "name": "GET receive",
```
aabmass, line 407: Don't have a ton of context on these tests, but this still doesn't seem to be following the HTTP semconv. What is `receive` here?

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
srikanthccv: Let me take a look at it.

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
srikanthccv: These are the internal spans created in the receive and send callable of the ASGI application https://asgi.readthedocs.io/en/latest/specs/main.html#applications. The name will be `{method} receive` or `{method} send` depending on the event. Does that clarify?

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
srikanthccv: I went through the code. It should be possible to keep the old and new name following the opt in mode. Let me know if we should make it opt-in.

tammy-baylis-swi: Hi! 👋 

There is a similar PR open for the aiohttp server instrumentation that will also introduce a breaking change to span name in order to follow semconv and reduce cardinality: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3896#discussion_r2500299139 . @aabmass should that PR also include an opt-in? cc @krnr

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
aabmass: Ah so these are in addition to the standard HTTP semconv, generally showing as child spans? SGTM

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
aabmass: IMO it would be better since the purpose of the opt-in was to prevent breaking people

Any objections to that though? Does it work for you @srikanthccv ?

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
srikanthccv: > Ah so these are in addition to the standard HTTP semconv, generally showing as child spans?

Yes

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
srikanthccv: Sounds good to me 👍

xrmx: Fine for me to gate the change under the semconv config but I think we're maybe a bit too cautious :)

PR #2652, opened 2024-07-02
Add capture to open file descriptors in system metrics instrumentation

### LINKED ISSUE #2562: looks like process.open_file_descriptor.count is not implemented in python

I do not see `process.open_file_descriptor.count` metric being collected in https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-system-metrics/src/opentelemetry/instrumentation/system_metrics/__init__.py. 

[This](https://opentelemetry.io/docs/specs/semconv/system/process-metrics/#metric-processopen_file_descriptorcount) says that it should be collected.

`process.open_file_descriptor.count` metric should be collected.

N/A
**Additional context**

N/A

emdneto:
# Description

Fixes #2562

## Type of change

# How Has This Been Tested?

- [X] tox 

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-system-metrics/src/opentelemetry/instrumentation/system_metrics/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-system-metrics/src/opentelemetry/instrumentation/system_metrics/__init__.py
@@ -399,7 +399,7 @@ class SystemMetricsInstrumentor(BaseInstrumentor):
 
         if "process.open_file_descriptor.count" in self._config:
             self._meter.create_observable_up_down_counter(
-                name=f"process.{self._python_implementation}.open_file_descriptor.count",
+                name="process.open_file_descriptor.count",
                 callbacks=[self._get_open_file_descriptors],
                 description="Number of file descriptors in use by the process.",
             )
--- a/instrumentation/opentelemetry-instrumentation-system-metrics/src/opentelemetry/instrumentation/system_metrics/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-system-metrics/src/opentelemetry/instrumentation/system_metrics/__init__.py
@@ -401,7 +401,7 @@ class SystemMetricsInstrumentor(BaseInstrumentor):
             self._meter.create_observable_up_down_counter(
                 name=f"process.{self._python_implementation}.open_file_descriptor.count",
                 callbacks=[self._get_open_file_descriptors],
-                description="Number of open file descriptors",
+                description="Number of file descriptors in use by the process.",
             )
 
     def _uninstrument(self, **__):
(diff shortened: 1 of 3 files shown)


xrmx (commented):
This requires a changelog entry

xrmx (approved):

instrumentation/opentelemetry-instrumentation-system-metrics/src/opentelemetry/instrumentation/system_metrics/__init__.py
```diff
@@ -395,9 +397,25 @@ def _instrument(self, **kwargs):
                 unit="switches",
             )
 
+        if "process.open_file_descriptor.count" in self._config:
+            self._meter.create_observable_up_down_counter(
+                name=f"process.{self._python_implementation}.open_file_descriptor.count",
+                callbacks=[self._get_open_file_descriptors],
+                description="Number of open file descriptors",
```
lzchen, line 404:
```suggestion
                description="Number of file descriptors in use by the process.",
```

instrumentation/opentelemetry-instrumentation-system-metrics/src/opentelemetry/instrumentation/system_metrics/__init__.py
```diff
@@ -395,9 +397,25 @@ def _instrument(self, **kwargs):
                 unit="switches",
             )
 
+        if "process.open_file_descriptor.count" in self._config:
+            self._meter.create_observable_up_down_counter(
+                name=f"process.{self._python_implementation}.open_file_descriptor.count",
```
xrmx, line 402: Taking a second look I don't think this is dependent on the python implementation though. Do we need it? Genuine question, no clue why we are adding these prefixes.

emdneto: Good question. If we look at [semconv here](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-semantic-conventions/src/opentelemetry/semconv/_incubating/metrics/process_metrics.py#L177) there's no reference to python implementation in the metric name.

lzchen: It does look like the original intent was to match the semantic conventions for [runtime environment metrics](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/runtime/README.md#metrics) where the runtime is prefixed. I believe the question is whether `process.*` metrics should be treated as runtime metrics or `process` metrics. To me it is mostly the latter so we should probably follow the semantic conventions for [process metrics](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/process-metrics.md#process-metrics) and remove the runtime prefix altogether.

lzchen: @emdneto 

For this pr, I suggest removing the prefix for file descriptor and maybe opening up an issue to remove the prefixes for everything else is people agree on this approach.

emdneto: I'll open the issue. Thanks

ocelotl (approved):

PR #2655, opened 2024-07-02
Add opentelemetry-util-http tests and lint to CI

emdneto:
# Description

This package has utils that are used in several http instrumentation and I think would be nice to also have it tests and lint running on CI


.github/workflows/lint.yml
```diff
@@ -70,6 +70,7 @@ jobs:
           - "propagator-ot-trace"
           - "resource-detector-container"
           - "sdk-extension-aws"
+          - "util-http"
```
emdneto, line 73: Rename to opentelemetry-util-http to match the package name?

lzchen (approved):

.github/workflows/lint.yml
lzchen: Not a huge deal. We use shorter terms like "distro" as well for other packages.

xrmx (approved):

PR #2658, opened 2024-07-02
Point pylint to the package root directories

### LINKED ISSUE #2657: pylint is not pointing to the root directory of every package
### Describe your environment

OS: (e.g, Ubuntu)
Python version: (e.g., Python 3.8.10)
Package version: (e.g., 0.46.0)

### What happened?

pylint is not pointing to the root directory of every package, this could lead to some files [not being checked](https://github.com/open-telemetry/opentelemetry-python/issues/3814#issuecomment-2204496561).

### Steps to Reproduce

Check `tox.ini`

### Expected Result

`pylint` points to the root directory of every package

### Actual Result

`pylint` does not point to the root directory of every package

### Additional context

_No response_

None

ocelotl:
Fixes #2657

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-grpc/tests/protobuf/test_server_pb2_grpc.py
+++ b/instrumentation/opentelemetry-instrumentation-grpc/tests/protobuf/test_server_pb2_grpc.py
@@ -12,6 +12,7 @@ class GRPCTestServerStub:
         Args:
             channel: A grpc.Channel.
         """
+        # pylint: disable=invalid-name
         self.SimpleMethod = channel.unary_unary(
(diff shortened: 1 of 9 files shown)


tammy-baylis-swi (approved):

xrmx:
Now it's only matter of fixing hundreds warnings :sweat_smile:

ocelotl:
> Now it's only matter of fixing hundreds warnings 😅

Right, :laughing: 

Apparently that happened because I forgot to tell `pylint` to use the configuration file, after I fixed that the amount of failures dropped to just 6 failed lint jobs.

Issue #2639, opened 2024-06-26
`redis.exception.WatchError` is not really an error


chrisguidry:
### What problem do you want to solve?

When upgrading `opentelemetry-instrumentation-redis` from 0.39b0 to 0.46b0, we picked up the (very nice) fixes to instrumentation of the async redis client.  However, all of our `arq` (https://github.com/samuelcolvin/arq/) workers started marking tons of traces as errors because they got a `WatchError` exception.

The `redis` `WatchError` is not really an error, it's a flow-control and concurrency-management mechanism (kind of like a `LockError`, etc).  The fact that we're marking all of those spans as errored introduces a lot of noise to something that's a totally reasonable pattern and use of redis.

Here's an example of how it's used for concurrency control: https://github.com/samuelcolvin/arq/blob/main/arq/connections.py#L178C20-L180

### Describe the solution you'd like

I'd recommend that we adjust the instrumentation to not mark spans as failed when they encounter `WatchError` (or to at least make this behavior configurable).  It does make sense to still mark them as events on the current span, though.

### Would you like to implement a fix?

None

Charlie-lizhihan:
May I take this one?

xrmx:
@Charlie-lizhihan sure

Charlie-lizhihan:
Hi, @xrmx @chrisguidry 
When an WatchError occur, the current instrumentation will set span as 
```
{
    "name": "SET",
    "context": {
        "trace_id": "0x878718c68c75dbc45f875358f21f1820",
        "span_id": "0x4302e43a40344ddd",
        "trace_state": "[]"
    },
    "kind": "SpanKind.CLIENT",
    "parent_id": null,
    "start_time": "2024-07-04T13:37:05.133151Z",
    "end_time": "2024-07-04T13:37:05.136076Z",
    "status": {
        "status_code": "ERROR",
        "description": "WatchError: Watched variable changed."
    },
    "attributes": {
        "db.statement": "SET ? ?",
        "db.system": "redis",
        "db.redis.database_index": 0,
        "net.peer.name": "localhost",
        "net.peer.port": 6379,
        "net.transport": "ip_tcp",
        "db.redis.pipeline_length": 1
    },
    "events": [
        {
            "name": "exception",
            "timestamp": "2024-07-04T13:37:05.136058Z",
            "attributes": {
                "exception.type": "redis.exceptions.WatchError",
                "exception.message": "Watched variable changed.",
[...]

PR #2668, opened 2024-07-05
Handle redis.exceptions.WatchError as a non-error event in instrumentation

zhihali:
# Description

Fixes # (issue)
When we face the WatchError when running redis instrumentation, Status Code: The status_code should be "UNSET" instead of "ERROR",
#2639

## Type of change

# How Has This Been Tested?

py312-test-instrumentation-redis

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-redis/tests/test_redis.py
```diff
@@ -311,3 +314,32 @@ def test_attributes_unix_socket(self):
...
+                    await redis_client.set("a", "bad")
+                    pipe.multi()
+                    await pipe.set("a", "1")
+                    await pipe.execute()
+            except WatchError:
+                pass
+
+        await redis_operations()
+
+        spans = self.memory_exporter.get_finished_spans()
```
emdneto, line 344: Are we expecting only one span here? If so, it's okay to assert the length of spans.

tox.ini
```diff
@@ -1206,6 +1206,7 @@ deps =
   docopt==0.6.2
   exceptiongroup==1.2.0
   flaky==3.7.0
+  fakeredis==2.23.3
```
emdneto, line 1209: Do docker tests fail without that?

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
```diff
@@ -212,9 +212,16 @@ def _traced_execute_pipeline(func, instance, args, kwargs):
...
-            response = func(*args, **kwargs)
-            if callable(response_hook):
-                response_hook(span, instance, response)
+
+            response = None
+            try:
+                response = func(*args, **kwargs)
+            except redis.WatchError:
+                span.set_status(StatusCode.UNSET)
+            finally:
```
xrmx, line 221: You can avoid the finally and save one level of indentation right?

tox.ini
xrmx: The import will fail, maybe handle the input failure and skip the test?

zhihali: I have removed this one and seems the CI still could pass

instrumentation/opentelemetry-instrumentation-redis/tests/test_redis.py
zhihali: you are right, we should have 3 spans, I will add more assert here

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
zhihali: yes, updated, thanks for pointing it out

tox.ini
emdneto: But the tests in docker-tests are like integration tests no?  I mean, we have a docker-compose.yml file which is deploying a real redis-cluster and seems in redis dockertests we are not mocking a redis

emdneto (commented):
Are we missing a test to check if any other error will have the `StatusCode.ERROR`?

zhihali:
> Are we missing a test to check if any other error will have the `StatusCode.ERROR`?

Yes, and updated. Thanks for the review

instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
```diff
@@ -281,7 +288,13 @@ async def _async_traced_execute_pipeline(func, instance, args, kwargs):
                 span.set_attribute(
                     "db.redis.pipeline_length", len(command_stack)
                 )
-            response = await func(*args, **kwargs)
+
+            response = None
+            try:
+                response = func(*args, **kwargs)
```
xrmx, line 294: you lost an await here, do we have tests for async?

zhihali: you are right! I should add await here. Thanks!

I have added the async test, but seems without await here, the test will still pass.

this remind me that I should add sync test of WatchError as well, will update it in newest commit.

xrmx (approved):

emdneto (approved):

chrisguidry:
Hi folks, shouldn't the `WatchError` be re-raised?  It's part of a redis flow control mechanism but this change will have the instrumentation swallowing the exception.

instrumentation/opentelemetry-instrumentation-redis/tests/test_redis.py
```diff
@@ -311,3 +316,113 @@ def test_attributes_unix_socket(self):
...
+            try:
+                redis_client = FakeRedis()
+                async with redis_client.pipeline(transaction=False) as pipe:
+                    await pipe.watch("a")
+                    await redis_client.set("a", "bad")
+                    pipe.multi()
+                    await pipe.set("a", "1")
+                    await pipe.execute()
+            except WatchError:
+                pass
```
chrisguidry, line 412: I'd recommend this instead, so that we ensure the `WatchError` is still raised

```suggestion
            with pytest.raises(WatchError):
                redis_client = FakeRedis()
                async with redis_client.pipeline(transaction=False) as pipe:
                    await pipe.watch("a")
                    await redis_client.set("a", "bad")
                    pipe.multi()
                    await pipe.set("a", "1")
                    await pipe.execute()
```

instrumentation/opentelemetry-instrumentation-redis/tests/test_redis.py
```diff
@@ -311,3 +316,113 @@ def test_attributes_unix_socket(self):
...
+            try:
+                redis_client = fakeredis.FakeStrictRedis()
+                pipe = redis_client.pipeline(transaction=True)
+                pipe.watch("a")
+                redis_client.set("a", "bad")  # This will cause the WatchError
+                pipe.multi()
+                pipe.set("a", "1")
+                pipe.execute()
+            except WatchError:
+                pass
```
chrisguidry, line 371:
```suggestion
            with pytest.raises(WatchError):
                redis_client = fakeredis.FakeStrictRedis()
                pipe = redis_client.pipeline(transaction=True)
                pipe.watch("a")
                redis_client.set("a", "bad")  # This will cause the WatchError
                pipe.multi()
                pipe.set("a", "1")
                pipe.execute()
```

xrmx:
@chrisguidry please open a new PR with the fix, you're commenting a PR already merged

PR #2672, opened 2024-07-07
Fix emitting logs when Logger propagation is disabled for a logger

oldium:
# Description

The logs are emited via `LoggingHandler` as a handler of the root logger. Unfortunatelly when there is a Logger instance with disabled propagation with `logger.propagate=False`, the root logger handler does not get called. Fix this by instrumenting `Logger.propagate` as a property attribute, so when the value changes to `False`, install the `LoggingHandler` to the logger automatically, and when the value changes to `True`, uninstall it.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Tested with unit tests. Tested with real project using `gunicorn` with `uvicorn` workers (`uvicorn.workers.UvicornWorker`) and `celery` task executor. Previously there were logs missing from gunicorn, uvicorn and celery tasks, now they are forwarded to OpenTelemetry collector.

- [x] No.

# Checklist:


oldium:
Added few more tests for user-added `LoggingHandler` - they are not touched by instrumentation.

oldium:
Updated changelog to refer to this pull request.

instrumentation/opentelemetry-instrumentation-logging/pyproject.toml
```diff
@@ -26,6 +26,7 @@ classifiers = [
 ]
 dependencies = [
   "opentelemetry-api ~= 1.12",
+  "opentelemetry-sdk ~= 1.12",
```
mariojonke, line 29: as per the spec it is not allowed for an instrumentation to take a dependency on the/any SDK. otherwise it wouldn't be possible to use the instrumentation with a different/3rd party SDK.
see: https://opentelemetry.io/docs/specs/otel/overview/#sdk
> Instrumentation authors MUST NOT directly reference any SDK package of any kind, only the API.

oldium: Thanks for review. I will try to deal with this via `LoggingProvider` then.

oldium: Hm. That would require to actually copy `LoggingHandler` from SDK. Do you have any idea on how to forward `logging.LogRecord` (i.e. not `opentelemetry._logs.LogRecord`) to OpenTelemetry without using SDK?

oldium: The one who is actually instrumenting the logging subsystem is SDK (it installs its `LoggingHandler` into root logger if `OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED` env variable is set), not the logging instrumentation. This seems to be wrong then. What about to move LoggingHandler to logging instrumentation?

oldium: Well, this also looks like a dead-end, SDK makes use of its own `LogRecord` within `LoggingHandler`, which extends API's `LogRecord` by a `resource` attribute. So please suggest how to deal with this.

oldium:
I have added a Proof-of-Concept commit to prove that `LoggingHandler` can be moved out of SDK:

 1. Root logger instrumentation is done in the logging instrumentation,
    not in the SDK.
 2. Logger instrumentation uses only `opentelemetry-api` package, nothing
    from the SDK.

There are several things that needs to be checked/finished:

 1. Ensure that resource attribute is added to `LogRecord` by SDK, when
    we are not able to do so in `LogginHandler` API.
 2. Fix SDK startup with `OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED`
    enabled for compatibility - run instrumentation manually (as the env
    var name suggests)?

Please suggest how to handle the logging instrumentation situation. I will be happy to do some implementation if necessary.

instrumentation/opentelemetry-instrumentation-logging/pyproject.toml
oldium: See POC commit for a proposed solution. Marking this conversation as resolved (for now).

lzchen:
@oldium 

A couple of questions as I seem to be confused about the issue and the proposed solution. 

1. Can you provide sample code in the description to show a logger that is configured in such a way that this bug is prevalent?
2. Why do we have to move the `LoggingHandler` out of the SDK? How does this address the original problem?
3. What does this issue have to do with the logging instrumentation? I though we were discussing the `LoggingHandler`?

oldium:
Any update, please? I can finish the PoC - just let me know if that is acceptable, I can prepare it. 😁

PR #2687, opened 2024-07-10
Generate workflow files with a Jinja template

### LINKED ISSUE #2686: Duplicate test running information in `tox.ini` and CI workflows
We have duplicate information about the tests we run in CI and in `tox.ini`.

For example, `tox.ini` specifies that a particular package is to be run using certain python versions:

```
    ; opentelemetry-instrumentation-asyncpg
    py3{8,9,10,11,12}-test-instrumentation-asyncpg
    ; ext-asyncpg intentionally excluded from pypy3
    lint-instrumentation-asyncpg
```
The equivalent information (not running this package with `pypy`) is stored in the Github workflow file:

```
        exclude:
          ...
          - python-version: pypy3
            package: "asyncpg"
```

There is nothing that guarantees these 2 sources of information match. This can lead to packages failing and not being detected in CI because they are not being run in CI at all, or to having python versions specified in `tox.ini` for a certain package that actually don't work but we don't realize that because we have excluded them in our workflow files.

ocelotl:
Fixes #2686

The idea behind this PR is to delete our current test and lint workflow files and replace them with automatically generated workflow files.

This automatically generated file gets its information directly from `tox.ini`, removing the duplication of test running information in `tox.ini` and in the Github workflow files.

To achieve this, I have added a Jinja2 template that is used to generate the new workflow file. The resulting workflow file does not use a Github matrix. This is intentional, a Github matrix only supports 256 jobs, which had forced us to have 2 almost identical workflow files. With this new approach we only need one huge file.

I also cleaned up the information that is displayed in the web browser, so jobs now have more user-friendly names.

I also added a check that makes sure that the automatically generated workflow file is up to date with every PR that modifies the automatically generated workflow files so that we don't miss adding any necessary changes to this file.

I also removed the step where we cache the `tox` environment. Since every `tox` run happens in a separate runner, I think there is no point in saving this cache.


tox.ini
```diff
@@ -349,7 +344,6 @@ envlist =
 
     ; opentelemetry-instrumentation-confluent-kafka
     py3{8,9,10,11,12}-test-instrumentation-confluent-kafka
-    pypy3-test-instrumentation-confluent-kafka
```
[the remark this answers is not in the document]
xrmx: And it looks like they need a specific version so fine to drop it.

tox.ini
```diff
@@ -171,7 +168,6 @@ envlist =
 
     ; opentelemetry-exporter-prometheus-remote-write
     py3{8,9,10,11,12}-test-exporter-prometheus-remote-write
```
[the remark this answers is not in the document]
xrmx: We can keep it with https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2692

.github/workflows/try_parser.py
```diff
@@ -0,0 +1,39 @@
+from pathlib import Path
+from tox.config.cli.parse import get_options
+from tox.session.state import State
+from tox.config.sets import CoreConfigSet
+from tox.config.source.tox_ini import ToxIni
+from collections import defaultdict
+from json import dumps
+
+tox_ini_path = Path(
+    "/home/tigre/github/ocelotl/opentelemetry-python-contrib/tox.ini"
```
xrmx, line 10: I think this will not work generally :sweat_smile:

aabmass:
Bit of nit/soft comment for consideration: it looks like you can use a github workflow to generate the matrix via JSON: https://www.kenmuse.com/blog/dynamic-build-matrices-in-github-actions/

This would avoid having a separate template with text munging and the script you added could instead output JSON. A lot of what the template does is "reimplementing" GHA features like variable interpolation and matrices. Any thoughts?

tox.ini
```diff
@@ -383,7 +379,9 @@ deps =
 
   ; FIXME: add coverage testing
   ; FIXME: add mypy testing
-allowlist_externals = sh
+allowlist_externals =
+  sh
+  sudo
```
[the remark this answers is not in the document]
ocelotl: Right, fixed

ocelotl:
> Bit of nit/soft comment for consideration: it looks like you can use a github workflow to generate the matrix via JSON: [kenmuse.com/blog/dynamic-build-matrices-in-github-actions](https://www.kenmuse.com/blog/dynamic-build-matrices-in-github-actions/)
> 
> This would avoid having a separate template with text munging and the script you added could instead output JSON. A lot of what the template does is "reimplementing" GHA features like variable interpolation and matrices. Any thoughts?

Ah, yes, I remembered this when I saw this uses `$GITHUB_OUTPUT`. This was actually my first approach. What I realized was that the matrix approach has a few disadvantages:

#. No control on job name: right now the jobs that show up in the browser are named something like **instrumentations-1 (py39, urllib3, ubuntu-20.04)**.

.github/workflows/generate_workflows_lib/hatch_build.py
```diff
@@ -0,0 +1,15 @@
...
+
+class CustomBuildHook(BuildHookInterface):
+
+    def initialize(self, version, build_data):
+
+        with open(
+            Path(__file__).parent.parent.parent.parent.joinpath("tox.ini")
+        ) as tox_ini_file_0:
+            with open(
+                Path(__file__).parent.joinpath("src/generate_workflows_lib/tox.ini"), "w"
```
ocelotl, line 13: This adds the contrib `tox.ini` file to the `generate_workflows_lib`. We use this `tox.ini` file [here](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2687/files#diff-732236323b5b142b6c9ebcf4f04ee39a55a6c0e3420c64f5e09d276a7c709cdeR251) in the `generate_contrib_workflows` function.

The `generate_contrib_workflows` function is called [in the core repo](https://github.com/open-telemetry/opentelemetry-python/pull/4041/files#diff-4776cd291072e79bc82b8eaea2f2dfac817ce69e8f3018d42c1ff68451d9beacR20).

By adding the `tox.ini` file from the contrib repo to the `generate_workflows_lib` package, the `generate_contrib_workflows` function has the information it needs to generate the contrib workflow file for the core repo. This saves us from having to clone the entire contrib repo just to get the contrib repo `tox.ini` file when running the core tests.

.github/workflows/generate_workflows_lib/src/generate_workflows_lib/test.yml.j2
```diff
@@ -0,0 +1,41 @@
...
+# This file is generated automatically by executing tox -e generate-workflows
+
+name: Test {{ file_number }}
+
+on:
+  push:
+    branches-ignore:
+    - 'release/*'
+  pull_request:
+env:
```
emdneto, line 11:
```suggestion

env:
```

.github/workflows/misc_0.yml
```diff
@@ -0,0 +1,125 @@
...
+
+name: Misc 0
+
+on:
+  push:
+    branches-ignore:
+    - 'release/*'
+  pull_request:
+env:
+  CORE_REPO_SHA: main
```
emdneto, line 12: I'm wondering if we need to specify a different SHA than the main. Do we need to update the template and generate all workflows again right?

.github/workflows/generate_workflows_lib/src/generate_workflows_lib/misc.yml.j2
```diff
@@ -0,0 +1,61 @@
...
+# This file is generated automatically by executing tox -e generate-workflows
+
+name: Misc {{ file_number }}
+
+on:
+  push:
+    branches-ignore:
+    - 'release/*'
+  pull_request:
+env:
```
emdneto, line 11:
```suggestion

env:
```

.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
```diff
@@ -0,0 +1,36 @@
...
+# This file is generated automatically by executing tox -e generate-workflows
+
+name: Lint {{ file_number }}
+
+on:
+  push:
+    branches-ignore:
+    - 'release/*'
+  pull_request:
+env:
```
emdneto, line 11:
```suggestion

env:
```

.github/workflows/generate_workflows_lib/src/generate_workflows_lib/contrib.yml.j2
```diff
@@ -0,0 +1,46 @@
...
+# This file is generated automatically by executing tox -e generate-workflows
+
+name: Contrib {{ file_number }}
+
+on:
+  push:
+    branches-ignore:
+    - 'release/*'
+  pull_request:
+env:
```
emdneto, line 11:
```suggestion

env:
```

.github/workflows/misc_0.yml
ocelotl: Yes, this is a valid concern I had forgottent to address. Not the template. We would need to manually change the SHA in the `.yml` file. I added a condition that will allow us to skip the `generate-workflows` test if there is a label that indicates so just like `Skip changelog` does.

xrmx (approved):
I am sympathetic to Aaron concerns regarding the obvious huge listing vs the compactness of the gh matrix but once we are generating the workflows if we see areas of improvements we can iterate on it.

emdneto (approved):

PR #2694, opened 2024-07-10
add sync and async test guide at contributing.md

### LINKED ISSUE #2684: add sync and async test guide at contributing.md

at contrib repo, usually the current test is sync tests, but we sometimes need to run async test. If we just add a async test in our current test class, the test will automatically skip the async test. This will lead to async test pass in anytime. 

an example could be this one:
Under redis instrumentation, I want to add an async test, to test if I run set under a watch process, what will happen. If I just add it under the class we already have, it will pass in any case.
```
class TestRedis(TestBase):
    def setUp(self):
        super().setUp()
        RedisInstrumentor().instrument(tracer_provider=self.tracer_provider)

    def tearDown(self):
        super().tearDown()
        RedisInstrumentor().uninstrument()

    async def redis_operations(self):
        r = redis.Redis()

        async with r.pipeline(transaction=False) as pipe:
            await pipe.watch("a")

            # Simulate a change by another client
            await r.set("a", "bad")

            pipe.multi()
            await pipe.set("a", "1")

            await pipe.execute()

    @pytest.mark.asyncio
    async def test_redis_operations(self):
        # Run the redis_operations function
        await self.redis_operations()

        # Retrieve the spans

        # Check that the last span has the correct properties
        assert spans[-1].status.status_code == trace.StatusCode.UNSET

        # Check for specific attributes in the span
        assert any(event.name == "WatchError" for event in spans[-1].events)
```

The solution of test issue is create an async test for it, and the new async test should iInherited from IsolatedAsyncioTestCase
we should add some guide at contributing guide, which could save much time for new rookie contributors(like myself) to debug

not sure if we should add some annotations at test file about this thing

### Additional Context

_No response_

Yes

zhihali:
# Description

fix #2684

## Type of change

# How Has This Been Tested?

do not need to

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -270,7 +270,7 @@ Below is a checklist of things to be mindful of when implementing a new instrume
 - Appropriate error handling
   - ex. <https://github.com/open-telemetry/opentelemetry-python-contrib/blob/2518a4ac07cb62ad6587dd8f6cbb5f8663a7e179/instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py#L220>
 - Isolate sync and async test
-  - For synchronous tests, the typical test class structure is `class TestXxxxXxxx(TestBase)`. However, if you want to write asynchronous tests, you should create a new class that inherits from `IsolatedAsyncioTestCase` instead. Adding asynchronous tests to a common test class can lead to tests passing without actually running, which can be misleading.
+  - For synchronous tests, the typical test case class is inherited from `opentelemetry.test.test_base.TestBase`. However, if you want to write asynchronous tests, the test case class should inherit also from `IsolatedAsyncioTestCase`. Adding asynchronous tests to a common test class can lead to tests passing without actually running, which can be misleading.
   - ex. <https://github.com/open-telemetry/opentelemetry-python-contrib/blob/60fb936b7e5371b3e5587074906c49fb873cbd76/instrumentation/opentelemetry-instrumentation-grpc/tests/test_aio_server_interceptor.py#L84>
 
 ## Expectations from contributors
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -269,6 +269,9 @@ Below is a checklist of things to be mindful of when implementing a new instrume
   - ex. <https://github.com/open-telemetry/opentelemetry-python-contrib/blob/2518a4ac07cb62ad6587dd8f6cbb5f8663a7e179/instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py#L234>
 - Appropriate error handling
   - ex. <https://github.com/open-telemetry/opentelemetry-python-contrib/blob/2518a4ac07cb62ad6587dd8f6cbb5f8663a7e179/instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py#L220>
+- Isolate sync and async test
+  - For synchronous tests, the typical test class structure is `class TestXxxxXxxx(TestBase)`. However, if you want to write asynchronous tests, you should create a new class that inherits from `IsolatedAsyncioTestCase` instead. Adding asynchronous tests to a common test class can lead to tests passing without actually running, which can be misleading.
```
xrmx, line 273:
```suggestion
  - For synchronous tests, the typical test case class is inherited from `opentelemetry.test.test_base.TestBase`. However, if you want to write asynchronous tests, the test case class should inherits also from `IsolatedAsyncioTestCase`. Adding asynchronous tests to a common test class can lead to tests passing without actually running, which can be misleading.
```

zhihali: Thanks!

emdneto (approved):

lzchen (approved):

PR #2702, opened 2024-07-12
Teach fastapi instrumentation about fastapi-slim

### LINKED ISSUE #2683: fastapi-slim support

FastAPI now includes a new [CLI dependency](https://github.com/tiangolo/fastapi/pull/11522) by default. A new package `fastapi-slim` was created which excludes the `fastapi-cli` dependency. We ideally want to use the new `fastapi-slim` package and the existing open telemetry fastapi instrumentation.

When trying to use this combination the open telemetry instrumentation fails to initialise because the expected package name is now different:
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/eb8e45695e88510a611d2cc04f14cdf56036968a/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/package.py#L16

```
'DependencyConflict: requested: "fastapi ~= 0.58" but found: "None"'
```

The open telemetry instrumentation for FastAPI should work with the `fastapi-slim` package and/or the `fastapi` package.

Can the FastAPI instrumentation be updated to work with either of the FastAPI packages or alternatively remove the check for the dependency completely?

### Additional Context

_No response_

None

zhihali:
# Description

Fixes #2683

## Type of change

# How Has This Been Tested?

test with running the fastapi locally with the updated code

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


zhihali:
This is a query for reviewers:

The current FastAPI instrumentation is set to `fastapi ~= 0.58.0`, meaning it supports only versions within the 0.58.X series. However, some other instruments, like HTTPX, allow for versions greater than a specified minimum, such as:

```
instruments = [
  "httpx >= 0.18.0",
]
```
In FastAPI, updating the specification to `fastapi >= 0.58.0` isn't straightforward due to the versioning changes outlined in the release history on [PyPI](https://pypi.org/project/fastapi/#history). FastAPI progressed from version 0.58 to 0.99, then jumped to 0.100.0, with the latest release being 0.111.0. In Python, 0.111.0 is numerically less than 0.58.0 😅.

Therefore, my questions are:

1. Does our FastAPI instrumentation currently only support versions in the 0.58.X series, excluding newer versions?
2. If not, should we consider changing the ~ operator to accommodate compatibility with newer versions of FastAPI?

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -265,6 +270,11 @@ def uninstrument_app(app: fastapi.FastAPI):
         app._is_instrumented_by_opentelemetry = False
 
     def instrumentation_dependencies(self) -> Collection[str]:
+        if find_spec("fastapi") is not None:
+            return (_fastapi,)
+        if find_spec("fastapi_slim") is not None:
+            return (_fastapi_slim,)
+        # If neither is installed, return both as potential dependencies
```
xrmx, line 277: Do you really need this?

instrumentation/opentelemetry-instrumentation-fastapi/pyproject.toml
```diff
@@ -34,9 +34,14 @@ dependencies = [
...
 [project.optional-dependencies]
 instruments = [
-  "fastapi ~= 0.58",
+    "fastapi >= 0.58.0",
+    "fastapi-slim ~= 0.111.0",
 ]
 
+[instruments-all]
+fastapi = ">= 0.58.0"
+fastapi-slim = "~ 0.111.0"
```
xrmx, line 43: Is this used at all?

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
zhihali: you are right, without this, the issue could be solved as well.

instrumentation/opentelemetry-instrumentation-fastapi/pyproject.toml
zhihali: you are right, without this, the issue could be solved as well.

zhihali:
@xrmx Thanks for the review, always appreciate your quick reply and guidance!

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
xrmx: Ah right, dependencies are OR'ed with the opentelemetry-bootstrap but are AND'ed when instrumenting :sweat:

xrmx (commented):
LGTM, I would just add a test requirements file with fastapi-slim installed so we check that we support that

instrumentation/README.md
```diff
@@ -19,7 +19,7 @@
 | [opentelemetry-instrumentation-django](./opentelemetry-instrumentation-django) | django >= 1.10 | Yes | experimental
 | [opentelemetry-instrumentation-elasticsearch](./opentelemetry-instrumentation-elasticsearch) | elasticsearch >= 6.0 | No | experimental
 | [opentelemetry-instrumentation-falcon](./opentelemetry-instrumentation-falcon) | falcon >= 1.4.1, < 4.0.0 | Yes | experimental
-| [opentelemetry-instrumentation-fastapi](./opentelemetry-instrumentation-fastapi) | fastapi ~= 0.58 | Yes | migration
+| [opentelemetry-instrumentation-fastapi](./opentelemetry-instrumentation-fastapi) | fastapi >= 0.58.0,fastapi-slim ~= 0.111.0 | Yes | experimental
```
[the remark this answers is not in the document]
zhihali: yeap, I forget to run the generate so lead to the check fail, thanks

tox.ini
```diff
@@ -548,6 +549,12 @@ commands_pre =
   fastapi: pip install opentelemetry-sdk@{env:CORE_REPO}\#egg=opentelemetry-sdk&subdirectory=opentelemetry-sdk
   fastapi: pip install opentelemetry-test-utils@{env:CORE_REPO}\#egg=opentelemetry-test-utils&subdirectory=tests/opentelemetry-test-utils
   fastapi: pip install -r {toxinidir}/instrumentation/opentelemetry-instrumentation-fastapi/test-requirements.txt
+  fastapi-slim: pip install opentelemetry-api@{env:CORE_REPO}\#egg=opentelemetry-api&subdirectory=opentelemetry-api
+  fastapi-slim: pip install opentelemetry-semantic-conventions@{env:CORE_REPO}\#egg=opentelemetry-semantic-conventions&subdirectory=opentelemetry-semantic-conventions
+  fastapi-slim: pip install opentelemetry-sdk@{env:CORE_REPO}\#egg=opentelemetry-sdk&subdirectory=opentelemetry-sdk
+  fastapi-slim: pip install opentelemetry-test-utils@{env:CORE_REPO}\#egg=opentelemetry-test-utils&subdirectory=tests/opentelemetry-test-utils
+  fastapi-slim: pip install -r {toxinidir}/instrumentation/opentelemetry-instrumentation-fastapi/test-requirements-slim.txt
```
emdneto, line 556: Are we missing running this on CI?

xrmx:
I would not add new tox environment, just add a test requirement for the
current fastapi one

Il lun 15 lug 2024, 16:12 Emídio Neto ***@***.***> ha scritto:

instrumentation/opentelemetry-instrumentation-fastapi/pyproject.toml
```diff
@@ -34,7 +34,8 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
-  "fastapi ~= 0.58",
+    "fastapi >= 0.58.0",
```
lzchen, line 37: I might've missed the discussion on this but why change from ~= to >=?

zhihali: Ah, after reading the doc and my commit history, I found I wrongly changed `fastapi ~= 0.58` to `fastapi ~= 0.58.0`, which lead to tox test fail. Then I thought the 0.58.0 is wrong, bcz it just support 0.58.X. So I changed it to `>=0.58.0`.

But in fact, `~=0.58` could support any version lower than 1.0, and current newest fastapi version is 0.111.1. So for now, `~=0.58` and `>=0.58.0` both works.  

I think to be more cautious and ensure compatibility within a known range, we should keep `~=0.58`

zhihali: tks for pointing this out!

zhihali:
> I would not add new tox environment, just add a test requirement for the current fastapi one

Hey @xrmx, I think we should create a new tox environment for this. The issue is that FastAPI-Slim uses the same import statement as FastAPI (import fastapi). This means we can't just add a new test file under the current FastAPI environment and expect it to work correctly with FastAPI-Slim.

To properly test FastAPI-Slim, we need to:

1. Create a new tox environment.
2. Update the requirements.txt to include FastAPI-Slim but not FastAPI.
3. Run the FastAPI instrumentation tests in this new environment.
We should also add this new environment to our CI pipeline.

xrmx:
@zhihali So the issue is that fastapi and fastapi-slim does not conflict with each other and so installing one will not remove the other. Right?

tox.ini
zhihali: yes, added. Thanks to point this our

zhihali:
> @zhihali So the issue is that fastapi and fastapi-slim does not conflict with each other and so installing one will not remove the other. Right?

@xrmx 
So, the original issue is some user would like to use the fastapi-slim, and use the otel-fastapi-instrumentation, but our instrumentation doesn't support it. I add the support for fastapi-slim. This should solve the issue. In this case, I supporse add a new test env and a CI for fastapi-slim is reasonable.

But for now, I think we solve the OR problem, for AND, this is a bit complex.
```
Python's import system will typically use the package that was installed last or the one that appears first in the Python path.
```
So it's defo not recommand to use both fastapi and fastapi-slim(since the import all be `import fastapi`). I think the current change will support AND, but it's not recommand to use it.

What's your opinion? Any further discussion is welcomed!

xrmx:
> > @zhihali So the issue is that fastapi and fastapi-slim does not conflict with each other and so installing one will not remove the other. Right?
> 
> @xrmx So, the original issue is some user would like to use the fastapi-slim, and use the otel-fastapi-instrumentation, but our instrumentation doesn't support it. I add the support for fastapi-slim. This should solve the issue. In this case, I supporse add a new test env and a CI for fastapi-slim is reasonable.
> 
> But for now, I think we solve the OR problem, for AND, this is a bit complex.
> 
> ```
> Python's import system will typically use the package that was installed last or the one that appears first in the Python path.
> ```
> 
> So it's defo not recommand to use both fastapi and fastapi-slim(since the import all be `import fastapi`). I think the current change will support AND, but it's not recommand to use it.
> 
> What's your opinion? Any further discussion is welcomed!

I was talking specifically to the tox environment, sorry for not being more clear. PR is fine for me

tox.ini
```diff
@@ -116,6 +116,7 @@ envlist =
 
     ; opentelemetry-instrumentation-fastapi
     py3{8,9,10,11,12}-test-instrumentation-fastapi
+    py3{8,9,10,11,12}-test-instrumentation-fastapi-slim
     pypy3-test-instrumentation-fastapi
```
xrmx, line 120:
```suggestion
    pypy3-test-instrumentation-fastapi
    pypy3-test-instrumentation-fastapi-slim
```

tox.ini
```diff
@@ -544,6 +545,12 @@ commands_pre =
   fastapi: pip install opentelemetry-sdk@{env:CORE_REPO}\#egg=opentelemetry-sdk&subdirectory=opentelemetry-sdk
   fastapi: pip install opentelemetry-test-utils@{env:CORE_REPO}\#egg=opentelemetry-test-utils&subdirectory=tests/opentelemetry-test-utils
   fastapi: pip install -r {toxinidir}/instrumentation/opentelemetry-instrumentation-fastapi/test-requirements.txt
+  fastapi-slim: pip install opentelemetry-api@{env:CORE_REPO}\#egg=opentelemetry-api&subdirectory=opentelemetry-api
+  fastapi-slim: pip install opentelemetry-semantic-conventions@{env:CORE_REPO}\#egg=opentelemetry-semantic-conventions&subdirectory=opentelemetry-semantic-conventions
+  fastapi-slim: pip install opentelemetry-sdk@{env:CORE_REPO}\#egg=opentelemetry-sdk&subdirectory=opentelemetry-sdk
+  fastapi-slim: pip install opentelemetry-test-utils@{env:CORE_REPO}\#egg=opentelemetry-test-utils&subdirectory=tests/opentelemetry-test-utils
+  fastapi-slim: pip install -r {toxinidir}/instrumentation/opentelemetry-instrumentation-fastapi/test-requirements-slim.txt
+
```
xrmx, line 553:
```suggestion
```

.github/workflows/instrumentations_1.yml
```diff
@@ -38,10 +38,13 @@ jobs:
           - "resource-detector-azure"
           - "resource-detector-container"
           - "util-http"
+          - "fastapi-slim"
         os: [ubuntu-20.04]
         exclude:
           - python-version: pypy3
             package: "prometheus-remote-write"
+          - python-version: pypy3
```
emdneto, line 46: It doesn't work with pypy?

emdneto: Could you please try if it works with the changes @xrmx suggested?

zhihali: yeah, I did and it works.

xrmx (approved):

CHANGELOG.md
```diff
@@ -57,6 +57,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#2726](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2726))
 
 ### Fixed
+
+- `opentelemetry-instrumentation-fastapi` Add dependency support for fastapi-slim
```
emdneto, line 61: This should be moved to the added section, no?

emdneto (approved):

PR #2708, opened 2024-07-15
Deprecate sklearn instrumentation

### LINKED ISSUE #2176: Drop sklearn instrumentation
We should no longer publish the [sklearn](https://pypi.org/project/sklearn/) package sice it has been deprecated.

ocelotl:
Fixes #2176


xrmx:
This is more like remove than deprecate, I was thinking that we may drop it when we drop python 3.8 maybe?

ocelotl:
> This is more like remove than deprecate, I was thinking that we may drop it when we remove support for python 3.8 maybe?

This is deprecation. We no longer will publish it, but the sklearn instrumentation packages that have been published so far will continue to exist. Supporting 3.8 or not is a topic not related to this PR, the reason we want to deprecate this instrumentation is because the `sklearn` package is [deprecated](https://pypi.org/project/sklearn/):

> Starting 2023 December 1st, trying to install the sklearn PyPI package will raise an error.

lzchen (approved):

lzchen:
> > This is more like remove than deprecate, I was thinking that we may drop it when we remove support for python 3.8 maybe?
> 
> This is deprecation. We no longer will publish it, but the sklearn instrumentation packages that have been published so far will continue to exist. Supporting 3.8 or not is a topic unrelated to this PR, the reason we want to deprecate this instrumentation is because the `sklearn` package is [deprecated](https://pypi.org/project/sklearn/):
> 
> > Starting 2023 December 1st, trying to install the sklearn PyPI package will raise an error.

@ocelotl 

Might want to include this in the CHANGELOG.md that this is deprecated at least.

@xrmx 

We do have precedence of simply deleting packages when they are being deprecated, although, we should probably be more transparent to users. https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1366

emdneto (approved):
Should we remove the package from the CI in the core repo?

xrmx:
> @xrmx
> 
> We do have precedence of simply deleting packages when they are being deprecated, although, we should probably be more transparent to users. #1366

You are breaking current opentelemetry-bootstrap users no? It won't install the package anymore.

ocelotl:
> @ocelotl
> 
> Might want to include this in the CHANGELOG.md that this is deprecated at least.

It is already added.

ocelotl:
> Should we remove the package from the CI in the core repo?

I'll do that :v:

PR #2711, opened 2024-07-16
feat(urllib3)!: refactor request hook parameters

eigenein:
# Description

Currently, in the `urllib3` instrumentation, there is no way for a request hook to inspect HTTP method and URL. This change implements forwarding the new parameters to request hooks.

Additionally, the documentation has been updated: both to represent the changes from this PR, and to fix the incorrect hook signatures.

## Type of change

# How Has This Been Tested?

The unit tests have been updated to verify moving the existing `body` parameter and the newly added parameters.

- [ ] Yes.
- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#2726](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2726))
 - `opentelemetry-instrumentation-fastapi` Add dependency support for fastapi-slim
   ([#2702](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2702))
+- `opentelemetry-instrumentation-urllib3` improve request_hook, replacing `headers` and `body` parameters with a single `request_info: RequestInfo` parameter that now contains the `method` and `url` ([#2711](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2711))
 
 ### Fixed
 
--- a/instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
@@ -47,14 +47,19 @@ The hooks can be configured as follows:
 
 .. code:: python
 
-    # `request` is an instance of urllib3.connectionpool.HTTPConnectionPool
-    def request_hook(span, request):
-        pass
+    def request_hook(
+        span: Span,
+        pool: urllib3.connectionpool.HTTPConnectionPool,
+        request_info: RequestInfo,
+    ) -> Any:
+        ...
 
-    # `request` is an instance of urllib3.connectionpool.HTTPConnectionPool
-    # `response` is an instance of urllib3.response.HTTPResponse
-    def response_hook(span, request, response):
-        pass
+    def response_hook(
+        span: Span,
+        pool: urllib3.connectionpool.HTTPConnectionPool,
+        response: urllib3.response.HTTPResponse,
+    ) -> Any:
+        ...
 
     URLLib3Instrumentor().instrument(
         request_hook=request_hook, response_hook=response_hook
@@ -81,6 +86,7 @@ API
 import collections.abc
 import io
 import typing
+from dataclasses import dataclass
 from timeit import default_timer
 from typing import Collection
 
@@ -135,14 +141,29 @@ from opentelemetry.util.http.httplib import set_ip_on_next_http_connection
 
 _excluded_urls_from_env = get_excluded_urls("URLLIB3")
 
+
+@dataclass
+class RequestInfo:
+    """Arguments that were passed to the ``urlopen()`` call."""
+
+    __slots__ = ("method", "url", "headers", "body")
+
+    # The type annotations here come from ``HTTPConnectionPool.urlopen()``.
+    method: str
+    url: str
+    headers: typing.Optional[typing.Mapping[str, str]]
+    body: typing.Union[
+        bytes, typing.IO[typing.Any], typing.Iterable[bytes], str, None
+    ]
+
+
 _UrlFilterT = typing.Optional[typing.Callable[[str], str]]
 _RequestHookT = typing.Optional[
     typing.Callable[
         [
             Span,
             urllib3.connectionpool.HTTPConnectionPool,
-            typing.Dict,
-            typing.Optional[str],
+            RequestInfo,
         ],
         None,
     ]
@@ -317,7 +338,16 @@ def _instrument(
             span_name, kind=SpanKind.CLIENT, attributes=span_attributes
         ) as span, set_ip_on_next_http_connection(span):
             if callable(request_hook):
-                request_hook(span, instance, headers, body)
+                request_hook(
+                    span,
+                    instance,
+                    RequestInfo(
+                        method=method,
+                        url=url,
+                        headers=headers,
+                        body=body,
+                    ),
+                )
             inject(headers)
             # TODO: add error handling to also set exception `error.type` in new semconv
             with suppress_http_instrumentation():
--- a/instrumentation/opentelemetry-instrumentation-urllib3/tests/test_urllib3_integration.py
+++ b/instrumentation/opentelemetry-instrumentation-urllib3/tests/test_urllib3_integration.py
@@ -27,7 +27,10 @@ from opentelemetry.instrumentation._semconv import (
     _HTTPStabilityMode,
     _OpenTelemetrySemanticConventionStability,
 )
-from opentelemetry.instrumentation.urllib3 import URLLib3Instrumentor
+from opentelemetry.instrumentation.urllib3 import (
+    RequestInfo,
+    URLLib3Instrumentor,
+)
 from opentelemetry.instrumentation.utils import (
     suppress_http_instrumentation,
     suppress_instrumentation,
@@ -42,6 +45,7 @@ from opentelemetry.semconv.attributes.url_attributes import URL_FULL
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.test.mock_textmap import MockTextMapPropagator
 from opentelemetry.test.test_base import TestBase
+from opentelemetry.trace import Span
 from opentelemetry.util.http import get_excluded_urls
 
 # pylint: disable=too-many-public-methods
@@ -521,10 +525,10 @@ class TestURLLib3Instrumentor(TestBase):
         self.assert_success_span(response, self.HTTP_URL)
 
     def test_hooks(self):
-        def request_hook(span, request, body, headers):
+        def request_hook(span, pool, request_info):
             span.update_name("name set from hook")
 
-        def response_hook(span, request, response):
+        def response_hook(span, pool, response):
             span.set_attribute("response_hook_attr", "value")
 
         URLLib3Instrumentor().uninstrument()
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -38,8 +38,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#2580](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2580))
 - Populate `{method}` as `HTTP` on `_OTHER` methods from scope for `asgi` middleware
   ([#2610](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2610))
-  - Populate `{method}` as `HTTP` on `_OTHER` methods from scope for `fastapi` middleware
-  ([#2682](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2682))
+- Populate `{method}` as `HTTP` on `_OTHER` methods from scope for `fastapi` middleware
```
eigenein, line 41: Note: incorrect indentation

instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
```diff
@@ -47,13 +47,21 @@ def strip_query_params(url: str) -> str:
 
 .. code:: python
 
-    # `request` is an instance of urllib3.connectionpool.HTTPConnectionPool
-    def request_hook(span, request):
```
eigenein, line 51: Note: the previous signature actually was: `def request_hook(span, instance, body)`

instrumentation/opentelemetry-instrumentation-urllib3/tests/test_urllib3_integration.py
```diff
@@ -358,6 +360,8 @@ def request_hook(span, request, headers, body):
 
         span = self.assert_span()
 
+        self.assertEqual(span.attributes["request_hook_method"], "POST")
+        self.assertEqual(span.attributes["request_hook_url"], "http://mock/status/200")
```
eigenein, line 364: Here I verify that it's actually working as expected

eigenein:
There are failed checks in `aiopg` integration that I haven't touched. Something's wrong with its `test-requirements.txt`:

```
ERROR: Could not find a version that satisfies the requirement install==1.3.5 (from versions: none)
ERROR: No matching distribution found for install==1.3.5
```

Perhaps, it should be `pip-install==1.3.5`?

emdneto:
> There are failed checks in `aiopg` integration that I haven't touched. Something's wrong with its `test-requirements.txt`:
> 
> ```
> ERROR: Could not find a version that satisfies the requirement install==1.3.5 (from versions: none)
> ERROR: No matching distribution found for install==1.3.5
> ```
> 
> Perhaps, it should be `pip-install==1.3.5`?

Fix for this is here See #2712

instrumentation/opentelemetry-instrumentation-urllib3/tests/test_urllib3_integration.py
```diff
@@ -358,6 +360,10 @@ def request_hook(span, request, headers, body):
 
         span = self.assert_span()
 
+        self.assertEqual(span.attributes["request_hook_method"], "POST")
+        self.assertEqual(
+            span.attributes["request_hook_url"], "http://mock/status/200"
+        )
```
eigenein, line 366: Here I verify that it's actually working as expected

CHANGELOG.md
```diff
@@ -44,8 +44,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+([#2682](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2682))
+- `urllib3`: add `method` and `url` parameters to the request hook
```
lzchen, line 49:
```suggestion
- `urllib3`: add `method` and `url` parameters to the request hook ([#2711](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2711))
```

instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
lzchen: I'm not too familiar with the inner workings of `urllib3` but does `HTTPConnectionPool` not have any of this information inside it (method, url, headers, body)? Would there be fields in it that would be useful, and if not, maybe we shouldn't even be including as part of the hooks. Most other instrumentations accept a `request` object of some sort which contains all information pertaining to the request itself.

CHANGELOG.md
lzchen, line 49: Nit: If you simply say "add" it wouldn't be a breaking change. I would just say "refactor to include `method` and `url`.

instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
eigenein: > I'm not too familiar with the inner workings of `urllib3` but does `HTTPConnectionPool` not have any of this information inside it (method, url, headers, body)?

Not really, [`HTTPConnectionPool`](https://urllib3.readthedocs.io/en/stable/reference/urllib3.connectionpool.html#urllib3.connectionpool.ConnectionPool) is just it, a connection pool. The instrumentation wraps [`urlopen`](https://urllib3.readthedocs.io/en/stable/reference/urllib3.connectionpool.html#urllib3.HTTPConnectionPool.urlopen) that accepts request parameters and uses a connection from the pool to send the request

> Would there be fields in it that would be useful, and if not, maybe we shouldn't even be including as part of the hooks

Indeed, it would make sense to me to remove the pool from the hook parameters. I just didn't want to do it without knowing why it's there in the first place, but I'm totally fine with removing it

> Most other instrumentations accept a `request` object of some sort which contains all information pertaining to the request itself

Indeed, it's just that the `urlopen` receives all this info as normal function parameters, and whoever made the initial implementation probably just forwarded them to the hook

How about this: I'd introduce our own `Request` class, put there the new `method` and `url`, and move the existing `headers` and `body` in there:

```python
@dataclass
class Request:
    method: str
    url: str
    headers: Mapping[str, str] | None
    body: bytes | IO[Any] | Iterable[bytes] | str | None
```

And then, the request hook becomes just `def request_hook(span: Span, request: Request)` with the possibility to extend `Request` in future without making it a breaking change. Thoughts?

emdneto: @eigenein we are doing something similar in httpx-instrumentation if you want take a look

lzchen: > And then, the request hook becomes just def request_hook(span: Span, request: Request) with the possibility to extend Request in future without making it a breaking change. Thoughts?

This sounds like a good approach to me. It would also be good to make the class exposed in the docs very clearly so users will know how to construct one fairly easily.

eigenein:
@lzchen Done and rebased, ready for review

I decided to keep the `pool` parameter because of [this note](https://urllib3.readthedocs.io/en/stable/reference/urllib3.connectionpool.html#urllib3.HTTPConnectionPool.urlopen):

> **headers** ([Mapping](https://docs.python.org/3/library/typing.html#typing.Mapping)[[str](https://docs.python.org/3/library/stdtypes.html#str), [str](https://docs.python.org/3/library/stdtypes.html#str)] | None) – Dictionary of custom headers to send, such as User-Agent, If-None-Match, etc. _**If None, pool headers are used.**_ If provided, these headers completely replace any pool-specific headers.

So, the hook may still want to access `pool.headers`

instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
```diff
@@ -47,14 +47,19 @@ def strip_query_params(url: str) -> str:
...
-        pass
-
-    # `request` is an instance of urllib3.connectionpool.HTTPConnectionPool
-    # `response` is an instance of urllib3.response.HTTPResponse
-    def response_hook(span, request, response):
-        pass
+    def request_hook(
+        span: Span,
+        pool: urllib3.connectionpool.HTTPConnectionPool,
+        request_info: RequestInfo,
```
lzchen, line 53: Nit: Anyway to include a link to the definition or include the fields necessary so it's easier for users to create one.

lzchen (approved):

CHANGELOG.md
```diff
@@ -46,8 +46,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#2580](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2580))
 - Populate `{method}` as `HTTP` on `_OTHER` methods from scope for `asgi` middleware
   ([#2610](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2610))
-  - Populate `{method}` as `HTTP` on `_OTHER` methods from scope for `fastapi` middleware
-  ([#2682](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2682))
+- Populate `{method}` as `HTTP` on `_OTHER` methods from scope for `fastapi` middleware
+([#2682](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2682))
+- `urllib3`: the request hook has been refactored: the `headers` and `body` parameters have been replaced by a single `request_info: RequestInfo` parameter that now contains `method` and `url` as well ([#2711](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2711))
```
emdneto, line 51:
```suggestion
- `opentelemetry-instrumentation-urllib3` improve request_hook, replacing `headers` and `body` parameters with a single `request_info: RequestInfo` parameter that now contains the `method` and `url` ([#2711](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2711))
```

instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
eigenein: Could you elaborate on why users would need to create one? In the current form, it's used to pass the info from the instrumentation down to the hook, the user code wouldn't need to create one in the first place

emdneto: I think @lzchen is talking about cases where users just copy and paste the example from the documentation; see https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/urllib3/urllib3.html. If we are providing the example with the type would be nice to have a reference to it, or just omit it from the example

eigenein: I believe, all the public members should be automagically rendered in the API section, right? And then, I thought, having the type hint would be beneficial. I'm not sure if it's even possible to make this type hint inside the snippet a proper link

lzchen: > the user code wouldn't need to create one in the first place

Apologies, instead of "create" I meant "they need to know what a `RequestInfo` consists of" for them to use it properly in their hook. As @emdneto mentioned, if a user were to just look at the doc example and copy and paste it, it would be convenient to immediately have the class definition so as to know how to use it. A simple comment would suffice. However, not a blocker.

eigenein:
Alright, I'm keeping the docs "as is". Adjusted the changelog and rebased. Ready for merge

PR #2715, opened 2024-07-16
HTTP semantic convention stability migration for urllib3

### LINKED ISSUE #2681: opentelemetry-instrumentation-urllib3: HTTP semantic convention stability migration
 
HTTP semantic convention stability migration for `opentelemetry-instrumentation-urllib3`. Here is the [plan](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/migration-guide.md)

Implement the semconv opt-in feature for urllib3

_No response_

### Additional Context

Part of this https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2453

None

emdneto:
# Description
Part of #2453 
Fixes #2681

# Checklist
- [X] Add semconv status as `migration` for urllib3 in instrumentations README
- [X] Populate `{method}` as `HTTP` when nonstandard method (sanitize_method)
- [X]  set request HTTP method attribute as `_OTHER` when nonstandard http method and for new semconv also set `http.request.method_original`  with the original nonstandard method.
- [X] Set schema_url based on the opt-in mode
- [X] Create histograms based on the opt-in mode (different metric names so no dup attributes)
- [X] Set metrics attributes based on the opt-in mode
- [X] Set span attributes based on the opt-in mode

At the current implementation, some of the required/recommended semconv span attributes aren't being set:
* `server.address` , `server.port`
* `network.protocol.verison`

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
@@ -304,8 +304,7 @@ def _instrument(
 
         span_name = _get_span_name(method)
         span_attributes = {}
-        # If the HTTP request method is not known to instrumentation,
-        # it MUST set the http.request.method attribute to _OTHER.
+
         _set_http_method(
             span_attributes,
             method,
@@ -320,13 +319,11 @@ def _instrument(
             if callable(request_hook):
                 request_hook(span, instance, headers, body)
             inject(headers)
-
             # TODO: add error handling to set `error.type` in new semconv
             with suppress_http_instrumentation():
                 start_time = default_timer()
                 response = wrapped(*args, **kwargs)
                 duration_s = default_timer() - start_time
-
             # set http status code based on semconv
             metric_attributes = {}
             _set_status_code_attribute(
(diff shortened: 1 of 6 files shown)


lzchen:
> At the current implementation, some of the required/recommended semconv span attributes aren't being set:

server.address , server.port
network.protocol.verison

Can probably create a separate issue for these.

instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
```diff
@@ -233,44 +302,60 @@ def instrumented_urlopen(wrapped, instance, args, kwargs):
...
+        _set_http_url(span_attributes, url, sem_conv_opt_in_mode)
 
         with tracer.start_as_current_span(
             span_name, kind=SpanKind.CLIENT, attributes=span_attributes
         ) as span, set_ip_on_next_http_connection(span):
             if callable(request_hook):
                 request_hook(span, instance, headers, body)
             inject(headers)
-
+            # TODO: add error handling to also set exception `error.type` in new semconv
```
lzchen, line 322: Maybe create a tracking issue?

lzchen (approved):

instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
emdneto: Sure. Will do that. We also have to do that for other instrumentations

PR #2719, opened 2024-07-17
add `error.type` to asgi-instrumentation attributes

emdneto:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?
- [X] tox
-  
- [X] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
@@ -549,23 +549,18 @@ class OpenTelemetryMiddleware:
         sem_conv_opt_in_mode = _OpenTelemetrySemanticConventionStability._get_opentelemetry_stability_opt_in_mode(
             _OpenTelemetryStabilitySignalType.HTTP,
         )
+        schema_url = _get_schema_url(sem_conv_opt_in_mode)
         self.app = guarantee_single_callable(app)
         self.tracer = (
             trace.get_tracer(
-                __name__,
-                __version__,
-                tracer_provider,
-                schema_url=_get_schema_url(sem_conv_opt_in_mode),
+                __name__, __version__, tracer_provider, schema_url=schema_url
             )
             if tracer is None
             else tracer
         )
         self.meter = (
             get_meter(
-                __name__,
-                __version__,
-                meter_provider,
-                schema_url=_get_schema_url(sem_conv_opt_in_mode),
+                __name__, __version__, meter_provider, schema_url=schema_url
             )
             if meter is None
             else meter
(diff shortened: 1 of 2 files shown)


instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -729,7 +725,20 @@ async def __call__(
                 )
 
                 await self.app(scope, otel_receive, otel_send)
+        except Exception as exc:  # pylint: disable=broad-except
+            exception = exc
         finally:
+            if exception is not None and _report_new(
```
pmcollins, line 731: Just a style thing, but I'm wondering if it would be worthwhile to break this method into functions for readability. Perhaps those functions could stand to be further modularized, but this would be a step in that direction. Something like
```
        try:
            await self.instrument_app(attributes, receive, scope, send, span, span_name)
        except Exception as exc:  # pylint: disable=broad-except
            exception = exc
        finally:
            await self.cleanup(active_requests_count_attrs, attributes, exception, scope, span, start, token)
```
Not a blocking comment, just a suggestion while we're in here.

lzchen: Another suggestion would also to wait and implement this for the various instrumentations that still require this (like flask, wsgi (for metrics), django, etc) and see if there are any commonalities we can refactor out. Difficult to say what we can abstract/modularize without seeing more implementation.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -729,7 +725,20 @@ async def __call__(
...
+                self._sem_conv_opt_in_mode
+            ):
+                _set_status(
+                    span=span,
+                    metrics_attributes=attributes,
+                    status_code=-1,
```
lzchen, line 737: The `-1` status code is a bit awkward. The current change is fine but I would like to brainstorm some ways we can abstract all of this logic out somehow. The following are probably required in some way or another and should be shared across all instrumentations (especially wsgi and asgi) : 1. Setting status code on span 2. Setting error attributes on span if recording and if status code is erroneous 3. Setting error attributes on metrics attributes if status code is erroneous 4. Setting error attributes on span/metrics in general.

emdneto: I wouldn't say I like it either. We are actually doing that in several places where we try to int the status_code, and if it fails, we set status_code as -1. Let me see what can I do here to improve the way we are doing this.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
emdneto: Agreed. Indeed, while reviewing the transition PRs I identified several instrumentations we are not handling exceptions and setting the exception in `error.type` even for the ones that are already migrated to the new semconv. I will mark this as draft for now while I think in a better way to handle this. Maybe we'll have a better idea on the commonalities after more transitions.

PR #2722, opened 2024-07-18
Re-add `opentelemetry-instrumentation-aiohttp-server` to release script

lzchen:
Due to the namespace of this package being released, we have successfully reclaimed the package name under `opentelemetry`.
Readding back into release script to finally release.


tammy-baylis-swi (approved):
Yay!! 🎆

ocelotl (approved):

PR #2726, opened 2024-07-22
Fix http clients method attribute in case of non standard http methods

emdneto:
# Description
We have the http.method attribute in spans for non-standard HTTP methods as HTTP instead of _OTHER . I identified instrumentations affected are requests, aiohttp-client and httpx

Right now, he have this in case of requests using nonstandard http methods, will generate span:
* Span Name: HTTP
* `http.method`: "HTTP"

And we must have:
* Span Name: HTTP
* `http.method`: "_OTHER"

For new semconv we must have in case of a request with method="NONSTANDARD":
* Span Name: HTTP
* `http.request.method`: "_OTHER"
* `http.request.method_original`: "NONSTANDARD"

See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/http.md and https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#name

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-aiohttp-client/src/opentelemetry/instrumentation/aiohttp_client/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-aiohttp-client/src/opentelemetry/instrumentation/aiohttp_client/__init__.py
@@ -132,10 +132,9 @@ _ResponseHookT = typing.Optional[
 
 
 def _get_span_name(method: str) -> str:
-    method = sanitize_method(method.upper().strip())
+    method = sanitize_method(method.strip())
     if method == "_OTHER":
         method = "HTTP"
-
     return method
 
 
@@ -230,8 +229,8 @@ def create_trace_config(
             trace_config_ctx.span = None
             return
 
-        http_method = params.method
-        request_span_name = _get_span_name(http_method)
+        method = params.method
+        request_span_name = _get_span_name(method)
         request_url = (
             remove_url_credentials(trace_config_ctx.url_filter(params.url))
             if callable(trace_config_ctx.url_filter)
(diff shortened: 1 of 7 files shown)


instrumentation/opentelemetry-instrumentation-aiohttp-client/src/opentelemetry/instrumentation/aiohttp_client/__init__.py
```diff
@@ -230,8 +229,8 @@ async def on_request_start(
             trace_config_ctx.span = None
             return
 
-        http_method = params.method
-        request_span_name = _get_span_name(http_method)
+        method = params.method
+        request_span_name = _get_span_name(method)
```
lzchen, line 233: Probably don't need this anymore.

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -188,13 +188,19 @@ def get_or_create_headers():
```
lzchen, line 188: Probably want to also remove `upper()` call in `get_default_span_name()` to be consistent.

lzchen, line 188: Should we also remove `upper()` call in `get_default_span_name` to be consistent?

instrumentation/opentelemetry-instrumentation-aiohttp-client/src/opentelemetry/instrumentation/aiohttp_client/__init__.py
emdneto: `request_span_name` is used as span name to start span

lzchen (approved):

PR #2733, opened 2024-07-24
allow a user-defined function for providing span attributes on engine/connect span begin

jochs:
# Description

Add `attr_provider` as an optional kwarg for `SQLAlchemyInstrumentor().instrument`. If attr_provider (a 0 argument callable) is provided, it will be evaluated before any calls to start_current_span within the engine or connection instrumentation. This allows users who have an attribute aware sampler to inject attributes into the sampler to control sample behavior.

Specifically, for my usecase, I have a subclass of the ParentBasedTraceIdRatio called OverrideableParentBasedTraceIdRatio which allows passing of an 'x-ignore-sample' attribute which forces the sample to emit a DROP decision when the parent is not already sampled. I would like to force all SqlAlchemy spans to be dropped unless the parent is sampled, which necessitates this change.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] New test in instrumentation/opentelemetry-instrumentation-sqlalchemy/tests/test_sqlalchemy.py

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
@@ -165,25 +165,26 @@ class SQLAlchemyInstrumentor(BaseInstrumentor):
 
         enable_commenter = kwargs.get("enable_commenter", False)
         commenter_options = kwargs.get("commenter_options", {})
+        attrs_provider = kwargs.get("attrs_provider")
 
         _w(
             "sqlalchemy",
             "create_engine",
             _wrap_create_engine(
-                tracer, connections_usage, enable_commenter, commenter_options
+                tracer, connections_usage, enable_commenter, commenter_options, attrs_provider
             ),
         )
         _w(
             "sqlalchemy.engine",
             "create_engine",
             _wrap_create_engine(
-                tracer, connections_usage, enable_commenter, commenter_options
+                tracer, connections_usage, enable_commenter, commenter_options, attrs_provider
             ),
         )
         _w(
             "sqlalchemy.engine.base",
             "Engine.connect",
-            _wrap_connect(tracer),
+            _wrap_connect(tracer, attrs_provider),
         )
         if parse_version(sqlalchemy.__version__).release >= (1, 4):
             _w(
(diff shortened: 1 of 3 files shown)


jochs:
hey @shalevr - do you think you'd be able to review this sometime soon? Happy to create an issue if you need additional context

lzchen:
> Specifically, for my usecase, I have a subclass of the ParentBasedTraceIdRatio called OverrideableParentBasedTraceIdRatio which allows passing of an 'x-ignore-sample' attribute which forces the sampler to emit a DROP decision when the parent is not already sampled. I would like to force all SqlAlchemy spans to be dropped unless the parent is sampled, which necessitates this change.

I'm a bit confused with the behavior of this custom sampler. Would you be able to provide a specific example? Can't you achieve this behavior with a custom [ParentBased](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/trace/sampling.py#L308) sampler?

jochs:
hey @lzchen sorry I missed this. Yes, I am getting the behavior with a custom ParentBased sampler, but the sample decision depends on the attributes, and I currently can't provide attributes to the OTEL-sqlalchemy spans.

The sampler code looks like this:

```
class OverrideableParentBasedTraceIdRatio(ParentBasedTraceIdRatio):
    @override
    def should_sample(
        self,
        parent_context: Optional[Context],
        trace_id: int,
        name: str,
        kind: Optional[SpanKind] = None,
        attributes: Optional[Attributes] = None,
        links: Optional[Sequence[Link]] = None,
        trace_state: Optional[TraceState] = None,
    ) -> SamplingResult:
        if attributes is not None:
            if not not attributes.get(X_SAMPLED):
                return SamplingResult(
                    decision=Decision.RECORD_AND_SAMPLE,
                    attributes=attributes,
                    trace_state=_get_parent_trace_state(parent_context),
                )
            parent_span_context = get_current_span(parent_context).get_span_context()
            if not parent_span_context.trace_flags.sampled and not not attributes.get(X_IGNORE_SAMPLE):
                return SamplingResult(
                    decision=Decision.DROP,
                    attributes=attributes,
                    trace_state=_get_parent_trace_state(parent_context),
                )
        return super().should_sample(parent_context, trace_id, name, kind, attributes, links, trace_state)
```

And the idea is that I want to sample X% of the time in general but if I pass in the relevant attributes I can force record and sample or force drop. This is generally helpful for hot library code where I don't want to pay the overhead of tracing unless a span is already active via app code. Currently I'm somewhat blocked on sqlalchemy instrumentation because I don't want to pay that overhead cost on 5% of all sqlalchemy function calls (which is what the ratio is set at in production).

Issue #2745, opened 2024-07-26
`http.target` missing from Django duration metric attributes in old semconv


alexmojaki:
In https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2624 I added the `http.target` attribute to duration metrics. In https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2714 by @lzchen it was accidentally removed when not opted in to the new semconv. This can be seen by adding `print(point.attributes["http.target"])` here:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/5a48824f4f6c09c1583fafc44ba9cdf1e4293569/instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py#L737-L742

Adding `SpanAttributes.HTTP_TARGET` to `_server_duration_attrs_old` fixes this particular case but breaks others, e.g. `http.target` starts appearing in starlette metrics with high cardinality values (i.e. the actual path, not the template).

I strongly suggest that these tests assert the actual concrete values of the metrics data as much as possible. Right now they're very lenient and tied to the implementation which allowed this regression to pass through.

lzchen:
@alexmojaki 

Apologies for the oversight. The instrumentations today support multiple "old" conventions (e.g. most instrumentations support v1.11 but others support v1.20 and others have a mixture of different attributes). In order to better consolidate the old semconvs and encourage migration to new semconv, we have decided to discourage adding any contributions for "old" semconv. [This](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2621) and [this](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2624) were made exceptions and the regression only occurred for django because we just happened to be working on the same change at similar times :). Hopefully with this new guidance these semantic  conventions mixups won't happen again.

I've put up a PR to fix the regression. I do realize the solution is a bit hacky but since this SHOULD be a one-off we don't really want to change `_server_duration_attrs_old ` as you have outlined above.

PR #2714, opened 2024-07-16
HTTP semantic convention stability migration for django

lzchen:
Part of https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2453

PR #2746, opened 2024-07-26
Re-add `http.target` to Django old sem conv server duration metric

lzchen:
Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2745

https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2714 introduced a regression outlined in the issue above. Going forward, we will probably try to discourage additional contributions that pertain to old HTTP semantic conventions to avoid this kind of mixup happening. We do not want to be supporting multiple "old" versions of the semantic conventions.


instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
```diff
@@ -806,7 +810,7 @@ def test_wsgi_metrics_both_semconv(self):
         _recommended_attrs = {
             "http.server.active_requests": active_count_both_attrs,
             "http.server.request.duration": _server_duration_attrs_new,
-            "http.server.duration": _server_duration_attrs_old,
+            "http.server.duration": _django_old_server_duration_attrs,
```
alexmojaki, line 813: I really think this should be more concrete to prevent other similar mistakes slipping through, with something like this:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/c45a6201e234f73322fcbcb858e906a55d349796/instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py#L577-L602

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -447,12 +447,19 @@ def process_response(self, request, response):
 def _parse_duration_attrs(
     req_attrs, sem_conv_opt_in_mode=_HTTPStabilityMode.DEFAULT
 ):
-    return _filter_semconv_duration_attrs(
+    # http.target to be included in old semantic conventions
+    target = None
+    if sem_conv_opt_in_mode is _HTTPStabilityMode.DEFAULT:
+        target = req_attrs.get(SpanAttributes.HTTP_TARGET)
```
alexmojaki, line 453:
```suggestion
        target = req_attrs.get(SpanAttributes.HTTP_ROUTE)
```

`target` in metrics ~ `route` in spans

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
lzchen: I agree. We probably would want to rewrite all instrumentation tests to better reflect the attributes actually expected instead of simply comparing with a static list. At least for this PR, it would be better to fix this regression separately from the larger effort of rewriting these tests and to have a separate PR to address that.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
lzchen: Not entirely sure what you mean by this? We parsing duration attributes so this pertains to metrics.

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
alexmojaki: It's genuinely hard right now to tell if this PR (or past versions of it) are doing the correct thing.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
alexmojaki: I think `req_attrs` had a bunch of span attributes as well. Looking more closely I think this was probably fine, and the change you just made feels better still. I'd be more confident that this won't break in future if (1) the tests were stricter and (2) span attributes already had `http.target` as they should.

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
lzchen: I've added some stricter tests to assert the actual values of expected attributes. Please take a look if they fall along the lines of what you'd expect.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
lzchen: I've added stricter tests that assert values and keys of the attributes instead of just the presence of keys that follow a static list. In regards to (2), I believe span attributes are populated by [wsgi](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py#L210) middleware and `http.target` is already [populated](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py#L348).

alexmojaki: Thanks, I do think that's an improvement. The span attributes stuff does push the scope here, but if you're willing, I think this would be even better:

```python
expected_attributes = {
    'http.request.method': 'GET',
    'http.method': 'GET',
    'http.url': 'http://testserver/traced/',
    'http.scheme': 'http',
    'url.scheme': 'http',
    'http.status_code': 200,
    'http.response.status_code': 200,
    'server.port': 80,
    'client.address': '127.0.0.1',
    'network.protocol.version': '1.1',
}

if DJANGO_2_2:
    expected_attributes['http.route'] = '^traced/'

self.assertEqual(span.attributes, expected_attributes)
```

That's an example for `test_traced_get_both_semconv`. Then it would be clear whether or not `http.target` is in the span attributes. Right now if it is then the tests don't show it.

emdneto (approved):
Discussed in slack, prob we need to review other tests as well, but LGTM. Thanks for addressing the main comments.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
lzchen: @alexmojaki 

This is a great point and we most likely need this scenario tested out.

I do not believe `http.target` was ever explicitly supported for `django` and only relies on the `wsgi` and `asgi` implementations of extracting this information. If `http.target` is missing, it simply would be a feature request of the django instrumentation and outside the scope of this pr. We also do not want to be including old semantic convention changes so the feature request would more accurately be "support url.path/url.query" in this case.

xrmx (approved):

lzchen:
@alexmojaki 

Will wait on a review from you before merging!

alexmojaki (approved):

PR #2748, opened 2024-07-29
add missing tests to urllib3 after semconv migration

emdneto:
# Description
Add missing tests to urllib3 instrumentation after semconv migration. I missed test both_semconv for metrics.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-urllib3/tests/test_urllib3_metrics.py
+++ b/instrumentation/opentelemetry-instrumentation-urllib3/tests/test_urllib3_metrics.py
@@ -26,6 +26,10 @@ from opentelemetry.instrumentation._semconv import (
     _OpenTelemetrySemanticConventionStability,
 )
 from opentelemetry.instrumentation.urllib3 import URLLib3Instrumentor
+from opentelemetry.semconv._incubating.metrics.http_metrics import (
+    HTTP_CLIENT_REQUEST_BODY_SIZE,
+    HTTP_CLIENT_RESPONSE_BODY_SIZE,
+)
 from opentelemetry.semconv.attributes.http_attributes import (
     HTTP_REQUEST_METHOD,
     HTTP_RESPONSE_STATUS_CODE,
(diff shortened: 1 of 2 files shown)


instrumentation/opentelemetry-instrumentation-urllib3/tests/test_urllib3_metrics.py
```diff
@@ -102,7 +111,9 @@ def test_basic_metrics(self):
             SpanAttributes.HTTP_SCHEME: "http",
         }
 
-        self.assertEqual(client_duration.name, "http.client.duration")
+        self.assertEqual(
+            client_duration.name, MetricInstruments.HTTP_CLIENT_DURATION
```
xrmx, line 115: Have to say I prefer to assert the exact string instead of a variable that can change its value over time

emdneto: Me too, and that's the way I did in the first version of the tests. But while writing it I thought referencing the variable from semconv package would  make better to notice the changes from old to new in metric names

emdneto: But I'm also fine with using the exact name

lzchen: We probably should rewrite all the sem conv tests to assert exact strings similar to [here](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2746#discussion_r1693506236). Can start with this pr and we can leave the other changes to other prs.

emdneto: Agreed. So while we are here I'll change the PR to be the refactor for urllib3 tests then to use plain string instead of referencing semconv symbols

lzchen (approved):

shalevr (approved):

xrmx (approved):

PR #2756, opened 2024-07-30
instrumentation/fastapi: fix fastapi-slim support

### LINKED ISSUE #2683: fastapi-slim support

xrmx:
# Description

In instrumentation_dependencies we need to use importlib.metadata.distribution because find_spec("fastapi") will return something even with just fastapi-slim installed.

Then need to fix the tox environment name otherwise all the fastapi dependencies will also be installed for fastapi-slim and tests would pass because they find fastapi installed.

Fixes #2683

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
@@ -176,6 +176,7 @@ from importlib.util import find_spec
 from typing import Collection
 
 import fastapi
+from pkg_resources import get_distribution
 from starlette.routing import Match
 
 from opentelemetry.instrumentation._semconv import (
@@ -285,10 +286,16 @@ class FastAPIInstrumentor(BaseInstrumentor):
         app._is_instrumented_by_opentelemetry = False
 
     def instrumentation_dependencies(self) -> Collection[str]:
+        # need to use get_distribution because find_spec("fastapi") will return
+        # something even with just fastapi-slim installed
+        try:
+            get_distribution("fastapi-slim")
+            return (_fastapi_slim,)
+        except Exception:  # pylint: disable=broad-exception-caught
+            pass
+
         if find_spec("fastapi") is not None:
             return (_fastapi,)
-        if find_spec("fastapi_slim") is not None:
-            return (_fastapi_slim,)
         # If neither is installed, return both as potential dependencies
         return _instruments
 
--- a/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
@@ -172,10 +172,10 @@ API
 ---
 """
 import logging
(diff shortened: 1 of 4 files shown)


instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -285,10 +286,16 @@ def uninstrument_app(app: fastapi.FastAPI):
         app._is_instrumented_by_opentelemetry = False
 
     def instrumentation_dependencies(self) -> Collection[str]:
+        # need to use get_distribution because find_spec("fastapi") will return
+        # something even with just fastapi-slim installed
+        try:
+            get_distribution("fastapi-slim")
```
emdneto, line 292: Wouldn't it be more reliable to use get_distribution for both?

xrmx: I don't know, but using it we can drop an import

emdneto (approved):

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -172,10 +172,10 @@ def client_response_hook(span: Span, scope: dict[str, Any], message: dict[str, A
 ---
 """
 import logging
-from importlib.util import find_spec
 from typing import Collection
 
 import fastapi
+from pkg_resources import get_distribution
```
lzchen, line 178: Not a big deal but is it a lot of work to use `importlib.metadata` instead since that is encouraged over `pkg_resources`

lzchen, line 178: Does `pkg_resources` come with the standard Python distribution or do we need to add a direct dependency for `setuptools` which includes `pkg_resources`?

lzchen (approved):

emdneto:
It seems the tests for fastapislim aren't being run in CI. Checked the job and is only installing the packages 😓

PR #2784, opened 2024-08-07
Sort labelset to ensure compatibility with thanos-receive

fstolba:
We're using thanos-receive which requires labels to be sorted lexicographically. Trying to submit a metric without first sorting the label list results in the write request being rejected, yielding the message `Out of order labels in the label set`.

This PR fixes this behaviour. 

Please note I only tested this on thanos-remote, could use testing on other implementations.

### DIFF
--- a/exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
+++ b/exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
@@ -258,7 +258,7 @@ class PrometheusRemoteWriteMetricsExporter(MetricExporter):
         timeseries = []
         for labels, samples in sample_sets.items():
             ts = TimeSeries()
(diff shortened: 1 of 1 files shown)


xrmx:
@fstolba You need to sign the CLA in order to have your changes integrated

dsarazin:
Hello guys,

i also faced this issue and confirm that the commit proposed by @fstolba fixes this timeseries error !

image

fstolba:
Sorry for the delay, CLA is now signed.

emdneto (commented):
I think this can happen not only for Thanos but also for every Prometheus-like backend. Maybe it's worth having a comment and a simple test to prevent someone from removing that. https://github.com/prometheus/prometheus/issues/11505 

Could you please add changelog and fix lint?

giuseka:
I tested the solution proposed in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2784 with my thanos receiver, and it works for me. I print the labels and the order is correct.

lzchen:
@fstolba 

Please add a CHANGELOG entry.

Issue #2744, opened 2024-07-26
opentelemetry-instrumentation-aws-lambda: code does not handle lack of expected environment variables


Kyle-sandeman-mrdfood:
### Describe your environment

OS: Docker (python3.12)
Python version: (e.g., Python 3.12.1)
Package version: (e.g., 0.48.0)

### What happened?

Please note that my application is NOT a Lambda or related in any way. I only have this package because it is a default in the bootstrap.

```
Instrumenting of aws-lambda failed
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/opentelemetry/instrumentation/auto_instrumentation/_load.py", line 87, in _load_instrumentors
    distro.load_instrumentor(entry_point, skip_dep_check=True)
  File "/usr/local/lib/python3.12/site-packages/opentelemetry/instrumentation/distro.py", line 64, in load_instrumentor
    instrumentor().instrument(**kwargs)
  File "/usr/local/lib/python3.12/site-packages/opentelemetry/instrumentation/instrumentor.py", line 114, in instrument
    result = self._instrument(  # pylint: disable=assignment-from-no-return
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/opentelemetry/instrumentation/aws_lambda/__init__.py", line 429, in _instrument
    ) = lambda_handler.rsplit(".", 1)
        ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'rsplit'
Failed to auto initialize opentelemetry
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py", line 39, in initialize
    _load_instrumentors(distro)
  File "/usr/local/lib/python3.12/site-packages/opentelemetry/instrumentation/auto_instrumentation/_load.py", line 91, in _load_instrumentors
    raise exc
  File "/usr/local/lib/python3.12/site-packages/opentelemetry/instrumentation/auto_instrumentation/_load.py", line 87, in _load_instrumentors
    distro.load_instrumentor(entry_point, skip_dep_check=True)
  File "/usr/local/lib/python3.12/site-packages/opentelemetry/instrumentation/distro.py", line 64, in load_instrumentor
[...]

PR #2786, opened 2024-08-08
opentelemetry-bootstrap: remove aws-lambda from default instrumentations

### LINKED ISSUE #2787: Remove aws-lambda instrumentation from the defaults instrumentation list
As discussed in [SIG (August 1, 2024)](https://docs.google.com/document/d/1CIMGoIOZ-c3-igzbd6_Pnxx1SjAkjwqoYSUWxPY8XIs/edit#heading=h.u68u6n2oon43) we agreed to remove aws-lambda instrumentation from the list of default instrumentation used in bootstrap. Instead of requiring everyone to opt-out, AWS Lambda users should opt-in.

More context: we had some cases of users facing issues in aws-lambda instrumentation even if they are not running anything on aws lambda, forcing them to have to uninstall the instrumentation package or disable the instrumentation using `OTEL_PYTHON_DISABLED_INSTRUMENTATIONS` environment variable:
https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2744
https://github.com/open-telemetry/opentelemetry-python-contrib/discussions/2769

emdneto:
# Description
Fixes #2787

As discussed in [SIG (August 1, 2024)](https://docs.google.com/document/d/1CIMGoIOZ-c3-igzbd6_Pnxx1SjAkjwqoYSUWxPY8XIs/edit#heading=h.u68u6n2oon43) we agreed to remove aws-lambda instrumentation from the list of default instrumentation used in bootstrap. Instead of requiring everyone to opt-out, AWS Lambda users should opt-in.

More context: we had some cases of users facing issues in aws-lambda instrumentation even if they are not running anything on aws lambda, forcing them to have to uninstall the instrumentation package: 
https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2744
https://github.com/open-telemetry/opentelemetry-python-contrib/discussions/2769

## Type of change


scripts/generate_instrumentation_bootstrap.py
```diff
@@ -53,12 +53,16 @@
     "bootstrap_gen.py",
 )
 
+packages_to_ignore = ["opentelemetry-instrumentation-aws-lambda"]
```
lzchen, line 56: This file is to generate the `bootstrap_gen.py` file used to autoinstrument using `bootstrap.py`. Any reason why we are removing it?

emdneto: Right. 

1. This file is used in `tox -e generate` command to generate the entries in bootstrap_gen.py. 

2. Yes, we are removing it because the LambdaInstrumentor only makes sense when running in Lambda runtime, so the users should opt-in. For example, to use the LambdaInstrumentor you should add and use this [wrapper](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/python/src/otel/otel_sdk/otel_wrapper.py#L52) and this lambda layer cares of installing the package, [see](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/python/src/otel/otel_sdk/requirements.txt#L4).

Options: 
1. (this PR) We can opt for removing it here, so the package won't be installed through bootstrap
2. We can also include the instrumentation-aws-lambda package in `package_to_exclude` in `_load_instrumentors` to not load the LambdaInstrumentor by default. But the question is: does it make sense to have the package installed and not have the respective instrumentor loaded?

emdneto: If we want to be less radical, we can just not load the instrumentor by default but continue to install the package as default through bootstrap. Wdyt? When we had the issue with the aws-lambda the workaround was: remove the package or just disable the instrumentation through environment variable `OTEL_PYTHON_DISABLED_INSTRUMENTATIONS`

lzchen: I think I'm fine with the PR as it is. Maybe jsut add a comment to outline what was discussed here and why we are exlcluding it?

lzchen (approved):

PR #2790, opened 2024-08-09
Implement new HTTP semantic convention opt-in for Falcon

### LINKED ISSUE #1532: Falcon exceptions should not always be masked as 500 and marked as error
In a Falcon app, if we raise an Exception of a type that is a subclass of a `falcon.HTTPError` or `falcon.HTTPStatus` it ends up being always marked as error and the trace status code set as 500 (Except if it is a 404).

All exceptions that are subclasses of `falcon.HTTPError` or `falcon.HTTPStatus` should not always be masked with 500 status code. Additionally, status codes in the 4xx range are not server errors and should not be marked as errors, which, from what I can read in the code are not, but since everything is being set as 500, they end up being marked as errors.

My suggestion would be something like checking the exception to see if it is a subclass of a `falcon.HTTPError` or `falcon.HTTPStatus` and try to extract its status code from there. Only if it wasn't possible we would follow into the current path.

```python
#instrumentation/falcon/__init__.py around like 499 could be something like:

if exc_type and not req_succeeded:
  if "HTTPNotFound" in exc_type.__name__:
    status = "404"
    reason = "NotFound"
  else:
    if isinstance(exc, falcon.HTTPError) or isinstance(exc, falcon.HTTPStatus):
      try:
        status = int(exc.title.split(" ")[0])
      except ValueError:
        status = "500"
    reason = f"{exc_type.__name__}: {exc}"
```

FilipNikolovski:
# Description 

This change updates the falcon instrumentor to adhere to the new HTTP semantic conventions, according to the migration plan outlined here:
- https://github.com/open-telemetry/semantic-conventions/tree/main/docs/http 
- https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2453

also fixes #1532 
## Type of change

# How Has This Been Tested?

Updated the tests to assert that the new span and duration attributes have been added.


instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
```diff
@@ -66,6 +70,7 @@ def setUp(self):
             {
                 "OTEL_PYTHON_FALCON_EXCLUDED_URLS": "ping",
                 "OTEL_PYTHON_FALCON_TRACED_REQUEST_ATTRS": "query_string",
+                "OTEL_SEMCONV_STABILITY_OPT_IN": "http/dup",
```
emdneto, line 73: Are we missing the tests for "http" and "default"?

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -243,6 +251,11 @@ class _InstrumentedFalconAPI(getattr(falcon, _instrument_app)):
     def __init__(self, *args, **kwargs):
         otel_opts = kwargs.pop("_otel_opts", {})
 
+        _OpenTelemetrySemanticConventionStability._initialize()
```
emdneto, line 254: I think we are missing some important points here:

# Checklist
- [X] Add semconv status as `migration` for urllib3 in instrumentations README
- [X] Populate `{method}` as `HTTP` when nonstandard method (sanitize_method)
- [X]  set request HTTP method attribute as `_OTHER` when nonstandard http method and for new semconv also set `http.request.method_original`  with the original nonstandard method.
- [X] Set schema_url based on the opt-in mode
- [X] Create histograms based on the opt-in mode (different metric names so no dup attributes)
- [X] Set metrics attributes based on the opt-in mode
- [X] Set span attributes based on the opt-in mode

emdneto (commented):
A big thanks for working on this one. Just some comments

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
FilipNikolovski: Thank you for the great feedback, I'll get started on this.

instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
FilipNikolovski: Updated the tests to test for "default", "http" and "http/dup".

xrmx:
@FilipNikolovski `tox -e generate` step is failing, adding ` _semconv_status = "migration"` in package.py should be enough

xrmx:
> > @FilipNikolovski `tox -e generate` step is failing, adding ` _semconv_status = "migration"` in package.py should be enough
> 
> Thanks for the feedback. I've updated the file, but that step is still failing 🤷. Not sure if I need to run `tox -e generate` locally or not?

You need to run it locally and commit the updated files so it match on CI

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -379,12 +411,36 @@ def _start_response(status, response_headers, *args, **kwargs):
...
-                duration_attrs[SpanAttributes.HTTP_STATUS_CODE] = (
-                    span.attributes.get(SpanAttributes.HTTP_STATUS_CODE)
+            duration_s = default_timer() - start
+            if self.duration_histogram_old:
+                duration_attrs = otel_wsgi._parse_duration_attrs(
+                    attributes, _HTTPStabilityMode.DEFAULT
                 )
-            duration = max(round((default_timer() - start) * 1000), 0)
-            self.duration_histogram.record(duration, duration_attrs)
+                if span.is_recording():
```
emdneto, line 419: Thanks for the work so far. I think we want status_code independent of tracing decisions—also, a test showing that a nonrecording span would produce metrics with the correct HTTP status code.

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
emdneto: LGTM. Left a small comment

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -243,6 +255,11 @@ class _InstrumentedFalconAPI(getattr(falcon, _instrument_app)):
     def __init__(self, *args, **kwargs):
         otel_opts = kwargs.pop("_otel_opts", {})
 
+        _OpenTelemetrySemanticConventionStability._initialize()
```
lzchen, line 258: I think [BaseInstrumentor](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/instrumentor.py#L112) already calls this.

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -379,12 +411,36 @@ def _start_response(status, response_headers, *args, **kwargs):
             exception = exc
             raise
         finally:
-            if span.is_recording():
-                duration_attrs[SpanAttributes.HTTP_STATUS_CODE] = (
-                    span.attributes.get(SpanAttributes.HTTP_STATUS_CODE)
+            duration_s = default_timer() - start
+            if self.duration_histogram_old:
```
lzchen, line 415: Nit: I think you can use [add_response_attributes](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py#L473) instead to populate duration attributes with status.

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
FilipNikolovski: On a second thought, I don't think we should set the status code for duration metrics at all, at least I don't see that we do in the other instrumentation wrappers. What are your thoughts on this?

emdneto: The attribute is conditionally required if there is a send/received status code. We are doing exactly the same thing for asgi

FilipNikolovski: I was able to solve this by adding the request attributes in the env. Now the status code is properly added to the metric attributes.

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
FilipNikolovski: Removed it from here.

emdneto:
> @xrmx running `tox -e generate` throws an error for me:
> 
That's weird, because `/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-sklearn` is deprecated for a while

FilipNikolovski:
> > @xrmx running `tox -e generate` throws an error for me:
> 
> That's weird, because `/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-sklearn` is deprecated for a while

Turns out I still had the sklearn directory :facepalm:, the command ran successfully after I deleted it.

emdneto (approved):

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
```diff
@@ -524,9 +524,7 @@ def process_response(self, req, resp, resource, req_succeeded=None):  # pylint:d
             if exc_type and not req_succeeded:
                 if "HTTPNotFound" in exc_type.__name__:
                     status = "404"
-                elif isinstance(exc, falcon.HTTPError) or isinstance(
-                    exc, falcon.HTTPStatus
-                ):
+                elif isinstance(exc, falcon.HTTPError, falcon.HTTPStatus):
```
xrmx:
```suggestion
                elif isinstance(exc, (falcon.HTTPError, falcon.HTTPStatus)):
```

isinstance has only two arguments :)

CHANGELOG.md
```diff
@@ -51,6 +51,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#2942](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2942))
 - `opentelemetry-instrumentation-click`: new instrumentation to trace click commands
   ([#2994](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2994))
+- `opentelemetry-instrumentation-falcon` Implement new HTTP semantic convention opt-in for Falcon
```
xrmx, line 54: You have to move this to unreleased above

FilipNikolovski: Ah thanks, i've made the update.

instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
```diff
@@ -345,47 +515,92 @@ def test_falcon_metrics(self):
...
-            "http.flavor": "1.1",
-            "http.server_name": "falconframework.org",
-        }
-        start = default_timer()
+    def test_falcon_metric_values_new_semconv(self):
+        number_data_point_seen = False
+        histogram_data_point_seen = False
+
         self.client().simulate_get("/hello/756")
-        duration = max(round((default_timer() - start) * 1000), 0)
```
emdneto, line 368: Maybe I missed something in the discussion, but why are you removing the duration itself check?

instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
```diff
@@ -125,23 +201,44 @@ def _test_method(self, method):
...
+            if self._has_fixed_http_target
+            else "/hello",
+            SpanAttributes.NET_PEER_PORT: 65133,
+            SpanAttributes.HTTP_FLAVOR: "1.1",
+            "falcon.resource": "HelloWorldResource",
+            SpanAttributes.HTTP_STATUS_CODE: 201,
+            SpanAttributes.HTTP_ROUTE: "/hello",
+        }
+        expected_attributes_new = {
+            SpanAttributes.HTTP_REQUEST_METHOD: method,
```
emdneto, line 222: In other instrumentations we are using the symbols from `opentelemetry.semconv.attributes.http_attributes`

instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
FilipNikolovski: Good catch, that probably got lost in the refactor of the tests, I can re-add that.

instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
FilipNikolovski: I'll make the change.

instrumentation/opentelemetry-instrumentation-falcon/src/opentelemetry/instrumentation/falcon/__init__.py
lzchen: Should be fixed with: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3148/files

FilipNikolovski: I've made the necessary updates.

instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
FilipNikolovski: I've re-added the checks.

instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
```diff
@@ -345,47 +530,119 @@ def test_falcon_metrics(self):
...
+                        self.assertEqual(
+                            metric.name, "http.server.request.duration"
+                        )
+                    else:
+                        self.assertEqual(
+                            metric.name, "http.server.active_requests"
+                        )
+                    data_points = list(metric.data.data_points)
+                    self.assertEqual(len(data_points), 1)
+                    for point in list(metric.data.data_points):
```
emdneto, line 591:
```suggestion
                    for point in data_points:
```

instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
```diff
@@ -345,47 +530,119 @@ def test_falcon_metrics(self):
...
+        self.client().simulate_get("/hello/756")
+        duration = max(default_timer() - start, 0)
+
+        metrics_list = self.memory_metrics_reader.get_metrics_data()
+        for resource_metric in metrics_list.resource_metrics:
+            for scope_metric in resource_metric.scope_metrics:
+                for metric in scope_metric.metrics:
+                    data_points = list(metric.data.data_points)
+                    self.assertEqual(len(data_points), 1)
+                    for point in list(metric.data.data_points):
```
emdneto, line 547:
```suggestion
                    for point in data_points:
```

instrumentation/opentelemetry-instrumentation-falcon/tests/test_falcon.py
```diff
@@ -305,6 +468,28 @@ def test_traced_not_recording(self):
             self.assertFalse(mock_span.set_attribute.called)
             self.assertFalse(mock_span.set_status.called)
 
+            metrics_list = self.memory_metrics_reader.get_metrics_data()
+            self.assertTrue(len(metrics_list.resource_metrics) != 0)
+
+            metrics_list = self.memory_metrics_reader.get_metrics_data()
```
emdneto, line 474:
```suggestion
```
remove this?

FilipNikolovski: done.

lzchen (approved):

Issue #2734, opened 2024-07-24
Sync hook used as async hook in opentelemetry-instrumentation-httpx causing TypeError


horw:
### Describe your environment

OS: Ubuntu
Python 3.10.12
Name: opentelemetry-instrumentation-httpx
Version: 0.46b0

### What happened?

code from lib
```python
request_hook = kwargs.get("request_hook")
response_hook = kwargs.get("response_hook")
async_request_hook = kwargs.get("async_request_hook", request_hook)
async_response_hook = kwargs.get("async_response_hook", response_hook
```
If you set up `request_hook`, it will also be used for `async_request_hook` if you don't initialize `async_request_hook`. This could lead to issues because you don't want to pass an async function to `request_hook`. If a non-async function is passed, it may not work as expected. It seems that there needs to be a wrapper around `request_hook` to check if it is not async and then wrap it accordingly. 

### Steps to Reproduce
```python
def hook(span, req):
    pass

HTTPXClientInstrumentor().instrument(request_hook=hook)
    
with Client() as client:
    client.post("http://localhost:9999/hello")
async with AsyncClient() as client:
    await client.post("http://localhost:9999/hello")
```

### Expected Result

Both requests completed

### Actual Result

```
await self._request_hook(span, request_info)
TypeError: object NoneType can't be used in 'await' expression
```

### Would you like to implement a fix?

None

shijiadong2022:
Can you assign this one to me?

xrmx:
@shijiadong2022 looking forward your PR

xrmx:
@shijiadong2022 have you started working on this?

shijiadong2022:
I am working on it and will submit a PR soon

xrmx:
@horw Could you please double check if the fix works for you? To have the desired output with the updated code you should  provide yourself the `async_request_hook` callback e.g. `HTTPXClientInstrumentor().instrument(request_hook=hook, async_request_hook=async_request_hook)`

PR #2794, opened 2024-08-13
fix Sync hook used as async hook in opentelemetry-instrumentation-httpx causing TypeError

shijiadong2022:
Fixes # (https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2734)


instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -731,7 +731,9 @@ def _instrument(self, **kwargs):
         self._original_async_client = httpx.AsyncClient
         request_hook = kwargs.get("request_hook")
         response_hook = kwargs.get("response_hook")
-        async_request_hook = kwargs.get("async_request_hook", request_hook)
```
lzchen, line 734: I'm wondering if we should remove this logic of defaulting to `request hook` if `async_request_hook` does not exist. It doesn't make much intuitive sense to me. We can then probably just add a check to see if the hook coming from `async_request_hook` is indeed an async function with `inspect.iscoroutinefunction` as part of the [inspect](https://docs.python.org/3/library/inspect.html) module and not assign it to `_InstrumentedAsyncClient._request_hook` if not.

emdneto: The easiest way is probably to leave it as None; it will fail during the callable check when hook is None. Also, this PR needs the async_response_hook fix as well. And of course, some tests would be nice to avoid problems with this again. Wdyt?

lzchen: > leave it as None if not set; it will fail during the callable check when the hook is None.

Yes this and checking if `async_request_hook` is actually `async` is probably sufficient, and as always, adding more tests are always welcome.

shijiadong2022: Thanks for your review. I will work on the changes.

lzchen:
Please also add a CHANGELOG entry :)

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -731,8 +732,16 @@ def _instrument(self, **kwargs):
...
-        async_response_hook = kwargs.get("async_response_hook", response_hook)
+        if iscoroutinefunction(request_hook):
+            async_request_hook = kwargs.get("async_request_hook", request_hook)
```
xrmx, line 736: From the previous review, I'm not sure we want this. I think the idea is to remove the default from the previous code here

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -731,8 +732,16 @@ def _instrument(self, **kwargs):
...
+            async_request_hook = kwargs.get("async_request_hook")
+        if iscoroutinefunction(response_hook):
+            async_response_hook = kwargs.get(
+                "async_response_hook", response_hook
+            )
+        else:
+            async_response_hook = kwargs.get("async_response_hook")
         if callable(request_hook):
             _InstrumentedClient._request_hook = request_hook
         if callable(async_request_hook):
```
xrmx, line 747: And check here if it is a coroutine

lzchen: I believe you'd just need the below:
```
async_request_hook = kwargs.get("async_request_hook")

if callable(async_request_hook) and iscoroutinefunction(async_request_hook):
     _InstrumentedAsyncClient._request_hook = async_request_hook
```

shijiadong2022: Thanks @xrmx and @lzchen for your feedback. I've tried the code you mentioned but the testAsyncInstrumentationIntegration test case will fail with the change. The reason is in this case, request_hook is actually an aysnc request hook which we cannot set to None. Please correct me if I misunderstood.

PR #2795, opened 2024-08-13
Fix psycopg2 instrument issue

### LINKED ISSUE #2522: `Psycopg2Instrumentor().instrument_connection` raises `AttributeError: 'psycopg2.extensions.connection' object has no attribute '_is_instrumented_by_opentelemetry'`
**Describe your environment**

```
opentelemetry-instrumentation-psycopg2==0.45b0
psycopg2-binary==2.9.9
```

**Steps to reproduce**

Use `Psycopg2Instrumentor().instrument_connection`. For example:

```python
import psycopg2
from opentelemetry.instrumentation.psycopg2 import Psycopg2Instrumentor

connection = psycopg2.connect(database='database', user='user', password='secret', host='0.0.0.0', port=5432)

Psycopg2Instrumentor().instrument_connection(connection)
```

and run a database with:

```
docker run --name postgres \
    -e POSTGRES_USER=user \
    -e POSTGRES_PASSWORD=secret \
    -e POSTGRES_DB=database \
    -p 5432:5432 -d postgres
```

```
  File "test.py", line 6, in <module>
    Psycopg2Instrumentor().instrument_connection(connection)
  File "/home/alex/work/logfire/.venv/lib/python3.12/site-packages/opentelemetry/instrumentation/psycopg2/__init__.py", line 164, in instrument_connection
    connection._is_instrumented_by_opentelemetry = False
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'psycopg2.extensions.connection' object has no attribute '_is_instrumented_by_opentelemetry'
```

**Additional context**

Here's where the error happens:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/46d2ce6acea9a1a6cb1a4d4c863077002f5f7d21/instrumentation/opentelemetry-instrumentation-psycopg2/src/opentelemetry/instrumentation/psycopg2/__init__.py#L164

The code seems tested here:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/46d2ce6acea9a1a6cb1a4d4c863077002f5f7d21/instrumentation/opentelemetry-instrumentation-psycopg2/tests/test_psycopg2_integration.py#L210

but it doesn't get the error because it's a mock connection.

qiuge615:
# Description

Currently there is issue with psycopg2, the instrumentation can not work as expected. Update the codes based on mysql and sqlite3 to resolve the issue. Will check the two properties _otel_orig_cursor_factory and _is_instrumented_by_opentelemetry in subsequent issue later.
Fix one issue in unit test, assign instrumented connection to cnx.

Fixes #2522

## Type of change

# How Has This Been Tested?

Tested with below command to run unit test
tox -e py310-test-opentelemetry-instrumentation-psycopg2

Tested traces with local environment and the traces are displayed without error.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-psycopg2/src/opentelemetry/instrumentation/psycopg2/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-psycopg2/src/opentelemetry/instrumentation/psycopg2/__init__.py
@@ -158,33 +158,40 @@ class Psycopg2Instrumentor(BaseInstrumentor):
         dbapi.unwrap_connect(psycopg2, "connect")
 
     # TODO(owais): check if core dbapi can do this for all dbapi implementations e.g, pymysql and mysql
-    @staticmethod
-    def instrument_connection(connection, tracer_provider=None):
-        if not hasattr(connection, "_is_instrumented_by_opentelemetry"):
-            connection._is_instrumented_by_opentelemetry = False
+    # TODO: comment out below codes to fix issue #2522, will try to set the properties later.
+    def instrument_connection(self, connection, tracer_provider=None):
+        # if not hasattr(connection, "_is_instrumented_by_opentelemetry"):
+        #     connection._is_instrumented_by_opentelemetry = False
 
-        if not connection._is_instrumented_by_opentelemetry:
-            setattr(
-                connection, _OTEL_CURSOR_FACTORY_KEY, connection.cursor_factory
-            )
-            connection.cursor_factory = _new_cursor_factory(
-                tracer_provider=tracer_provider
-            )
-            connection._is_instrumented_by_opentelemetry = True
-        else:
-            _logger.warning(
-                "Attempting to instrument Psycopg connection while already instrumented"
-            )
-        return connection
+        # if not connection._is_instrumented_by_opentelemetry:
+        #     setattr(
+        #         connection, _OTEL_CURSOR_FACTORY_KEY, connection.cursor_factory
+        #     )
+        #     connection.cursor_factory = _new_cursor_factory(
+        #         tracer_provider=tracer_provider
+        #     )
+        #     connection._is_instrumented_by_opentelemetry = True
+        # else:
+        #     _logger.warning(
+        #         "Attempting to instrument Psycopg connection while already instrumented"
+        #     )
+        # return connection
+        return dbapi.instrument_connection(
+            __name__,
+            connection,
+            self._DATABASE_SYSTEM,
+            self._CONNECTION_ATTRIBUTES,
+            version=__version__,
+            tracer_provider=tracer_provider,
+        )
 
     # TODO(owais): check if core dbapi can do this for all dbapi implementations e.g, pymysql and mysql
-    @staticmethod
-    def uninstrument_connection(connection):
+    def uninstrument_connection(self, connection):
         connection.cursor_factory = getattr(
             connection, _OTEL_CURSOR_FACTORY_KEY, None
         )
 
-        return connection
+        return dbapi.uninstrument_connection(connection)
 
 
 # TODO(owais): check if core dbapi can do this for all dbapi implementations e.g, pymysql and mysql
--- a/instrumentation/opentelemetry-instrumentation-psycopg2/tests/test_psycopg2_integration.py
+++ b/instrumentation/opentelemetry-instrumentation-psycopg2/tests/test_psycopg2_integration.py
@@ -235,7 +235,7 @@ class TestPostgresqlIntegration(TestBase):
     # pylint: disable=unused-argument
     def test_uninstrument_connection_with_instrument_connection(self):
         cnx = psycopg2.connect(database="test")
-        Psycopg2Instrumentor().instrument_connection(cnx)
+        cnx = Psycopg2Instrumentor().instrument_connection(cnx)
         query = "SELECT * FROM test"
         cursor = cnx.cursor()
         cursor.execute(query)


instrumentation/opentelemetry-instrumentation-psycopg2/src/opentelemetry/instrumentation/psycopg2/__init__.py
```diff
@@ -158,33 +158,40 @@ def _uninstrument(self, **kwargs):
         dbapi.unwrap_connect(psycopg2, "connect")
 
     # TODO(owais): check if core dbapi can do this for all dbapi implementations e.g, pymysql and mysql
-    @staticmethod
-    def instrument_connection(connection, tracer_provider=None):
-        if not hasattr(connection, "_is_instrumented_by_opentelemetry"):
```
lzchen, line 163: I don't think this is an actual solution to the issue. We have this logic here to prevent double instrumentation of the same `connection` object. @qiuge615 Are you looking to simply unblock yourself for now or are you looking to find an actual fix for why `connection` is throwing exceptions when being assigned properties?

qiuge615: @lzchen Thanks for your comments. I updated the codes with previous logic, use DatabaseApiIntegration to instrument connection and use "isinstance(connection, ObjectProxy)" to check if the connection already instrumented, at last use dbapi to create TracedConnectionProxy. Tested locally, it can instrument connection and display warning message if double instrumentation. If this approach makes sense to you, I will update PR accordingly, thanks.

lzchen: @qiuge615 

Did you update the code or are you planning to update the code because I don't see any new commit.

qiuge615: I did not push the code yet. I will push the changes for review. In the meanwhile if you have any suggestions or comments, please let me know, I will update accordingly, thank you.

lzchen: @qiuge615 

It will be easier to make suggestions + comments once the changes are pushed for review thanks.

qiuge615: @lzchen I updated the code, please review and share your comments, thanks.

xrmx (commented):
I think this is missing a test showing what is currently not working

qiuge615:
Thanks for your suggestion, I will add a test case for it.

qiuge615:
> I think this is missing a test showing what is currently not working
Added test case for this. It will be failed when running with previous codes, since there is no property _is_instrumented_by_opentelemetry in connection object. Please review and let me know if there is anything I need to update, thanks.

instrumentation/opentelemetry-instrumentation-psycopg2/tests/test_psycopg2_integration.py
```diff
@@ -63,6 +64,17 @@ def cursor(self):
     def get_dsn_parameters(self):  # pylint: disable=no-self-use
         return {"dbname": "test"}
 
+    def __setattr__(self, name, value):
```
xrmx, line 67: Instead of doing this I think adding something like `self.assertTrue(foo._is_instrument_by_opentelemetry is True)` would be a lot more explicit

qiuge615: Thanks for your reply. How about I add "self.assertFalse(hasattr(cnx, "_is_instrumented_by_opentelemetry"))" in test case. It will be failed with existing code in main branch, but pass with the changes. If this makes sense to you, I will update the code and push.

xrmx: Yeah, that would work too.

qiuge615: Thank you for your reply, I will make the changes and push shortly.

lzchen: I guess we don't have tests right now to test for duplicate instrumentation. I think since we are changing this to depend on the existence of `ObjectProxy` we should test that at least (since we are removing the reliance on `_is_instrumented_by_opentelemetry` )

qiuge615: @xrmx @lzchen I updated PR, could you please help to review and suggest if there is anything I need to change? Thanks.

instrumentation/opentelemetry-instrumentation-psycopg2/src/opentelemetry/instrumentation/psycopg2/__init__.py
```diff
@@ -158,24 +159,25 @@ def _uninstrument(self, **kwargs):
...
-            setattr(
-                connection, _OTEL_CURSOR_FACTORY_KEY, connection.cursor_factory
-            )
-            connection.cursor_factory = _new_cursor_factory(
-                tracer_provider=tracer_provider
-            )
-            connection._is_instrumented_by_opentelemetry = True
-        else:
+    def instrument_connection(self, connection, tracer_provider=None):
+        if isinstance(connection, ObjectProxy):
```
lzchen, line 163: I'm still not sure this solves the original issue (we still don't know WHY the error is being generated when we are trying to access `_is_instrumented_by_opentelemetry ` field but this current change is fine as a short-term solution. Please add a TODO or follow-up issue to investigate.

lzchen:
@qiuge615 

Please allow for maintainer edits so we can rebase for you.

tammy-baylis-swi:
Hi @qiuge615 are you still working on this PR?

tammy-baylis-swi:
Closing as resolved by https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4257

PR #2802, opened 2024-08-16
feat: add ability to optionally disable internal HTTP send and receive spans

### LINKED ISSUE #831: FastApi auto-instrumentation send extra spans for a single API execution
Hi open-telemetry community,

I have a python project with fastapi auto-instrumentation which export the spans to zipkin server, and I have a simple "GET" API which returns a "hello". But I notice it generates 3 spans instead of 1.

image

As you can see, the top one is related to the request I send out. But the other 2 child spans, I am not sure why they are there. 

**Describe your environment**
python = 3.8
opentelemetry-api = 1.7.1
opentelemetry-sdk = 1.7.1
opentelemetry-exporter-zipkin = 1.7.1
opentelemetry-instrumentation-fastapi = 0.26b1
fastapi = 0.70.0
uvicorn = 0.15.0

**Steps to reproduce**
```
from opentelemetry.exporter.zipkin.json import ZipkinExporter
from opentelemetry.sdk.resources import Resource
from fastapi import FastAPI
import uvicorn

app = FastAPI()

zipkin_exporter = ZipkinExporter(
    endpoint="https://bolt-msl-zipkin-server.eks-dev-us-east-1.container.spglobal.com:30443/api/v2/spans",
)
span_processor = SimpleSpanProcessor(zipkin_exporter)
trace_provider = TracerProvider(resource=Resource.create({"service.name": "open-telemetry-test"}))
trace_provider.add_span_processor(span_processor)

@app.get('/opentelemetry-fastapi')
def about():
   return "hello"


FastAPIInstrumentor.instrument_app(app,
                                   tracer_provider=trace_provider)

if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", port=8000)
```

Here is the REST request I made:
```
curl --location --request GET 'http://localhost:8000/opentelemetry-fastapi'
```

Here is the JSON span metadata:
```
[{
	"traceId": "7e5511d0c28934bf480bc7f28ba2d34d",
	"parentId": "76122e06995e41b6",
	"id": "a920c88a2a64c8e2",
	"name": "/opentelemetry-fastapi http send",
	"timestamp": 1639091853270289,
	"duration": 1002,
	"localEndpoint": {
		"serviceName": "open-telemetry-test"
	},
	"tags": {
		"otel.library.name": "opentelemetry.instrumentation.asgi",
		"otel.library.version": "0.26b1",
		"service.name": "open-telemetry-test",
		"telemetry.sdk.version": "1.7.1",
		"type": "http.response.body"
	},
	"debug": true
}, {
	"traceId": "7e5511d0c28934bf480bc7f28ba2d34d",
	"id": "76122e06995e41b6",
	"kind": "SERVER",
	"name": "/opentelemetry-fastapi",
	"timestamp": 1639091853034614,
	"duration": 289682,
	"localEndpoint": {
		"serviceName": "open-telemetry-test"
	},
	"tags": {
		"correlation-id": "111222333444555",
		"http.flavor": "1.1",
		"http.host": "127.0.0.1:8000",
		"http.method": "GET",
		"http.route": "/opentelemetry-fastapi",
		"http.scheme": "http",
		"http.server_name": "localhost:8000",
		"http.status_code": "200",
		"http.target": "/opentelemetry-fastapi",
		"http.url": "http://127.0.0.1:8000/opentelemetry-fastapi",
		"http.user_agent": "PostmanRuntime/7.28.4",
		"net.host.port": "8000",
		"net.peer.ip": "127.0.0.1",
		"net.peer.port": "55402",
		"otel.library.name": "opentelemetry.instrumentation.asgi",
		"otel.library.version": "0.26b1",
		"region": "us-east-1",
		"service.name": "open-telemetry-test",
		"telemetry.sdk.version": "1.7.1"
	},
	"debug": true
}, {
	"traceId": "7e5511d0c28934bf480bc7f28ba2d34d",
	"parentId": "76122e06995e41b6",
	"id": "edd83d58575f0811",
	"name": "/opentelemetry-fastapi http send",
	"timestamp": 1639091853053617,
	"localEndpoint": {
		"serviceName": "open-telemetry-test"
	},
	"tags": {
		"http.status_code": "200",
		"otel.library.name": "opentelemetry.instrumentation.asgi",
		"otel.library.version": "0.26b1",
		"service.name": "open-telemetry-test",
		"telemetry.sdk.version": "1.7.1",
		"type": "http.response.start"
	},
	"debug": true
}]
```

Exactly one span related to the API get executed.

Besides the span relates to the API got executed, there are 2 extra spans are generated.

**Additional context**

I am suspecting this is related to the ASGI interfaces since the extra span has "send" in it which related to sending response back. But shouldn't it be a single "send" instead two?

I feel like there could some missing link regarding certain concept. I am looking forward to hearing from your suggestions.

Thanks for your time!

omgitsaheadcrab:
# Description

This PR addresses the issue where it is not possible to disable internal HTTP send and receive spans. For many members of the community these do not provide any value and only add noise to the traces. Other mechanisms of filtering them out, such as wrapping span processors or exporters are clunky. 

Providing environment variables, namely: `OTEL_PYTHON_ASGI_EXCLUDE_SEND_SPAN` and `OTEL_PYTHON_ASGI_EXCLUDE_RECEIVE_SPAN` allows users to disable one or both of these with no additional code.

Fixes #831

## Type of change

# How Has This Been Tested?

- [X] TestAsgiApplication.test_exclude_internal_spans added to test logic
- [X Integration tests performed manually using the environment variables

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
@@ -185,6 +185,11 @@ will replace the value of headers such as ``session-id`` and ``set-cookie`` with
 Note:
     The environment variable names used to capture HTTP headers are still experimental, and thus are subject to change.
 
+Internal HTTP spans
+*****************************
+Internal HTTP send and receive spans are added by default. These can optionally be excluded by setting the boolean environment variables
+``OTEL_PYTHON_ASGI_EXCLUDE_SEND_SPAN`` and ``OTEL_PYTHON_ASGI_EXCLUDE_RECEIVE_SPAN`` to ``true``.
+
 API
 ---
 """
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
@@ -487,7 +487,7 @@ def get_default_span_details(scope: dict) -> Tuple[str, dict]:
 
 
 def _collect_target_attribute(
-    scope: typing.Dict[str, typing.Any]
+    scope: typing.Dict[str, typing.Any],
 ) -> typing.Optional[str]:
     """
     Returns the target path as defined by the Semantic Conventions.
@@ -814,6 +814,7 @@ class OpenTelemetryMiddleware:
     def _get_otel_receive(self, server_span_name, scope, receive):
         if self.exclude_receive_span:
             return receive
+
         @wraps(receive)
         async def otel_receive():
             with self.tracer.start_as_current_span(
@@ -858,7 +859,6 @@ class OpenTelemetryMiddleware:
             elif message["type"] == "websocket.send":
                 status_code = 200
 
-            # Conditional send_span creation
             if not self.exclude_send_span:
                 with self.tracer.start_as_current_span(
                     " ".join((server_span_name, scope["type"], "send"))
@@ -867,7 +867,11 @@ class OpenTelemetryMiddleware:
                         self.client_response_hook(send_span, scope, message)
 
                     if send_span.is_recording():
-                        send_span.set_attribute("asgi.event.type", message["type"])
+                        if message["type"] == "http.response.start":
+                            expecting_trailers = message.get("trailers", False)
+                        send_span.set_attribute(
+                            "asgi.event.type", message["type"]
+                        )
                         if status_code:
                             set_status_code(
                                 send_span,
@@ -876,40 +880,43 @@ class OpenTelemetryMiddleware:
                                 self._sem_conv_opt_in_mode,
                             )
 
-            # Server span logic always applied
-            if server_span.is_recording() and "headers" in message:
-                if server_span.kind == trace.SpanKind.SERVER:
-                    custom_response_attributes = (
-                        collect_custom_headers_attributes(
-                            message,
-                            self.http_capture_headers_sanitize_fields,
-                            self.http_capture_headers_server_response,
-                            normalise_response_header_name,
-                        )
-                        if self.http_capture_headers_server_response
-                        else {}
-                    )
-                    if len(custom_response_attributes) > 0:
-                        server_span.set_attributes(custom_response_attributes)
+                        if (
+                            server_span.is_recording()
+                            and server_span.kind == trace.SpanKind.SERVER
+                            and "headers" in message
+                        ):
+                            custom_response_attributes = (
+                                collect_custom_headers_attributes(
+                                    message,
+                                    self.http_capture_headers_sanitize_fields,
+                                    self.http_capture_headers_server_response,
+                                    normalise_response_header_name,
+                                )
+                                if self.http_capture_headers_server_response
+                                else {}
+                            )
+                            if len(custom_response_attributes) > 0:
+                                server_span.set_attributes(
+                                    custom_response_attributes
+                                )
 
-                if status_code:
-                    # We record metrics only once
-                    set_status_code(
-                        server_span,
-                        status_code,
-                        duration_attrs,
-                        self._sem_conv_opt_in_mode,
-                    )
+            if status_code:
+                set_status_code(
+                    server_span,
+                    status_code,
+                    duration_attrs,
+                    self._sem_conv_opt_in_mode,
+                )
 
-                propagator = get_global_response_propagator()
-                if propagator:
-                    propagator.inject(
-                        message,
-                        context=set_span_in_context(
-                            server_span, trace.context_api.Context()
-                        ),
-                        setter=asgi_setter,
-                    )
+            propagator = get_global_response_propagator()
+            if propagator:
+                propagator.inject(
+                    message,
+                    context=set_span_in_context(
+                        server_span, trace.context_api.Context()
+                    ),
+                    setter=asgi_setter,
+                )
 
             content_length = asgi_getter.get(message, "content-length")
             if content_length:
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
@@ -191,9 +191,11 @@ API
 
 from __future__ import annotations
 
+import os
 import typing
 import urllib
 from collections import defaultdict
+from distutils.util import strtobool
 from functools import wraps
 from timeit import default_timer
 from typing import Any, Awaitable, Callable, DefaultDict, Tuple
@@ -225,6 +227,10 @@ from opentelemetry.instrumentation._semconv import (
     _set_http_url,
     _set_http_user_agent,
     _set_status,
+)
+from opentelemetry.instrumentation.asgi.environment_variables import (
+    OTEL_PYTHON_ASGI_EXCLUDE_RECEIVE_SPAN,
+    OTEL_PYTHON_ASGI_EXCLUDE_SEND_SPAN,
 )
 from opentelemetry.instrumentation.asgi.types import (
     ClientRequestHook,
@@ -527,6 +533,8 @@ class OpenTelemetryMiddleware:
             the current globally configured one is used.
         meter_provider: The optional meter provider to use. If omitted
             the current globally configured one is used.
+        exclude_receive_span: Optional flag to exclude the http receive span from the trace.
+        exclude_send_span: Optional flag to exclude the http send span from the trace.
     """
 
     # pylint: disable=too-many-branches
@@ -545,6 +553,8 @@ class OpenTelemetryMiddleware:
         http_capture_headers_server_request: list[str] | None = None,
         http_capture_headers_server_response: list[str] | None = None,
         http_capture_headers_sanitize_fields: list[str] | None = None,
+        exclude_receive_span: bool = False,
+        exclude_send_span: bool = False,
     ):
         # initialize semantic conventions opt-in if needed
         _OpenTelemetrySemanticConventionStability._initialize()
@@ -650,6 +660,12 @@ class OpenTelemetryMiddleware:
                 )
             )
             or []
+        )
+        self.exclude_receive_span = exclude_receive_span or strtobool(
+            os.getenv(OTEL_PYTHON_ASGI_EXCLUDE_RECEIVE_SPAN, "false")
+        )
+        self.exclude_send_span = exclude_send_span or strtobool(
+            os.getenv(OTEL_PYTHON_ASGI_EXCLUDE_SEND_SPAN, "false")
         )
 
     # pylint: disable=too-many-statements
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/environment_variables.py
```diff
@@ -0,0 +1,24 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Exclude the http send span from the tracing feature in Python.
+"""
+
+OTEL_PYTHON_ASGI_EXCLUDE_SEND_SPAN = "OTEL_PYTHON_ASGI_EXCLUDE_SEND_SPAN"
```
lzchen, line 19: I would prefer not introducing new environment variables and only have this supported manually. Would this be enough to cover your use case?

omgitsaheadcrab: Unfortunately no. I use fastapi and in that instrumentor's implementation this class isn't injectable. A further PR would be required to remedy that, and I guess also for any other frameworks where that's the case.

omgitsaheadcrab: Oh ugh, it's even worse. The `add_middleware` method takes a class not an instance, that would require a fair amount of refactoring. What is your worry about adding env vars if the default behavior remains the same?

lzchen: @omgitsaheadcrab 

> What is your worry about adding env vars if the default behavior remains the same?

Not a huge worry. The way you are doing it seems fairly unintrusive and we already have precedence for configuration capturing headers through env var.

> I use fastapi and in that instrumentor's implementation this class isn't injectable

Can't you simply add another configuration option [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py#L275)? Is the refactoring that much?

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -651,6 +666,12 @@ def __init__(
             )
             or []
         )
+        self.exclude_receive_span = exclude_receive_span or strtobool(
+            os.getenv(OTEL_PYTHON_ASGI_EXCLUDE_RECEIVE_SPAN, "false")
```
lzchen, line 670: Is it possible for multiple instrumentations that depend on `asgi` middleware to be used (like [starlette](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/__init__.py#L226) and [fastapi](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py#L263)? What if a user wants different internal http span configurations per instrumentation? Should we care about this scenario or does this seem not likely?

omgitsaheadcrab: That's fair, the env vars would certainly be global. If you wanted different behavior for different instrumentations then you'd need to configure them individually.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/environment_variables.py
omgitsaheadcrab: Ah, I'd misunderstood how add_middleware works. Happy to remove the env vars and make a small PR adding the config options to the fastapi instrumentor.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
lzchen: As you've mentioned in the above [conversation](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2802#discussion_r1722003261), adding this as an explicit configuration setting in each instrumentation might be the way to go for now instead of `asgi` middleware. Feel free to change your PR just for `fastapi` :)

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -545,6 +547,8 @@ def __init__(
         http_capture_headers_server_request: list[str] | None = None,
         http_capture_headers_server_response: list[str] | None = None,
         http_capture_headers_sanitize_fields: list[str] | None = None,
+        exclude_receive_span: bool = False,
+        exclude_send_span: bool = False,
```
xrmx, line 551: Question: what do you think on adding something a list of possible values instead, e.g. `excluded_span: list[Literal["receive", "send"]] | None` ?

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -832,74 +841,82 @@ def _get_otel_send(
...
-                                self.http_capture_headers_server_response,
-                                normalise_response_header_name,
-                            )
-                            if self.http_capture_headers_server_response
-                            else {}
+
+            status_code = None
+            if message["type"] == "http.response.start":
+                status_code = message["status"]
+                expecting_trailers = message.get("trailers", False)
```
xrmx, line 848: Is this another bugfix?

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
omgitsaheadcrab: For the sake of minimizing method arguments I guess, but really you'll still need to convert that into a boolean for each. Happy to make the change if you and @lzchen agree on it.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -222,7 +222,7 @@ class FastAPIInstrumentor(BaseInstrumentor):
 
     @staticmethod
     def instrument_app(
-        app: fastapi.FastAPI,
```
omgitsaheadcrab, line 225: Sphinx autodoc was unhappy with the app arg in the docstring. I tried explicitly typing it within the docstring, adding the mock import to sphinx's conf - nothing helped. Removing the typing here fixed it :man_shrugging:

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
lzchen: I don't have a strong opinion on this.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -832,74 +840,81 @@ def _get_otel_send(
...
-                propagator = get_global_response_propagator()
-                if propagator:
-                    propagator.inject(
-                        message,
-                        context=set_span_in_context(
-                            server_span, trace.context_api.Context()
-                        ),
-                        setter=asgi_setter,
-                    )
+                        if (
```
lzchen, line 871: See as these are in the `if not self.exclude_send_span:` if block, are we not performing the below logic anymore if `self._exclude_send_span` is `True`?

omgitsaheadcrab: You're right, server span logic should always be executed. They were previously needlessly nested which threw me

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
omgitsaheadcrab: Alrighty, made the change

omgitsaheadcrab:
`tox -e lint` didn't give the same result as `tox -elint-instrumentation-asgi`.. I've refactored it by extracting the send_span and server_span logic to helpers rather than just disabling the pylint rule. Looks cleaner now too.

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
```diff
@@ -572,6 +572,27 @@ def test_background_execution(self):
...
+            (True, True, ["GET / http receive", "GET / http send"]),
+            (False, True, ["GET / http send"]),
+            (True, False, ["GET / http receive"]),
+            (False, False, []),
+        ]
+        for exclude_receive_span, exclude_send_span, excluded_spans in cases:
+            app.exclude_receive_span = exclude_receive_span
+            app.exclude_send_span = exclude_send_span
+            self.send_default_request()
+            span_list = self.memory_exporter.get_finished_spans()
```
xrmx, line 591: At least for the cases where we expect a span we should be adding an assert that there are spans otherwise if this always returns an empty span list the test will pass

xrmx (commented):
Next time please better shut up pylint warnings on methods size and do the separation in helpers separately to ease review

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
```diff
@@ -572,6 +572,28 @@ def test_background_execution(self):
...
+            assert span_list
```
xrmx, line 592:
```suggestion
            self.assertTrue(span_list)
```

xrmx (approved):

lzchen (approved):

lzchen:
@omgitsaheadcrab 

Changes look good to us! I think you have a few tests that are failing as a result of your changes.

omgitsaheadcrab:
Ugh! I ran it with assert locally and it worked, didn't check the self.assertTrue suggestion. I'm away from home on leave now so I'll see if I can fix whatever broke from another pc 🤔

omgitsaheadcrab:
@xrmx @lzchen I'm back from annual leave now and I've updated the test. Apologies for letting this sit so long!

PR #2814, opened 2024-08-22
Fix `http.host` and `net.peer.ip` new http semconv mapping

lzchen:
`http.host` old semantic convention should map to different new semantic convention depending on if they are `client` or `server` type telemetry. See [spec](https://github.com/open-telemetry/semantic-conventions/blob/v1.27.0/docs/attributes-registry/http.md#deprecated-http-attributes) for details.

Also updates `net.peer.ip` -> `client.address` logic to be conditional only if it doesn't exist already. This is outlined in the [migration doc](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/http-migration.md#common-attributes-across-http-client-and-server-spans) net.peer.ip -> net.sock.peer.addr → network.peer.address.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-instrumentation-tornado` Handle http client exception and record exception info into span
   ([#2563](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2563))
 - `opentelemetry-instrumentation` fix `http.host` new http semantic convention mapping to depend on `kind` of span
-  ([2814](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2814))
+  ([#2814](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2814))
 
 ## Version 1.26.0/0.47b0 (2024-07-23)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#2792](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2792))
 - `opentelemetry-instrumentation-tornado` Handle http client exception and record exception info into span
   ([#2563](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2563))
+- `opentelemetry-instrumentation` fix `http.host` new http semantic convention mapping to depend on `kind` of span
+  ([2783](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2783))
 
 ## Version 1.26.0/0.47b0 (2024-07-23)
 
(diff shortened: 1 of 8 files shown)


opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -310,27 +326,32 @@ def _set_http_target(result, target, path, query, sem_conv_opt_in_mode):
...
 
 
-def _set_http_peer_ip(result, ip, sem_conv_opt_in_mode):
+def _set_http_host_server(result, host, sem_conv_opt_in_mode):
     if _report_old(sem_conv_opt_in_mode):
-        set_string_attribute(result, SpanAttributes.NET_PEER_IP, ip)
+        set_string_attribute(result, SpanAttributes.HTTP_HOST, host)
     if _report_new(sem_conv_opt_in_mode):
-        set_string_attribute(result, CLIENT_ADDRESS, ip)
+        set_string_attribute(result, CLIENT_ADDRESS, host)
```
emdneto, line 333: This change wasn't too clear for me, but after taking a look [here](https://github.com/open-telemetry/semantic-conventions/blob/main/model/registry/deprecated/network.yaml) it makes sense.
>use `server.address` on client spans and `client.address` on server spans.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -310,27 +326,32 @@ def _set_http_target(result, target, path, query, sem_conv_opt_in_mode):
...
+# net.sock.peer.addr -> client.socket.address for server spans (TODO) AND client.address if missing
+# https://github.com/open-telemetry/semantic-conventions/blob/v1.21.0/CHANGELOG.md#v1210-2023-07-13
+# https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/http-migration.md#common-attributes-across-http-client-and-server-spans
+def _set_http_peer_ip_server(result, ip, sem_conv_opt_in_mode):
     if _report_old(sem_conv_opt_in_mode):
-        set_int_attribute(result, SpanAttributes.NET_PEER_PORT, port)
+        set_string_attribute(result, SpanAttributes.NET_PEER_IP, ip)
     if _report_new(sem_conv_opt_in_mode):
-        set_int_attribute(result, CLIENT_PORT, port)
+        # Only populate if not already populated
```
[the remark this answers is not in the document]
lzchen: Yes the `test_behavior_with_scope_server_as_none_new_semconv` tests already cover this behavior. Notice how the `client.address` field remains as `0.0.0.0` instead of being overwritten to `127.0.0.1`.

CHANGELOG.md
```diff
@@ -40,6 +40,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#2792](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2792))
 - `opentelemetry-instrumentation-tornado` Handle http client exception and record exception info into span
   ([#2563](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2563))
+- `opentelemetry-instrumentation` fix `http.host` new http semantic convention mapping to depend on `kind` of span
+  ([2814](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2814))
```
emdneto, line 44:
```suggestion
  ([#2814](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2814))
```

emdneto (approved):

PR #2823, opened 2024-08-25
httpx: fix handling of async hooks

### LINKED ISSUE #2734: Sync hook used as async hook in opentelemetry-instrumentation-httpx causing TypeError

xrmx:
# Description

Don't default to sync hooks for async hooks, then check that they are actually async. Needed to fix current tests that were testing only the sync path because tests are inherited from a base class :broken_heart: .

Fixes #2734
Supersedes #2794

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,25 +7,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-## Added
+### Added
 
 - `opentelemetry-instrumentation-kafka-python` Instrument temporary fork, kafka-python-ng
   inside kafka-python's instrumentation
   ([#2537](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2537))
 
-## Breaking changes
+### Breaking changes
 
 - `opentelemetry-bootstrap` Remove `opentelemetry-instrumentation-aws-lambda` from the defaults instrumentations
   ([#2786](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2786))
 
-## Fixed
+### Fixed
 
 - `opentelemetry-instrumentation-httpx` fix handling of async hooks
   ([#2823](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2823))
 - `opentelemetry-instrumentation-system-metrics` fix `process.runtime.cpu.utilization` values to be shown in range of 0 to 1
-  ([2812](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2812))
+  ([#2812](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2812))
 - `opentelemetry-instrumentation-fastapi` fix `fastapi` auto-instrumentation by removing `fastapi-slim` support, `fastapi-slim` itself is discontinued from maintainers
-  ([2783](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2783))
+  ([#2783](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2783))
 - `opentelemetry-instrumentation-aws-lambda` Avoid exception when a handler is not present.
   ([#2750](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2750))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Fixed
 
+- `opentelemetry-instrumentation-httpx` fix handling of async hooks
+  ([2823](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2823))
```
emdneto, line 24:
```suggestion
  ([#2823](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2823))
```

instrumentation/opentelemetry-instrumentation-httpx/tests/test_httpx_integration.py
```diff
@@ -1214,3 +1226,36 @@ def test_basic_multiple(self):
...
+            tracer_provider=self.tracer_provider,
+            async_request_hook=_request_hook,
+        )
+        client = self.create_client()
+        result = self.perform_request(self.URL, client=client)
+
+        self.assertEqual(result.text, "Hello!")
+        span = self.assert_span()
+        self.assertEqual(span.name, "GET")
+        HTTPXClientInstrumentor().uninstrument()
```
emdneto, line 1261: Are we missing out on testing the issue behavior? In cases we have only request_hook, it shouldn't generate an error anymore:

For example, this should fail without your fix:
```
            HTTPXClientInstrumentor().instrument(
                tracer_provider=self.tracer_provider,
                request_hook=_request_hook,
            )
            client = self.create_client()
            result = self.perform_request(self.URL, client=client)

            self.assertEqual(result.text, "Hello!")
            span = self.assert_span()
            self.assertEqual(span.name, "GET" + self.URL)
            HTTPXClientInstrumentor().uninstrument()
```
Will produce the error: 

```
        with self._tracer.start_as_current_span(
            span_name, kind=SpanKind.CLIENT, attributes=span_attributes
        ) as span:
            exception = None
            if callable(self._request_hook):
>               await self._request_hook(span, request_info)
E               TypeError: object NoneType can't be used in 'await' expression
```

xrmx: We can be more defensive but I think the regression was to use a sync function in the async path under the hood. Given that we have two different parameters for sync and async hooks using a sync function with an async client I think is a progamming error more than an issue of the instrumentation.

xrmx: Maybe we can add a check when using it too

emdneto: Right. Taking a second look at the tests I'm fine with

emdneto (approved):

PR #2849, opened 2024-09-04
Update README.rst

blueswen:
The URL link is incorrect; it should be a repository in the open-telemetry account.

# Description

The URL link `https://github.com/flands/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap_gen.py` should be `https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap_gen.py`.

## Type of change

# How Has This Been Tested?

It’s only a documentation update, so there is no need to test.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/opentelemetry-instrumentation/README.rst
+++ b/opentelemetry-instrumentation/README.rst
@@ -40,7 +40,7 @@ to figure out which instrumentation packages the user might want to install. By 
 prints out a list of the default and detected instrumentation packages that can be added to a
 requirements.txt file. It also supports installing the packages when run with
 :code:`--action=install` or :code:`-a install` flag. All default and detectable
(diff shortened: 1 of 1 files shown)


zhihali (approved):
I agree this should be updated. I've compared the two versions, and the opentelemetry one is more up to date. Here are some key differences for the approvers:

Instrumentation Versions: opentelemetry uses 0.49b0.dev, while the personal repo uses 0.44b0.dev.
Library Versions:
confluent-kafka: opentelemetry supports up to 2.4.0, personal repo up to 2.2.0.
elasticsearch: opentelemetry starts at 6.0, personal repo from 2.0.
Additional Libraries: opentelemetry includes kafka-python-ng, which isn't in the personal repo.

xrmx:
Good catch!

Issue #2861, opened 2024-09-11
Multiple system metrics return negative values when using DELTA temporality.


bjrara:
### Describe your environment

OS: CentOS
Python version: 3.9.16
Package version: OTEL 1.25.0

### What happened?

The following metrics are found with negative values when configuring to export using DELTA temporality.
* system.network.connections
* process.runtime.cpython.gc_count
* process.runtime.cpython.memory
* process.runtime.cpython.thread_count
* process.open_file_descriptor.count

### Steps to Reproduce

```
temporality_dict: Dict[type, AggregationTemporality] = {}
for typ in [
    Counter,
    UpDownCounter,
    ObservableCounter,
    ObservableCounter,
    ObservableUpDownCounter,
    ObservableGauge,
    Histogram,
]:
    temporality_dict[typ] = AggregationTemporality.DELTA

set_meter_provider(MeterProvider([PeriodicExportingMetricReader(ConsoleMetricExporter(preferred_temporality=temporality_dict))], resource=resource))
SystemMetricsInstrumentor().instrument()
```

### Expected Result

Metrics generated with non-negative value

### Actual Result

Metrics generated with negative value

### Additional context

```
Sep 11 03:35:10: Descriptor:
Sep 11 03:35:10:      -> Name: process.runtime.cpython.memory
Sep 11 03:35:10:      -> Description: Runtime cpython memory
Sep 11 03:35:10:      -> Unit: bytes
Sep 11 03:35:10:      -> DataType: Sum
Sep 11 03:35:10:      -> IsMonotonic: false
Sep 11 03:35:10:      -> AggregationTemporality: Delta
Sep 11 03:35:10: NumberDataPoints #0
Sep 11 03:35:10: Data point attributes:
Sep 11 03:35:10:      -> type: Str(rss)
Sep 11 03:35:10: StartTimestamp: 2024-09-11 03:34:09.999252322 +0000 UTC
Sep 11 03:35:10: Timestamp: 2024-09-11 03:35:10.010880332 +0000 UTC
Sep 11 03:35:10: Value: -1081344
Sep 11 03:35:10: NumberDataPoints #1
Sep 11 03:35:10: Data point attributes:
Sep 11 03:35:10:      -> type: Str(vms)
Sep 11 03:35:10: StartTimestamp: 2024-09-11 03:34:09.999252322 +0000 UTC
[...]

PR #2865, opened 2024-09-11
Fix instrument types in system metrics

bjrara:
# Description

This PR fixes the instrument types when creating the system metrics.

Metrics that are impacted:
* system.network.connections:
  Collected from [psutil.net_connections](https://psutil.readthedocs.io/en/latest/#psutil.net_connections) is a snapshot of the current state of network connections, thus should be delta and is not monotonic.
* process.runtime.cpython.gc_count
  Collected from [python gc](https://docs.python.org/3/library/gc.html) is the current collection counts,  thus should be delta and is not monotonic.
* process.runtime.cpython.memory
  Collected from [psutil.Process.memory_info](https://psutil.readthedocs.io/en/latest/#psutil.Process.memory_info) is a snapshot of the process’s memory usage at the time of the call, thus should be delta and is not monotonic.
* process.runtime.cpython.thread_count
  Collected from [psutil.Process.num_threads](https://psutil.readthedocs.io/en/latest/#psutil.Process.num_threads) is the number of threads currently used by this process (non cumulative).
* process.open_file_descriptor.count
  Collected from [psutil.Process.num_fds](https://psutil.readthedocs.io/en/latest/#psutil.Process.num_fds) is the number of file descriptors currently opened by this process (non cumulative).

Change the instrument types for the metrics above to `observable_gauge` for correction.

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2861

## Type of change

# How Has This Been Tested?

* Tested locally using console exporter configured with `delta` temporality for all the metric types.

- [ ] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#2860](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2860))
 - `opentelemetry-instrumentation-aiokafka` Add instrumentor and auto instrumentation support for aiokafka
   ([#2082](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2082))
+- `opentelemetry-instrumentation-system-metrics` Fix instrument types in system metrics
+  ([#2865](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2865))
 
 ## Version 1.27.0/0.48b0 ()
 
--- a/instrumentation/opentelemetry-instrumentation-system-metrics/src/opentelemetry/instrumentation/system_metrics/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-system-metrics/src/opentelemetry/instrumentation/system_metrics/__init__.py
@@ -331,11 +331,10 @@ class SystemMetricsInstrumentor(BaseInstrumentor):
             )
 
         if "system.network.connections" in self._config:
-            self._meter.create_observable_up_down_counter(
+            self._meter.create_observable_gauge(
                 name="system.network.connections",
                 callbacks=[self._get_system_network_connections],
                 description="System network connections",
-                unit="connections",
             )
 
         if "system.thread_count" in self._config:
@@ -346,7 +345,7 @@ class SystemMetricsInstrumentor(BaseInstrumentor):
             )
 
         if "process.runtime.memory" in self._config:
-            self._meter.create_observable_up_down_counter(
+            self._meter.create_observable_gauge(
                 name=f"process.runtime.{self._python_implementation}.memory",
                 callbacks=[self._get_runtime_memory],
                 description=f"Runtime {self._python_implementation} memory",
@@ -367,15 +366,14 @@ class SystemMetricsInstrumentor(BaseInstrumentor):
                     "The process.runtime.gc_count metric won't be collected because the interpreter is PyPy"
                 )
             else:
-                self._meter.create_observable_counter(
+                self._meter.create_observable_gauge(
                     name=f"process.runtime.{self._python_implementation}.gc_count",
                     callbacks=[self._get_runtime_gc_count],
                     description=f"Runtime {self._python_implementation} GC count",
-                    unit="bytes",
                 )
 
         if "process.runtime.thread_count" in self._config:
-            self._meter.create_observable_up_down_counter(
+            self._meter.create_observable_gauge(
                 name=f"process.runtime.{self._python_implementation}.thread_count",
                 callbacks=[self._get_runtime_thread_count],
                 description="Runtime active threads count",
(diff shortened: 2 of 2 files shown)


xrmx:
Don't have much clue about metrics but the generated semantic conventions package provides so helper functions to creates these metrics, and they return `UpDownCounter`, e.g. `create_process_open_file_descriptor_count`

emdneto (commented):
From specs:

I want to measure something (by reporting an absolute value):

* If the measurement values are [non-additive](https://opentelemetry.io/docs/specs/otel/metrics/supplementary-guidelines/#additive-property), use an [Asynchronous Gauge](https://opentelemetry.io/docs/specs/otel/metrics/api/#asynchronous-gauge).

If the measurement values are [additive](https://opentelemetry.io/docs/specs/otel/metrics/supplementary-guidelines/#additive-property):
* If the value is monotonically increasing - use an [Asynchronous Counter](https://opentelemetry.io/docs/specs/otel/metrics/api/#asynchronous-counter).
* If the value is NOT monotonically increasing - use an [Asynchronous UpDownCounter](https://opentelemetry.io/docs/specs/otel/metrics/api/#asynchronous-updowncounter).

So I think it's right no?

bjrara:
@xrmx @emdneto Thanks for taking a look at the PR.

The issue was not caused by UpDownCounter, but the temporality used by **Observable**UpDownCounter.

From the definition, [ObservableUpDownCounter is cumulative](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/metrics/_internal/aggregation.py#L1173-L1181), while these metrics are delta. 

```
if isinstance(instrument, ObservableUpDownCounter):
        return _SumAggregation(
            attributes,
            instrument_is_monotonic=False,
            instrument_aggregation_temporality=(
                AggregationTemporality.CUMULATIVE
            ),
            start_time_unix_nano=start_time_unix_nano,
        )
```

Change them from to UpDownCounter should also fix the issue. However, `create_up_down_counter` isn't asynchronous, so I choose to use ObservableGauge. If you have better suggestions on how to fix the issue properly with UpDownCounter, please let me know.

emdneto (commented):
Taking a second look, the issue makes sense.  According to the specs, UpDownCounter is suitable because the metrics are additive and not monotonically increasing, but in practice, for our asynchronous measurements (connections, memory, thread_count, etc), the cumulative temporality of the ObservableUpDownCounter doesn't seem to fit in that situation. We want to report an absolute value (current snapshot) that is not monotonic and non-cumulative. 

In [java](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/b51cedf925e4f2395715c8a49e261dc8a251738e/jmx-metrics/src/integrationTest/java/io/opentelemetry/contrib/jmxmetrics/target_systems/JvmTargetSystemIntegrationTest.java#L70) it's using ObservableGauge.

@bjrara, would you be open to adding a failing test showing how this is fixing the issue with the cumulative temporality? 

There's also another issue: The specs do not reference the `gc_count` metric, but I totally agree with that change (from async counter to async gauge).

xrmx:
If there is already another language going out of spec we should open an issue in the semconv repo, no?

bjrara:
> Taking a second look, the issue makes sense. According to the specs, UpDownCounter is suitable because the metrics are additive and not monotonically increasing, but in practice, for our asynchronous measurements (connections, memory, thread_count, etc), the cumulative temporality of the ObservableUpDownCounter doesn't seem to fit in that situation. We want to report an absolute value (current snapshot) that is not monotonic and non-cumulative.
> 
> In [java](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/b51cedf925e4f2395715c8a49e261dc8a251738e/jmx-metrics/src/integrationTest/java/io/opentelemetry/contrib/jmxmetrics/target_systems/JvmTargetSystemIntegrationTest.java#L70) it's using ObservableGauge.
> 
> @bjrara, would you be open to adding a failing test showing how this is fixing the issue with the cumulative temporality?
> 
> There's also another issue: The specs do not reference the `gc_count` metric, but I totally agree with that change (from async counter to async gauge).

Thanks for your feedback. Sure, checking how to add a failing test on the issue.

emdneto:
As discussed in SIG (09/26) -- @bjrara will try using the LastValue aggregation in a View to configure the SDK to output this metric as a gauge. If it doesn't work, we should bring the issue to the semconv SIG.

bjrara:
> As discussed in SIG (09/26) -- @bjrara will try using the LastValue aggregation in a View to configure the SDK to output this metric as a gauge. If it doesn't work, we should bring the issue to the semconv SIG.

The proposed solution works for me. Thanks to @emdneto and @aabmass for the suggestions. I will update the issue with the summary why we believe the existing instrument types are right, and how to work around to report metrics to systems that only accept delta.

PR #2879, opened 2024-09-23
feat: Add path parameter rendering feature

jagerzhang:
# Description

In the current version, path parameters are not rendered properly in the span name (e.g., `GET /api/{userId}`), leading to confusion for users who expect to see the relevant path information. This issue particularly affects scenarios where dynamic paths are generated.

To address this, a new optional boolean parameter, render_path_parameters, has been added to enable or disable the rendering of path parameters in the span name:
 
```python
otel_fastapi.FastAPIInstrumentor().instrument_app(app, render_path_parameters=True)
```

With this change, the spanName `GET /api/{userId}` will be recorded as `GET /api/123456`.

## Type of change

# How Has This Been Tested?

Unit tests have been added to ensure that path parameters render correctly in different scenarios.

- [x] `TestFastAPIRenderPathParameters` in ’tests/test_fastapi_instrumentation_wrapped.py`

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


emdneto (commented):
Thanks for the PR. 

Some specs guidance:
>HTTP span names SHOULD be {method} {target} if there is a (low-cardinality) target available. The {target} SHOULD be [http.route](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/) for HTTP Server spans. In that case, `http.route` is the the matched route, that is, the path template in the format used by the respective server framework (e.g., /users/:userID?; {controller}/{action}/{id?}) and the route attribute should have low-cardinality.

So, a genuine question: by providing that support are we not allowing high-cardinality span names?

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -1839,3 +1839,50 @@ def test_custom_header_not_present_in_non_recording_span(self):
         self.assertEqual(200, resp.status_code)
         span_list = self.memory_exporter.get_finished_spans()
         self.assertEqual(len(span_list), 0)
+
+
+class TestFastAPIRenderPathParameters(TestBaseFastAPI):
```
emdneto, line 1844: this can't be a new test in any of other test base class? For example, TestWrappedApplication

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -1839,3 +1839,50 @@ def test_custom_header_not_present_in_non_recording_span(self):
...
+            app, render_path_parameters=True)
+        client = TestClient(app)
+
+        # Make a request to the endpoint with a path parameter
+        response = client.get("/user/johndoe")
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), {"username": "johndoe"})
+
+        # Retrieve the spans generated
+        spans = self.memory_exporter.get_finished_spans()
```
emdneto, line 1867: Are we missing assert the span.name?

jagerzhang:
> Thanks for the PR.
> 
> Some specs guidance:
> 
> > HTTP span names SHOULD be {method} {target} if there is a (low-cardinality) target available. The {target} SHOULD be [http.route](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/) for HTTP Server spans. In that case, `http.route` is the the matched route, that is, the path template in the format used by the respective server framework (e.g., /users/:userID?; {controller}/{action}/{id?}) and the route attribute should have low-cardinality.
> 
> So, a genuine question: by providing that support are we not allowing high-cardinality span names?

Thanks for the CR and your response.

Indeed, there can be issues with high cardinality, which is why this feature is designed to be disabled by default. It can be enabled as needed in specific scenarios, ensuring that it does not affect regular usage.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
jagerzhang: Sorry, this is my first time working with OpenTelemetry and this project, so I'm not very familiar with it. I have moved the test code to `test_fastapi_instrumentation_wrapped.py`. Could you please take another look and see if there are any other issues?

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
jagerzhang: Assertions have been added. Could you please take another look?

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
emdneto: No worries. I mean under an existent test class instead of creating a new one

jagerzhang: I understand what you mean, I have edited it.

jagerzhang:
Hi, Can it be merged now?

tammy-baylis-swi:
Thanks for contributing and discussing!

There is a breakdown of the same userId example in the table [here](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.37.0/specification/trace/api.md#span):

> <p dir="auto" style="box-sizing: border-box; margin-top: 0px; margin-bottom: var(--base-size-16); color: rgb(31, 35, 40); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, &quot;Noto Sans&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">For example, here are potential span names for an endpoint that gets a hypothetical account information:</p><markdown-accessiblity-table data-catalyst="" style="box-sizing: border-box; display: block; color: rgb(31, 35, 40); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, &quot;Noto Sans&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">
Span Name | Guidance
-- | --
get | Too general
get_account/42 | Too specific
get_account | Good, and account_id=42 would make a nice Span attribute
get_account/{accountId} | Also good (using the "HTTP route")

</markdown-accessiblity-table>

My recommendation would be to follow and include userId as a span attribute instead of part of the span name, and prioritize generality over human readability.

jagerzhang:
> Thanks for contributing and discussing!
> 
> There is a breakdown of the same userId example in the table [here](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.37.0/specification/trace/api.md#span):
> 
> > For example, here are potential span names for an endpoint that gets a hypothetical account information:
> 
> Span Name	Guidance
> get	Too general
> get_account/42	Too specific
> get_account	Good, and account_id=42 would make a nice Span attribute
> get_account/{accountId}	Also good (using the "HTTP route")
> My recommendation would be to follow and include userId as a span attribute instead of part of the span name, and prioritize generality over human readability.

Got it.

This may be mainly because our usage scenarios are quite special. For example, we use the following API routing:

`/jobs/{biz_name}/{module}/{function}`

To call specific business logic, we hope that span name can uniquely identify and track the execution process or results of specific business logic, such as

```shell
/jobs/foo_biz/foo_module/foo_func
/jobs/bar_biz/bar_module/bar_func
```

We hope there are two span records.

Moreover, this PR feature is turned off by default, and can only be enabled by actively setting parameters when needed. I think it does not pose a challenge to the standard, and I hope it can be considered.

Of course, everyone felt that it did not meet the standards, so I had to use callback functions such as request_hook to achieve the goal.

```python
def fastapi_server_request_hook(span, request):
    method = request.get("method")
    path = request.get("path")
    span.update_name(f"{method} {path}")

```

tammy-baylis-swi:
Hi @jagerzhang , yes that's an interesting use case. I suggest continuing to use the request hooks, either for setting span name like your example, or for setting one or more custom span attributes that your backend can use to filter. It could be a single "full http route" attribute, or by one each for `biz_name`, `module`, `function`.

jagerzhang:
> Hi @jagerzhang , yes that's an interesting use case. I suggest continuing to use the request hooks, either for setting span name like your example, or for setting one or more custom span attributes that your backend can use to filter. It could be a single "full http route" attribute, or by one each for `biz_name`, `module`, `function`.

ok, I closing this pr now.

Issue #2870, opened 2024-09-13
opentelemetry-instrumentation-system-metrics: collecting fails on windows because of missing num_fds


jenshnielsen:
### Describe your environment

OS: Windows 11
Python version: Any but specifically 3.11
Package version: (e.g., 0.48b0) of opentelemetry-instrumentation-system-metrics

### What happened?

Enabling system metrics with the default configuration result in errors of the form `AttributeError: 'Process' object has no attribute 'num_fds'` during metrics collection on windows

This metrics was implemented in #2652 which was first part of 0.480b of opentelemetry-instrumentation-system-metrics

### Steps to Reproduce

```python
from opentelemetry.sdk.metrics import MeterProvider
from opentelemetry.sdk.metrics.export import ConsoleMetricExporter, PeriodicExportingMetricReader
from opentelemetry import metrics
exporter = ConsoleMetricExporter()
reader = PeriodicExportingMetricReader(exporter=exporter)
provider = MeterProvider(metric_readers=[reader])
metrics.set_meter_provider(provider)
from opentelemetry.instrumentation.system_metrics import SystemMetricsInstrumentor
SystemMetricsInstrumentor().instrument()
```

And wait for metrics to be collected and you will see an error as below

### Expected Result

No errors during metrics collection. Either metric is off by default or implemented differently on window

### Actual Result

Error during metrics collection

```
Callback failed for instrument process.open_file_descriptor.count.
Traceback (most recent call last):
  File "C:\Users\jenielse\Miniconda3\envs\myenv\Lib\site-packages\opentelemetry\sdk\metrics\_internal\instrument.py", line 137, in callback
    for api_measurement in callback(callback_options):
  File "C:\Users\jenielse\Miniconda3\envs\myenv\Lib\site-packages\opentelemetry\instrumentation\system_metrics\__init__.py", line 415, in _get_open_file_descriptors
    self._proc.num_fds(),
    ^^^^^^^^^^^^^^^^^^
AttributeError: 'Process' object has no attribute 'num_fds'
```

Since psutils does not implement num_fds on windows

### Would you like to implement a fix?

None

xrmx:
Thanks for reporting, that means that we are not running tests for this instrumentation on windows I guess :sweat:

shijiadong2022:
Will a check of sys.platform be sufficient OR need to collect this metric on Windows?

jenshnielsen:
[...]

PR #2895, opened 2024-10-05
Fix collecting fails on windows because of missing num_fds for opentelemetry-instrumentation-system-metrics

shijiadong2022:
Fixes # ([2870](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2870))


xrmx:
Care to add a test that exercise this?

PR #2897, opened 2024-10-09
Add DB-API instrumentor support for MySQL driver sqlcommenting

tammy-baylis-swi:
# Description

Updates the shared DB-API instrumentor for continued support of psycopg2/psycopg instrumentor sqlcommenting, and adds prerequisite logic for support of MySQL driver instrumentor sqlcommenting. Motivation is to implement sqlcommenting later for [mysql-connector](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-mysql), [mysqlclient](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-mysqlclient), [pymysql](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-pymysql) instrumentors. Getting the names and versions of these modules is not consistent, therefore several "if/elif" added.

Fixes a small bug ("breaking change"?) where sqlcomment for both psycopg2/psycopg query always includes the KV `db_driver='psycopg2<VERSION>`. Now it will include either `db_driver='psycopg2<VERSION>` or `db_driver='psycopg<VERSION>` respectively.

Adds an optimization to do most of `calculate_commenter_data` at init of `DatabaseApiIntegration`, instead of at every `CursorTracer.traced_execution`.

Fixes [#2902](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2902) partially -- other PRs to follow

## Type of change

# How Has This Been Tested?

Updated and added unit tests. Used local installations of `opentelemetry-instrumentation-dbapi` and upstream dependencies with downstream instrumentors of mysql-connector, mysqlclient, pymysql, psycopg2, psycopg, sqlalchemy on apps that use respective db client drivers/ORM to query MySQL/PostgreSQL with general logs enabled to check sqlcomments.

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -275,6 +280,60 @@ def __init__(
...
+        self,
+    ):
+        commenter_data = {}
+        if not self.enable_commenter:
+            return commenter_data
+
+        try:
+            db_driver = self.connect_module.__name__
+        except AttributeError:
+            db_driver = "unknown"
```
emdneto, line 295: Maybe this instead?
```suggestion
        db_driver = getattr(self.connect_module, '__name__', 'unknown')
```

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -275,6 +280,60 @@ def __init__(
...
+
+        db_version = ""
+        if db_driver in _DB_DRIVER_ALIASES:
+            db_version = util_version(_DB_DRIVER_ALIASES[db_driver])
+        else:
+            try:
+                db_version = self.connect_module.__version__
+            except AttributeError:
+                db_version = "unknown"
```
emdneto, line 304: Wdyt separate this block in a dedicated function, like: `_get_db_version(db_driver_alias)`

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -275,6 +280,60 @@ def __init__(
...
+
+        commenter_data = {
+            "db_driver": f"{db_driver}:{db_version.split(' ')[0]}",
+            "dbapi_threadsafety": self.connect_module.threadsafety,
```
emdneto, line 308: Do we have default values for this in case of missing attributes?

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
tammy-baylis-swi: Good idea! Changed in [c461544](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2897/commits/c461544d226437e552a1c6ffdabf8bf406fb9f38)

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
tammy-baylis-swi: Sure! Changed in [c461544](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2897/commits/c461544d226437e552a1c6ffdabf8bf406fb9f38)

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
tammy-baylis-swi: I don't think we do, and I just copied this from what was previously in place. I'll look into making this better.

tammy-baylis-swi: In [daf3a73](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2897/commits/daf3a739aa545f39af6d34523b06f64c96e3002a) I've added fallbacks, a comment, and an additional test.

As far as I know, all of psycopg2, mysql-connector, etc database drivers that we're instrumenting/preparing to instrument here should be [PEP 249](https://peps.python.org/pep-0249/#globals) compliant and have the 3 globals for threadsafety, apilevel, and paramstyle. But if they don't, we set default of `"1.0"` for apilevel [because](https://peps.python.org/pep-0249/#apilevel)

> If not given, a DB-API 1.0 level interface should be assumed.

My interpretation is that the others don't have defaults, so we use `"unknown"` to prevent crash from AttributeError.

Let me know if that works!

emdneto (approved):

PR #2909, opened 2024-10-16
httpx: rewrite patching to use wrapt instead of subclassing client

### LINKED ISSUE #2364: HttpX instrumentation does not work on classes that extend httpx client
Becuase the HttpX instrumentation changes the httpx.client class, it does not work on classes that are defined on import (even if the class is only instantiated after instrumentation). This means that as soon as OpenAI created an extension of the HttpX client, the httpx instrumentation [stopped working](https://github.com/openai/openai-python/issues/1144). This could be fixed in OpenAI by defining the class at runtime:

```python
class SyncAPIClient(BaseClient[httpx.Client, Stream[Any]]):
    ...
    def __init__(
        ...
    ) -> None:
        ...
        # Define at runtime
        class SyncHttpxClientWrapper(httpx.Client):
            def __del__(self) -> None:
                try:
                    self.close()
                except Exception:
                    pass
        self._client = http_client or SyncHttpxClientWrapper(
            base_url=base_url,
            # cast to a valid type because mypy doesn't understand our type narrowing
            timeout=cast(Timeout, timeout),
            proxies=proxies,
            transport=transport,
            limits=limits,
            follow_redirects=True,
        )
```

This could also be solved by instrumenting httpx even before importing any library that uses httpx. However, I think these restrictions mean that the HttpX instrumentation is too fragile. We need to improve it so that it works intuitively for all such scenarios.

**Describe your environment**
Windows
opentelemetry-api                      1.23.0
opentelemetry-instrumentation          0.44b0
opentelemetry-instrumentation-httpx    0.44b0
opentelemetry-instrumentation-openai   0.14.1
opentelemetry-sdk                      1.23.0
opentelemetry-semantic-conventions     0.44b0
opentelemetry-semantic-conventions-ai  0.0.23
opentelemetry-util-http                0.44b0

**Steps to reproduce**
```python
from openai import OpenAI # 1.x
import httpx
from opentelemetry.instrumentation.openai import OpenAIInstrumentor

url = "https://www.example.org/"
     response = client.get(url)

OpenAIInstrumentor().instrument()

client = OpenAI() # 1.x
completion = client.chat.completions.create( # 1.x
  model="gpt-3.5-turbo",
  messages=[
    {"role": "system", "content": "You are a poetic assistant, skilled in explaining complex programming concepts with creative flair."},
    {"role": "user", "content": "Compose a poem that explains the concept of recursion in programming."}
  ]
)
print(completion.choices[0].message)

input()
```

api.openai.com POST should be captured. Note that this is separate from the openai.chat span captured by the openai instrumentation.

Only the httpx example span and openai.chat spans are collected.

**Additional context**

### LINKED ISSUE #2609: Subclassing httpx causing `isinstance` checks failed
### Describe your environment

OS: Windows-10-10.0.22631-SP0
Python version: Python 3.12.3
Package version: opentelemetry-instrumentation-httpx="0.46b0"

### What happened?

I instrumented httpx **after some httpx clients being initialized**, and then `isinstance(client, Client)` failed.

I admit that instrumenting code should ideally be placed before any other code, but I hope that even if it is not done this way, it will result in no errors if possible.

Related to https://github.com/pydantic/logfire/issues/264

### Steps to Reproduce

```py
from httpx import Client
client = Client()
from httpx import Client
isinstance(client, Client)
```

### Expected Result

True

### Actual Result

False

### Additional context

Maybe we can use an non-subclassing approach, like patching methods on the original class.

Being open to

xrmx:
# Description

Porting of httpx instrumentation to patch async transport methods instead of substituting the client. That is because the current approach will instrument httpx by instantianting another client with a custom transport class and this will race with code already subclassing. This one uses wrapt to patch the default httpx transport classes.

A few changes of behavior to discuss:
- we are patching the two default httpx transport classes, if you subclass something else it still won't be instrumented, still instrument_client should work
- global uninstrument() will remove instrumentation from client instrumented with instrument_client but I think that's legit
- would be nice to deprecate the client subclassing thing and remove a ton of code

There is still a couple of FIXME that would like to get a second opinion.

Fixes #2364 
Fixes #2609

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
@@ -736,8 +736,18 @@ class HTTPXClientInstrumentor(BaseInstrumentor):
         tracer_provider = kwargs.get("tracer_provider")
         self._request_hook = kwargs.get("request_hook")
         self._response_hook = kwargs.get("response_hook")
-        self._async_request_hook = kwargs.get("async_request_hook")
-        self._async_response_hook = kwargs.get("async_response_hook")
+        _async_request_hook = kwargs.get("async_request_hook")
+        self._async_request_hook = (
+            _async_request_hook
+            if iscoroutinefunction(_async_request_hook)
+            else None
+        )
+        _async_response_hook = kwargs.get("async_response_hook")
+        self._async_response_hook = (
+            _async_response_hook
+            if iscoroutinefunction(_async_response_hook)
+            else None
+        )
 
         _OpenTelemetrySemanticConventionStability._initialize()
         self._sem_conv_opt_in_mode = _OpenTelemetrySemanticConventionStability._get_opentelemetry_stability_opt_in_mode(
@@ -829,7 +839,7 @@ class HTTPXClientInstrumentor(BaseInstrumentor):
                     span.set_attribute(
                         ERROR_TYPE, type(exception).__qualname__
                     )
-                raise exception.with_traceback(exception.__traceback__)
+                raise exception
 
         return response
 
@@ -898,7 +908,7 @@ class HTTPXClientInstrumentor(BaseInstrumentor):
                     span.set_attribute(
                         ERROR_TYPE, type(exception).__qualname__
                     )
-                raise exception.with_traceback(exception.__traceback__)
+                raise exception
 
         return response
 
@@ -930,6 +940,19 @@ class HTTPXClientInstrumentor(BaseInstrumentor):
             )
             return
 
+        # FIXME: sharing state in the instrumentor instance maybe it's not that great, need to pass tracer and semconv to each
+        # instance separately
+        _OpenTelemetrySemanticConventionStability._initialize()
+        self._sem_conv_opt_in_mode = _OpenTelemetrySemanticConventionStability._get_opentelemetry_stability_opt_in_mode(
+            _OpenTelemetryStabilitySignalType.HTTP,
+        )
+        self._tracer = get_tracer(
+            __name__,
+            instrumenting_library_version=__version__,
+            tracer_provider=tracer_provider,
+            schema_url=_get_schema_url(self._sem_conv_opt_in_mode),
+        )
+
         if iscoroutinefunction(request_hook):
             self._async_request_hook = request_hook
             self._request_hook = None
@@ -950,6 +973,13 @@ class HTTPXClientInstrumentor(BaseInstrumentor):
                 "handle_request",
                 self._handle_request_wrapper,
             )
+            for transport in client._mounts.values():
+                # FIXME: check it's not wrapped already?
+                wrap_function_wrapper(
+                    transport,
+                    "handle_request",
+                    self._handle_request_wrapper,
+                )
             client._is_instrumented_by_opentelemetry = True
         if hasattr(client._transport, "handle_async_request"):
             wrap_function_wrapper(
@@ -957,6 +987,13 @@ class HTTPXClientInstrumentor(BaseInstrumentor):
                 "handle_async_request",
                 self._handle_async_request_wrapper,
             )
+            for transport in client._mounts.values():
+                # FIXME: check it's not wrapped already?
+                wrap_function_wrapper(
+                    transport,
+                    "handle_async_request",
+                    self._handle_async_request_wrapper,
+                )
             client._is_instrumented_by_opentelemetry = True
 
     @staticmethod
@@ -970,7 +1007,11 @@ class HTTPXClientInstrumentor(BaseInstrumentor):
         """
         if hasattr(client._transport, "handle_request"):
             unwrap(client._transport, "handle_request")
+            for transport in client._mounts.values():
+                unwrap(transport, "handle_request")
             client._is_instrumented_by_opentelemetry = False
         elif hasattr(client._transport, "handle_async_request"):
             unwrap(client._transport, "handle_async_request")
+            for transport in client._mounts.values():
+                unwrap(transport, "handle_async_request")
             client._is_instrumented_by_opentelemetry = False
--- a/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
@@ -194,6 +194,7 @@ API
 import logging
 import typing
 from asyncio import iscoroutinefunction
+from functools import partial
 from types import TracebackType
 
 import httpx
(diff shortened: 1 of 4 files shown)


instrumentation/opentelemetry-instrumentation-httpx/tests/test_httpx_integration.py
```diff
@@ -741,8 +743,10 @@ def create_proxy_transport(self, url: str):
 
         def setUp(self):
             super().setUp()
-            HTTPXClientInstrumentor().instrument()
             self.client = self.create_client()
+            HTTPXClientInstrumentor().instrument_client(self.client)
+
+        def tearDown(self):
```
xrmx, line 749: This is the change of behavior, before uninstrument() did not touch already instrumented clients

instrumentation/opentelemetry-instrumentation-httpx/tests/test_httpx_integration.py
```diff
@@ -1010,13 +1012,13 @@ def test_uninstrument_client_with_proxy(self):
...
                 client._mounts.values(),
                 2,
-                (SyncOpenTelemetryTransport, AsyncOpenTelemetryTransport),
             )
 
             HTTPXClientInstrumentor().uninstrument_client(client)
             result = self.perform_request(self.URL, client=client)
 
             self.assertEqual(result.text, "Hello!")
+            # FIXME: this does fail if uninstrument() has been called before and is a change of behaviour from before
```
xrmx, line 1021:
```suggestion
```

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -731,44 +735,211 @@ def _instrument(self, **kwargs):
...
-        # Intentionally using a private attribute here, see:
-        # https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2538#discussion_r1610603719
-        httpx.Client = httpx._api.Client = _InstrumentedClient
-        httpx.AsyncClient = _InstrumentedAsyncClient
+            if iscoroutinefunction(async_response_hook)
+            else None
+        )
+
+        _OpenTelemetrySemanticConventionStability._initialize()
+        sem_conv_opt_in_mode = _OpenTelemetrySemanticConventionStability._get_opentelemetry_stability_opt_in_mode(
```
lzchen, line 755: Nice catch with this.

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -788,67 +959,90 @@ def instrument_client(
...
+            _OpenTelemetryStabilitySignalType.HTTP,
+        )
+        tracer = get_tracer(
+            __name__,
+            instrumenting_library_version=__version__,
+            tracer_provider=tracer_provider,
+            schema_url=_get_schema_url(sem_conv_opt_in_mode),
+        )
+
+        if iscoroutinefunction(request_hook):
```
lzchen, line 980: Not sure if we want to introduce this behavior (auto setting request hook to async if it is indeed an async function). What happens if they are both are provided as an async function? Wouldn't it be confusing if `request_hook` overrides `async_request_hook`?

instrumentation/opentelemetry-instrumentation-httpx/tests/test_httpx_integration.py
lzchen: I believe this is fine.

lzchen (approved):
Some small comments but overall lgtm

emdneto (approved):

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
xrmx: This was already the behavior of `instrument_client`, agree that it's not great that it behaves differently than plain `instrument`. See old code at https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2909/files/f0880e65765f01f761bd3c2ff691e6414c53d4f2#diff-eed55d62a8d5f01b055c1bb2d63934c48cfd767416c03826f2c19d4442a76e7bL839

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -731,44 +735,211 @@ def _instrument(self, **kwargs):
...
+                        request_info,
+                        ResponseInfo(status_code, headers, stream, extensions),
+                    )
+
+            if exception:
+                if span.is_recording() and _report_new(sem_conv_opt_in_mode):
+                    span.set_attribute(
+                        ERROR_TYPE, type(exception).__qualname__
+                    )
+                raise exception
```
xrmx, line 862:
```suggestion
                raise exception.with_traceback(exception.__traceback__)
```

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -731,44 +735,211 @@ def _instrument(self, **kwargs):
...
```
xrmx, line 937:
```suggestion
                raise exception.with_traceback(exception.__traceback__)
```

tammy-baylis-swi (approved):

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -788,67 +959,90 @@ def instrument_client(
...
+                partial(
+                    self._handle_request_wrapper,
+                    tracer=tracer,
+                    sem_conv_opt_in_mode=sem_conv_opt_in_mode,
                     request_hook=request_hook,
                     response_hook=response_hook,
+                ),
+            )
+            for transport in client._mounts.values():
+                # FIXME: check it's not wrapped already?
```
xrmx, line 1007:
```suggestion
```

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -788,67 +959,90 @@ def instrument_client(
...
+                partial(
+                    self._handle_async_request_wrapper,
+                    tracer=tracer,
+                    sem_conv_opt_in_mode=sem_conv_opt_in_mode,
+                    async_request_hook=async_request_hook,
+                    async_response_hook=async_response_hook,
+                ),
             )
+            for transport in client._mounts.values():
+                # FIXME: check it's not wrapped already?
```
xrmx, line 1033:
```suggestion
```

instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -731,44 +735,211 @@ def _instrument(self, **kwargs):
...
+                if span.is_recording() and _report_new(sem_conv_opt_in_mode):
+                    span.set_attribute(
+                        ERROR_TYPE, type(exception).__qualname__
+                    )
+                raise exception.with_traceback(exception.__traceback__)
+
+        return response
+
     def instrument_client(
+        self,
```
alexmojaki, line 942: Why is this no longer a staticmethod? Seems like a significant breaking change that wasn't documented in the release notes

robotadam: I raised this as an issue in #2998 and it was fixed in #3003.

PR #2919, opened 2024-10-21
opentelemetry-instrumentation: add unwrapping from dotted paths strings

xrmx:
# Description

Make it possible to express object to unwrap as dotted module paths strings. This helps in avoiding side effects or race conditions with other instrumentations if we do importing too early.

While at it add tests also for current functionality.

## Type of change

# How Has This Been Tested?

- [x] tox

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
@@ -14,6 +14,7 @@
 
 import urllib.parse
 from contextlib import contextmanager
+from importlib import import_module
 from re import escape, sub
 from typing import Dict, Iterable, Sequence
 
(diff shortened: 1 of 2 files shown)


opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -83,10 +84,27 @@ def http_status_to_status_code(
 def unwrap(obj, attr: str):
```
emdneto, line 84:
```suggestion
def unwrap(obj: Union[object, str], attr: str):
```
Maybe we can type this too?

xrmx: Added, thanks

emdneto (approved):

PR #2921, opened 2024-10-22
Introducing ruff as linter and formatter (basic setup)

### LINKED ISSUE #2449: Investigate better tools for linting
There may be better tools for linting, investigate that.

emdneto:
# Description

Same as https://github.com/open-telemetry/opentelemetry-python/pull/4223, but added ruff to `eachdist.py` utility script as well in `lint`and `format` commands.

Closes #2449


scripts/eachdist.py
```diff
@@ -724,16 +726,15 @@ def format_args(args):
...
         args.dry_run,
-        ("black", "--config", f"{root_dir}/pyproject.toml", "."),
+        ("ruff", "check", "--fix"),
         cwd=format_dir,
         check=True,
     )
     runsubprocess(
         args.dry_run,
-        ("isort", "--settings-path", f"{root_dir}/.isort.cfg", "--profile", "black", "."),
+        ("ruff", "format"),
```
xrmx, line 737: Have you checked it is finding the config file at the root of the dir?

emdneto: Yes! 

You can run in any directory with verbose mode `ruff format instrumentation/opentelemetry-instrumentation-wsgi -v` for example

```
[2024-10-22][13:11:17][ruff::resolve][DEBUG] Using configuration file (via parent) at: opentelemetry-python-contrib/pyproject.toml
```
>Similar to [ESLint](https://eslint.org/docs/latest/use/configure/configuration-files#cascading-configuration-objects), Ruff supports hierarchical configuration, such that the "closest" config file in the directory hierarchy is used for every individual file, with all paths in the config file (e.g., exclude globs, src paths) being resolved relative to the directory containing that config file. 

https://docs.astral.sh/ruff/configuration/#config-file-discovery

emdneto (commented):
Actual changed files https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2921/files/5f384996d9a0a8d65ea72b14012be29019457fd4

pyproject.toml
```diff
@@ -1,11 +1,41 @@
...
+]
+ignore = [
+  "E501", # line-too-long
+]
+
+[tool.ruff.lint.per-file-ignores]
+"docs/**/*.*" = ["A001"]
+
+[tool.ruff.lint.isort]
+detect-same-package = false # to not consider instrumentation packages as first-party
```
emdneto, line 32: Had to add this parameter https://docs.astral.sh/ruff/settings/#lint_isort_detect-same-package
to tell ruff to avoid considering instrumentation packages as first-party. One other option is to use known-third-party with all packages we instrument (e.g, known-third-party = ["sqlalchemy", "httpx", ...]

xrmx (approved):
~This switches opentelemetry from third to first party but~ LGTM

lzchen (approved):

PR #2922, opened 2024-10-22
Fix unit in system metrics

bjrara:
# Description

Convert system metric to use UCUM to follow [OpenTelemetry semantic conventions](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/)

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#2874](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2874))
 - `opentelemetry-instrumentation-confluent-kafka` Fix to allow `topic` to be extracted from `kwargs` in `produce()` 
   ([#2901])(https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2901)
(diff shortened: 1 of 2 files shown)


xrmx:
It can't hurt I guess, but aren't the names of the metrics out of spec?

xrmx:
There is at least one other "bytes" usage, please update that too

bjrara:
> There is at least another "bytes" usage, please update that too

Fixed it. Thank you!

bjrara:
> It can't hurt I guess, but aren't the names of the metrics out of spec?

They aren't. But I guess this is the legacy we will need to carry until we find a deprecation path. For metric units fixes, I believe it is a reasonable change because the OTEL dependency services won't expect the units as `bytes` or `seconds` since it doesn't conform to the convention.

emdneto (approved):
LGTM since we are not changing metric names, ~but worried about this comment in semconv~:

~>SHOULD NOT adopt any breaking changes from document until the system semantic conventions are marked stable. Conventions include, but are not limited to, attributes, metric names, and unit of measure.~

~We are including unit of measure changes~

PR #2925, opened 2024-10-26
OpenAI - update semantic conventions to the latest version, write logs based events instead of span events

lmolkova:
# Description

- Update OpenAI semantic conventions to the 1.28.0 
  - changes span events to log based events 
  - adds a few new openai-specific attributes
  - adds opt-in flag for event contents
- Add tests

TODOs:
- [x] we need opt-in flag for contents 
- [x] changelog

## Type of change

# How Has This Been Tested?

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-openai-v2/tests/conftest.py
```diff
@@ -58,7 +81,8 @@ def instrument():
 
 @pytest.fixture(scope="session", autouse=True)
 def uninstrument():
-    OpenAIInstrumentor().uninstrument()
+    # OpenAIInstrumentor().uninstrument()
```
lmolkova, line 84: TODO: This was running before the test, will need to figure out how to do clean up right

CONTRIBUTING.md
```diff
@@ -227,7 +227,7 @@ tox -e py312-test-instrumentation-aiopg
 Some of the tox targets install packages from the [OpenTelemetry Python Core Repository](https://github.com/open-telemetry/opentelemetry-python) via pip. The version of the packages installed defaults to the main branch in that repository when tox is run locally. It is possible to install packages tagged with a specific git commit hash by setting an environment variable before running tox as per the following example:
 
 ```sh
-CORE_REPO_SHA=c49ad57bfe35cfc69bfa863d74058ca9bec55fc3 tox
+CORE_REPO_SHA=966750aad4ba2a9c123b4f14785958ffe50b54ae tox
```
xrmx, line 230: I'm not sure we want to keep this up to date :sweat_smile:

instrumentation/opentelemetry-instrumentation-openai-v2/tests/conftest.py
xrmx: How does it fail?

xrmx: To me the correct way to handle this would be to do it per test:

```
@pytest.fixture(autouse=True)
def instrument():
    instrumentor = OpenAIInstrumentor()
    instrumentor.instrument()

    yield instrumentor

    instrumentor.uninstrument()
```

instrumentation/opentelemetry-instrumentation-openai-v2/tests/test_chat_completions.py
```diff
@@ -102,56 +356,385 @@ def test_chat_completion_streaming(exporter, openai_client):
...
+        log.log_record.attributes[GenAIAttributes.GEN_AI_SYSTEM]
+        == GenAIAttributes.GenAiSystemValues.OPENAI.value
     )
-    completion_event = list(
-        filter(
-            lambda event: event.name == "gen_ai.content.completion",
-            events,
-        )
+    assert json.dumps(log.log_record.body, sort_keys=True) == json.dumps(
+        expected_content, sort_keys=True
```
xrmx, line 634:
```suggestion
    assert dict(log.log_record.body) == expected_content
```

Have you tried something like this? It's cheaper and the comparison is specialized so key order does not matter and you get the key that is different highlighted already

instrumentation/opentelemetry-instrumentation-openai-v2/tests/test_chat_completions.py
```diff
@@ -102,56 +356,385 @@ def test_chat_completion_streaming(exporter, openai_client):
...
+    response = openai_client.chat.completions.create(
+        messages=messages_value,
+        model=llm_model_value,
+        tool_choice="auto",
+        tools=[get_current_weather_tool_definition()],
+        stream=True,
+        stream_options={"include_usage": True},
+    )
+
+    finish_reason = None
```
xrmx, line 546: Since these are recorded shouldn't we just hardcode the numbers later instead? this way we are copying the implementation that may hide any issue.

instrumentation/opentelemetry-instrumentation-openai-v2/tests/conftest.py
lmolkova: thanks! will try!  how it failed - the instrumentation was disabled (since uninstrument was called before test started),

CONTRIBUTING.md
lmolkova: I need a few features from the not-yet-released core (semconv 1.28.0 and event SDK for tests). I think I can work around both needs by hardcoding attribute names and implementing custom event SDK in tests.

but out of curiosity, how do we normally handle things like this - wait for the core repo to release?

A related question - it seems we last released core two months ago - is a new release planned any time soon or should I do the workarounds?

instrumentation/opentelemetry-instrumentation-openai-v2/tests/test_chat_completions.py
lmolkova: thanks, will try!

CONTRIBUTING.md
xrmx: By default it should use main so you won't need to touch anything usually

instrumentation/opentelemetry-instrumentation-openai-v2/tests/test_chat_completions.py
lmolkova: if we hardcode everything, then it'd become a problem if we ever need to re-record. From my experience working with test recordings the need comes every once in a while.

CONTRIBUTING.md
lzchen: > but out of curiosity, how do we normally handle things like this - wait for the core repo to release?

The SHA points to a commit in the core repo (usually unreleased when you change it) to pull dev changes from the other repository that is needed for the current PR. As @xrmx has mentioned, USUALLY you can set it to `main` to pick up the latest changes from core. In the off chance that you make two simultaneous PRs in both core and contrib that need changes from each other, this is when we usually set the SHA to specific commits of the respective PRs.

> A related question - it seems we last released core two months ago - is a new release planned any time soon or should I do the workarounds?

We will probably do a release soon once this gets merged.

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/__init__.py
```diff
@@ -64,12 +65,19 @@ def _instrument(self, **kwargs):
             __name__,
             "",
             tracer_provider,
-            schema_url=Schemas.V1_27_0.value,
+            schema_url=Schemas.V1_28_0.value,
+        )
+        event_provider = kwargs.get("event_provider")
```
lzchen, line 70: Should this be `event_logger_provider`?

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/patch.py
```diff
@@ -25,97 +28,76 @@
...
     return traced_method
 
 
 @silently_fail
-def _set_input_attributes(span, attributes):
-    for field, value in attributes.items():
-        set_span_attribute(span, field, value)
-
+def _set_response_attributes(span, result, event_logger: EventLogger):
+    if not span.is_recording():
```
lzchen, line 87: Do we need this check if it already exists in `chat_completions_create`

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
```diff
@@ -43,69 +48,138 @@ def wrapper(*args, **kwargs):
...
+    body = {
+        "index": choice.index,
+        "finish_reason": choice.finish_reason or "error",
+    }
+
+    if choice.message:
+        message = {
+            "role": choice.message.role
+            if choice.message and choice.message.role
+            else "assistant"
```
lzchen, line 166: > Conditionally Required: if available and if not equal to assistant

Does that mean if it IS equal to "assistant", we do not populate this attribute? From the sem conv, it's not obvious to me that "assistant" is the default value if nothing is there.

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
```diff
@@ -43,69 +48,138 @@ def wrapper(*args, **kwargs):
...
+        }
+        tool_calls = extract_tool_calls(choice.message)
+        if tool_calls:
+            message["tool_calls"] = tool_calls
+        content = get_property_value(choice.message, "content")
```
lzchen, line 171: NIT: Put a TODO for opt-in functionality?

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
```diff
@@ -43,69 +48,138 @@ def wrapper(*args, **kwargs):
...
+        if content:
+            message["content"] = content
+        body["message"] = message
+
+    return Event(
+        name="gen_ai.choice",
```
lzchen, line 177: Nit: Should we have a collective place where we can refer to different event names (constants.py?) or might be good to add to sem conv.

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
```diff
@@ -43,69 +48,138 @@ def wrapper(*args, **kwargs):
...
+    tool_calls = get_property_value(item, "tool_calls")
     if tool_calls is None:
         return
 
     calls = []
     for tool_call in tool_calls:
         tool_call_dict = {
-            "id": getattr(tool_call, "id", ""),
-            "type": getattr(tool_call, "type", ""),
+            "id": get_property_value(tool_call, "id"),
```
lzchen, line 72: Could these possibly be `None`? Should we skip a call if any of these fields don't exist?

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
```diff
@@ -43,69 +48,138 @@ def wrapper(*args, **kwargs):
...
+        if func:
+            tool_call_dict["function"] = {}
+
+            name = get_property_value(func, "name")
+            if name:
+                tool_call_dict["function"]["name"] = name
+
+            arguments = get_property_value(func, "arguments")
+            if arguments:
+                tool_call_dict["function"]["arguments"] = arguments.replace(
```
lzchen, line 86: From the definition of [Function object](https://github.com/open-telemetry/semantic-conventions/blob/v1.28.0/docs/gen-ai/gen-ai-events.md#function-object), it seems like `arguments` can be `AnyValue` so `replace` would probably error out in those cases. Perhaps we simply leave the value as is?

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
```diff
@@ -43,69 +48,138 @@ def wrapper(*args, **kwargs):
...
+        return obj.get(property_name, None)
+
+    return getattr(obj, property_name, None)
+
+
+def message_to_event(message, span_ctx):
+    attributes = {
+        GenAIAttributes.GEN_AI_SYSTEM: GenAIAttributes.GenAiSystemValues.OPENAI.value
+    }
+    role = get_property_value(message, "role")
```
lzchen, line 127: Just curious, are there `role` values that are not currently defined in sem conv? (system, user, assistant, tool, etc)?

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/patch.py
```diff
@@ -144,27 +132,63 @@ def _set_response_attributes(span, result):
...
+                idx, tool_call.id, tool_call.function.name
+            )
+        self.tool_calls_buffers[idx].append_arguments(
+            tool_call.function.arguments
+        )
+
+
 class StreamWrapper:
     span: Span
-    response_id: str = ""
```
lzchen, line 149: I'm not too familiar with the concept of `Stream`. How does it conceptually map to a `Choice` or number of `Choice`s?

CONTRIBUTING.md
xrmx: > We will probably do a release soon once this gets merged.

We can release earlier since this instrumentation will be released separately on its own

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
lmolkova: image

https://github.com/open-telemetry/semantic-conventions/blob/v1.28.0/docs/gen-ai/gen-ai-events.md#assistant-event

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
lmolkova: `arguments` is a string in openai REST api and client.  Semconv leave it undefined for the future. I can check if it's a sting - I don't want to have new lines in the logs, so replacing is kinda important

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
lmolkova: not that I'm aware of, but I believe users can pass any role and it may work in some edge cases or with some non-OpenAI backends (a lot of backends are Open-AI compatible on the REST level and this instrumentation can be used with any of them).

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/patch.py
lmolkova: You can ask model to return a completion (one or more choices) using regular HTTP response body or you can ask it to stream tokens one-by-one - https://platform.openai.com/docs/api-reference/making-requests - in which case model returns the HTTP response and then writes tokens to the response stream using  SSE events. Check out  test recordings to see what happens there with and without streaming.

I.e. streaming affects how the response is received, but not the contents (choices).

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
lmolkova: good catch, removed default

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
lmolkova: added check for none, we should not really skip anything - let's just translate what we have into event

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
```diff
@@ -27,6 +28,18 @@
...
+    "OTEL_INSTRUMENTATION_OPENAI_CAPTURE_MESSAGE_CONTENT"
+)
+
+
+def is_content_enabled() -> bool:
+    capture_content = environ.get(
+        OTEL_INSTRUMENTATION_OPENAI_CAPTURE_MESSAGE_CONTENT, None
+    )
+
+    return bool(capture_content)
```
lzchen, line 39: NIt: Any truthy value will return true (any non-empty string would return true). Do we want that or do we want explicitly "true" string like we have for other [env vars](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/__init__.py#L132). Not a strong opinion though.

lzchen (approved):
LGTM!

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
lmolkova: great catch, thanks!

karthikscale3 (approved):
looking good. nice work @lmolkova

alizenhom:
looking good, thanks for doing this @lmolkova

alizenhom (approved):

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
```diff
@@ -12,117 +12,162 @@
...
 
-    # Return an empty string if none of the above conditions are met
-    else:
-        return ""
+def is_content_enabled() -> bool:
+    capture_content = environ.get(
+        OTEL_INSTRUMENTATION_OPENAI_CAPTURE_MESSAGE_CONTENT, None
+    )
+
+    return capture_content.lower() == "true" if capture_content else False
```
xrmx, line 40:
```suggestion
        OTEL_INSTRUMENTATION_OPENAI_CAPTURE_MESSAGE_CONTENT, "false"
    )

    return capture_content.lower() == "true"
```

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
```diff
@@ -12,117 +12,162 @@
...
+        if tool_call_id:
+            body["id"] = tool_call_id
+
+    return Event(
+        name=f"gen_ai.{role}.message",
+        attributes=attributes,
+        body=body if body else None,
+        trace_id=span_ctx.trace_id,
+        span_id=span_ctx.span_id,
+        trace_flags=span_ctx.trace_flags,
```
xrmx, line 128: These should be set by default at emit() time, are you adding them explicitly because they do not match?

xrmx (approved):

instrumentation/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
lmolkova: great catch! Changed to implicit context, except the streaming case where the span associated with the stream is not current after the original method returns.

PR #2990, opened 2024-11-11
Re-add testpypi publish step

lzchen:
We have reclaimed the TestPypi account credentials so readding into release process.


xrmx (approved):

emdneto (commented):
Is this a duplicate of #2991? Should it be re-added in -core instead?

PR #2984, opened 2024-11-08
Add Support for `Async` openai instrumentation

alizenhom:
# Description
- Add wrapper around `AsyncCompletions.create` to instrument async openai functions
- Adjust `StreamWrapper` to include `__anext__`, `__aiter__` & `__aenter__`

## Type of change
- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?
- [x] `tox -e py312-test-instrumentation-openai-v2-0`
- [x] `tox -e py312-test-instrumentation-openai-v2-1`

# Does This PR Require a Core Repo Change?
- [x] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [x] Followed the style guidelines of this project
- [x] Changelogs have been updated
- [x] Unit tests have been added
- [ ] Documentation has been updated

PR #2993, opened 2024-11-11
opentelemetry-instrumentation-openai-v2: scrub cookie from tests

codefromthecrypt:
# Description

This scrubs test data and rewords token replacements corresponding to canonical ENV variables.

Before, we had inconsistent test data where most sensitive properties were scrubbed, but not all. The main thing this does is scrub out the request cookie, so that folks don't accidentally leak it in new PRs. After that, it helps clarify the source of sensitive data by making the replacement tokens match naming conventions of OpenAI ENV variables.

Also, before we scrubbed "api-key", but that is only set when using `AzureClient`, which this package doesn't use, yet. When it does, we should probably scrub more (as the URLs include sensitive project-specific details). If we end up supporting `AzureClient`, I can help in the scrub config for that.

Noticed the cookie problem in #2984
See https://github.com/openai/openai-python/blob/646a579cdb305a9d3fba6c5f9a96011c5e2c2882/src/openai/_client.py#L98-L100

## Type of change

# How Has This Been Tested?

I deleted all test yaml and ran the test two ways to make sure it recreates correctly, with no cookies in the yaml
- [x] `tox -e py312-test-instrumentation-openai-v2-0`
- [x] `tox -e py312-test-instrumentation-openai-v2-1`

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/cassettes/test_chat_completion_tool_calls_no_content.yaml
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/cassettes/test_chat_completion_tool_calls_no_content.yaml
@@ -42,6 +42,8 @@ interactions:
       - application/json
       accept-encoding:
       - gzip, deflate
+      authorization:
+      - Bearer test_openai_api_key
       connection:
       - keep-alive
       content-length:
@@ -51,7 +53,7 @@ interactions:
       host:
       - api.openai.com
       user-agent:
-      - OpenAI/Python 1.26.0
+      - OpenAI/Python 1.54.3
       x-stainless-arch:
       - arm64
       x-stainless-async:
@@ -61,7 +63,9 @@ interactions:
       x-stainless-os:
       - MacOS
       x-stainless-package-version:
-      - 1.26.0
+      - 1.54.3
+      x-stainless-retry-count:
+      - '0'
       x-stainless-runtime:
       - CPython
       x-stainless-runtime-version:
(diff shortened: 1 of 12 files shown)


instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/cassettes/test_chat_completion_tool_calls_no_content.yaml
```diff
@@ -218,19 +226,20 @@ interactions:
...
       - gzip, deflate
+      authorization:
+      - Bearer test_openai_api_key
       connection:
       - keep-alive
       content-length:
       - '746'
       content-type:
       - application/json
       cookie:
```
codefromthecrypt, line 237: here's the intended and most important impact

instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/conftest.py
```diff
@@ -66,7 +66,12 @@ def openai_client():
 @pytest.fixture(scope="module")
 def vcr_config():
     return {
-        "filter_headers": ["authorization", "api-key"],
+        "filter_headers": [
+            ("cookie", "test_cookie"),
```
codefromthecrypt, line 70: here's the replacement and the way to add substitutions on request side (similar to how we do in response side)

codefromthecrypt (commented):
notes

alizenhom:
lgtm @codefromthecrypt

alizenhom (approved):

codefromthecrypt:
Thanks @alizenhom can you put the skip changelog label on for me? I don't have karma for that

karthikscale3 (approved):

xrmx (approved):
Good catch!

Issue #2995, opened 2024-11-12
Add openai instrumentation to opentelemetry-bootstrap


lzchen:
### What problem do you want to solve?

[bootstrap_gen.py](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap_gen.py) does not include the new [openai instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation-genai/opentelemetry-instrumentation-openai-v2) due to it being in a different folder. Update the scripts to allow automatic installing through `opentelemetry-bootstrap -a install`, which should automatically install the instrumentation if the user has `openai` installed.

PR #2996, opened 2024-11-12
Add openai instrumentation to opentelemetry-bootstrap

lzchen:
Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2995

Note this does NOT include openai instrumentation as a default installation but will install it if user has `openai` installed.


scripts/otel_packaging.py
```diff
@@ -21,14 +21,24 @@
...
 def get_instrumentation_packages():
-    for pkg in sorted(os.listdir(instrumentations_path)):
+    pkg_paths = []
+    for pkg in os.listdir(instrumentations_path):
         pkg_path = os.path.join(instrumentations_path, pkg)
         if not os.path.isdir(pkg_path):
             continue
+        pkg_paths.append(pkg_path)
+    for pkg in os.listdir(genai_instrumentations_path):
+        pkg_path = os.path.join(genai_instrumentations_path, pkg)
```
karthikscale3, line 35: nice, I guess this logic bootstraps the necessary packages?

karthikscale3 (approved):

scripts/otel_packaging.py
lzchen: @karthikscale3 

Yes! It scans the packages a user has that can be instrumented and installs the relevant instrumentation if exists. Usually the first step before running zero-code mechanisms like `opentelemetry-instrument`.

nirga (approved):

codefromthecrypt:
Thanks, so this adds the dep "opentelemetry-instrumentation-openai-v2", which users could also do manually if they don't use opentelemetry-bootstrap to setup the project. Sounds great!

lzchen:
@codefromthecrypt  yup! It all helps with the zero-code story.

Issue #795, opened 2021-11-09
FastAPI/Starlette - Inconsistent placement of the OpenTelemetryMiddleware


phillipuniverse:
The `Starlette` application class (which is the parent of the `FastAPI` application class) hardcodes the `ServerErrorMiddleware` to appear as the [very first middleware](https://github.com/encode/starlette/blob/6af5c515e0a896cbf3f86ee043b88f6c24200bcf/starlette/applications.py#L68-L92). This means that if there is an error in some downstream middleware, handling the request with `ServerErrorMiddleware` will not have an active trace.

Also, if you happen to do something like this:

```python
app = FastAPI(...)

app.add_middleware(...)
app.add_middlware(....)
```

The additional middleware added after the first `__init__` of `FastAPI` will also not have an active trace established.

There is a relatively easy way around this to ensure that the OpenTelemetry middleware is always first:

```python
class _InstrumentedFastAPI(fastapi.FastAPI):
    _tracer_provider = None
    _excluded_urls = None
    _server_request_hook: _ServerRequestHookT = None
    _client_request_hook: _ClientRequestHookT = None
    _client_response_hook: _ClientResponseHookT = None

    def build_middleware_stack(self) -> ASGIApp:
        original_mw_app = super().build_middleware_stack()
        return OpenTelemetryMiddleware(
            original_mw_app,
            excluded_urls=_InstrumentedFastAPI._excluded_urls,
            default_span_details=_get_route_details,
            server_request_hook=_InstrumentedFastAPI._server_request_hook,
            client_request_hook=_InstrumentedFastAPI._client_request_hook,
            client_response_hook=_InstrumentedFastAPI._client_response_hook,
            tracer_provider=_InstrumentedFastAPI._tracer_provider,
        )
```

FastAPI version - 0.70.0
Starlette version - 0.16.0
Python version - 3.9

phillipuniverse:
The workaround for now is to add an `opentelemetry_pre_instrument` entrypoint. Here is my patching function:

```python
import opentelemetry
import typing
from starlette.types import ASGIApp
from opentelemetry.instrumentation.asgi import OpenTelemetryMiddleware
from opentelemetry.instrumentation.fastapi import _InstrumentedFastAPI
from opentelemetry.instrumentation.fastapi import _get_route_details


def patch_middleware_ordering():
    """
    Replaces the default OpenTelemetry instrumentation class with one that is assured to always be added to the very
[...]

PR #3012, opened 2024-11-15
fastapi: fix wrapping of middlewares

adriangb:
MRE:

```python
import anyio
import fastapi
import httpx
import uvicorn
from opentelemetry.sdk.resources import SERVICE_NAME, Resource

resource = Resource(attributes={SERVICE_NAME: '🐛'})

traceProvider = TracerProvider(resource=resource)
traceProvider.add_span_processor(processor)
trace.set_tracer_provider(traceProvider)

app = fastapi.FastAPI()

@app.get('/ok')
async def ok() -> int:
    return 123

@app.get('/error')
async def error() -> int:
    raise Exception('An error occurred!')

async def main() -> None:
    config = uvicorn.Config(app=app, port=9999, log_config=None)
    server = uvicorn.Server(config)

    async with anyio.create_task_group() as tg:
        tg.start_soon(server.serve)
        await anyio.sleep(1)  # Wait for the server to start
        async with httpx.AsyncClient(base_url='http://localhost:9999') as client:
            response = await client.get('/ok')
            assert response.status_code == 200, f'{response.status_code}: {response.text}'
            response = await client.get('/error')
            assert response.status_code == 500,  f'{response.status_code}: {response.text}'
            tg.cancel_scope.cancel()


if __name__ == '__main__':
    anyio.run(main)
```

With this change it shows up properly

Run this and you'll see that there is no `"http.status_code": 500` in the logs.


Kludex (approved):

codefromthecrypt (approved):
nice comment and test!

xrmx:
@adriangb Please take a look at the failing tests

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -280,21 +284,40 @@ def instrument_app(
...
+                    http_capture_headers_server_request=http_capture_headers_server_request,
+                    http_capture_headers_server_response=http_capture_headers_server_response,
+                    http_capture_headers_sanitize_fields=http_capture_headers_sanitize_fields,
+                    exclude_spans=exclude_spans,
+                )
+                # Wrap in an outer layer of ServerErrorMiddleware so that any exceptions raised in OpenTelemetryMiddleware
+                # are handled.
+                # This should not happen unless there is a bug in OpenTelemetryMiddleware, but if there is we don't want that
+                # to impact the user's application just because we wrapped the middlewares in this order.
+                stack = ServerErrorMiddleware(stack)
```
lzchen, line 313: Just to clarify, is the purpose of this to handle exceptions generated from the `OpenTelemetryMiddleware` itself or unhandled exceptions thrown from the request?

Kludex:
fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/795

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
adriangb: Yep! If there is a bug in `OpenTelemetryMiddleware` without this double wrapping then the server might not send a 500 to the client and instead abruptly disconnect. In practice I believe every ASGI server I know of does also have it's own handling of unhandled exceptions such that this would not be the case, but I'm not sure we want to rely on that + there would still be a noticeable change (different response content at least, maybe different headers, etc.)

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -170,7 +170,8 @@ def setUp(self):
         self._instrumentor = otel_fastapi.FastAPIInstrumentor()
         self._app = self._create_app()
         self._app.add_middleware(HTTPSRedirectMiddleware)
-        self._client = TestClient(self._app)
+        self._client = TestClient(self._app, base_url="https://testserver:443")
```
adriangb, line 173: This was the root cause of tests failures. It turns out _every request was being made twice because it was getting redirected to https_. Before this PR that wasn't being instrumented correctly, so this was not being caught! I think that's another major bug this PR is fixing.

adriangb:
> @adriangb Please take a look at the failing tests

See https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3012#discussion_r1859485948

adriangb:
@Kludex could you update the Starlette instrumentations after we finish up this PR? I'm assuming they have the same bugs.

Kludex:
The pipeline has been running for some hours... 👀

vanyae-cqc:
Hello! Just wondering if you'd expect this to enable getting the `trace_id` in a FastAPI `exception_handler`? 

This has been discussed in more detail here: https://github.com/open-telemetry/opentelemetry-python/issues/3477, but currently OpenTelemetryMiddleware's context seems to be removed by the time an exception gets to the `ServerErrorMiddleware`. 

Would be great to enable users to report back a `trace_id` when they get an unexpected 500 status response.  Have tried to verify this PR provides this by running it locally but haven't had any luck - could be something wrong in my setup though.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -280,21 +284,40 @@ def instrument_app(
...
+                    exclude_spans=exclude_spans,
+                )
+                # Wrap in an outer layer of ServerErrorMiddleware so that any exceptions raised in OpenTelemetryMiddleware
+                # are handled.
+                # This should not happen unless there is a bug in OpenTelemetryMiddleware, but if there is we don't want that
+                # to impact the user's application just because we wrapped the middlewares in this order.
+                app = ServerErrorMiddleware(app)
+                return app
+
+            app._original_build_middleware_stack = app.build_middleware_stack
```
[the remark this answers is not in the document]
adriangb: I'd rather not. This is simpler and works just fine.

adriangb:
Fixed issue with test shutdown hanging. All checks passing now.

xrmx (commented):
I trust you on the fastapi knowledge but would be nice to add a test to check that the change does what it's supposed to fix

adriangb:
Added back and updated the branch

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -226,9 +230,9 @@ def instrument_app(
         server_request_hook: ServerRequestHook = None,
         client_request_hook: ClientRequestHook = None,
         client_response_hook: ClientResponseHook = None,
-        tracer_provider=None,
-        meter_provider=None,
-        excluded_urls=None,
+        tracer_provider: TracerProvider | None = None,
```
lzchen, line 233: Nit: Could you use `typing.Optional` to be consistent with most other places in the code?

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
lzchen: Since asgi middleware does not have [exception handling](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L731) of requests on it's own, wouldn't legitimate exceptions that are raised from the request itself (that we would want to collect telemetry from) be swallowed up by the new middleware? Or does `ServerErrorMiddleware` ONLY handle `500`s? Apologies, I'm not very familiar with the inner workings of it.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
Kludex: Other places are out dated.

What should happen is to start switching the other places.

adriangb: Hard agreement

xrmx: For context the `from __future__ import annotations` let us use this syntax

xrmx:
ICYMI the added test is red

adriangb:
Thanks. I hadn't understood how the tests are parametrized and that `_create_app` is overridden in a subclass.

adriangb:
@xrmx @lzchen can we get this merged please?

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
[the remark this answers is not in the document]
Kludex: @xrmx genuine question: what would be the benefit of that in this case?

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -170,9 +171,14 @@ def setUp(self):
...
+        # run the lifespan, initialize the middleware stack
+        # this is more in-line with what happens in a real application when the server starts up
+        self._exit_stack = ExitStack()
+        self._exit_stack.enter_context(self._client)
```
Kludex, line 178: The `build_middleware_stack` this PR patches is called on the startup event - which the `TestClient` only runs if called within a context manager, that's why @adriangb is using `ExitStack` here. FYI

Kludex (approved):

lzchen:
@adriangb 

I believe I still have an outstanding [question](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3012#discussion_r1868146252)

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
adriangb: `ServerErrorMiddleware` catches unhandled exceptions and turns them into 500 responses. If any exception has made it up that far it's going to be a 500 response. I'm not sure if that's what your asking. Maybe you could give an example test case to integrate and see how it behaves.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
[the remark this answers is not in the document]
xrmx: > @xrmx genuine question: what would be the benefit of that in this case?

wrapt monkeypatching tend to generally work better than shuffling classes under the hood, had to move to wrapt in httpx instrumentation because otherwise the instrumentation did not patch stuff loaded before the load of the instrumentation.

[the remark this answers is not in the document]
adriangb: Unless there's a specific reason I don't think introducing more complexity and code to be executed is helpful here. There are plenty of tests checking that the monkey patching is working correctly. This is also not a method that users would ever call, in fact as someone who's [responsible for significant changes to this very method](https://github.com/encode/starlette/commit/51c1de1839ee187bb20de5748fca23baa37369c6) I would have made it a private method if it had not already been public for years.

adriangb:
@lzchen do you have any more concerns that block this PR?

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -280,21 +284,40 @@ def instrument_app(
...
+                    http_capture_headers_server_request=http_capture_headers_server_request,
+                    http_capture_headers_server_response=http_capture_headers_server_response,
+                    http_capture_headers_sanitize_fields=http_capture_headers_sanitize_fields,
+                    exclude_spans=exclude_spans,
+                )
+                # Wrap in an outer layer of ServerErrorMiddleware so that any exceptions raised in OpenTelemetryMiddleware
+                # are handled.
+                # This should not happen unless there is a bug in OpenTelemetryMiddleware, but if there is we don't want that
+                # to impact the user's application just because we wrapped the middlewares in this order.
+                app = ServerErrorMiddleware(app)
```
alexmojaki, line 313: If the user defined a custom 500 handler, it won't be used here. The snippets in https://github.com/open-telemetry/opentelemetry-python-contrib/issues/795#issuecomment-2396611991 handle that.

Also it's possible to have an error here from a user's `server_request_hook`, not just OTEL bugs.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -280,21 +284,40 @@ def instrument_app(
...
-                http_capture_headers_server_response=http_capture_headers_server_response,
-                http_capture_headers_sanitize_fields=http_capture_headers_sanitize_fields,
-                exclude_spans=exclude_spans,
+            # Instead of using `app.add_middleware` we monkey patch `build_middleware_stack` to insert our middleware
+            # as the outermost middleware.
+            # Otherwise `OpenTelemetryMiddleware` would have unhandled exceptions tearing through it and would not be able
+            # to faithfully record what is returned to the client since it technically cannot know what `ServerErrorMiddleware` is going to do.
+
+            def build_middleware_stack(self: Starlette) -> ASGIApp:
+                app = type(self).build_middleware_stack(self)
```
alexmojaki, line 293:
```suggestion
                app = app._original_build_middleware_stack()
```

alexmojaki: that feels easier to understand, and is more compatible with someone else doing similar patching

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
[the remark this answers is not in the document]
alexmojaki: > had to move to wrapt in httpx instrumentation because otherwise the instrumentation did not patch stuff loaded before the load of the instrumentation.

I'm very curious how wrapt could have made a difference in terms of the order of patching/loading.

But `@functools.wraps(app.build_middleware)` could be easily added to do most of what wrapt does.

Kludex:
Before the last SIG meeting people asked what was missing for the release. I pointed out this PR.

My request was ignored. I want to point out that a checklist may be useful if things get ignored on the meeting.

xrmx:
> Before the last SIG meeting people asked what was missing for the release. I pointed out this PR.

Correct, I did ask in that SIG

> My request was ignored. I want to point out that a checklist may be useful if things get ignored on the meeting.

It's not that your request was ignored, it's that people were busy with other things and haven't had time for reviewing this and others PRs.

CHANGELOG.md
```diff
@@ -49,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-instrumentation-requests` always record span status code in duration metric
   ([#3323](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3323))
 
+
```
xrmx, line 54:
```suggestion
```

emdneto (commented):
Running the MRE in the description, I was able to have spans in console when configuring it with:

```
...
app.middleware_stack = app.build_middleware_stack()
...
```
Is that the original intention?

adriangb:
> Running the MRE in the description, I was able to have spans in the console only when configuring it with:
> 
> ```
> ...
> FastAPIInstrumentor.instrument_app(app)
> app.middleware_stack = app.build_middleware_stack()
> ...
> ```
> 
> Is that the original intention?

I don't think you should ever be doing `app.middleware_stack = app.build_middleware_stack()`. Is this on this branch or on `main`?

emdneto:
> That doesn't seem fair to me. It was asked, and the release was made anyway. Why asking if that would be ignored?
> 
> But my point is more about the release process not having a proper checklist.

On this branch. main is working fine. When running the example as-is it didn't work for me

adriangb:
> > That doesn't seem fair to me. It was asked, and the release was made anyway. Why asking if that would be ignored?
> > But my point is more about the release process not having a proper checklist.
> 
> On this branch. main is working fine. When running the example as-is it didn't show any spans on console

I can't reproduce.

Here's a self-contained script that you can run by [installing uv](https://docs.astral.sh/uv/getting-started/installation/), downloading this script as `test.py` and running it with `uv run test.py`


```python
# /// script
# dependencies = [
#   "anyio==4.8.0",
#   "fastapi==0.115.11",
#   "httpx==0.28.1",
#   "uvicorn==0.34.0",
#   "opentelemetry-instrumentation-asgi@git+https://github.com/adriangb/opentelemetry-python-contrib.git@fix-asgi-middleware#subdirectory=instrumentation/opentelemetry-instrumentation-asgi",
#   "opentelemetry-instrumentation-fastapi@git+https://github.com/adriangb/opentelemetry-python-contrib.git@fix-asgi-middleware#subdirectory=instrumentation/opentelemetry-instrumentation-fastapi",
# ]
# requires-python = ">=3.12"
# ///

import anyio
import fastapi
import httpx
import uvicorn

from opentelemetry.sdk.resources import SERVICE_NAME, Resource
from opentelemetry.sdk.trace.export import (
    BatchSpanProcessor,
    ConsoleSpanExporter,
)

resource = Resource(attributes={SERVICE_NAME: "🐛"})

traceProvider = TracerProvider(resource=resource)
traceProvider.add_span_processor(processor)
trace.set_tracer_provider(traceProvider)

app = fastapi.FastAPI()

@app.get("/ok")
async def ok() -> int:
    return 123

@app.get("/error")
async def error() -> int:
    raise Exception("An error occurred!")

async def main() -> None:
    config = uvicorn.Config(app=app, port=9999, log_config=None)
    server = uvicorn.Server(config)

    async with anyio.create_task_group() as tg:
        tg.start_soon(server.serve)
        await anyio.sleep(1)  # Wait for the server to start
        async with httpx.AsyncClient(base_url="http://localhost:9999") as client:
            response = await client.get("/ok")
            assert response.status_code == 200, f"{response.status_code}: {response.text}"
            response = await client.get("/error")
            assert response.status_code == 500,  f"{response.status_code}: {response.text}"
            tg.cancel_scope.cancel()


if __name__ == "__main__":
    anyio.run(main)
```

I get the following output:

```
{
    "name": "GET /ok http send",
    "context": {
        "trace_id": "0x148c6b2049ca2a044737aed0f4e8f876",
        "span_id": "0xf40c7fda836726f5",
        "trace_state": "[]"
    },
    "kind": "SpanKind.INTERNAL",
    "parent_id": "0x01f2a256269b08a1",
    "start_time": "2025-03-12T14:54:50.389038Z",
    "end_time": "2025-03-12T14:54:50.389058Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "asgi.event.type": "http.response.start",
        "http.status_code": 200
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "service.name": "\ud83d\udc1b"
        },
        "schema_url": ""
    }
}
{
    "name": "GET /ok http send",
    "context": {
        "trace_id": "0x148c6b2049ca2a044737aed0f4e8f876",
        "span_id": "0x39dc19df63ef97f2",
        "trace_state": "[]"
    },
    "kind": "SpanKind.INTERNAL",
    "parent_id": "0x01f2a256269b08a1",
    "start_time": "2025-03-12T14:54:50.389153Z",
    "end_time": "2025-03-12T14:54:50.389161Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "asgi.event.type": "http.response.body"
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "service.name": "\ud83d\udc1b"
        },
        "schema_url": ""
    }
}
{
    "name": "GET /ok",
    "context": {
        "trace_id": "0x148c6b2049ca2a044737aed0f4e8f876",
        "span_id": "0x01f2a256269b08a1",
        "trace_state": "[]"
    },
    "kind": "SpanKind.SERVER",
    "parent_id": null,
    "start_time": "2025-03-12T14:54:50.388863Z",
    "end_time": "2025-03-12T14:54:50.389209Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "http.scheme": "http",
        "http.host": "127.0.0.1:9999",
        "net.host.port": 9999,
        "http.flavor": "1.1",
        "http.target": "/ok",
        "http.url": "http://127.0.0.1:9999/ok",
        "http.method": "GET",
        "http.server_name": "localhost:9999",
        "http.user_agent": "python-httpx/0.28.1",
        "net.peer.ip": "127.0.0.1",
        "net.peer.port": 64160,
        "http.route": "/ok",
        "http.status_code": 200
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "service.name": "\ud83d\udc1b"
        },
        "schema_url": ""
    }
}
{
    "name": "GET /error http send",
    "context": {
        "trace_id": "0x1658c9e569deca83ee28cb9ab16b5b60",
        "span_id": "0x09ec6892057ae7bc",
        "trace_state": "[]"
    },
    "kind": "SpanKind.INTERNAL",
    "parent_id": "0x321c4c0918154a05",
    "start_time": "2025-03-12T14:54:50.390320Z",
    "end_time": "2025-03-12T14:54:50.390332Z",
    "status": {
        "status_code": "ERROR"
    },
    "attributes": {
        "asgi.event.type": "http.response.start",
        "http.status_code": 500
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "service.name": "\ud83d\udc1b"
        },
        "schema_url": ""
    }
}
{
    "name": "GET /error http send",
    "context": {
        "trace_id": "0x1658c9e569deca83ee28cb9ab16b5b60",
        "span_id": "0x36820fbcbf7b747a",
        "trace_state": "[]"
    },
    "kind": "SpanKind.INTERNAL",
    "parent_id": "0x321c4c0918154a05",
    "start_time": "2025-03-12T14:54:50.390403Z",
    "end_time": "2025-03-12T14:54:50.390408Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "asgi.event.type": "http.response.body"
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "service.name": "\ud83d\udc1b"
        },
        "schema_url": ""
    }
}
{
    "name": "GET /error",
    "context": {
        "trace_id": "0x1658c9e569deca83ee28cb9ab16b5b60",
        "span_id": "0x321c4c0918154a05",
        "trace_state": "[]"
    },
    "kind": "SpanKind.SERVER",
    "parent_id": null,
    "start_time": "2025-03-12T14:54:50.390225Z",
    "end_time": "2025-03-12T14:54:50.390444Z",
    "status": {
        "status_code": "ERROR"
    },
    "attributes": {
        "http.scheme": "http",
        "http.host": "127.0.0.1:9999",
        "net.host.port": 9999,
        "http.flavor": "1.1",
        "http.target": "/error",
        "http.url": "http://127.0.0.1:9999/error",
        "http.method": "GET",
        "http.server_name": "localhost:9999",
        "http.user_agent": "python-httpx/0.28.1",
        "net.peer.ip": "127.0.0.1",
        "net.peer.port": 64160,
        "http.route": "/error",
        "http.status_code": 500
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "service.name": "\ud83d\udc1b"
        },
        "schema_url": ""
    }
}
```

adriangb:
Looks like that script doesn't work because of a bug in uv: https://github.com/astral-sh/uv/issues/12141.

Anyway @emdneto this PR doesn't fundamentally change anything that would cause _no_ telemetry to be emitted, and as test prove telemetry is emitted.

@xrmx @lzchen looks like another release was cut without including this. What do I need to do to get this merged? It's been months.

emdneto:
> > > That doesn't seem fair to me. It was asked, and the release was made anyway. Why asking if that would be ignored?
> > > But my point is more about the release process not having a proper checklist.
> > 
> > 
> > On this branch. main is working fine. When running the example as-is it didn't show any spans on console
> 
> I can't reproduce.
> 
> Here's a self-contained script that you can run by [installing uv](https://docs.astral.sh/uv/getting-started/installation/), downloading this script as `test.py` and running it with `uv run test.py`
> 

Sure. I was running the MRE on the root of `opentelemetry-python-contrib` with `uv run mre.py`, and uv got the setup from the repo, installing fastapi==0.64.0. I noticed for the fastapi version <= 0.90.0, the span is missing in the console, but for >= 0.91.0, the span appears.

xrmx:
> @xrmx @lzchen looks like another release was cut without including this. What do I need to do to get this merged? It's been months.

@adriangb In case you missed it you have a comment from Emidio saying that this breaks instrumentation for old fastapi versions we declare to support but no update from your side.

alexmojaki:
Does CI not test fastapi?

adriangb:
@alexmojaki I'd guess not based on https://github.com/open-telemetry/opentelemetry-python-contrib/blob/139d7871684810f3543e4df9376373c463362e65/instrumentation/opentelemetry-instrumentation-fastapi/test-requirements.txt#L8

emdneto (commented):
I think now it's just matter of run `tox -e generate` and commit the changes and tests will pass.

ArvidMartensRenson:
It seems that if your error handler is creating a log and exception originates from within you app, it will be exported twice. Once with the correct trace_id and span_id and once with no trace_id and span_id. I would expect that if an exception is handled by the inner ServerErrorMiddleware it should not be handled by the outer one again.

Only solution I can think of is tracking the traceback but that seems not future proof

adriangb:
> It seems that if your error handler is creating a log and exception originates from within you app, it will be exported twice. Once with the correct trace_id and span_id and once with no trace_id and span_id. I would expect that if an exception is handled by the inner ServerErrorMiddleware it should not be handled by the outer one again.
> 
> Only solution I can think of is tracking the traceback but that seems not future proof

Can you provide an MRE / test case and we can address from there?

xrmx:
@adriangb Please run `tox -e generate` and commit the updated README

adriangb:
Looks like we have an issue with version compatibility with Starlette now: https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/14343599524/job/40208648035?pr=3012

I can do one of:
1. Bump the Starlette dep with no other code changes.
2. Bump the Starlette dep and apply the same fix there (it needs it).

emdneto:
> Looks like we have an issue with version compatibility with Starlette now: https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/14343599524/job/40208648035?pr=3012
> 
> I can do one of:
> 
> 1. Bump the Starlette dep with no other code changes.
> 2. Bump the Starlette dep and apply the same fix there (it needs it).

Which version do you have in mind to bump? Did you see https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3317 ?

lzchen (approved):
@adriangb 

LGTM. Sorry for the late review and thanks for hanging in there with the contribution.

As for fixing the Starlette versioning conflict, we will try to get this PR in asap so you don't have to play catch up with Starlette/FastAPI releases.

adriangb:
> As for fixing the Starlette versioning conflict, we will try to get this PR in asap so you don't have to play catch up with Starlette/FastAPI releases.

Just to be clear, does this mean I need to do anything or are you saying you'll handle it / will merge this PR as is and deal with Starlette as a followup before next release?

ArvidMartensRenson:
> > It seems that if your error handler is creating a log and exception originates from within you app, it will be exported twice. Once with the correct trace_id and span_id and once with no trace_id and span_id. I would expect that if an exception is handled by the inner ServerErrorMiddleware it should not be handled by the outer one again.
> > Only solution I can think of is tracking the traceback but that seems not future proof
> 
> Can you provide an MRE / test case and we can address from there?

If you use this example with this middleware wrapper, you'll notice that the logger in the exception handler is executed twice. Once when the span is active and once when it was already ended. 

`import logging
from fastapi import FastAPI, Response
import uvicorn

from opentelemetry._logs import set_logger_provider
from opentelemetry.sdk._logs import LoggerProvider, LoggingHandler
from opentelemetry.sdk._logs.export import BatchLogRecordProcessor, ConsoleLogExporter

provider = LoggerProvider()
set_logger_provider(provider)
exporter = ConsoleLogExporter()
provider.add_log_record_processor(BatchLogRecordProcessor(exporter))

app = FastAPI()

LOGGER = logging.getLogger(__name__)
LOGGER.setLevel(logging.DEBUG)
LOGGER.addHandler(LoggingHandler())

@app.get("/")
def get_server_exception():
    raise Exception("Server Exception")

@app.exception_handler(Exception)
def exception_handler(request, exc):
    LOGGER.error("Exception occurred", exc_info=False)
    return Response(content="An error occurred", status_code=500)

if __name__ == "__main__":
    uvicorn.run(app)`

adriangb:
I haven't had a chance to look again. Maybe one of the maintainers here can look? I agree this is a critical issue.

outergod:
I did some debugging into what's going on, and this is my findings. Since you are the maintainer of Starlette @adriangb, maybe you can help me understand the semantics?

In `starlette.applications:build_middleware_stack(self)`, if a previously registered exception handler is found in `self.exception_handlers` that was registered with a key of `500` or `Exception`, it gets registered as `error_handler` for `ServerErrorMiddleware` (instead of being added as one of many `exception_handlers` for `ExceptionMiddleware`. 

When the patched FastAPI instrumentation creates the `outer_server_error_middleware` to deal with exceptions from the instrumentation itself, it appears to re-use that potentially previously registered, user-provided error handler through `handler=inner_server_error_middleware.handler`, if I understand correctly. That means, the user-provide handler is evoked a second time, also printing log statements again, this time without an active OTEL context.

My intuitive solution would be to *not* pass in the original `handler` to `ServerErrorMiddleware` and let it use the default one, or provide a FastAPI instrumentation specific one.

WDYT?

adriangb:
That sounds right to me! I think it's okay to not pass any handler. Really the second layer is there just as a failsafe, we could even remove it given that every ASGI server I know also handles uncaught exceptions.

outergod:
Would you like me to rebase and implement your suggested change (no outer exception handler) and provide a PR to your branch? Or do you think you can find the time to do it yourself?

adriangb:
You did the investigation you deserve the commit if you want to make a PR! I'd say just make the PR to this branch and we can rebase the whole thing, it'll be easier

outergod:
https://github.com/adriangb/opentelemetry-python-contrib/pull/2

outergod:
I've just signed the CLA, any way to re-trigger the check?

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -1032,6 +1072,26 @@ def client_response_hook(send_span, scope, message):
...
+    if req_name == "fastapi":
+        # TODO: Value now matters
+        return "0.92"
+    raise PackageNotFoundError()
+
+
+def mock_version_without_fastapi(*args, **kwargs):
+    raise PackageNotFoundError()
+
+
```
emdneto, line 1094: I think this isn't used anymore
```suggestion

```

emdneto:
@adriangb CI should be OK now

outergod:
Anything else in the way of merging now?

emdneto (approved):
@adriangb I fixed the changelog conflict and added this note:
`opentelemetry-instrumentation-fastapi`: Drop support for FastAPI versions earlier than `0.92`

adriangb:
I gave it a shot, lets see what CI says

xrmx:
@adriangb we merged a PR stopping using SpanAttributes in favor of semconv._incubating.attributes.*

alexmojaki (approved):

alexmojaki:
Does @codefromthecrypt need to approve?

xrmx:
@alexmojaki github was saying so but your approval was good enough :sweat_smile:

alexmojaki:
Thanks everyone for getting this across!

adriangb:
Amazing folks thanks so much!!

outergod:
Finally! Thanks to everyone involved, especially @adriangb!

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -284,21 +289,56 @@ def instrument_app(
...
+                ):  # usually true
+                    outer_server_error_middleware = ServerErrorMiddleware(
+                        app=otel_middleware,
+                    )
+                else:
+                    # Something else seems to have patched things, or maybe Starlette changed.
+                    # Just create a default ServerErrorMiddleware.
+                    outer_server_error_middleware = ServerErrorMiddleware(
+                        app=otel_middleware
+                    )
```
alexmojaki, line 331: these branches are the same now

outergod: I feel stupid for not realizing this when I changed the code

PR #3022, opened 2024-11-18
fix(httpx): check if mount transport is None before wrap

### LINKED ISSUE #3020: HTTPXClientInstrumentor.instrument_client fails when NO_PROXY is set
### Describe your environment

OS: Ubuntu in docker
Python version: 3.11
Package version: 0.49b2

### What happened?

If the environment variable `NO_PROXY` is set `HTTPXClientInstrumentor.instrument_client(client)` will fail when wrapping the mounted transport with `wrap_function_wrapper` because the transport is `None` for that mount.

### Steps to Reproduce

```shell
export NO_PROXY=http://example.com
``` 

```python
import httpx
    HTTPXClientInstrumentor.instrument_client(client)
```

### Expected Result

No exception, client is instrumented.

### Actual Result

Error in `wrapt/patches.py", line 46, in lookup_attribute`

```
AttributeError: 'NoneType' object has no attribute 'handle_request'
```

### Additional context

I had tests that failed only in CircleCI and I traced it to the presence of the environment variable `NO_PROXY`. HTTPX makes a mount for each entry in `NO_PROXY`, but with transport set to `None`. 
https://www.python-httpx.org/environment_variables/#no_proxy

CircleCI is where I found this, because CircleCI sets this env var in its base `cimg` images, probably to ensure curl works. They set it to `127.0.0.1,localhost,circleci-internal-outer-build-agent`.

The affected code is
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/54e684aa6985536c8807567c592ea60751d62162/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py#L1008 and
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/54e684aa6985536c8807567c592ea60751d62162/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py#L1033

The easiest fix appears to be to skip a mount if `transport` is `None`. 

There are two workarounds:
1. Use `HTTPXClientInstrumentor().instrument()` instead of `instrument_client`.
2. Set `trust_env=False` when creating the `httpx.Client` or `httpx.AsyncClient`. `httpx` won't use the env variables at all, which affects proxies and ssl key definitions.

None

emdneto:
# Description

Fixes #3020

## Type of change

# How Has This Been Tested?

- [X] tox

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-httpx/tests/test_httpx_integration.py
+++ b/instrumentation/opentelemetry-instrumentation-httpx/tests/test_httpx_integration.py
@@ -766,14 +766,14 @@ class BaseTestCases:
             self.assertEqual(len(mounts), num_mounts)
             for transport in mounts:
                 with self.subTest(transport):
+                    if transport is None:
+                        continue
                     if transport_type:
                         self.assertIsInstance(
                             transport,
                             transport_type,
                         )
                     else:
-                        if transport is None:
-                            continue
                         handler = self.get_transport_handler(transport)
                         self.assertTrue(
                             isinstance(handler, ObjectProxy)
@@ -1017,7 +1017,7 @@ class BaseTestCases:
             self.assert_proxy_mounts(
                 client._mounts.values(),
                 3,
-                (httpx.HTTPTransport, httpx.AsyncHTTPTransport, type(None)),
+                (httpx.HTTPTransport, httpx.AsyncHTTPTransport),
             )
             HTTPXClientInstrumentor.instrument_client(client)
             result = self.perform_request(self.URL, client=client)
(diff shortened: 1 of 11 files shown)


codefromthecrypt (approved):

instrumentation/opentelemetry-instrumentation-httpx/tests/test_httpx_integration.py
```diff
@@ -725,7 +725,6 @@ def test_client_mounts_with_instrumented_transport(self):
                 spans[1].attributes[SpanAttributes.HTTP_URL], https_url
             )
 
-    @mock.patch.dict("os.environ", {"NO_PROXY": ""}, clear=True)
```
xrmx, line 728: This is useful for having tests separated from the environment in which they are run

emdneto: Right. I missed that we can have conflicts with the local environment variables from the host

xrmx (approved):
Thanks!

tammy-baylis-swi (approved):

PR #3035, opened 2024-11-21
infra: configure generate-workflows to generate core-contrib-tests

emdneto:
Issue summary in https://github.com/open-telemetry/opentelemetry-python/pull/4302


.github/workflows/generate_workflows_lib/src/generate_workflows_lib/core_contrib_test.yml.j2
```diff
@@ -27,12 +28,6 @@ jobs:
           repository: open-telemetry/opentelemetry-python-contrib
           ref: ${% raw %}{{ env.CONTRIB_REPO_SHA }}{% endraw %}
 
-      - name: Checkout core repo @ SHA - ${% raw %}{{ github.sha }}{% endraw %}
```
emdneto, line 30: I'm not 100% sure but I think this has no use since we install packages in tox pointing to git+url@CORE_REPO_SHA

xrmx: It should be fine, grepping the checkout dir `opentelemetry-python-core` returns nothing other than this

xrmx (approved):

tammy-baylis-swi (approved):

lzchen (approved):

Issue #3038, opened 2024-11-22
Tox docs build fails in release branches


aabmass:
Because we do not have a way to override the specific branch in docs-requirements: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/5c5fc732fb0792f3aad3582cabdf2f290c95c541/docs-requirements.txt#L7-L10

This causes dependency conflict in semantic-conventions package, for example [this workflow run](https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/11863258829/job/33064347167) from https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3016:

```
The conflict is caused by:
    The user requested opentelemetry-semantic-conventions 0.50b0.dev0 (from git+https://github.com/open-telemetry/opentelemetry-python.git#egg=opentelemetry-semantic-conventions&subdirectory=opentelemetry-semantic-conventions)
    opentelemetry-sdk 1.29.0.dev0 depends on opentelemetry-semantic-conventions==0.50b0.dev
    opentelemetry-instrumentation 0.49b1 depends on opentelemetry-semantic-conventions==0.49b1

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip to attempt to solve the dependency conflict

docs: exit 1 (20.20 seconds) /home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib> python -I -m pip install pytest -r /home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/docs-requirements.txt -c /home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/dev-requirements.txt pid=1748
  docs: FAIL code 1 (20.64 seconds)
  evaluation failed :( (21.99 seconds)
```

emdneto:
@aabmass In 1.29.0 release we didn't have any issues with docs CI because we are skipping it for opentelemetrybot 😅 but I opened a [test PR]( https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/12307144773/job/34350245605?pr=3101 ) to validate everything is working as expected. I think we won't have problems anymore after #3039, so if you're good with the results, we can close the issue ✌🏻

aabmass:
Awesome thank you!

PR #3039, opened 2024-11-22
docs: separate readthedocs requirements in rtd-requirements.txt to avoid conflicts during CI

emdneto:
# Description

Trial to address #3038 
For RTD, it'll use a separate requirements.txt .With that, it's expected that during releases CI, it proceeds with the correct pip install of core packages in tox.


rtd-requirements.txt
```diff
@@ -0,0 +1,7 @@
+# Need to install the api/sdk in the venv for autodoc. Used by .readthedocs.yml
+opentelemetry-api @ git+https://github.com/open-telemetry/opentelemetry-python.git@main#egg=opentelemetry-api&subdirectory=opentelemetry-api
```
xrmx, line 2: I'm confused how this is supposed to work on release branches?

emdneto: Are we publishing docs from the releases branch?  

The idea here is to separate RTD requirements from tox Docs CI requirements. In CI, we run `tox—e docs`, which will install core packages using the `CORE_REPO_SHA` environment variable. We usually update the SHA during releases to point to the respective release branch in the core, right? So, this will make Docs CI pass in contrib.

xrmx: uhm maybe this is a thing only in core, sorry for the noise

xrmx (approved):

emdneto (commented):
Just in case I ran against release branch and it's passing: https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/12039647691/job/33567897205

Without this change: https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/12039729872/job/33568215527?pr=3051

lzchen (approved):

PR #3060, opened 2024-12-02
docs: add missing autodoc for click and aiohttp-server

emdneto:
# Description
Add missing docs for opentelemetry-instrumentation-click and opentelemetry-instrumentation-aiohttp-server


emdneto (commented):
Probably would be nice to have a CI job to check if we have docs for all available instrumentation

emdneto (commented):
Some documentations are broken with wrong rst blocks, specially related to kafka

tammy-baylis-swi (approved):

xrmx (approved):
Thanks, will need to update CONTRIBUTING.md with this

PR #3068, opened 2024-12-04
Add type hints to dbapi

Kludex:
There are runtime changes i.e. I took out the proxy classes from the inside of the function. This is necessary to have correct type hints.


Kludex:
Found the issue: https://wrapt.readthedocs.io/en/latest/wrappers.html#proxy-object-attributes

Kludex:
What's this docs thing?

emdneto (commented):
I think to fix docs CI you need to append the class  `opentelemetry.instrumentation.dbapi.TracedConnectionProxy` in [nitpick-exceptions.ini](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/docs/nitpick-exceptions.ini) file.

Kludex:
> I think to fix docs CI you need to append the class `opentelemetry.instrumentation.dbapi.TracedConnectionProxy` in [nitpick-exceptions.ini](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/docs/nitpick-exceptions.ini?rgh-link-date=2024-12-10T23%3A38%3A23Z) file.

What's that `nitpick...` for? I don't want this class to be documented. 🤔

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -55,22 +57,34 @@
...
 from opentelemetry.util._importlib_metadata import version as util_version
 
+if TYPE_CHECKING:
+
+    class wrapt:  # pylint: disable=invalid-name
+        class ObjectProxy:
+            def __init__(self, *args: Any, **kwargs: Any) -> None:
+                pass
+else:
+    import wrapt
```
Kludex, line 67: This messed a bit with sphinx... 😓

Kludex: I'm on the process of adding type hints to wrapt, so I've removed this logic here...

ref.: https://github.com/GrahamDumpleton/wrapt/pull/279

Kludex: Actually.. It was not this.. But I've reverted anyway, better to wait for wrapt..

emdneto (approved):
Thanks! I think would be nice a changelog in that case although there are no overall behavior break. QQ: Do we need `py.typed` here as well?

Kludex:
> Thanks! I think would be nice a changelog in that case although there are no overall behavior break. QQ: Do we need `py.typed` here as well?

Yeah, makes sense. I've added the changelog entry more aligned with the fact that I moved the `Traced` classes to the module level.

emdneto (approved):

lzchen (approved):

PR #3084, opened 2024-12-10
Add type hints to BaseInstrumentor

Kludex:


aabmass (approved):

codefromthecrypt (approved):
curious why only to this type? Is it mainly to lower risk?

I'm new to python and thought this was a python 3.10 change but as tests pass must be all good. Maybe someone can teach me why `Optional[str]` -> `str | None` passes on all versions we test!

Kludex:
This project is not very familiar with modern typing. I'm trying to be mindful and create minimal PRs to spread the knowledge, and move faster.

When you use `from __future__ import annotations`, at runtime Python will not look at the type hints, which allow us to use notation from future Python versions. Note that even if the notation is only introduced on newer versions, the Python parser can still parse the code, so we don't have any issue.

codefromthecrypt:
@Kludex hero level explanation. thanks for teaching me!

xrmx (approved):

Issue #2056, opened 2023-11-12
Circular Import Issue Caused by AwsLambdaInstrumentor Manual Instrumentation


sam-mosleh:
**Environment:**
- AWS Lambda runtime: Python 3.11
- Library version: opentelemetry-instrumentation-aws-lambda==0.42b0

**Steps to reproduce**
```python
from opentelemetry.instrumentation.aws_lambda import AwsLambdaInstrumentor

AwsLambdaInstrumentor().instrument()

def lambda_handler(event, context):
    return "OK"
```

**What is the expected behavior?**
The `AwsLambdaInstrumentor` should allow manual instrumentation without causing circular imports, ensuring the proper functioning of the project.

**What is the actual behavior?**
Circular import is triggered

**Additional context**
```json
{
  "errorMessage": "partially initialized module 'test' has no attribute 'lambda_handler' (most likely due to a circular import)",
  "errorType": "AttributeError",
  "requestId": "",
  "stackTrace": [
    "  File \"/var/lang/lib/python3.11/importlib/init.py\", line 126, in import_module\n    return _bootstrap._gcd_import(name[level:], package, level)\n",
    "  File \"<frozen importlib._bootstrap>\", line 1204, in _gcd_import\n",
    "  File \"<frozen importlib._bootstrap>\", line 1176, in _find_and_load\n",
    "  File \"<frozen importlib._bootstrap>\", line 1147, in _find_and_load_unlocked\n",
    "  File \"<frozen importlib._bootstrap>\", line 690, in _load_unlocked\n",
    "  File \"<frozen importlib._bootstrap_external>\", line 940, in exec_module\n",
    "  File \"<frozen importlib._bootstrap>\", line 241, in _call_with_frames_removed\n",
    "  File \"/var/task/test.py\", line 3, in <module>\n    AwsLambdaInstrumentor().instrument()\n",
    "  File \"/opt/python/opentelemetry/instrumentation/instrumentor.py\", line 108, in instrument\n    result = self._instrument(  # pylint: disable=assignment-from-no-return\n",
    "  File \"/opt/python/opentelemetry/instrumentation/aws_lambda/init.py\", line 462, in _instrument\n    _instrument(\n",
    "  File \"/opt/python/opentelemetry/instrumentation/aws_lambda/init.py\", line 403, in _instrument\n    wrap_function_wrapper(\n",
[...]

PR #3085, opened 2024-12-10
docs: move `AwsLambdaInstrumentor().instrument()` to after the lambda function

Kludex:
https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2056#issuecomment-2531619571

`tox -e docs` doesn't work...

```bash
        File "/private/var/folders/q1/wn12m6td0tx0dy1s84xzrxbm0000gn/T/pip-install-nycnab7m/mysqlclient_22dcb881261f49e7b709df7ab13fc6ed/setup_posix.py", line 31, in mysql_config
          raise OSError("{} not found".format(_mysql_config_path))
      OSError: mysql_config not found
      [end of output]
```


xrmx (approved):

emdneto:
@Kludex I think `tox—e docs` needs some packages installed at the host level, like mysql_config. When running at CI, it works without any issue because it already has the package, so it's not a problem with the command itself.

Kludex:
> mysql_config

I got that, but where's the instruction?

xrmx:
> > mysql_config
> 
> I got that, but where's the instruction?

It's one PR away!

PR #3099, opened 2024-12-12
Prevent RuntimeError for Flask on context.detach with the same token

abo-abo:
# Description

This happens for me when using werkzeug's debug=True. In this case, the token was already detached. But when I enter an expression into the debug console, we attempt to detach the token once more.

This causes a RuntimeError, as expected: https://peps.python.org/pep-0567/#contextvars-contextvar.

But if we set _ENVIRON_TOKEN to None after we detach, this won't happen.

## Type of change

# How Has This Been Tested?

- Create a simple Flask app with instrumented spans
- Raise an error in one of the routes and open the debugger web page
- Click the shell icon and enter the debugger PIN
- Try to examine any variables in the context
    - doing so will produce a RuntimeError in the logs, since the token was already used to detach the context on the original request

- [ ] Yes. - Link to PR:
- [X] No.


xrmx:
We need the CLA signed in order to get things accepted.

xrmx:
@abo-abo would it be possible to write a test for asserting that the fix work?

PR #2648, opened 2024-06-30
HTTP semantic convention stability migration for httplib

zhihali:
# Description

Part of this https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2453

Fixes # (issue)

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

`tox -e py312-test-util-http `

# Does This PR Require a Core Repo Change?

- [x] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [x] Followed the style guidelines of this project
- [x] Changelogs have been updated
- [x] Unit tests have been added
- [ ] Documentation has been updated

PR #3104, opened 2024-12-13
Don't normalize headers with "hyphen -> underscore"

Kludex:


util/opentelemetry-util-http/src/opentelemetry/util/http/__init__.py
```diff
@@ -184,12 +184,12 @@ def remove_url_credentials(url: str) -> str:
...
 def normalise_request_header_name(header: str) -> str:
-    key = header.lower().replace("-", "_")
+    key = header.lower()
     return f"http.request.header.{key}"
 
 
 def normalise_response_header_name(header: str) -> str:
-    key = header.lower().replace("-", "_")
+    key = header.lower()
     return f"http.response.header.{key}"
```
Kludex, line 193: Those are the relevant changes. The rest is just to comply with this.

Kludex:
I don't know what's wrong with the docs... 😢

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
```
emdneto, line 14: I think move to Breaking change

Kludex: Done. 🙏

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
```diff
@@ -139,11 +139,11 @@ def client_response_hook(span, future):
 It is recommended that you should give the correct names of the headers to be captured in the environment variable.
 Response header names captured in tornado are case insensitive. So, giving header name as ``CUStomHeader`` in environment variable will be able capture header with name ``customheader``.
 
-The name of the added span attribute will follow the format ``http.response.header.<header_name>`` where ``<header_name>`` being the normalized HTTP header name (lowercase, with - characters replaced by _ ).
+The name of the added span attribute will follow the format ``http.response.header.<header_name>`` where ``<header_name>``being the normalized HTTP header name (lowercase).
```
emdneto, line 142:
```suggestion
The name of the added span attribute will follow the format ``http.response.header.<header_name>`` where ``<header_name>`` being the normalized HTTP header name (lowercase).
```

Kludex:
Thanks @emdneto

lzchen:
Relevant pr in semconv: https://github.com/open-telemetry/semantic-conventions/pull/369

Would this breaking change fall under semconv migration? @emdneto @aabmass @xrmx  wdyt?

emdneto:
> Relevant pr in semconv: [open-telemetry/semantic-conventions#369](https://github.com/open-telemetry/semantic-conventions/pull/369)
> 
> Would this breaking change fall under semconv migration? @emdneto @aabmass @xrmx wdyt?

@lzchen Taking a second look: yes. We do still have an open PR to migrate util-http: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2648 . In that case, I think the problem is we are breaking already migrated instrumentation, like django, flask.

emdneto:
@Kludex Since we are breaking users of migrated instrumentation to support opt-in for stable semantic conventions, it would also be necessary to include this as part of it. So the change must take in consideration the `_sem_conv_opt_in_mode` specified by the user. If the user opt-in for the stable semantic convention, it should return the header with `-`, otherwise by default, keep the underscore `_`. [See](https://opentelemetry.io/docs/specs/semconv/non-normative/http-migration/)

PR #3106, opened 2024-12-13
httpx: instrument transport instead of client

Kludex:


instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
```diff
@@ -890,7 +890,9 @@ def instrument_client(
                 that is called right before the span ends
         """
 
-        if getattr(client, "_is_instrumented_by_opentelemetry", False):
+        if getattr(
+            client._transport, "_is_instrumented_by_opentelemetry", False
+        ):
             _logger.warning(
```
alexmojaki, line 896: We shouldn't log a warning if someone instruments a new client with the same transport. So I think we need to keep the attribute and warning on the client, but silently do nothing if the transport is already instrumented.

xrmx:
I think the title does not match the code, you are moving the flag that signals that the transport has already been instrumented to the transport itself.

Kludex:
> I think the title and description do not match the code, you are moving the flag that signals that the transport has already been instrumented to the transport itself.

Right. The PR started as what the title suggests, and I modified the code given this comment: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3106#discussion_r1884417021

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- `opentelemetry-instrumentation-httpx` Instrument transport instead of HTTPX client
```
Kludex, line 14:
```suggestion
- `opentelemetry-instrumentation-httpx` Add the `is_instrumented` flag to the transport
```

Kludex:
@xrmx Should be better now. 🙏

xrmx (approved):

PR #3111, opened 2024-12-14
prepare semconv utilities to support database stability opt-in

emdneto:
# Description

This PR prepares our _semconv utility suite, which was used to implement semantic convention stability migration for HTTP instrumentation to support database instrumentation.

Related: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2453


opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -135,50 +135,70 @@ class _OpenTelemetrySemanticConventionStability:
...
-                        http_opt_in = _HTTPStabilityMode.HTTP
-                _OpenTelemetrySemanticConventionStability._OTEL_SEMCONV_STABILITY_SIGNAL_MAPPING[
-                    _OpenTelemetryStabilitySignalType.HTTP
-                ] = http_opt_in
-                _OpenTelemetrySemanticConventionStability._initialized = True
+        with cls._lock:
+            if cls._initialized:
+                return
+            # Users can pass in comma delimited string for opt-in options
+            # Only values for http stability are supported for now
```
emdneto, line 142:
```suggestion
            # Only values for http and database stability are supported for now
```

opentelemetry-instrumentation/tests/test_semconv.py
```diff
@@ -0,0 +1,222 @@
...
+            span=span,
+            metrics_attributes=metrics_attributes,
+            status_code=500,
+            status_code_str="500",
+            server_span=True,
+            sem_conv_opt_in_mode=_StabilityMode.HTTP_DUP,
+        )
+        # Verify both old and new conventions are emitted
+        span.set_attribute.assert_any_call("http.status_code", 500)
+        span.set_attribute.assert_any_call("http.response.status_code", 500)
```
tammy-baylis-swi, line 185: These tests are helpful, thank you!

It would be great to see a similar `database/dup` implementation but maybe that's better for a different set of PRs? For example, if `database/dup`then I'd expect `set_attribute` of both [db.statement](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-semantic-conventions/src/opentelemetry/semconv/trace/__init__.py#L423-L426) (old) and db.query.text/db.query.summary (new). But I don't think we've updated [semconv.attributes](https://github.com/open-telemetry/opentelemetry-python/tree/main/opentelemetry-semantic-conventions/src/opentelemetry/semconv/attributes) with this yet

opentelemetry-instrumentation/tests/test_semconv.py
```diff
@@ -0,0 +1,222 @@
...
+    @semconv_stability_mode("database/dup")
+    def test_database_dup_mode(self):
+        self.assertEqual(
+            _OpenTelemetrySemanticConventionStability._get_opentelemetry_stability_opt_in_mode(
+                _OpenTelemetryStabilitySignalType.DATABASE
+            ),
+            _StabilityMode.DATABASE_DUP,
+        )
+
+    @semconv_stability_mode("database,http")
```
tammy-baylis-swi, line 88: Could we also have tests for mixed config, e.g. `database,http/dup` and `database_dup/http`

opentelemetry-instrumentation/tests/test_semconv.py
```diff
@@ -0,0 +1,222 @@
...
+
+class TestOpenTelemetrySemConvStability(TestCase):
+    @semconv_stability_mode("")
+    def test_default_mode(self):
+        self.assertEqual(
+            _OpenTelemetrySemanticConventionStability._get_opentelemetry_stability_opt_in_mode(
+                _OpenTelemetryStabilitySignalType.HTTP
+            ),
+            _StabilityMode.DEFAULT,
+        )
```
tammy-baylis-swi, line 50: It will also be `_StabilityMode.DEFAULT` for `_OpenTelemetryStabilitySignalType.DATABASE`, is that correct?

tammy-baylis-swi:
Overall this looks good to me and lines up with [spec](https://github.com/open-telemetry/semantic-conventions/blob/294d2dbf00dc679fba70ad48c89bb8e3dce72f23/docs/database/README.md?plain=1#L31) for `database/dup` as part of `OTEL_SEMCONV_STABILITY_OPT_IN` comma-separated list. I just had some questions about db attribute switches and test cases.

opentelemetry-instrumentation/tests/test_semconv.py
emdneto: I think we can do that as per as we migrate instrumentation to use the stability opt_in. Wdyt? Right now is hard to write tests because I didn't read all the changes 😅

opentelemetry-instrumentation/tests/test_semconv.py
emdneto: Yes, sure. Will add

opentelemetry-instrumentation/tests/test_semconv.py
emdneto: That's correct. For any kind of signal type, we have only one default, which stands for "" or the absence of the environment variable.

opentelemetry-instrumentation/tests/test_semconv.py
tammy-baylis-swi: That works! This is good for the contrib config part

opentelemetry-instrumentation/tests/test_semconv.py
emdneto: Added. Thanks!

tammy-baylis-swi (approved):
Thanks Emídio!

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -135,50 +135,74 @@ class _OpenTelemetrySemanticConventionStability:
...
-                    # http/dup takes priority over http
-                    if _HTTPStabilityMode.HTTP_DUP.value in opt_in_list:
-                        http_opt_in = _HTTPStabilityMode.HTTP_DUP
-                    elif _HTTPStabilityMode.HTTP.value in opt_in_list:
-                        http_opt_in = _HTTPStabilityMode.HTTP
-                _OpenTelemetrySemanticConventionStability._OTEL_SEMCONV_STABILITY_SIGNAL_MAPPING[
-                    _OpenTelemetryStabilitySignalType.HTTP
-                ] = http_opt_in
-                _OpenTelemetrySemanticConventionStability._initialized = True
+        with cls._lock:
```
emdneto, line 138: Important changes here

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -367,10 +391,11 @@ def _set_status(
     status_code: int,
     status_code_str: str,
     server_span: bool = True,
-    sem_conv_opt_in_mode: _HTTPStabilityMode = _HTTPStabilityMode.DEFAULT,
+    sem_conv_opt_in_mode: _StabilityMode = _StabilityMode.DEFAULT,
 ):
     if status_code < 0:
-        metrics_attributes[ERROR_TYPE] = status_code_str
+        if _report_new(sem_conv_opt_in_mode):
```
emdneto, line 397: Not related change, but when working on this noticed we miss this if. ERROR_TYPE is required only for stable semconv.

lzchen (approved):

Issue #3107, opened 2024-12-13
Support sqlcomment-in-`db.statement` as an opt-in for db driver/ORM instrumentors


tammy-baylis-swi:
### What problem do you want to solve?

Some platforms that use the Python db client component instrumentors don't need `db.statement` population with sqlcomment -- see this [issue comment](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2938#issuecomment-2541134100). Currently this always happens as long as sqlcommenting is enabled, e.g.:

```
Attributes:
     -> db.system: Str(postgresql)
     -> db.name: Str(world-db)
     -> db.statement: Str(SELECT * FROM city WHERE id = %s /*db_driver='psycopg2%%3A2.9.9',traceparent='00-acbdad825030e4fc58260b2bd9861eb3-451dac2c63520272-01'*/)
     -> db.user: Str(world)
     -> net.peer.name: Str(postgres-world-db)
     -> net.peer.port: Int(5432)
```

### Describe the solution you'd like

Implement support of a config so that Python db client component instrumentors only add sqlcomment to `db.statement` when opted in. Default should not include the comment.

Changes are required in:
- [x] DB-API integration: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3115
- [x] sqlalchemy instrumentor: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3112
- [x] psycopg2 instrumentor: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3121
- [x] psycopg instrumentor: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3121
- [x] mysqlclient instrumentor: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3121
- [x] PyMySQL instrumentor: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3121

### Describe alternatives you've considered

This is a short-term solution. Longer term, we need to start with a general semconv upgrade of the db driver/ORM instrumentors: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2453. Part of this is moving from assignment of `db.statement` to `db.query.text`.

### Additional Context

Will append to changes made for https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2936 and https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2938

### Would you like to implement a fix?

Yes

tammy-baylis-swi:
[...]

Issue #2938, opened 2024-10-29
Include sqlcomment in db.statement span attribute from SQLAlchemy instrumentation 


tammy-baylis-swi:
### What problem do you want to solve?

SQLAlchemy instrumentation's opt-in sqlcommenting feature ([doc](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/sqlalchemy/sqlalchemy.html#sqlcommenter)) works to insert OTel context into MySQL or PostgreSQL general logs, e.g. this when used with mysql-connector driver:

```
2024-10-29T22:09:35.469520Z	  240 Query	SELECT city.id, city.name, city.district, city.population, city.countrycode
FROM city
WHERE city.id = '1818' /*db_driver='mysqlconnector',db_framework='sqlalchemy%%3A2.0.36',traceparent='00-3b654158a48190e458f6c4e15e125814-b62c6e3e794e9f05-01'*/
```

For correlation of db queries to corresponding query spans, it would be convenient to also include the sqlcomment in the db.statement attribute of those spans, e.g.

```
InstrumentationScope opentelemetry.instrumentation.sqlalchemy 0.49b0.dev
Span #0
    Trace ID       : 3b654158a48190e458f6c4e15e125814
    Parent ID      : ed65cfad7167aecd
    ID             : 69d58000ef348179
    Name           : connect
    Kind           : Client
    Start time     : 2024-10-29 22:09:35.42945621 +0000 UTC
    End time       : 2024-10-29 22:09:35.466973419 +0000 UTC
    Status code    : Unset
    Status message :
Attributes:
     -> net.peer.name: Str(mysql-world-db)
     -> net.peer.port: Int(3306)
     -> db.name: Str(world)
     -> db.user: Str(world)
     -> db.system: Str(mysql)
Span #1
    Trace ID       : 3b654158a48190e458f6c4e15e125814
    Parent ID      : ed65cfad7167aecd
    ID             : b62c6e3e794e9f05
    Name           : SELECT world
    Kind           : Client
    Start time     : 2024-10-29 22:09:35.469197169 +0000 UTC
    End time       : 2024-10-29 22:09:35.469996085 +0000 UTC
    Status code    : Unset
    Status message :
Attributes:
     -> db.statement: Str(SELECT city.id, city.name, city.district, city.population, city.countrycode
FROM city
[...]

PR #3112, opened 2024-12-16
SQLAlchemy: db.statement inclusion of sqlcomment as opt-in

tammy-baylis-swi:
# Description

Supports new, optional kwarg `enable_attribute_commenter` at SQLAlchemy instrumentation to opt into `db.statement` span attribute inclusion of sqlcomment.

Partially addresses https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3107

## Type of change

Breaking because `db.statement` span attribute inclusion of sqlcomment currently always happens if sqlcommenter enabled. It'll be opt-in with this update, so might break for those users relying on the feature introduced in 0.50b0.

# How Has This Been Tested?

- [x] `tox -e py312-test-instrumentation-sqlalchemy-0`
- [x] `tox -e py312-test-instrumentation-sqlalchemy-1`
- [x] `tox -e py312-test-instrumentation-sqlalchemy-2`
- [x] Instrumentation of a test app to check span attributes and mysql server logs when `enable_attribute_commenter=True` vs unset

- [X] No.

# Checklist:


tammy-baylis-swi:
cc'ing @alexmojaki , feel free to comment/review!

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -251,13 +263,22 @@ def _before_cur_exec(
...
                 span.set_attribute(SpanAttributes.DB_STATEMENT, statement)
                 span.set_attribute(SpanAttributes.DB_SYSTEM, self.vendor)
                 for key, value in attrs.items():
                     span.set_attribute(key, value)
 
+                if (
+                    self.enable_commenter
+                    and not self.enable_attribute_commenter
+                ):
+                    statement = _add_sql_comment(statement, **commenter_data)
```
shalevr, line 280: Why do we need the _add_sql_comment after the set_attribute statement?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -251,13 +263,22 @@ def _before_cur_exec(
...
+
```
shalevr, line 281: Line 267 and this line look the same both have self.enable_commenter=True, but on line 267, self.enable_attribute_commenter=True, while in this condition, self.enable_attribute_commenter=False. So, what's the difference? Can we call _add_sql_comment once to make it clearer

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
tammy-baylis-swi: @shalevr ! At this condition, user has enabled sqlcommenting but has not opted into "enable attribute commenter". So `db.statement` get written first without sqlcomment, then sqlcomment is added to the query statement before execution.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
tammy-baylis-swi: I'll see how I can rearrange and add comments to make it more clear!

Currently L247 checks `self.enable_commenter=True` and L267 checks `self.enable_attribute_commenter=True`. If so, we `_add_sql_comment` first so that both `db.statement` attribute and executed query statement include the sqlcomment.

This next part is unclear to me too now 😅 If at L247 `self.enable_commenter=False`, then `db.statement` gets set first without sqlcomment. Then at L276 we check again?! Not sure how the tests are passing.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -218,6 +230,32 @@ def _operation_name(self, db_name, statement):
...
+        commenter_data = {
+            k: v
+            for k, v in commenter_data.items()
+            if self.commenter_options.get(k, True)
+        }
+        return commenter_data
+
+    def _set_db_client_span_attributes(self, span, statement, attrs) -> None:
+        """Uses statement and attrs to set attributes of provided Otel span"""
+        span.set_attribute(SpanAttributes.DB_STATEMENT, statement)
```
emdneto, line 254: Prob isn't related, but shouldn't these attributes be provided at span creation?  For sampling decisions.

emdneto: If so, maybe we can pass them at span creation and convert this method to a staticmethod that just update the statement with the statement with commenter_data.

emdneto: If so, maybe we can pass them at span creation and convert this method to a staticmethod that just update the statement with the statement with commenter_data.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
```diff
@@ -65,6 +65,28 @@
 ::
 Enabling this flag will add traceparent values /*traceparent='00-03afa25236b8cd948fa853d67038ac79-405ff022e8247c46-01'*/
 
+SQLComment in span attribute
+****************************
+If sqlcommenter is enabled, you can optionally configure SQLAlchemy instrumentation to append sqlcomment to query span attribute for convenience of your platform.
```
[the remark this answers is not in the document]
tammy-baylis-swi: Yes that's a fine idea. Added in [3b97e64](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3112/commits/3b97e641cee718ede49052a10e86096b9c2500c2)

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
tammy-baylis-swi: The calls to `set_attribute` in this instrumentor generally have been set to only happen if `span.is_recording()`, which for this (new) helper is at [L272](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3112/files#diff-4bf18e7d840238995e3eb1152d6bf0a3f56ce6ab6f2bcf1726ecbac7a79afedbR272). This condition was introduced in https://github.com/open-telemetry/opentelemetry-python/issues/1057 to improve performance, so I think that part should be kept in. Then regarding staticmethod, I'm not sure if it'll be possible but maybe I miss something. Thoughts? 🙂

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
```diff
@@ -65,6 +65,30 @@
...
+        enable_commenter=True,
+        commenter_options={},
+        enable_attribute_commenter=True,
+    )
+
+
+For example,
+::
+
+    Invoking engine.execute("select * from auth_users") will lead to sql query "select * from auth_users" but when SQLCommenter and attribute_commenter is enabled
```
lzchen, line 86:
```suggestion
    Invoking `engine.execute("select * from auth_users")` will lead to sql query "select * from auth_users" but when SQLCommenter and `attribute_commenter` is enabled
```

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
```diff
@@ -65,6 +65,30 @@
...
+If sqlcommenter is enabled, you can optionally configure SQLAlchemy instrumentation to append sqlcomment to query span attribute for convenience of your platform.
+
+.. code:: python
+
+    from opentelemetry.instrumentation.sqlalchemy import SQLAlchemyInstrumentor
+
+    SQLAlchemyInstrumentor().instrument(
+        enable_commenter=True,
+        commenter_options={},
+        enable_attribute_commenter=True,
```
lzchen, line 79: If the span attribute is specifically only `db.statement` perhaps the config name should reflect that?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
```diff
@@ -65,6 +65,30 @@
 ::
 Enabling this flag will add traceparent values /*traceparent='00-03afa25236b8cd948fa853d67038ac79-405ff022e8247c46-01'*/
 
+SQLComment in span attribute
+****************************
+If sqlcommenter is enabled, you can optionally configure SQLAlchemy instrumentation to append sqlcomment to query span attribute for convenience of your platform.
```
lzchen, line 70:
```suggestion
If sqlcommenter is enabled, you can further configure SQLAlchemy instrumentation to append sqlcomment to the `db.statement` span attribute for convenience of your platform.
```

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
```diff
@@ -65,6 +65,30 @@
...
+    )
+
+
+For example,
+::
+
+    Invoking engine.execute("select * from auth_users") will lead to sql query "select * from auth_users" but when SQLCommenter and attribute_commenter is enabled
+    the query will get appended with some configurable tags like "select * from auth_users /*tag=value*/;" for both server query and `db.statement` span attribute.
+
+Warning: capture of sqlcomment in ``db.statement`` or ``db.query.text`` may have high cardinality without platform normalization. See `Semantic Conventions for database spans <https://opentelemetry.io/docs/specs/semconv/database/database-spans/#generating-a-summary-of-the-query-text>`_ for more information.
```
lzchen, line 89: Since we are still in old semantic conventions, I would omit mentioning of `db.query.text`  for now.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -218,6 +230,32 @@ def _operation_name(self, db_name, statement):
...
+    def _get_commenter_data(self, conn) -> dict:
+        """Calculate sqlcomment contents from conn and configured options"""
+        commenter_data = {
+            "db_driver": conn.engine.driver,
+            # Driver/framework centric information.
+            "db_framework": f"sqlalchemy:{sqlalchemy.__version__}",
+        }
+
+        if self.commenter_options.get("opentelemetry_values", True):
+            commenter_data.update(**_get_opentelemetry_values())
```
lzchen, line 242: Related to this [comment](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2938#issuecomment-2541134100) it seems strange to have trace context related information in a span attribute. I am okay with sqlcommenting as part of the attribute but maybe we can omit the OT values?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
tammy-baylis-swi: Addressed in [2c8dbe3](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3112/commits/2c8dbe353e34d8dcf0db7c0e8e3211921c7dfb9c)

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
tammy-baylis-swi: Capturing Otel context in `db.statement` is a convenience for checking that sqlcommenter is enabled with the current config, and checking that the expected trace context was injected into the query statement sent from db client app to to the db server. The latter is an instrumentation checkup; duplicate `traceparent` can be sqlcommented if both SQLAlchemy (ORM) and db driver (e.g. pymysql/psycopg2) are enabled unintentionally and this could mess with the client-server correlation.

It's not a requirement since all can both be checked in the mysql/postgresql server general logs. But diagnosing issues is quicker with `db.statement` inclusion of full sqlcomment, e.g. looking at span attributes in a local otel-collector exporting to console versus accessing a mysql server and opening its logs. Or e.g. looking at spans exported to a platform versus getting creds and accessing a cloud-hosted db server.

lzchen: Not sure I fully understand

> e.g. looking at span attributes in a local otel-collector exporting to console

Wouldn't the otel-collector be exporting span related information as well? Can't we find the traceid, spanid, etc information there bundled with the span attributes?

tammy-baylis-swi: > Can't we find the traceid, spanid, etc information there bundled with the span attributes?

Yes, and this helps with `checking that the expected trace context was injected into the query statement` to make sure spanid of the span matches spanid of the sqlcomment.

The sqlcomment in span attribute can also contain db_driver. So if there is a spanid mismatch, then the span attribute can provide info about which instrumentor might be injecting the wrong spanid due to user config or due to bug in code.

tammy-baylis-swi:
@lzchen , @xrmx : regarding [this comment](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3112#discussion_r1894367232):

Would it be all right if we worked on getting this particular PR merged first (plus https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3115 and https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3121) to stop this feature from being On all the time in the next release?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
emdneto: Not a blocker. Just asked because I saw in semconv `The following attributes can be important for making sampling decisions and SHOULD be provided at span creation time (if provided at all):` https://opentelemetry.io/docs/specs/semconv/database/database-spans/

emdneto (approved):

lzchen (approved):

PR #3113, opened 2024-12-17
bug_fix(1477): type handling in _add_sql_comment

### LINKED ISSUE #1477: The sqlcommenter cannot take non string queries
We observe this issue using the sqlcommenter with the psycopg2 instrumentor. The commenter doesn't support all types of query execution in psycopg2.

**Steps to reproduce**
cursor.execute(sql.SQL("SELECT 1"))

Both `cursor.execute(sql.SQL("SELECT 1"))` or `cursor.execute("SELECT 1")` should work.
The sqlcommenter should be able to work with the psycopg2.sql.SQL() objects (which is a subclass of [psycopg2.sql.Composable](https://www.psycopg.org/docs/sql.html#psycopg2.sql.Composable)) passed into the `cursor.execute()`.

`cursor.execute(sql.SQL("SELECT 1"))` fails while `cursor.execute("SELECT 1")` works.
The sqlcommenter assumes that a Python `string` is passed in, and it will [try to call .rstrip()](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/25a6535f26ab6cb4f85a8c23da4e42fa50e218f9/opentelemetry-instrumentation/src/opentelemetry/instrumentation/sqlcommenter_utils.py#L25) on the query object. In case of a composable that is passed in, it would blow up and [eventually caught](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/25a6535f26ab6cb4f85a8c23da4e42fa50e218f9/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py#L462) in the execption.

aryabharat:
# Description

This PR fixes an issue with SQL comment generation, the current implementation fails when processing non-string SQL queries, particularly those from psycopg2's SQL utilities.
Modified _add_sql_comment() function to handles different SQL query types by converting incoming sql queries to strings
Preserves the original query if comment generation fails

Fixes #1477

## Type of change


xrmx (changes requested):
Thanks for the PR, could you please add a test so we are sure that the code is fixing the reported issue?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/sqlcommenter_utils.py
```diff
@@ -22,6 +22,8 @@ def _add_sql_comment(sql, **meta) -> str:
     """
     meta.update(**_add_framework_tags())
     comment = _generate_sql_comment(**meta)
+    # converting to str to handle any type errors
+    sql = str(sql)
```
Kludex, line 26: I think the type of `sql` should be `str` here, and we should fix the callers.

aryabharat: Is this the [caller](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/25a6535f26ab6cb4f85a8c23da4e42fa50e218f9/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py) function?

aryabharat: Hi @tammy-baylis-swi thanks for the reply.
Just one question, on handling the [composable object](https://www.psycopg.org/docs/sql.html#psycopg2.sql.Composable)
```
For handling SQL inputs:
If the input is a psycopg2.sql.Composable object (not a string), 
convert it to a string using the as_string() method.

Example:
    if isinstance(sql_input, sql.Composable):
        sql_string = sql_input.as_string(connection)
```

Should i handle like this?

tammy-baylis-swi: Yes, it's generally good if you can follow the psycopg2 docs and it fixes the type errors.

aryabharat: I have update the code, can you please check if this is the fine then will add the test case.

tammy-baylis-swi: Sorry @aryabharat I must have mixed up PRs in haste. Using that `psycopg2` utility will work within Psycopg2Instrumentor, but adding the usage/dependency might not work for the places where the changes are being made (DB-API, SQLAlchemy, Django). Please could you try these instead:

For SQLAlchemy, their [documentation](https://docs.sqlalchemy.org/en/20/faq/sqlexpressions.html#how-do-i-render-sql-expressions-as-strings-possibly-with-bound-parameters-inlined) says to use `str()`, so please switch to that. It's like the change you originally proposed, but in the Sqlalchemy instrumentor instead.

For DB-API and Django it can depend on the db driver being used. (DB-API instrumentor more of a shared utility rather; Django is an ORM). Sometimes it is `psycopg2` but user could be using `mysqlclient` instead, for example. Using `str()` _might_ work just as well, but should be tested for different cases.

aryabharat: Hi @tammy-baylis-swi , thanks for your help and sorry for the delay.

What I have done for SQLAlchemy is to use str() directly. 

However, in the case of DB-API and Django with `psycopg2,` we cannot use str(), as it returns an unformatted query. So, I handled the psycopg2 case by checking the function as_string().

For `mysqlclient`, I handled it by using str() only.

Please let me, if you have any inputs.

tammy-baylis-swi: Thanks for pushing! I think those approaches look ok and the comments help. Please could you also update the unit tests.

aryabharat: @tammy-baylis-swi I have added test cases. Please let me know if you need any changes.

tammy-baylis-swi:
Thanks @aryabharat . Please could you update CHANGELOG.md with this PR?

CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `opentelemetry-instrumentation-dbapi` Add support for non string query and composable object.
+  ([#3113](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3113))
+- `opentelemetry-instrumentation-django` Add support for non string query and composable object.
+  ([#3113](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3113))
+- `opentelemetry-instrumentation-sqlalchemy` Add support for non string query.
+  ([#3113](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3113))
```
tammy-baylis-swi, line 21: @aryabharat I suggest making this one line in the log and moving it under the **Fixed** subheader below:

```suggestion
- `opentelemetry-instrumentation-dbapi`, `opentelemetry-instrumentation-django`, `opentelemetry-instrumentation-sqlalchemy`: Fix sqlcomment for non string query and composable object.
  ([#3113](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3113))
```

aryabharat:
@tammy-baylis-swi Please let me what all changes are required.

tammy-baylis-swi (approved):
Thanks again! Lgtm

xrmx (approved):

aryabharat:
@xrmx @tammy-baylis-swi can we merge this?

aryabharat:
Hi @Kludex can we merge this?

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -498,6 +498,11 @@ def _update_args_with_added_sql_comment(self, args, cursor) -> tuple:
             args_list = list(args)
             self._capture_mysql_version(cursor)
             commenter_data = self._get_commenter_data()
+            # Convert sql statement to string, handling  psycopg2.sql.Composable object
+            if hasattr(args_list[0], "as_string"):
+                args_list[0] = args_list[0].as_string(cursor.connection)
```
Kludex, line 503: Is this path reached by both psycopg and psycopg2?

aryabharat: yes it will be reached by both:
https://www.psycopg.org/docs/sql.html#psycopg2.sql.Composable.as_string

aryabharat: @Kludex

Kludex:
Please stop pinging me.

I do not have bandwidth to review this again. I've shared my concerns already.

aryabharat:
> Please stop pinging me.
> 
> I do not have bandwidth to review this again. I've shared my concerns already.

I have replied to your comment.

aryabharat:
@xrmx Please have a look, if this is not accepted then we can close this PR.
Thanks.

xrmx:
@aryabharat Please stop pinging people, this will be merged eventually.

emdneto (approved):

lzchen (approved):

PR #3116, opened 2024-12-18
Add type checker

Kludex:
This PR adds type checker to `tox`.

### PyRight Vs MyPy

There's a lot of discussion around about which one to use. I don't think it matters much which one to choose. There are points where Python itself is not specific, and then the type checkers have freedom to choose the behavior they want... Example:

- `mypy`: On strict mode, you need to add `-> None` on the return statement.
- `pyright`: infers that is `None`, so you don't need to add `-> None`.

PyRight is easier for VSCode users, me included... 👀

About the ecosystem... I think most of them use mypy, but [Pydantic](https://github.com/pydantic/pydantic/blob/a915c7cd717be767b37867e7ff4bd69be2d4c627/pyproject.toml#L259) for example, uses PyRight.

If you look at the projects I maintain (starlette, uvicorn), I use mypy on them, but I have to say that is a bit more practical to have the type checker match my IDE.

Anyway... I hope this is not an issue, and we can start with this setup.


dev-requirements.txt
```diff
@@ -1,6 +1,7 @@
 pylint==3.0.2
 httpretty==1.1.4
 mypy==0.931
```
Kludex, line 3: This can be removed. I didn't because I don't know who is using it...

tox.ini
```diff
@@ -963,3 +963,14 @@ deps =
   pre-commit
 commands =
   pre-commit run --color=always --all-files {posargs}
+
+[testenv:typecheck]
+description = type checker
```
Kludex, line 968:
```suggestion
```

Kludex:
cc @emdneto (asked about this on PM) and @lzchen (since you been happy with the type hints PRs 👀 👍)

instrumentation/opentelemetry-instrumentation-threading/src/opentelemetry/instrumentation/threading/__init__.py
Kludex, line 1: I started with this instrumentation because it was the easiest I found to prove that the setup works.

xrmx:
I don't have any preference but I think you missed that we are using mypy in opentelemetry-python

tox.ini
```diff
@@ -963,3 +963,13 @@ deps =
   pre-commit
 commands =
   pre-commit run --color=always --all-files {posargs}
+
+[testenv:typecheck]
```
emdneto, line 967: I would say we can use the name of the tool here, like mypy or pyright. For example: [.github/workflows/lint_0.yml](https://github.com/open-telemetry/opentelemetry-python/blob/main/.github/workflows/misc_0.yml#L55)

dev-requirements.txt
emdneto: I think there's no usage on this repo. But we are using mypy in core + a basic setup of pyright with typeCheckingMode off.

tox.ini
Kludex: But... The `lint` jobs only run `pylint`, and the `test` jobs run `pytest`. 😅

Kludex:
I was checking opentelemetry-python, and it seems that pyright is also present there?

https://github.com/open-telemetry/opentelemetry-python/blob/17782492f5cec2e093fcd02e983d7f5e638b7cca/tox.ini#L349-L358

A bit more into it... Seems like if `pyright` looks at the pyproject, the type checker is disabled... 🤔 

https://github.com/open-telemetry/opentelemetry-python/blob/17782492f5cec2e093fcd02e983d7f5e638b7cca/pyproject.toml#L5-L12

Kludex:
I took some time trying to set up mypy to see how that would look like, and it just took too much of my time, and I was not able to make it work as I wanted...

I'm happy to close this if people are strong on mypy, but I'm not willing to spend more time trying it.

Kludex:
I think the mypy setup on `opentelemetry-python` is not taking in consideration all local packages.

It would be great if people could share opinions on this, I want to continue adding type hints, but without regressions.

emdneto (approved):
Thanks. Discussed this at SIG (01/09) and agreed to move forward with pyright in contrib and include some packages to be checked. After that, also adopt pyright at core as well.

nit: can we include the packages you've added type hints?

Kludex:
Not yet, cause I've mainly added type hints on public APIs. I need to add more stuff to them that I was not very strict.

dev-requirements.txt
emdneto, line 3:
```suggestion
```
remove this this then?

.github/workflows/misc_0.yml
aabmass, line 1: I think you need to add this to [the Jinja template](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/9515bc64efd65892a7e4307a13f4044e84dd2779/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/misc.yml.j2) instead and then regenerate this file with `tox -e generate-workflows`.

aabmass (approved):
Very exciting, thank you!

.github/workflows/misc_0.yml
emdneto: I think he already did that; For new misc jobs we just ned to add in tox and run `tox -e generate-workflows` and it will update the misc_0.yml file

pyproject.toml
```diff
@@ -39,3 +39,18 @@ known-third-party = [
   "opencensus",
 ]
 
+# https://github.com/microsoft/pyright/blob/main/docs/configuration.md#type-check-rule-overrides
+[tool.pyright]
+typeCheckingMode = "strict"
+reportUnnecessaryTypeIgnoreComment = true
+reportMissingTypeStubs = false
```
emdneto, line 46:
```suggestion
reportMissingTypeStubs = false
pythonVersion = "3.8"
```
to run with the lowest supported version?

Kludex: This is not needed.

emdneto: If not set it will run with the default in CI, like 3.11 no?

Kludex: Ah, I see. Where's that version coming from?

emdneto: From jinja2 templates which generates the workflows.

.github/workflows/misc_0.yml
aabmass: Sorry yup, I forgot it pulls from the tox file.

xrmx (approved):

tjordahl:
@Kludex - This change prevent mypy from running on users of the opentelemetry-instrumentation-threading library with the follow error:
```
venv/lib/python3.10/site-packages/opentelemetry/instrumentation/threading/__init__.py:132: error: Ellipses cannot accompany other argument types in function type signature  [syntax]
Found 1 error in 1 file (errors prevented further checking)
```
This is a serious problem for my project as convincing mypy not to check this file is difficult bordering on impossible.

tjordahl:
Opened the issue https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3242
CC: @Kludex @lzchen

PR #3117, opened 2024-12-18
Populate db.statement when Composable passed to cursor.execute

oparaskos:
# Description

We were getting warnings such as 
```
[...] WARNING: opentelemetry.attributes: Invalid type Composed for attribute 'db.statement' value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
```

We're using aiopg and its autoinstrumentor `db.statement` isn't being populated because we're passing in the SQL() object
the psycopg2 cursor wrapper supports this but the aiopg only seems to support string and bytes.

This adds support for psycopg.sql.Composable objects, by calling as_string() which is similar to the solution in the psycopg autoinstrumentors.

Fixes # (issue not created)

## Type of change

# How Has This Been Tested?

- [x]  tox -e py312-test-instrumentation-aiopg

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

N/A
N/A


xrmx:
@oparaskos Hello, thanks for the PR, any reason it's in draft?

oparaskos:
> @oparaskos Hello, thanks for the PR, any reason it's in draft?

In the GitHub app it looked like there was a failing pipeline stage, so I marked it as draft til I could look at it. Though now it looks like there isn't a failing step so I don't know what happened there.
I also still haven't had a chance to test it in a real scenario beyond the unit tests.

lucianoVideam:
Hello, we are experiencing the same issue with psycopg using psycopg.sql.SQL format for the statement. 

```
async def check_db_working(conn: psycopg.AsyncConnection) -> bool:
    async with conn.cursor() as cursor:
        query_str = sql.SQL('SELECT 1')
        await cursor.execute(query_str)
        [...]
```

it would be possible to avoit getting the warngin and parse the sb.statement also in psycopg.sql.SQL format? we are using the `opentelemetry-instrumentation-psycopg` module.

Thanks.

Current warning: `Invalid type SQL for attribute 'db.statement' value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types`

PR #3110, opened 2024-12-13
Add type hints to Redis

Kludex:

PR #3118, opened 2024-12-18
Fix typing for redis

lzchen:
Followup to https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3110

I believe [Pep 0604](https://peps.python.org/pep-0604/) which introduces union typing syntax is only supported for Python 3.10+. Since all instrumentations still support 3.8+, I believe we still need to adhere to the explicit `Optional` and `Union` typings.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-redis/src/opentelemetry/instrumentation/redis/__init__.py
@@ -159,7 +159,8 @@ _FIELD_TYPES = ["NUMERIC", "TEXT", "GEO", "TAG", "VECTOR"]
 
 
 def _set_connection_attributes(
-    span: Span, conn: Union[AsyncRedisInstance, RedisInstance],
+    span: Span,
+    conn: Union[AsyncRedisInstance, RedisInstance],
 ) -> None:
     if not span.is_recording() or not hasattr(conn, "connection_pool"):
         return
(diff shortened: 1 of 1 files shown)


Kludex (changes requested):
@lzchen This is not needed.

The `from __future__ import annotations` on the first line makes sure Python will not look at the type hints at runtime - Which allow us to use type hints from more modern versions.

Also... The pipeline for Python 3.8 and 3.9 is passing... 😅

samuelcolvin:
@lzchen seriously?!

samuelcolvin:
@lzchen this change is not necessary.

As of [pep-0563](https://peps.python.org/pep-0563/) introduced in Python 3.7, the magic import

``py
```

causes Python to not evaluate type hints, this in turn means we can use new syntax like `|` unions even with older versions of Python. Static type checkers like mypy and pyright can still read and interpret these type annotations, so there's no drawback.

Many popular python packages like Pydantic, FastAPI, rich, cryptography use this technique.

Even Microsoft (who I believe you work for) make extensive use of `from __future__ import annotations` to adopt modern typing while supporting older versions of python:

In the most downloaded python package from Microsoft [`azure-core`](https://github.com/Azure/azure-sdk-for-python), there are 82 references to `from __future__ import annotations`, indeed [`azure-sdk-for-python/doc/dev
/static_type_checking.md`](https://github.com/Azure/azure-sdk-for-python/blob/57de74af5c31b3e71ffd0bc37a4d1a22aebd0561/doc/dev/static_type_checking.md?plain=1#L134) includes this:

> *Note: Supported in Python 3.9+. For <3.9, You must include `from __future__ import annotations` import to be able to pass in generic `list[str]` as a type hint rather than `typing.List[str]`.

emdneto (commented):
Thanks for your concern, everybody. We already discussed this offline in Slack yesterday, and there's already an agreement to drop the PR. So, no need to worry about ✌🏻

Issue #3065, opened 2024-12-03
Library for uploading blobs as part of instrumentation #GenAI #MultiModal


michaelsafyan:
### What problem do you want to solve?

Many (if not most) observability backends are not built to accept arbitrarily large blobs.

However, GenAI observability introduces a use case where users want to log full prompt/response data. In "multi-modal" use cases where large files (PDFs, PNGs, JPEGs, etc.) are involved, users still want these to be recorded. An example use case: the user prompts an LLM with a PDF and asks a question about it, asking to generate an infographic summarizing the essential details; the LLM responds with summary text as well as a JPEG infographic summarizing the salient points.  GenAI Semantic conventions are aiming towards recording the prompt/response details in event bodies (which end up in logs), but many logging backends are not capable of receiving these large blob payloads.

In [OpenLLMetry](https://github.com/traceloop/openllmetry), this has been addressed with an [`ImageUploader`](https://github.com/traceloop/openllmetry/blob/8b1f29249e7094dcfd7319a85766a8ac6557f698/packages/traceloop-sdk/traceloop/sdk/images/image_uploader.py), which is very specific to the Traceloop backend. It is likely that generic instrumentation of GenAI frameworks (such as moving frameworks from OpenLLMetry to the OTel Python repo) may require a more generic alternative to provide such capability/functionality.

### Describe the solution you'd like

At a high-level, I would like to separate this out into:

 1. **A consumption interface** that is aimed at implementors of instrumentation packages.
 2. **A producer interface/library** that is aimed at those trying to provide a storage backend for such blobs.
 3. **Default implementations** that use the above to provide useful out-of-the-box functionality

### NOT_UPLOADED

A constant used as a response when failing to upload.

```
NOT_UPLOADED = '/dev/null'
```

### Blob

Encapsulates the raw payload together with associated properties:

```
    class Blob(object):

        def __init__(self, raw_bytes, content_type=None, labels=None):
            ...

        @property
        def raw_bytes(self) -> bytes:
            ...
  
        @property
        def content_type(self) -> Optional[str]:
            ...

         @property
         def labels(self) -> dict:
             ... 
```

### get_blob_uploader

A function used in instrumentation to retrieve the uploader for a certain context/usage:

```
   def get_blob_uploader(use_case: Optional[str] = None) -> BlobUploader:
       ...
```

[...]

PR #3122, opened 2024-12-19
[DRAFT] BlobUploader utilities to enable handling of large data in instrumentation

michaelsafyan:
# Draft

This is a draft of the solution related to https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3065 . This is not yet ready for review.

# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

### DIFF
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/blob.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/blob.py
@@ -1,0 +1,72 @@
+import base64
+
+class Blob(object):
+    """Represents an opaque binary object and associated metadata.
+    
+    This object conteptually has the following properties:
+
+      - raw_bytes: the actual data (payload) of the Blob
+      - content_type: metadata about the content type (e.g. "image/jpeg")
+      - labels: key/value data that can be used to identify and contextualize
+         the object such as {"trace_id": "...", "span_id": "...", "filename": ...}
+    """
+
+    def __init__(self, raw_bytes: bytes, content_type: Optional[str]=None, labels: Optional[dict]=None):
+        """Initialize the blob with an explicit set of properties.
+        
+        Args:
+          raw_bytes: the required payload
+          content_type: the MIME type describing the type of data in the payload
+          labels: additional key/value data about the Blob
+        """
+        self._raw_bytes = _raw_bytes
+        self._content_type = content_type
+        self._labels = labels or {}
+
+    @staticmethod
+    def from_data_uri(cls, uri: str, labels: Optional[dict]=None) -> Blob:
+        """Instantiate a blob from a 'data:...' URI.
+
+        Args:        
+            uri: A URI in the 'data:' format. Supports a subset of 'data:' URIs
+              that encode the data with the 'base64' extension and that include
+              a content type. Should work with any normal 'image/jpeg', 'image/png',
+              'application/pdf', 'audio/aac', and many others. DOES NOT SUPPORT
+              encoding data as percent-encoded text (no "base64").
+    
+            labels: Additional key/value data to include in the constructed Blob.
+        """
+        if not uri.startswith('data:'):
+            raise ValueError('Invalid "uri"; expected "data:" prefix. Found: "{}"'.format(uri))
+        if not ';base64,' in uri:
+            raise ValueError('Invalid "uri"; expected ";base64," section. Found: "{}"'.format(uri))
+        data_prefix_len = len('data:')
+        after_data_prefix = uri[data_prefix_len:]
+        if ';' not in after_data_prefix:
+            raise ValueError('Invalid "uri"; expected ";" in URI. Found: "{}"'.format(uri))
+        content_type, remaining = after_data_prefix.split(';', 1)
+        while not remaining.startswith('base64,'):
+            _, remaining = remaining.split(';', 1)
+        assert remaining.startswith('base64,')
+        base64_len = len('base64,')
+        base64_encoded_content = remaining[base64_len:]
+        try:
+            raw_bytes = base64.standard_b64decode(base64_encoded_content)
+        except ValueError:
+            raw_bytes = base64.urlsafe_b64decode(base64_encoded_content)
+        return Blob(raw_bytes, content_type=content_type, labels=labels)
+
+    @property
+    def raw_bytes(self) -> bytes:
+        """Returns the raw bytes (payload) of this Blob."""
+        return self._raw_bytes
+
+    @property
+    def content_type(self) -> Optional[str]:
+        """Returns the content type (or None) of this Blob."""
+        return self._content_type
+
+    @property
+    def labels(self) -> dict:
+        """Returns the key/value metadata of this Blob."""
+        return _frozendict(self._labels)
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/blob_uploader.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/blob_uploader.py
@@ -1,0 +1,13 @@
+"""Defines an interface for performing asynchronous blob uploading."""
+
+import abc
+
+from opentelemetry.instrumentation._blobupload.api.constants import NOT_UPLOADED
+from opentelemetry.instrumentation._blobupload.api.blob import Blob
+
+class BlobUploader(abc.ABC):
+    """Pure abstract base class representing a component that does blob uploading."""
+    
+    @abc.abstractmethod
+    def upload_async(self, blob: Blob) -> str:
+        return NOT_UPLOADED
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/content_type.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/content_type.py
@@ -1,0 +1,24 @@
+"""Provides utilities for automatic content-type detection."""
+
+# Helper used to handle the possibility of optional 'magic' dependency
+# being unavailable for guessing the MIME type of raw bytes.
+class _FallBackModule(object):
+    """Class that is shaped like the portion of 'magic' we need."""
+
+    def from_buffer(self, raw_bytes):
+        """Fallback, subpar implementation of 'from_buffer'."""
+        return 'application/octet-stream'
+
+
+# Set up '_module' to either use 'magic' or the fallback.
+_module = _FallBackModule()
+try:
+    import magic
+    _module = magic
+except ImportError:
+    pass
+
+
+def detect_content_type(raw_bytes: bytes) -> str:
+    """Attempts to infer the content type of the specified data."""
+    return _module.from_buffer(raw_bytes)
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/labels.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/labels.py
@@ -1,0 +1,29 @@
+"""Provides utilities for providing basic identifying labels for blobs."""
+
+def generate_labels_for_span(trace_id, span_id) -> dict:
+    """Returns metadata for a span."""
+    return {
+        'otel_type': 'span',
+        'trace_id': trace_id,
+        'span_id': span_id
+    }
+
+
+def generate_labels_for_event(trace_id, span_id, event_name) -> dict:
+    """Returns metadata for an event."""
+    result = generate_labels_for_span(trace_id, span_id)
+    result.update({
+        'otel_type': 'event',
+        'event_name': event_name,
+    })
+    return result
+
+
+def generate_labels_for_span_event(trace_id, span_id, event_name, event_index) -> dict:
+    """Returns metadata for a span event."""
+    result = generate_labels_for_event(trace_id, span_id, event_name)
+    result.update({
+        'otel_type': 'span_event',
+        'event_index': event_index,
+    })
+    return result
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/utils/simple_blob_uploader_adaptor.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/utils/simple_blob_uploader_adaptor.py
@@ -1,0 +1,49 @@
+from concurrent.futures import Executor, ProcessPoolExecutor
+
+from opentelemetry.instrumentation._blobupload.api import Blob
+from opentelemetry.instrumentation._blobupload.api import BlobUploader
+from opentelemetry.instrumentation._blobupload.api import detect_content_type
+from opentelemetry.instrumentation._blobupload.utils.simple_blob_uploader import SimpleBlobUploader
+
+
+def _with_content_type(blob: Blob) -> Blob:
+    if blob.content_type is not None:
+        return blob
+    content_type = detect_content_type(blob.raw_bytes)
+    return Blob(blob.raw_bytes, content_type=content_type, labels=blob.labels)
+
+
+def _UploadAction(object):
+
+    def __init__(self, simple_uploader, uri, blob):
+        self._simple_uploader = simple_uploader
+        self._uri = uri
+        self._blob = blob
+    
+    def __call__(self):
+        self._simple_uploader.upload_sync(self._uri, self._blob)
+
+
+def _create_default_executor():
+    return ProcessPoolExecutor()
+
+
+class _SimpleBlobUploaderAdaptor(BlobUploader):
+
+    def __init__(self, simple_uploader, executor=None):
+        self._simple_uploader = simple_uploader
+        self._executor = executor or _create_default_executor()
+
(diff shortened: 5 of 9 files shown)


opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/blob_uploader.py
```diff
@@ -0,0 +1,16 @@
+"""Defines an interface for performing asynchronous blob uploading."""
+
+import abc
```
haf, line 3: Why not use a modern `Protocol`?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/content_type.py
```diff
@@ -0,0 +1,28 @@
+"""Provides utilities for automatic content-type detection."""
+
+
+# Helper used to handle the possibility of optional 'magic' dependency
+# being unavailable for guessing the MIME type of raw bytes.
+class _FallBackModule(object):
```
haf, line 6: Why do you inherit object? https://stackoverflow.com/a/45062077/63621

michaelsafyan:
Ran `tox`, but this concerningly attempted to run `sudo`; I quit the tool at that point. Will file a bug for this behavior.

michaelsafyan:
Looks like I'm already getting some review comments, so will convert from DRAFT to READY.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/blob_uploader.py
michaelsafyan: Thanks for the early review.

The primary reason was that I was not familiar with `Protocol` (I learned Python a really, really long time ago, so my Python knowledge is not fully current).

Having now read up on this topic, I am somewhat hesitant to switch to a `Protocol` because it does not require explicit inheritance. This might be my bias (coming primarily from C++ and Java land), but while there are cases where duck typing is justifiable, I don't think that should be the default (and should be strongly justified). For example, the presence of  a compatible `upload_async` signature does not necessarily mean that the semantics of that function are intended to match that of this interface ... explicit inheritance makes it clearer that the derived object will adhere to the Liskov Substitution Principle. And it also makes it clearer how things are tied together.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/content_type.py
michaelsafyan: Removed.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/blob_uploader.py
haf: 1. It's statically typed duck typing; it's type-checked against the protocol where the protocol interface is expected, alike how interfaces are in Go.
2. From my understanding of reading PIPs and googling around, ABC is good if you want to create value objects, or if you need a hierarchy of objects
3. you're trying to assert on behaviour rather than object hierarchy, and you also don't care about `isinstance` checks, which would be another reason for ABC
4. you also don't use any template methods or other reused logic or create any base class invariants

=> IMO it should be a protocol.

haf: 1. It's statically typed duck typing; it's type-checked against the protocol where the protocol interface is expected, alike how interfaces are in Go.
2. From my understanding of reading PIPs and googling around, ABC is good if you want to create value objects, or if you need a hierarchy of objects
3. you're trying to assert on behaviour rather than object hierarchy, and you also don't care about `isinstance` checks, which would be another reason for ABC
4. you also don't use any template methods or other reused logic or create any base class invariants

=> IMO it should be a protocol.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/blob.py
```diff
@@ -0,0 +1,125 @@
...
+            (isinstance(o, Blob)) and
+            (self.raw_bytes == o.raw_bytes) and
+            (self.content_type == o.content_type) and
+            (self.labels == o.labels)
+        )
+
+    def __repr__(self):
+        params = [repr(self._raw_bytes)]
+        if self._content_type is not None:
+            params.append("content_type={}".format(repr(self._content_type)))
```
samuelcolvin, line 121:
```suggestion
            params.append(f"content_type={self._content_type!r}")
```
you don't need to call repr outside the formatting! This would work with `.format()` too.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/blob.py
```diff
@@ -0,0 +1,125 @@
...
+        if self._labels:
+            params.append("labels={}".format(json.dumps(self._labels, sort_keys=True)))
```
samuelcolvin, line 123: why are we JSON formatting? this will be much slower, and not lead to the correct repr.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/blob.py
```diff
@@ -0,0 +1,125 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import base64
+import json
+from types import MappingProxyType as _frozendict
+from typing import Mapping, Optional
+
+
+class Blob:
```
samuelcolvin, line 21: If you implemented this as a dataclass (perhaps frozen), you could avoid having to implement:
* most of `__init__`
* the three `@property` methods
* `__eq__` probably
* `__repr__`

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/blob_uploader.py
```diff
@@ -0,0 +1,30 @@
...
+from opentelemetry.instrumentation._blobupload.api.constants import (
+    NOT_UPLOADED,
+)
+
+
+class BlobUploader(abc.ABC):
+    """Pure abstract base class representing a component that does blob uploading."""
+
+    @abc.abstractmethod
+    def upload_async(self, blob: Blob) -> str:
```
samuelcolvin, line 29: is there any support for async upload methods?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/content_type.py
```diff
@@ -0,0 +1,43 @@
...
+
+"""Provides utilities for automatic content-type detection."""
+
+
+# Helper used to handle the possibility of optional 'magic' dependency
+# being unavailable for guessing the MIME type of raw bytes.
+class _FallBackModule:
+    """Class that is shaped like the portion of 'magic' we need."""
+
+    def from_buffer(self, raw_bytes, mime=True):
```
samuelcolvin, line 24:
```suggestion
    def from_buffer(self, raw_bytes: bytes, mime: bool = True):
```
?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/labels.py
```diff
@@ -0,0 +1,48 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Provides utilities for providing basic identifying labels for blobs."""
+
+
+def generate_labels_for_span(trace_id: str, span_id: str) -> dict:
```
samuelcolvin, line 18:
```suggestion
def generate_labels_for_span(trace_id: str, span_id: str) -> dict[str, str]:
```
But shouldn't this be a typeddict?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/labels.py
```diff
@@ -0,0 +1,48 @@
...
+    """Returns metadata for a span."""
+    return {"otel_type": "span", "trace_id": trace_id, "span_id": span_id}
+
+
+def generate_labels_for_event(
+    trace_id: str, span_id: str, event_name: str
+) -> dict:
```
samuelcolvin, line 25:
```suggestion
) -> dict[str, str]:
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/api/labels.py
```diff
@@ -0,0 +1,48 @@
...
+            "otel_type": "event",
+            "event_name": event_name,
+        }
+    )
+    return result
+
+
+def generate_labels_for_span_event(
+    trace_id: str, span_id: str, event_name: str, event_index: int
+) -> dict:
```
samuelcolvin, line 39:
```suggestion
) -> dict[str, str]:
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/backend/google/gcs/_gcs_impl.py
```diff
@@ -0,0 +1,155 @@
...
+    SimpleBlobUploader,
+    blob_uploader_from_simple_blob_uploader,
+)
+
+_logger = logging.getLogger(__name__)
+
+GcsClient: TypeAlias = _gcs_client_wrapper.GcsClientType
+
+
+def _path_for_span(trace_id, span_id):
```
samuelcolvin, line 36:
```suggestion
def _path_for_span(trace_id: str, span_id: str) -> str:
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/backend/google/gcs/_gcs_impl.py
```diff
@@ -0,0 +1,155 @@
...
+    if not trace_id or not span_id:
+        return ""
+    return "traces/{}/spans/{}".format(trace_id, span_id)
+
+
+def _path_for_event(trace_id, span_id, event_name):
```
samuelcolvin, line 42:
```suggestion
def _path_for_event(trace_id: str, span_id: str, event_name: str) -> str:
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/backend/google/gcs/_gcs_impl.py
```diff
@@ -0,0 +1,155 @@
...
+    if not event_name:
+        return ""
+    span_path = _path_for_span(trace_id, span_id)
+    if not span_path:
+        return ""
+    return "{}/events/{}".format(span_path, event_name)
+
+
+def _path_for_span_event(trace_id, span_id, event_index):
```
samuelcolvin, line 51:
```suggestion
def _path_for_span_event(trace_id: str, span_id: str, event_index) -> str:
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/backend/google/gcs/_gcs_impl.py
```diff
@@ -0,0 +1,155 @@
...
+    if event_index is None:
+        return ""
+    span_path = _path_for_span(trace_id, span_id)
+    if not span_path:
+        return ""
+    return "{}/events/{}".format(span_path, event_index)
+
+
+def _path_segment_from_labels(labels):
```
samuelcolvin, line 60:
```suggestion
def _path_segment_from_labels(labels: Mapping[str, str]) -> str:
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/backend/google/gcs/_gcs_impl.py
```diff
@@ -0,0 +1,155 @@
...
+    """Returns a path segment based on blob label metadata.
+
+    This aims to return paths like:
+
+       'traces/12345/spans/56789'
+       'traces/12345/spans/56789/events/0'
+       'traces/12345/spans/56789/events/some.event.name'
```
samuelcolvin, line 67: what happens if we want to include some kind of customer or project reference in the path?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/backend/google/gcs/_gcs_impl.py
```diff
@@ -0,0 +1,155 @@
...
+        return _path_for_span(trace_id, span_id)
+    elif signal_type == "event":
+        return _path_for_event(trace_id, span_id, event_name)
+    elif signal_type == "span_event":
+        return _path_for_span_event(trace_id, span_id, event_index)
+
+
+class _SimpleGcsBlobUploader(SimpleBlobUploader):
+
+    def __init__(self, prefix: str, client:Optional[GcsClient]=None):
```
samuelcolvin, line 89:
```suggestion
    def __init__(self, prefix: str, client: Optional[GcsClient] = None):
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/utils/simple_blob_uploader_adaptor.py
```diff
@@ -0,0 +1,138 @@
...
+)
+
+_logger = logging.getLogger(__name__)
+
+
+def _with_content_type(blob: Blob) -> Blob:
+    """Returns a variant of the Blob with the content type auto-detected if needed."""
+    if blob.content_type is not None:
+        return blob
+    content_type = detect_content_type(blob.raw_bytes)
```
samuelcolvin, line 47: can't we infer the content type from the labels, instead of inspecting the bytes?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/utils/simple_blob_uploader_adaptor.py
```diff
@@ -0,0 +1,138 @@
...
+    return _default_executor
+
+
+class _SimpleBlobUploaderAdaptor(BlobUploader):
+    """Implementation of 'BlobUploader' wrapping a 'SimpleBlobUploader'.
+
+    This implements the core of the function 'blob_uploader_from_simple_blob_uploader'.
+    """
+
+    def __init__(self, simple_uploader: SimpleBlobUploader, executor: Optional[Executor]=None):
```
samuelcolvin, line 112:
```suggestion
    def __init__(self, simple_uploader: SimpleBlobUploader, executor: Optional[Executor] = None):
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_blobupload/utils/simple_blob_uploader_adaptor.py
```diff
@@ -0,0 +1,138 @@
...
+        self._simple_uploader = simple_uploader
+        self._executor = executor or _get_or_create_default_executor()
+
+    def upload_async(self, blob: Blob) -> str:
+        full_blob = _with_content_type(blob)
+        uri = self._simple_uploader.generate_destination_uri(full_blob)
+        self._do_in_background(_UploadAction(self._simple_uploader, uri, full_blob))
+        return uri
+
+    def _do_in_background(self, action):
```
samuelcolvin, line 122:
```suggestion
    def _do_in_background(self, action: _UploadAction) -> None:
```
?

samuelcolvin (changes requested):
I've review most of the code and suggested type hints, and a few simple improvements like f-strings.

But more generally I suggest this should be fundamentially reconsidered:

The obvious and most scalable way to implement blob uploads today is using pre-signed URLs.

The idea would be:
1. the client makes a get or post request to an endpoint provided by the backend with the file type (optional), name (optional) and size (you could implement support for more specific attributes like width/height etc)
2. The backend returns a URL (most likely an S3 style pre-signed URL), and a reference
3. the client posts the data to that URL, raises and error if the response is not 2XX
4. the client stores the reference in the OTEL data

There are lots of advantages of this approach IMHO:
* it means the client needs to implement ZERO logic related to different providers and object stores, it just gets a URL and posts data to it
* this pre-signed url approach is already implement by S3, GCS and ever other S3 compatible object store, so it should be pretty easy for backends to implement
* if backends want to implement things differently, they can, the client logic is completely independent of the signing method, destination URL etc.

michaelsafyan:
Apologies for the delay. 

Full transparency: this is being deprioritized behind work to add instrumentation for the GenAI SDK (github.com/googleapis/python-genai).

I will probably not be able to pick this work up again until that is completed.

PR #3124, opened 2024-12-20
Use UV to install all needed packages locally

Kludex:
This is specifically to have a single virtual environment with all dependencies to make it easier to develop here.


pyproject.toml
```diff
@@ -1,11 +1,39 @@
...
+[tool.uv]
+package = false  # https://docs.astral.sh/uv/reference/settings/#package
+
+# https://docs.astral.sh/uv/reference/settings/#sources
+[tool.uv.sources]
+opentelemetry-api = { git = "https://github.com/open-telemetry/opentelemetry-python", branch = "main", subdirectory = "opentelemetry-api" }
+opentelemetry-sdk = { git = "https://github.com/open-telemetry/opentelemetry-python", branch = "main", subdirectory = "opentelemetry-sdk" }
+opentelemetry-semantic-conventions = { git = "https://github.com/open-telemetry/opentelemetry-python", branch = "main", subdirectory = "opentelemetry-semantic-conventions" }
+opentelemetry-instrumentation = { workspace = true }
+opentelemetry-instrumentation-httpx = { workspace = true }
```
Kludex, line 25: I've only added httpx.

Kludex: I added more, since I was working on adding type hints on other packages.

emdneto (commented):
Loved this. I would say other than speed-up local development, this paves the way to adopt uv in opentelemetry-python. 

I think it will be necessary to add some instructions on how to use that. But just a note for reviewers:

With a simple command like `uv sync` you can have all packages installed locally. If you want to install only one package you can run: `uv sync --package opentelemetry-instrumentation-sqlalchemy` for example.

xrmx:
> Loved this. I would say other than speed-up local development, this paves the way to adopt uv in opentelemetry-python.
> 
> I think it will be necessary to add some instructions on how to use that. But just a note for reviewers:
> 
> With a simple command like `uv sync` you can have all packages from the project installed locally. If you want to install only one package you can run: `uv sync --package opentelemetry-instrumentation-sqlalchemy` for example.

Sure but aren't we adding another source of dependabot warnings? I'm fine on adding the metadata to the pyproject for people using uv but aren't we duplicating what we use tox for?

Kludex:
> Sure but aren't we adding another source of dependabot warnings?

Unfortunately/Fortunately, Dependabot doesn't see the `uv.lock` yet.

> I'm fine on adding the metadata to the pyproject for people using uv but isn't this duplicating what we use tox for?

Not really. The idea here is to have a single virtual environment that can be used by the IDE, and to simplify local development (running some scripts locally to test some integrations, etc). The tox is used to run the tests with multiple different environments, it doesn't really create this virtual environment.

idan-rahamim-lendbuzz:
Is it possible that `opentelemetry-bootstrap --action=install` would install with UV/Poetry instead of pip?

Kludex:
> Is it possible that `opentelemetry-bootstrap --action=install` would install with UV/Poetry instead of pip?

This is not related with this PR, please create an issue.

pyproject.toml
```diff
@@ -1,11 +1,149 @@
...
+# https://docs.astral.sh/uv/reference/settings/#sources
+[tool.uv.sources]
+opentelemetry-api = { git = "https://github.com/open-telemetry/opentelemetry-python", branch = "main", subdirectory = "opentelemetry-api" }
+opentelemetry-sdk = { git = "https://github.com/open-telemetry/opentelemetry-python", branch = "main", subdirectory = "opentelemetry-sdk" }
+opentelemetry-semantic-conventions = { git = "https://github.com/open-telemetry/opentelemetry-python", branch = "main", subdirectory = "opentelemetry-semantic-conventions" }
+opentelemetry-test-utils = { git = "https://github.com/open-telemetry/opentelemetry-python", branch = "main", subdirectory = "tests/opentelemetry-test-utils" }
+opentelemetry-exporter-prometheus-remote-write = { workspace = true }
+opentelemetry-exporter-richconsole = { workspace = true }
+opentelemetry-instrumentation = { workspace = true }
+opentelemetry-instrumentation-aio-pika = { workspace = true }
```
aabmass, line 80: Followup up on SIG discussion, are you planning to update this PR to generate the list from a template? 

Also looks like it may not be necessary in the future https://github.com/astral-sh/uv/issues/7143

pyproject.toml
```diff
@@ -1,11 +1,149 @@
+# Used for local development, not to be published.
```
aabmass, line 1: Should we add this? https://docs.astral.sh/uv/guides/publish/#preparing-your-project-for-packaging:~:text=If%20you%20have,be%20accidentally%20published.

pyproject.toml
Kludex: Can you point me where I'm supposed to do that?

Kludex: Also, is it blocking? 👀

pyproject.toml
Kludex: You mean the classifier? It's not needed because it's not a package.

uv.lock
```diff
@@ -0,0 +1,4316 @@
+version = 1
```
xrmx, line 1: Question: given that if you are gonna test an instrumentation you need to install some libraries to instrument shouldn't we just ignore the uv.lock file since it will go out of sync with the one we have here?

pyproject.toml
aabmass: @lzchen is this blocking? this is what you mentioned in the SIG call

emdneto: I think it can be an improvement not to get the list of packages out-of-date. Maybe we can have the `uv.toml` file generated from a jinja2 template. Was that the original idea?

lzchen: It's not blocking, maybe just a suggestion for a future PR so we don't have to manage so many files when creating new instrumentations.

aabmass: Tracking in https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3189

uv.lock
aabmass, line 1: I think you may need to regenerate the lockfile, it fails for me unless i regenerate with

```
      AttributeError: module 'falcon' has no attribute '__version__'

      hint: This usually indicates a problem with the package or the build environment.
  help: `falcon` (v3.1.1) was included because `opentelemetry-python-contrib` (v0.0.0) depends on `opentelemetry-instrumentation-falcon[instruments]`
        (v0.51b0.dev0) which depends on `falcon`
```

aabmass (approved):

emdneto (commented):
@Kludex could you please write a guide on how people can use this setup? Maybe a small section in CONTRIBUTING.md is enough.

emdneto (approved):

codefromthecrypt (approved):

uv.lock
aabmass: `uv sync`

Kludex: That updates the lock file.

I'll write some notes on how to use it.

Kludex: You'd need to do `uv sync --frozen`, which doesn't update the lockfile.

xrmx:
Looks like the uv.lock is confusing ruff? Also curious aws xray propagator failing.

Kludex:
> Looks like the uv.lock is confusing ruff? Also curious aws xray propagator test failures.

No...

```
[testenv:ruff]
basepython: python3
deps =
  -c {toxinidir}/dev-requirements.txt
  pre-commit
commands =
  pre-commit run --color=always --all-files {posargs}
```

Calling the test ruff but running pre-commit doesn't help :eyes:

Kludex:
> Also curious aws xray propagator test failures.

It seems it's looking to the `uv.lock`.

Kludex:
> > Also curious aws xray propagator test failures.
> 
> It seems it's looking to the `uv.lock`.

@emdneto we talked about this at some point. Do you remember why?

pyproject.toml
```diff
@@ -1,11 +1,149 @@
...
+opentelemetry-propagator-ot-trace = { workspace = true }
+opentelemetry-util-http = { workspace = true }
+
+# https://docs.astral.sh/uv/reference/settings/#workspace
+[tool.uv.workspace]
+members = [
+  "instrumentation/*",
+  "exporter/*",
+  "opentelemetry-instrumentation",
+  "propagator/*",
```
emdneto, line 138: @Kludex According to [documentation](https://docs.astral.sh/uv/concepts/projects/workspaces/#workspace-sources): 
>Within a workspace, dependencies on workspace members are facilitated via [tool.uv.sources](https://docs.astral.sh/uv/concepts/projects/dependencies/), so since we have tests pointing to pinned versions of API (as example 1.16 for awsxray propagator) uv is trying to install it using the source at Git instead of PyPI

emdneto: Same will happen for all packages that are tested using pinned released versions of api/sdk in test requirements (example: gen-ai stuff)

uv.lock
emdneto: uv sync --frozen is failing for me too

pyproject.toml
emdneto: Instead of workspaces maybe we can have this for sources:
```
opentelemetry-propagator-aws-xray = { path = "./propagator/opentelemetry-propagator-aws-xray", editable = true }
opentelemetry-propagator-ot-trace = { path = "./propagator/opentelemetry-propagator-ot-trace", editable = true  }
```

Kludex:
I dropped xray from the list of packages. It seems the problem is that on the tests, we try to install a package that is listed in the pyproject, but on the pyproject it references a commit, and in the tox it references a version.

I asked on their discord to see if I can do something about it.

---

But the test suite has their own virtual environments, so it's all good.

emdneto:
> I dropped xray from the list of packages. It seems the problem is that on the tests, we try to install a package that is listed in the pyproject, but on the pyproject it references a commit, and in the tox it references a version.
> 
> I asked on their discord to see if I can do something about it.
> 
> But the test suite has their own virtual environments, so it's all good.

Just a note: The same will happen for all packages that are tested using pinned released versions of api/sdk in test requirements e.g., gen-ai stuff

pyproject.toml
emdneto: All good, just commented below as a note.

uv.lock
aabmass: I think this is resolved

aabmass:
🥳

xrmx:
@Kludex Could you please see if #3282 helps instead of dropping xray?

PR #3128, opened 2024-12-20
Improve type hints in `opentelemetry.instrumentation.utils`

Kludex:


opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -44,9 +44,9 @@
 
 
 def extract_attributes_from_object(
-    obj: any, attributes: Sequence[str], existing: Dict[str, str] = None
+    obj: Any, attributes: Sequence[str], existing: Dict[str, str] | None = None
```
Kludex, line 49: This is the most important change. `any` cannot be used as `Any`.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py
```diff
@@ -81,7 +83,7 @@ def http_status_to_status_code(
     return StatusCode.ERROR
 
 
-def unwrap(obj: Union[object, str], attr: str):
+def unwrap(obj: object, attr: str):
```
Kludex, line 86: There's a type of redundancy here, since `str` is an `object`.

But on the first line of this function, the type checker is able to infer if it's a `str` or not.

xrmx: The union has been added for humans more than for type checkers

xrmx (approved):

emdneto (approved):

lzchen (approved):

PR #3130, opened 2024-12-20
Support PEP 561 to `opentelemetry-instrumentation-urllib3`

Kludex:


instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
```diff
@@ -151,32 +155,16 @@ class RequestInfo:
...
-            urllib3.connectionpool.HTTPConnectionPool,
-            RequestInfo,
-        ],
-        None,
-    ]
+    headers: Mapping[str, str] | None
+    body: bytes | typing.IO[typing.Any] | typing.Iterable[bytes] | str | None
+
+
+_UrlFilterT = Optional[Callable[[str], str]]
```
xrmx, line 162: Is the Optional kept on purpose?

Kludex: Yep. You can't do the `|` at runtime on Python 3.8 and 3.9.

I can do the following instead:

```suggestion
_UrlFilterT: TypeAlias = 'Callable[[str], str] | None`
```

Both are equivalent. But you need to say it's a `TypeAlias`, otherwise it's interpreted as string.

instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
```diff
@@ -83,15 +83,19 @@ def response_hook(
...
+
 import collections.abc
 import io
 import typing
 from dataclasses import dataclass
 from timeit import default_timer
-from typing import Collection
+from typing import Any, Callable, Collection, Mapping, Optional
 
 import urllib3.connectionpool
```
xrmx, line 95: Maybe update callers of this while we are it? It looks like we are only using `urllib3.connectionpool.HTTPConnectionPool`

instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
xrmx: Thanks for the explanation

instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
Kludex: You mean the `wrap` calls? I didn't want to do it because it may have some runtime unexpected behavior that I'm not aware. Should I change it as well?

emdneto (approved):

instrumentation/opentelemetry-instrumentation-urllib3/src/opentelemetry/instrumentation/urllib3/__init__.py
aabmass: @xrmx any follow up here?

PR #3131, opened 2024-12-20
Support PEP 561 to `opentelemetry-instrumentation-urllib`

Kludex:


instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/version.py
```diff
@@ -13,5 +13,3 @@
 # limitations under the License.
 
 __version__ = "0.51b0.dev"
-
-_instruments = tuple()
```
xrmx, line 17: What's wrong with this?

Kludex: Oh, it's not used. There's a `_instruments` on the `package.py`. This is the `version.py`.

instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/package.py
```diff
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
 
-_instruments = tuple()
+_instruments: tuple[str, ...] = tuple()
```
Kludex, line 17: See here.

lzchen (approved):

instrumentation/opentelemetry-instrumentation-urllib/src/opentelemetry/instrumentation/urllib/version.py
xrmx: :rofl: , good catch!

xrmx (approved):

PR #3134, opened 2024-12-20
add official support to Python 3.13

emdneto:
# Description

py313 issues
- [X] sqlalchemy -- https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3139
- [x] grpc -- https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3140
- [x] asyncpg -- https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3146
- [x] Celery -- https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3145
- [x] psycopg -- https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3147
- [x] psycopg2 -- https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3147
- [x] aiopg -- https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3147
- [x] falcon - https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3086
- [x] cassandra -- https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3152
- [x] prometheus-remote-write -- https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3157  
- [X] httpx -- fixed in the PR itself (tox.ini)
- [x] django -- fixed in the PR itself (tox.ini)

The only instrumentation we are not running py313 is pyramid. Will open a track issue for that:
- [x] pyramid - Disabled py313 for now. See https://github.com/Pylons/pyramid/pull/3767

Closes https://github.com/open-telemetry/opentelemetry-python/issues/4211 
Closes https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2724

PR #3140, opened 2024-12-21
grpc: bump grpcio version in test-requirements

emdneto:
# Description

# Description
Address CI errors from https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3134 to support python3.13


tox.ini
```diff
@@ -258,8 +258,8 @@ envlist =
     ; opentelemetry-instrumentation-grpc
     ; The numbers at the end of the environment names
     ; below mean these dependencies are being used:
-    ; 0: grpcio==1.62.0
-    ; 1: grpcio==1.63.0
```
xrmx, line 262: Maybe keep 1.62.0 and test it with older versions instead? AFAIR we are testing it because an api changed in 1.63.0

emdneto: Good point. Let me update it

emdneto: -0 and -1 works fines for py312. But for py313 only works using grpc >= 1.66.2.

xrmx (approved):

PR #3145, opened 2024-12-23
celery: remove deprecated import of ContextManager from typing

emdneto:
# Description

This PR fixes the errors from celery CI in #3134: `No name 'ContextManager' in module 'typing'`


instrumentation/opentelemetry-instrumentation-celery/test-requirements-1.txt
```diff
@@ -1,6 +1,6 @@
 amqp==5.2.0
 asgiref==3.8.1
-billiard==4.2.0
+billiard==4.2.1
```
Kludex, line 3: Does this change something?

instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/utils.py
```diff
@@ -12,15 +12,20 @@
...
+from typing import TYPE_CHECKING, Optional, Tuple
 
 from celery import registry  # pylint: disable=no-name-in-module
 from celery.app.task import Task
 
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.trace import Span
 
+if TYPE_CHECKING:
+    from contextlib import AbstractContextManager
```
Kludex, line 27: It will work outside the `TYPE_CHECKING` anyway here, it was introduced on Python 3.6. But if you want to be explicit what is only for typing, it's okay.

Kludex (approved):

instrumentation/opentelemetry-instrumentation-celery/test-requirements-1.txt
emdneto: Nope. They just released some fixes to work on py313 like: `Adapt for logging lock internal changes in Python3.13`

xrmx (approved):

instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/utils.py
emdneto: I want to be explicit that is only for typing. Thanks!!

PR #3154, opened 2024-12-27
docs: add missing import to aws-lambda

Kludex:


instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -55,6 +55,7 @@ def lambda_handler(event, context):
 .. code:: python
 
     from opentelemetry.instrumentation.aws_lambda import AwsLambdaInstrumentor
+    from opentelemetry.propagate import get_global_textmap
```
emdneto, line 58: I think the `custom_event_context_extractor` must be on the `Usage` part of the documentation. Wdyt?

emdneto (approved):

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
Kludex: I just got annoyed that the import was not there. I have no opinions regarding the sections.

xrmx (approved):

lzchen (approved):

PR #3155, opened 2024-12-27
enforce future annotations

Kludex:
- **Enforce `from __future__ import annotations`**
- **Enforce `from __future__ import annotations`**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,14 +12,16 @@ output-format = "concise"
 # https://docs.astral.sh/ruff/linter/#rule-selection
 # pylint: https://github.com/astral-sh/ruff/issues/970
 select = [
-  "I",   # isort
-  "F",   # pyflakes
-  "E",   # pycodestyle errors
-  "W",   # pycodestyle warnings
-  "PLC", # pylint convention
-  "PLE", # pylint error
-  "Q",   # flake8-quotes
-  "A",   # flake8-builtins
+  "I",   # https://docs.astral.sh/ruff/rules/#isort-i
+  "F",   # https://docs.astral.sh/ruff/rules/#pyflakes-f
+  "E",   # https://docs.astral.sh/ruff/rules/#error-e
+  "W",   # https://docs.astral.sh/ruff/rules/#warning-w
+  "PLC", # https://docs.astral.sh/ruff/rules/#convention-c
+  "PLE", # https://docs.astral.sh/ruff/rules/#error-e_1
+  "Q",   # https://docs.astral.sh/ruff/rules/#flake8-quotes-q
+  "A",   # https://docs.astral.sh/ruff/rules/#flake8-builtins-a
+  "UP",  # https://docs.astral.sh/ruff/rules/#pyupgrade-up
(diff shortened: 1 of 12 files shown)


pyproject.toml
```diff
@@ -12,14 +12,16 @@ output-format = "concise"
...
+  "I",   # https://docs.astral.sh/ruff/rules/#isort-i
+  "F",   # https://docs.astral.sh/ruff/rules/#pyflakes-f
+  "E",   # https://docs.astral.sh/ruff/rules/#error-e
+  "W",   # https://docs.astral.sh/ruff/rules/#warning-w
+  "PLC", # https://docs.astral.sh/ruff/rules/#convention-c
+  "PLE", # https://docs.astral.sh/ruff/rules/#error-e_1
+  "Q",   # https://docs.astral.sh/ruff/rules/#flake8-quotes-q
+  "A",   # https://docs.astral.sh/ruff/rules/#flake8-builtins-a
+  "UP",  # https://docs.astral.sh/ruff/rules/#pyupgrade-up
+  "FA"   # https://docs.astral.sh/ruff/rules/#flake8-future-annotations-fa
```
Kludex, line 24: Those two are the additions from this PR.

instrumentation/opentelemetry-instrumentation-aiopg/src/opentelemetry/instrumentation/aiopg/aiopg_integration.py
```diff
@@ -34,8 +36,8 @@ class AiopgIntegration(DatabaseApiIntegration):
     async def wrapped_connection(
         self,
         connect_method: typing.Callable[..., typing.Any],
-        args: typing.Tuple[typing.Any, typing.Any],
-        kwargs: typing.Dict[typing.Any, typing.Any],
+        args: tuple[typing.Any, typing.Any],
```
Viicos, line 39: Maybe next step would be to standardize typing imports (e.g. `import typing; var: typing.Any` vs `from typing import Any; var: Any`), and also how `typing_extensions` should be used. I usually see two patterns:
- Unconditionally import from `typing_extensions` if it is introduced in a newer Python version. We need to make sure these imports are updated whenever a Python version is dropped.
- Define a typing compat module, where new typing constructs are conditionally imported from `typing` or `typing_extensions`. Easier to update when a Python version is dropped.

Viicos (approved):
Assuming otel does not make use of any types at runtime, this should be ok

Kludex:
If people give me their thumbs up, I'm happy to rebase this.

codefromthecrypt (approved):
I love this

Kludex:
If any changelog is needed, please let me know. :)

emdneto (commented):
Maybe we can start using [.git-blame-ignore-revs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view) for ruff changes that touches several files

emdneto (approved):

Kludex:
Can I get someone to merge this? Just because it's something I would really need to keep rebasing otherwise, because of the number of the files it affects.

PR #3161, opened 2025-01-01
Basic instrumentation for Bedrock runtime Converse api

xrmx:
# Description

This is a basic implementation extending the botocore instrumentation with a bedrock-runtime specific extension. This follows genai semconv but is not implementing metrics or events yet.

This is still missing some more tests, ~some docs~, support for tools, events, metrics and streaming but can be discussed since this would be a genai insrumentation outside instrumentation-genai dir and looking very different than the others.

This is a run of the shipped example traced:

```
dotenv run -- opentelemetry-instrument --traces_exporter=console --logs_exporter=none --metrics_exporter=none python converse.py

This is a short poem on OpenTelemetry:

OpenTelemetry, a tool for tracing,
A tool for tracing and monitoring,
It helps to identify and fix bugs,
And make sure the system is secure.
{
    "name": "chat amazon.titan-text-lite-v1",
    "context": {
        "trace_id": "0x801260c37f04079cb7be221fd66cee25",
        "span_id": "0x2805303fc6f3bb1e",
        "trace_state": "[]"
    },
    "kind": "SpanKind.CLIENT",
    "parent_id": null,
    "start_time": "2025-01-15T14:26:05.379154Z",
    "end_time": "2025-01-15T14:26:07.062142Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "rpc.system": "aws-api",
        "rpc.service": "Bedrock Runtime",
        "rpc.method": "Converse",
        "aws.region": "eu-central-1",
        "gen_ai.system": "aws.bedrock",
        "gen_ai.request.model": "amazon.titan-text-lite-v1",
        "gen_ai.operation.name": "chat",
        "aws.request_id": "6f43aecb-ba21-46b5-be38-65fa656d70bb",
        "retry_attempts": 0,
        "http.status_code": 200,
        "gen_ai.usage.input_tokens": 12,
        "gen_ai.usage.output_tokens": 48,
        "gen_ai.response.finish_reasons": [
            "end_turn"
        ]
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "telemetry.sdk.version": "1.29.0",
            "service.name": "opentelemetry-python-bedrock",
            "telemetry.auto.version": "0.50b0"
        },
        "schema_url": ""
    }
}
```

### Random notes

- AWS exposes all the APIs via the boto3 library, botocore is a dependency of boto3 that contains the doc of the services from where the code is then generated at runtime
- Even if there are many different services, even if you write different instrumentations everyone will wrap the same botocore function
- botocore instrumentation has the notion of extensions that are a way to register some callback for a specific AWS service that would be then called by the common code
- Given the point above so just one span would be created using the same common tracer
- For some reason the botocore instrumentation suppresses http instrumentation so we are not getting http spans
- boto instrumentation is for the official client before the introduction of the current boto3; boto3sqs instrumentation is something specific to boto3 sqs client

## Type of change

# How Has This Been Tested?

- [x] tox -e py310-test-instrumentation-botocore

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-botocore/tests/cassettes/test_converse_with_content.yaml
```diff
@@ -0,0 +1,93 @@
...
+          "stopSequences": [
+            "|"
+          ]
+        }
+      }
+    headers:
+      Content-Length:
+      - '170'
+      Content-Type:
+      - !!binary |
```
codefromthecrypt, line 28: hmm weird... is this due to http/2 or something?

codefromthecrypt (approved):
Thanks for getting this started, I'm marking approved despite you mentioning that there are some work to backfill here. Mainly, I approve of the direction, and here are my answers to comments you made in the PR desc.

### On genai directory or here in boto: agree to do it in boto

Even though there is a "genai" directory, I agree this makes most sense handled as an extension with the rest of the boto instrumentation, especially to avoid problems as genai directory is released independently of the rest of contrib.

### On `invoke_model` or `converse`: do both

It could be best to support both from the start as it is easier to migrate folks that way.

* openinference instruments both `invoke_model` and `converse`
* openlit and langtrace instruments `converse`
* openllmetry instruments `invoke_model`

Basically, we don't want resistance for folks moving to the new canonical instrumentation unless `invoke_model` is a dead api.

karthikscale3 (approved):
Thanks for starting this

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/__init__.py
```diff
@@ -32,6 +32,7 @@ def loader():
 
 
 _KNOWN_EXTENSIONS = {
+    "bedrock-runtime": _lazy_load(".bedrock", "_BedrockRuntimeExtension"),
```
lmolkova, line 35: something brought up during the GenAI sig call on 1/9:

Do we have different tracers/meters/loggers instrumentation scope names for boto?

We should, this allows users to:
- differentiate what telemetry is about
- once `Tracer|Meter|LoggerConfig` and declarative config lands in python, users would be able to disable specific scopes

If it's not the case and requires some changes for everything in boto, it's probably best to do it as a follow up.

xrmx: At the moment there is one common tracer, meter, whatever per instrumentation. To make things clear the concept of "extensions" is not from boto3 or botocore but is specific to this instrumentation. It's what the original authors (I guess) created to handle the case where we have a *single* patched function that creates span for a myriad of different APIs.

xrmx: Something like this

```
diff --git a/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py b/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
index 0481b248..97e867eb 100644
--- a/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py
@@ -86,7 +86,7 @@ from botocore.endpoint import Endpoint
 from botocore.exceptions import ClientError
 from wrapt import wrap_function_wrapper
 
-from opentelemetry.instrumentation.botocore.extensions import _find_extension
+from opentelemetry.instrumentation.botocore.extensions import _find_extension, _list_extensions
 from opentelemetry.instrumentation.botocore.extensions.types import (
     _AwsSdkCallContext,
 )
@@ -123,12 +123,9 @@ class BotocoreInstrumentor(BaseInstrumentor):
 
         # pylint: disable=attribute-defined-outside-init
-        self._tracer = get_tracer(
-            __name__,
-            __version__,
-            kwargs.get("tracer_provider"),
-            schema_url="https://opentelemetry.io/schemas/1.11.0",
-        )
+
+        # tracers are lazy initialized per-extension in _get_tracer
+        self._tracers = {}
 
         self.request_hook = kwargs.get("request_hook")
         self.response_hook = kwargs.get("response_hook")
@@ -149,6 +146,24 @@ class BotocoreInstrumentor(BaseInstrumentor):
             self._patched_endpoint_prepare_request,
         )
 
+    @property
+    def _get_tracer(self, extension):
+        # this is a multiplexer in order to have a tracer per extension
+        extension = _get_extension_name(call_context)
+        if extension is None:
+            extension = __name__
+        tracer = self._tracers.get(extension)
+        if tracer:
+            return tracer
+
+        self._tracers[extension] = get_tracer(
+            __name__,
+            __version__,
+            kwargs.get("tracer_provider"),
+            schema_url="https://opentelemetry.io/schemas/1.11.0",
+        )
+        return self._tracers[extension]
+
         unwrap(BaseClient, "_make_api_call")
         unwrap(Endpoint, "prepare_request")
@@ -189,7 +204,8 @@ class BotocoreInstrumentor(BaseInstrumentor):
 
         _safe_invoke(extension.extract_attributes, attributes)
 
-        with self._tracer.start_as_current_span(
+        tracer = self._get_tracer(extension)
+        with tracer.start_as_current_span(
             call_context.span_name,
             kind=call_context.span_kind,
             attributes=attributes,
diff --git a/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/__init__.py b/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/__init__.py
index c4624aba..7dc9b63e 100644
--- a/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/__init__.py
@@ -14,6 +14,7 @@
 
 import importlib
 import logging
+from typing import Sequence
 
 from opentelemetry.instrumentation.botocore.extensions.types import (
     _AwsSdkCallContext,
@@ -40,9 +41,13 @@ _KNOWN_EXTENSIONS = {
 }
 
 
+def _get_extension_name(call_context: __AwsSdkCallContext) -> str:
+    return__KNOWN_EXTENSIONS.get(call_context.service)
+
+
 def _find_extension(call_context: _AwsSdkCallContext) -> _AwsSdkExtension:
     try:
-        loader = _KNOWN_EXTENSIONS.get(call_context.service)
+        loader = __get_extension_name(call_context)
         if loader is None:
             return _AwsSdkExtension(call_context)
```

instrumentation/opentelemetry-instrumentation-botocore/tests/cassettes/test_converse_with_content.yaml
xrmx: I don't think so, botocore should use urllib3 as http library and I've recorded with an old-ish `urllib3==1.26.19`. urllib3 does not have http/2 support yet

Issue #3169, opened 2025-01-06
Make span attributes json parseable for celery


shivanshuraj1333:
### Describe your environment

OS: (e.g, Ubuntu)
Python version: (e.g., Python 3.8.10)
Package version: (e.g., 0.46.0)

### What happened?

Currently, the "delivery_info" attribute in celery is a string and not a map (or even a json).
The attribute is transformed into a string, which makes it hard to parse, and one has to use regex instead.

### Steps to Reproduce

Run the celery instrumentation application from https://github.com/shivanshuraj1333/celery-opentelemetry-instrumentation/

### Expected Result

attribute name: celetry.delivery_info

option 1: return a python dict
{
    'exchange': '',
    'routing_key': 'queue2',
    'priority': 0,
    'redelivered': False
}

option 2: return a valid json
{
  "exchange": "",
  "routing_key": "queue2",
  "priority": 0,
  "redelivered": false
}

### Actual Result

attribute name: celetry.delivery_info
value: {'exchange': '', 'routing_key': 'queue2', 'priority': 0, 'redelivered': False}

### Would you like to implement a fix?

None

PR #3170, opened 2025-01-06
Use python dict instead of string as attribute value to enable parsing

shivanshuraj1333:
Fixes: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3169

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/utils.py
+++ b/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/utils.py
@@ -95,8 +95,6 @@ def set_attributes_from_context(span, context):
                     SpanAttributes.MESSAGING_DESTINATION, routing_key
                 )
 
-            value = str(value)
-
         elif key == "id":
             attribute_name = SpanAttributes.MESSAGING_MESSAGE_ID
 


instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/utils.py
```diff
@@ -95,8 +95,6 @@ def set_attributes_from_context(span, context):
                     SpanAttributes.MESSAGING_DESTINATION, routing_key
                 )
 
-            value = str(value)
```
xrmx, line 98: I think this does match with the value an attributes can have https://opentelemetry.io/docs/concepts/signals/traces/#attributes

shivanshuraj1333: Yes, I opened [this](https://github.com/open-telemetry/opentelemetry-python/issues/4380) and realized that it is intentional.

Right now, since we do `value = str(value)` the map becomes tricky to parse (filtering based on key).

How about if I flatten the map and then stringily (or convert to corresponding datatype) the values?
```python
celery.delivery_info ⇒ `{'exchange': '', 'routing_key': 'distilled-guidance-dpo-25-10k-v1', 'priority': 0, 'redelivered': False}`
```
would become
```python
celery.delivery_info.exchange: "routing_key"
celery.delivery_info.routing_key: "distilled-guidance-dpo-25-10k-v1"
celery.delivery_info.priority: 0
celery.delivery_info.redelivered: false
```

xrmx: Flattening the map would indeed be better I think than going out of spec :)

shivanshuraj1333: I have updated the changes to add span attributes with flattened map. PTAL, thanks!

CHANGELOG.md
```diff
@@ -13,7 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `opentelemetry-instrumentation-botocore` Add support for GenAI user events and lazy initialize tracer
```
xrmx, line 16: I think you mis-merged things here

shivanshuraj1333: my bad, fixed it

PR #3184, opened 2025-01-13
Run pyright on Vertex AI instrumentation

aabmass:
# Description

Follow up to https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3116, adds typechecking to the vertexai instrumentation. This instrumentation is WIP and only has boilerplate so far, so this is very simple.

The interesting part is how to make dependencies needed for typechecking available to pyright.

## Type of change

# How Has This Been Tested?

- [ ] `tox -e typecheck`

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/__init__.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/__init__.py
@@ -39,7 +39,7 @@ API
 ---
 """
 
-from typing import Collection
+from typing import Any, Collection
 
 from opentelemetry._events import get_event_logger
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
@@ -52,7 +52,7 @@ class VertexAIInstrumentor(BaseInstrumentor):
     def instrumentation_dependencies(self) -> Collection[str]:
         return _instruments
 
-    def _instrument(self, **kwargs):
+    def _instrument(self, **kwargs: Any):
         """Enable VertexAI instrumentation."""
         tracer_provider = kwargs.get("tracer_provider")
         _tracer = get_tracer(
@@ -70,5 +70,5 @@ class VertexAIInstrumentor(BaseInstrumentor):
         )
         # TODO: implemented in later PR
 
-    def _uninstrument(self, **kwargs) -> None:
+    def _uninstrument(self, **kwargs: Any) -> None:
         """TODO: implemented in later PR"""
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,10 +48,13 @@ pythonVersion = "3.8"
 reportPrivateUsage = false  # Ignore private attributes added by instrumentation packages.
 # Add progressively instrumentation packages here.
 include = [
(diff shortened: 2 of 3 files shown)


typecheck-requirements.txt
```diff
@@ -0,0 +1,5 @@
+# Contains libraries or stub requirements which are used in the typechecked code. Pyright needs
+# these available to check the code.
+
+# Used by instrumentation-genai/opentelemetry-instrumentation-vertexai
```
aabmass, line 4: I guess we'll just keep adding to this file more packages opt in to typechecking. I couldn't come up with a more automated way to achieve this.

Maybe we can use the `uv.lock` file from https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3124?

aabmass (commented):
@Kludex would appreciate your review

typecheck-requirements.txt
Kludex: What do people use the `dev-requirements.txt` for? 🤔 Can we add this there?

Yeah, we should use `[dependency-groups]` on the pyproject to simplify the installation of dependencies. When that PR is merged, we can simplify the requirements files.

aabmass: > What do people use the `dev-requirements.txt` for? 🤔 Can we add this there?

It's used in the tox file a few places to install tooling e.g. in `tox -e ruff`. This bloats things a bit since it pulls in all the transitive deps, especially once we have more libraries being typechecked:

```console
$ du -sh .tox/ruff/
12M	.tox/ruff/
$ du -sh .tox/typecheck/
164M	.tox/typecheck/
```

Kludex: That's heavy... 👀

Kludex (approved):

typecheck-requirements.txt
Kludex: I don't like adding new `requirements.txt`, but all good. 👍

Kludex: Actually, can you do this instead on the tox.ini typecheck:
```
  {toxinidir}/instrumentation-genai/opentelemetry-instrumentation-vertexai[instruments]
```
I'll organize the dependencies on the main pyproject when the uv PR is merged.

aabmass: This would also install the package itself. Does Pyright behave correctly when the code your are type checking is also installed into the virtualenv?

aabmass: I updated to do this but still curious your thoughts. I know that Mypy can behave differently if the code being checked is installed in virtualenv vs. specified as arguments vs. added to PYTHONPATH

Kludex: That should be seen as a package anyway. If other package internally references (imports) that package, it should consider it as it is locally.

We want local type checked folders to see other folders in the repository as packages.

Kludex (approved):

typecheck-requirements.txt
aabmass: OK I'm not really sure how pyright is supposed to work with monorepos. I saw some evidence in the docs that it understands `src/` layouts.

But this seems to be working for now so we can revisit if needed 👍

emdneto (approved):

PR #2667, opened 2024-07-05
testing uv

emdneto:

PR #3185, opened 2025-01-14
Introducing tox-uv 

emdneto:
# Description
I was working on #2667 and realized our tox setup was not prepared to adopt tox-uv, so after some rework I'm opening this PR with less intrusive changes to tox.ini file. So basically, this is a new version of #2667

uv makes pip install faster. It works like this: `uv pip install` and we can have tests running faster locally and in CI. The tool is from the creators of ruff. 

The main improvement I see here is the Dependency caching and fast resolution of packages during pip install. We can benefit from this for local development to avoid slow tests, mainly the git clones in contrib tests.

This PR also introduces the usage of [tox-uv](https://github.com/tox-dev/tox-uv) which is a tox plugin that supports venv runner with uv.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,6 +59,12 @@ To install `tox`, run:
 pip install tox
 ```
 
+You can also run tox with `uv` support. By default [tox.ini](./tox.ini) will automatically create a provisioned tox environment with `tox-uv`, but you can install it at host level:
+
+```sh
+pip install tox-uv
+```
+
 You can run `tox` with the following arguments:
 
 * `tox` to run all existing tox commands, including unit tests for all packages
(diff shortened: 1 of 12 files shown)


tox.ini
```diff
@@ -1,4 +1,6 @@
 [tox]
+requires =
```
emdneto, line 2: Basically, this means: If any of these dependencies are not satisfied will automatically create a provisioned tox environment that does not have this issue, and run the tox command within that environment.

xrmx (approved):

Kludex (approved):
For curiosity, did you check how much performance benefit there is here?

xrmx:
> For curiosity, did you check how much performance benefit there is here?

Completely unscientific but looks like on CI two test workflows out of three are a few minutes faster

Kludex:
Unscientific is good. I was lazy to look. 😐👍

Few minutes is great.

CONTRIBUTING.md
```diff
@@ -59,6 +59,12 @@ To install `tox`, run:
 pip install tox
 ```
 
+You can also run tox with `uv` support. By default [tox.ini](./tox.ini) will automatically create a provisioned tox environment with `tox-uv`, but you can install it at host level:
+
+```sh
+pip install tox-uv
```
codefromthecrypt, line 65: mind adding an example of using `uv` to run tests in a single package, or is this just plumbing under tox, so no uv based entrypoints..

xrmx: This is plumbing for tox

codefromthecrypt: thx

lzchen (approved):

PR #3190, opened 2025-01-15
fix (ci): Allow opentelemetry-python forks to use CI

jsoref:
# Description

Adds `CORE_REPO` and `CONTRIB_REPO` inputs and corresponding environment variables to allow people to use CI in forks of https://github.com/open-telemetry/opentelemetry-python

## Type of change

# How Has This Been Tested?

- [x] [Test A](https://github.com/check-spelling-sandbox/opentelemetry-python/pull/1)
       image

There is a corresponding change that should be made but it it can be done after this is merged.

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

### DIFF
--- a/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
+++ b/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
@@ -10,7 +10,9 @@ on:
   pull_request:
 
 env:
+  CORE_REPO: open-telemetry/opentelemetry-python
   CORE_REPO_SHA: main
+  CONTRIB_REPO: open-telemetry/opentelemetry-python-contrib
   CONTRIB_REPO_SHA: main
   PIP_EXISTS_ACTION: w
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -242,9 +242,12 @@ tox -e py312-test-instrumentation-aiopg
 
 ### Testing against a different Core repo branch/commit
 
-Some of the tox targets install packages from the [OpenTelemetry Python Core Repository](https://github.com/open-telemetry/opentelemetry-python) via pip. The version of the packages installed defaults to the main branch in that repository when tox is run locally. It is possible to install packages tagged with a specific git commit hash by setting an environment variable before running tox as per the following example:
+Some of the tox targets install packages from the [OpenTelemetry Python Core Repository](https://github.com/open-telemetry/opentelemetry-python) via pip.
+The version of the packages installed defaults to the main branch in that repository when tox is run locally.
+It is possible to install packages tagged with a specific git commit hash (optionally in a specific fork) by setting these environment variables before running tox as per the following example:
 
 ```sh
+CORE_REPO=check-spelling-sandbox/opentelemetry-python
 CORE_REPO_SHA=c49ad57bfe35cfc69bfa863d74058ca9bec55fc3 tox
 ```
 
--- a/tox.ini
+++ b/tox.ini
@@ -682,10 +682,11 @@ allowlist_externals =
   sh
 
 setenv =
+  ; override CORE_REPO via env variable when testing in forks
+  CORE_REPO=git+https://github.com/{env:CORE_REPO:open-telemetry/opentelemetry-python}.git@{env:CORE_REPO_SHA}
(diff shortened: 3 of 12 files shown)


.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
```diff
@@ -10,7 +10,9 @@ on:
   pull_request:
 
 env:
+  CORE_REPO: open-telemetry/opentelemetry-python
   CORE_REPO_SHA: main
+  CONTRIB_REPO: open-telemetry/opentelemetry-python-contrib
```
aabmass, line 15: Does this need to be`${{ github.repository }}` so users can run tests of contrib forks too? Just going off of symmetry here

jsoref: hmm, it probably needs to be:
```suggestion
  CONTRIB_REPO: ${% raw %}{{ inputs.CONTRIB_REPO || 'open-telemetry/opentelemetry-python-contrib' }}{% endraw %}
```

tox.ini
```diff
@@ -682,10 +682,11 @@ allowlist_externals =
   sh
 
 setenv =
+  ; override CORE_REPO via env variable when testing in forks
+  CORE_REPO=git+https://github.com/{env:CORE_REPO:open-telemetry/opentelemetry-python}.git@{env:CORE_REPO_SHA}
```
emdneto, line 686: I would say to avoid those changes, maybe just use `GITHUB_REPOSITORY` runtime environment variable? Like
```suggestion
  CORE_REPO=git+https://github.com/{env:GITHUB_REPOSITORY:open-telemetry/opentelemetry-python}.git@{env:CORE_REPO_SHA}
```
At CI this will point to the current owner and repository name, so when running in your fork, it should match your check-spelling-sandbox/opentelemetry-python

.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
emdneto: @aabmass I think this is not needed, because checkout action already does the trick for us, including forks.

CONTRIBUTING.md
```diff
@@ -242,9 +242,12 @@ tox -e py312-test-instrumentation-aiopg
 
 ### Testing against a different Core repo branch/commit
 
-Some of the tox targets install packages from the [OpenTelemetry Python Core Repository](https://github.com/open-telemetry/opentelemetry-python) via pip. The version of the packages installed defaults to the main branch in that repository when tox is run locally. It is possible to install packages tagged with a specific git commit hash by setting an environment variable before running tox as per the following example:
+Some of the tox targets install packages from the [OpenTelemetry Python Core Repository](https://github.com/open-telemetry/opentelemetry-python) via pip.
+The version of the packages installed defaults to the main branch in that repository when tox is run locally.
+It is possible to install packages tagged with a specific git commit hash (optionally in a specific fork) by setting these environment variables before running tox as per the following example:
 
 ```sh
+CORE_REPO=check-spelling-sandbox/opentelemetry-python
```
emdneto, line 250:
```suggestion
GITHUB_REPOSITORY=your-fork/opentelemetry-python tox
```
Based on my previous comment

.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
jsoref: In principle, I believe, this could be needed if someone was trying to test a change to their opentelemetry-python repository that needed changes to their opentelemetry-python-contrib repository at the same time (which is more or less what we're looking at in https://github.com/open-telemetry/opentelemetry-python/pull/4388 where things conceptually didn't work for https://github.com/jsoref/opentelemetry-python/actions/runs/12793563878 because the callee doesn't yet have the required feature).

I'm not saying it's working, but I suspect a thing like it would be needed. I'm also not entirely certain it would always work/would work in this case (I think you can't really programmatically change a called workflow target, so in this specific case it wouldn't, but for a normal case, of code as opposed to workflow dependencies, I think it'd be useful).

Issue #3179, opened 2025-01-09
Remove protobuf restriction on opentelemetry-exporter-prometheus-remote-write


michaelschuett-tomtom:
### What problem do you want to solve?

Currently if I want to use opentelemetry-exporter-prometheus-remote-write i am restricted to a 4 year old version of protobuf that is about to be EOL it also doesn't work with version of protoc which generates protobufs with imports that do not exist in the 4.21 version.

You will see something similar to this.

```
  File "<stdin>", line 1, in <module>
  File "/opt/anaconda3/lib/python3.11/site-packages/internal/client.py", line 8, in <module>
    from internal.feature import Feature, FeatureBuilder
  File "/opt/anaconda3/lib/python3.11/site-packages/internal/feature.py", line 3, in <module>
    from idl.model.feature_pb2 import FeatureProto, FeatureStatus
  File "/opt/anaconda3/lib/python3.11/site-packages/idl/model/feature_pb2.py", line 9, in <module>
    from google.protobuf import runtime_version as _runtime_version
```

https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/exporter/opentelemetry-exporter-prometheus-remote-write

### Describe the solution you'd like

Allow for newer versions of protobuf or possibly some other solution exists that I am not aware of but currently pulling in any python dep that holds back the protobuf version to some very old one is a pain because you need to go and setup older tooling that might not be readily available on your distro.

### Would you like to implement a fix?

I can after we have a clear path but I think I am likely missing some context for why we are pinning to such an old version. Possibly this is a much bigger issue with the entire project needing to upgrade the protobuf version.

michaelschuett-tomtom:
Through testing I have also found that using the newer version of 5.29.2 with this package works just so it seems it is backwards compatible with the code that is already generated with an older version of protobuf.

aabmass:
In fact, we test against protobuf 4.25.3 which is less than a year old https://github.com/open-telemetry/opentelemetry-python-contrib/blob/2480f4e846286df1d0ae0a6a0bc7293bc9c88813/exporter/opentelemetry-exporter-prometheus-remote-write/test-requirements.txt#L12

Regardless, I'm ok if you want to upgrade it to protobuf 5.x since we already did that for OTLP

michaelschuett-tomtom:
[...]

PR #3219, opened 2025-01-29
upgrade opentelemetry-exporter-prometheus-remote-write to use protobuf 5.26

michaelschuett-tomtom:
# Description

This change for my case it to allow us to use this in a project that needs to use protobuf 5.29. I have tested in that project that everything still works as expected so the version limit is only based on the restriction in the produced pypi package not on any code limitation.

I originally was looking at saying the package uses =~5.29 however this conflicts with another random package I had installed `streamlit 1.30.0 requires protobuf<5,>=3.20, but you have protobuf 5.29.3 which is incompatible`. Which made me think it might be nice to let people use all known working versions of protobuf with this package instead of arbitrarily deciding what one is correct. 

Fixes [#3179 ](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3179)

## Type of change

# How Has This Been Tested?

I tested locally by running
```
$ PACKAGE_NAME=opentelemetry-exporter-prometheus-remote-write VERSION=1.0 ./scripts/build_a_package.sh
$ pip install --force-reinstall ./dist/opentelemetry_exporter_prometheus_remote_write-0.51b0.dev0-py3-none-any.whl
```

- [ ] Yes. - Link to PR: 
- [ ] No.

I don't think so?

# Checklist:


emdneto (commented):
Hey: thanks for the PR. I think we should at least add a test using protobuf 5, although tests pass and there is [no guarantee for cross-version runtime](https://protobuf.dev/support/cross-version-runtime-guarantee/). What we did in opentelemetry-proto was to support only one major version, protobuf5. 

@aabmass do you think we should wait for the protobuf Q12025 release for this kind of issue?

aabmass:
> Which made me think it might be nice to let people use all known working versions of protobuf with this package instead of arbitrarily deciding what one is correct.

It's not arbitrary like @emdneto mentioned, protobuf makes it very clear that this is not officially supported. This PR lets the user decide but I fear people will come here when things break and expect us to fix it. I would prefer to do what @emdneto mentioned and just support protobuf 5. It's been out for almost a year now.

Is there a not a newer streamlit version that works with protobuf 5?

michaelschuett-tomtom:
> > Which made me think it might be nice to let people use all known working versions of protobuf with this package instead of arbitrarily deciding what one is correct.
> 
> It's not arbitrary like @emdneto mentioned, protobuf makes it very clear that this is not officially supported. This PR lets the user decide but I fear people will come here when things break and expect us to fix it. I would prefer to do what @emdneto mentioned and just support protobuf 5. It's been out for almost a year now.
> 
> Is there a not a newer streamlit version that works with protobuf 5?

I am completely fine with just having 5.x supported. steamlit isn't a specific issue for me but just something I ran into when testing in a few different environments. I'll update this to be only protobuf 5.x.

michaelschuett-tomtom:
https://protobuf.dev/support/version-support/#python picked 5.26 based on this page.

michaelschuett-tomtom:
@emdneto more than happy to write some tests although I am not sure exactly what kind of test to write here to improve coverage. I ran `tox -e test-exporter-prometheus-remote-write` and it passed outside of one test that looks like it needs a local server spun up `HTTPConnectionPool(host='victoria', port=8428)`.

I do see that we are loading the protobufs in the tests `opentelemetry.exporter.prometheus_remote_write.gen.types_pb2` so any breakage at least around imports changing should be caught.

aabmass:
If we're just doing protobuf 5, I think just the existing test is ok

aabmass (approved):

aabmass (changes requested):
Actually I think we need to regenerate the protobuf code too?

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
```diff
@@ -122,9 +122,9 @@ def test_parse_metric(metric, prom_rw):
         "bool_value": True,
     }
 
-    assert (
-        len(metric.data.data_points) == 1
-    ), "We can only support a single datapoint in tests"
+    assert len(metric.data.data_points) == 1, (
```
michaelschuett-tomtom, line 125: ruff lint did this which was part of the generate-proto-py script.

michaelschuett-tomtom:
> Actually I think we need to regenerate the protobuf code too?

Just regenerated with protoc 29.3

emdneto:
> Actually I think we need to regenerate the protobuf code too?

Right. There's a script in the `opentelemetry-exporter-prometheus-remote-write/proto` directory called `generate-proto-py.sh` @michaelschuett-tomtom did you use that script? If so, could you please create a requirement.txt  or a note in the README file with the pinned version of grpcio-tools you've installed to run the script?

exporter/opentelemetry-exporter-prometheus-remote-write/tests/test_prometheus_remote_write_exporter.py
michaelschuett-tomtom: not sure how this happened. I ran format from the root of the repo and then `tox -e ruff` and it switched back.

emdneto (approved):
LGTM with Changelog entry

michaelschuett-tomtom:
> LGTM with Changelog entry

Was thinking to add this under Breaking changes in the main CHANGELOG.md since upon update in the case someone has pinned protobuf to a specific version it will cause issues... But this is also true for any dependency update so I'm not 100% sure if this should go under another area.

michaelschuett-tomtom:
These small PRs always turn out to have the most foot guns 😄 I was just testing some more locally to make sure I don't break anything for people and was running the tests with different versions of presumably supported protobuf versions as defined in the requirements (~=5.26). Because I generated with 29.3 protobuf will blow up with anything that isn't python protobuf version 5.29.3 or later.

Although we could just say this is the min version we should likely be nice to people since they are all still supported. I'll regenerate using v26 of protoc so that we actually support ~=5.26.

Learning on the go here although i have used protobuf before I haven't had to care much about versions since I controled client and server and generate them at the same time.

emdneto:
> https://protobuf.dev/support/cross-version-runtime-guarantee.

Yeah. In that case, if we want to support starting from 5.26 we must re-generate it with protoc 26. You can also add a second test-requirements.txt file with protobuf 5.29, which should also work. With that we are sure 5.26 and newer runtime versions are covered (which is guaranteed according to [this](https://protobuf.dev/support/cross-version-runtime-guarantee/))

michaelschuett-tomtom:
Done, updated docs and change log.

michaelschuett-tomtom:
Hey @aabmass sorry to ping you. Do you mind taking a look at this one again when you have a chance.

CHANGELOG.md
```diff
@@ -69,6 +69,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking changes
 
+- `opentelemetry-exporter-prometheus-remote-write` updated protobuf required version from 4.21 to 5.26 and regenerated protobufs ([#3219](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3219))
```
emdneto, line 72:
```suggestion
- `opentelemetry-exporter-prometheus-remote-write` updated protobuf required version from 4.21 to 5.26 and regenerated protobufs 
 ([#3219](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3219))
```

aabmass:
> Although we could just say this is the min version we should likely be nice to people since they are all still supported. I'll regenerate using v26 of protoc so that we actually support ~=5.26.
>
> Learning on the go here although i have used protobuf before I haven't had to care much about versions since I controled client and server and generate them at the same time.

Yes it's a total PIA :) thank you for helping out on this!

aabmass (approved):
LGTM thank you for manually testing things

PR #3225, opened 2025-02-03
Provide advisory histogram boundaries when creating OpenAI metrics

### LINKED ISSUE #3235: OpenAI instrumentation: set explicit histogram boundaries when creating metrics
With https://github.com/open-telemetry/opentelemetry-python/pull/4361 (once 1.30 is released, we should be able to simplify user configuration A LOT by setting histogram boundaries according to semconv.

Adding work item to track it.

lmolkova:
Fixes #3235

Update: 
I overhauled this PR and it now leverages advisory boundaries - no need for users to configure anything and most users should have good experience with default boundaries. So no OpenAI-specific example is necessary.


gyliu513:
LGTM

instrumentation-genai/opentelemetry-instrumentation-openai-v2/examples/manual/main.py
```diff
@@ -24,6 +34,56 @@
     BatchSpanProcessor(OTLPSpanExporter())
 )
 
+# configure metrics
```
codefromthecrypt, line 37: As below are choosing boundaries, it could help in the example to explain why we are doing this, or at least what we are changing it, to.

For example, "# configure metrics for requests that are skewed slower/faster than the default bucket boundaries"
I'm not saying this is the right sentence, but we can take advantage of this to teach people why we are overriding the defaults. Otherwise, folks may just copy/paste this and now know why.

codefromthecrypt (approved):

aabmass (approved):

codefromthecrypt (approved):
Thanks, I guess this also applies to the auto-instrumentation, but also might water down how it looks, as we'd need to add code to the zero code ;)

thanks for linking to the issue

instrumentation-genai/opentelemetry-instrumentation-openai-v2/examples/manual/main.py
```diff
@@ -24,6 +34,66 @@
...
+#
+# The bucket boundaries are defined as follows:
+# For `gen_ai.client.token.usage`: [1, 4, 16, 64, 256, 1024, 4096, 16384, 65536, 262144, 1048576, 4194304, 16777216, 67108864]
+# For `gen_ai.client.operation.duration`: [0.01, 0.02, 0.04, 0.08, 0.16, 0.32, 0.64, 1.28, 2.56, 5.12, 10.24, 20.48, 40.96, 81.92]
+#
+# If your application benefits from different bucket boundaries, you can update them as needed.
+views = [
+    View(
+        instrument_name="gen_ai.client.token.usage",
+        aggregation=ExplicitBucketHistogramAggregation(
```
codefromthecrypt, line 54: IIUC what we are doing is avoiding the default boundaries being used both for token usage and also duration.

From [here](https://github.com/open-telemetry/opentelemetry-python/blob/4a271f946a23060bf99f7d6c0aeaad63cd8c764d/opentelemetry-sdk/src/opentelemetry/sdk/metrics/_internal/aggregation.py#L448-L464)
```python
_DEFAULT_EXPLICIT_BUCKET_HISTOGRAM_AGGREGATION_BOUNDARIES: Sequence[float] = (
    0.0,
    5.0,
    10.0,
    25.0,
    50.0,
    75.0,
    100.0,
    250.0,
    500.0,
    750.0,
    1000.0,
    2500.0,
    5000.0,
    7500.0,
    10000.0,
)
```

So, this default might be ok for tokens for some apps, but probably not reasonable for durations which are in seconds (except maybe in reasoning models ;)). Correct me if I'm wrong!

xrmx (approved):

codefromthecrypt (approved):
thank you so much

xrmx (approved):
Thanks!

instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/test_chat_metrics.py
```diff
@@ -77,8 +110,11 @@ def test_chat_completion_metrics(
         None,
     )
     assert duration_metric is not None
-    assert duration_metric.data.data_points[0].sum > 0
-    assert_all_metric_attributes(duration_metric.data.data_points[0])
+
+    duration_metric = duration_metric.data.data_points[0]
+    assert duration_metric.sum > 0
+    assert_all_metric_attributes(duration_metric)
+    assert duration_metric.explicit_bounds == _DURATION_BUCKETS
```
lmolkova, line 117:
```suggestion

    duration_point = duration_metric.data.data_points[0]
    assert duration_point .sum > 0
    assert_all_metric_attributes(duration_point )
    assert duration_point .explicit_bounds == _DURATION_BUCKETS
```

instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/test_chat_metrics.py
```diff
@@ -77,8 +110,11 @@ def test_chat_completion_metrics(
         None,
     )
     assert duration_metric is not None
-    assert duration_metric.data.data_points[0].sum > 0
-    assert_all_metric_attributes(duration_metric.data.data_points[0])
+
+    duration_point = duration_metric.data.data_points[0]
+    assert duration_point .sum > 0
+    assert_all_metric_attributes(duration_point )
+    assert duration_point .explicit_bounds == _DURATION_BUCKETS
```
lmolkova, line 117:
```suggestion
    assert duration_point.sum > 0
    assert_all_metric_attributes(duration_point)
    assert duration_point.explicit_bounds == _DURATION_BUCKETS
```

instrumentation-genai/opentelemetry-instrumentation-openai-v2/examples/manual/requirements.txt
```diff
@@ -2,4 +2,4 @@ openai~=1.57.3
 
 opentelemetry-sdk~=1.29.0
 opentelemetry-exporter-otlp-proto-grpc~=1.29.0
```
emdneto, line 4: bump example to 1.30.0 here as well?

lmolkova: great catch, thanks!

emdneto (approved):

PR #3226, opened 2025-02-03
Record content events regardless of span sampling decision

lmolkova:
Fixes # 3217

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

### DIFF
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/patch.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/patch.py
@@ -56,11 +56,8 @@ def chat_completions_create(
             attributes=span_attributes,
             end_on_exit=False,
         ) as span:
-            if span.is_recording():
-                for message in kwargs.get("messages", []):
-                    event_logger.emit(
-                        message_to_event(message, capture_content)
-                    )
+            for message in kwargs.get("messages", []):
+                event_logger.emit(message_to_event(message, capture_content))
 
             start = default_timer()
             result = None
@@ -76,6 +73,12 @@ def chat_completions_create(
                     _set_response_attributes(
                         span, result, event_logger, capture_content
                     )
+                if getattr(result, "choices", None):
+                    for choice in result.choices:
+                        event_logger.emit(
+                            choice_to_event(choice, capture_content)
+                        )
+
                 span.end()
                 return result
 
@@ -114,11 +117,8 @@ def async_chat_completions_create(
             attributes=span_attributes,
             end_on_exit=False,
         ) as span:
-            if span.is_recording():
-                for message in kwargs.get("messages", []):
-                    event_logger.emit(
-                        message_to_event(message, capture_content)
-                    )
+            for message in kwargs.get("messages", []):
+                event_logger.emit(message_to_event(message, capture_content))
 
             start = default_timer()
             result = None
@@ -134,6 +134,12 @@ def async_chat_completions_create(
                     _set_response_attributes(
                         span, result, event_logger, capture_content
                     )
+                if getattr(result, "choices", None):
+                    for choice in result.choices:
+                        event_logger.emit(
+                            choice_to_event(choice, capture_content)
+                        )
+
                 span.end()
                 return result
 
@@ -229,9 +235,6 @@ def _set_response_attributes(
 
     if getattr(result, "choices", None):
         choices = result.choices
-        for choice in choices:
-            event_logger.emit(choice_to_event(choice, capture_content))
-
         finish_reasons = []
         for choice in choices:
             finish_reasons.append(choice.finish_reason or "error")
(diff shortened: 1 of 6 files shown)


instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/test_async_chat_completions.py
```diff
@@ -633,6 +633,55 @@ async def test_async_chat_completion_multiple_tools_streaming_no_content(
     )
 
 
+@pytest.mark.vcr()
+@pytest.mark.asyncio()
+async def test_async_chat_completion_streaming_sample_out(
```
codefromthecrypt, line 638: note for later (so feel free to resolve), we can re-use cassettes when the server's response is uninteresting vs another code path. This often exists between "async" and not.

I've made a fixture like this to handle it, which makes the diffs a lot shorter

```python
@pytest.fixture
def vcr_cassette_name(request):
    """
    Strips `_async` from the test function name as they use the same data.
    """
    # Get the name of the test function
    test_name = request.node.name

    # Remove '_async' from the test name
    cassette_name = test_name.replace("_async", "")

    return cassette_name
```

codefromthecrypt, line 638: dismiss if "sampled out" is a common phrase

```suggestion
async def test_async_chat_completion_streaming_unsampled(
```

instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/test_chat_completions.py
```diff
@@ -659,6 +659,48 @@ def test_chat_completion_multiple_tools_streaming_no_content(
     )
 
 
+@pytest.mark.vcr()
+def test_chat_completion_with_content_span_sampled_out(
```
codefromthecrypt, line 663: again, close out if "sampled out" is common

```suggestion
def test_chat_completion_with_content_span_unsampled(
```

codefromthecrypt (approved):
I think the owners know after this change we'll get the conversion overhead always, and that can be quite large if there's a large context.


Follow-up notes:

There's little to do about this while we still depend on event_logger. Singe event_logger is deprecated, there's no sense in adding a sampling function to it. Since it could be wrapped, testing if it is noop may have limited benefit.

When things move to a logger base api, it could be possible to optimize this to check if 
* active span is sampled
* a corresponding log scope is enabled
in either case, do the overhead and when neither, don't.

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/patch.py
```diff
@@ -76,6 +73,12 @@ def traced_method(wrapped, instance, args, kwargs):
                     _set_response_attributes(
                         span, result, event_logger, capture_content
                     )
+                if getattr(result, "choices", None):
+                    for choice in result.choices:
```
aabmass, line 77: small nit to cut down on nesting
```suggestion
                for choice in getattr(result, "choices", []):
```

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/patch.py
```diff
@@ -134,6 +134,12 @@ async def traced_method(wrapped, instance, args, kwargs):
                     _set_response_attributes(
                         span, result, event_logger, capture_content
                     )
+                if getattr(result, "choices", None):
```
aabmass, line 137: same nit as above

aabmass (approved):

instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/test_async_chat_completions.py
lmolkova: I like the idea, but  it won't apply to this PR - I test no-streaming+sync and streming+async, so recordings are different.

lmolkova:
> I think the owners know after this change we'll get the conversion overhead always, and that can be quite large if there's a large context.
> 
> Follow-up notes:
> 
> There's little to do about this while we still depend on event_logger. Singe event_logger is deprecated, there's no sense in adding a sampling function to it. Since it could be wrapped, testing if it is noop may have limited benefit.
> 
> When things move to a logger base api, it could be possible to optimize this to check if
> 
> * active span is sampled
> * a corresponding log scope is enabled
> 
> in either case, do the overhead and when neither, don't.

There are a bunch of spec-level changes in discussions that would make it all possible. For now the best we can do in python is to implement logger.is_enabled that'd return false if logs SDK is not configured - created https://github.com/open-telemetry/opentelemetry-python/issues/4410 to track.

xrmx (approved):

PR #3234, opened 2025-02-03
[release/v1.30.x-0.51bx] Prepare release 1.30.0/0.51b0

opentelemetrybot:
Prepare release 1.30.0/0.51b0.


instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "2.1b0.dev"
+__version__ = "0.51b0"
```
aabmass, line 15: Can you revert this change? Should be versioned separately

aabmass (approved):
1 changed needed otherwise LGTM

PR #3253, opened 2025-02-10
fixed: asyncpg connection params are a namedtuple

privatwolke:
Follow-up on the apparently abbandonned https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2114.
Closes #2114

# Description

The asyncpg instrumentation attempts to fall back on using the database name as the span name in case the first argument to the instrumented method is falsey.

Any attempt to do so results in errors simliar to this (truncated traceback):
```
  File "/opt/venv/lib/python3.12/site-packages/asyncpg/pool.py", line 597, in fetch
    return await con.fetch(
           ^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/opentelemetry/instrumentation/asyncpg/__init__.py", line 154, in _do_execute
                                   ^^^^^^^^^^
AttributeError: 'ConnectionParameters' object has no attribute 'get'
```

This has probably never worked since asyncpg defines the `_params` attribute as an instance of `ConnectionParams`
(https://github.com/MagicStack/asyncpg/blob/master/asyncpg/connection.py#L62) which is a NamedTuple instance and thus don't define `get`. The proper way of safely accessing properties on a NamedTuple is using `getattr`.

The only case that I've actually found which triggers this branch is if the supplied query is an empty string. This is something that causes an `AttributeError` for `Connection.execute` but is fine for `fetch()`, `fetchval()`, `fetchrow()` and `executemany()`.

The tests have been expanded to check these cases. Also, more status code validation has been added where it was missing.

## Type of change

# How Has This Been Tested?

- [x] docker-tests for asyncpg instrumentation extended and passed

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
@@ -150,8 +150,10 @@ class AsyncPGInstrumentor(BaseInstrumentor):
 
     async def _do_execute(self, func, instance, args, kwargs):
         exception = None
-        params = getattr(instance, "_params", {})
-        name = args[0] if args[0] else params.get("database", "postgresql")
+        params = getattr(instance, "_params", None)
+        name = (
+            args[0] if args[0] else getattr(params, "database", "postgresql")
+        )
 
         try:
             # Strip leading comments so we get the operation name.
@@ -185,11 +187,11 @@ class AsyncPGInstrumentor(BaseInstrumentor):
     async def _do_cursor_execute(self, func, instance, args, kwargs):
         """Wrap cursor based functions. For every call this will generate a new span."""
         exception = None
-        params = getattr(instance._connection, "_params", {})
+        params = getattr(instance._connection, "_params", None)
         name = (
             instance._query
             if instance._query
-            else params.get("database", "postgresql")
+            else getattr(params, "database", "postgresql")
         )
 
         try:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - `opentelemetry-instrumentation-redis` Add missing entry in doc string for `def _instrument`
   ([#3247](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3247))
+- `opentelemetry-instrumentation-asyncpg` Fix fallback for empty queries.
+  ([#3253](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3253))
 
 ## Version 1.30.0/0.51b0 (2025-02-03)
 
--- a/tests/opentelemetry-docker-tests/tests/asyncpg/test_asyncpg_functional.py
+++ b/tests/opentelemetry-docker-tests/tests/asyncpg/test_asyncpg_functional.py
@@ -1,5 +1,7 @@
 import asyncio
 import os
+from collections import namedtuple
+from unittest.mock import patch
 
 import asyncpg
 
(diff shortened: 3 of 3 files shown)


instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
```diff
@@ -150,8 +150,8 @@ def _uninstrument(self, **__):
 
     async def _do_execute(self, func, instance, args, kwargs):
         exception = None
-        params = getattr(instance, "_params", {})
-        name = args[0] if args[0] else params.get("database", "postgresql")
+        params = getattr(instance, "_params")
+        name = args[0] if args[0] else getattr(params, "database", "postgresql")
```
tammy-baylis-swi, line 154: Nice. Could this also be updated in [_do_cursor_execute](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/317f7a6150d73f89959e8ec44c6d9f599977d289/instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py#L189-L193)?

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
```diff
@@ -150,8 +150,8 @@ def _uninstrument(self, **__):
 
     async def _do_execute(self, func, instance, args, kwargs):
         exception = None
-        params = getattr(instance, "_params", {})
```
tammy-baylis-swi, line 153: Would we still want to keep the default to prevent other AttributeError?

tammy-baylis-swi:
Thanks @privatwolke for revisiting this especially with the added test coverage.

To get [this check](https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/13242594700/job/36974933833?pr=3253) to pass, please could you run `tox -e ruff` on your local then commit the changes.

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
privatwolke: Yup, fixed!

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
privatwolke: Done!

privatwolke:
@tammy-baylis-swi - thanks for the quick review! I've updated the PR with the suggested changes and also added additional tests for the `Cursor` instrumentation. Code formatting should now pass as well ✨

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
tammy-baylis-swi: Thanks! We should still use the original default of `{}` instead of `None`, else the `getattr` call will give AttributeError for NoneType object if there's no `_params`.

privatwolke: Not really - `getattr`works just fine on `None`:
```py
>>> getattr(None, '_params', 'bar')
'bar'
```

That means that if there is no `_params`, we would have `params = None`. After that we do `getattr(params, "database", "postgresql")` which is evaluated to `getattr(None, "database", "postgresql")` which finally evaluates to `"postgresql"`.

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
```diff
@@ -150,8 +150,10 @@ def _uninstrument(self, **__):
...
-        name = args[0] if args[0] else params.get("database", "postgresql")
+        params = getattr(instance, "_params", None)
+        name = (
+            args[0] if args[0] else getattr(params, "database", "postgresql")
```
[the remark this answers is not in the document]
privatwolke: TBH, I don't think that this code is possible to hit without doing some pretty shady things to the internals of asyncpg. It would mean that asyncpg has changed in a way that either `Connection._params` is not there anymore or that the `ConnectionParams` namedtuple has no `database` member anymore.

Even if you leave the `database` kwarg blank when you initiate the connection, it'll fall back on the username as the default database. In that case `database` is still a member of the `ConnectionParams` ([and has a value](https://github.com/MagicStack/asyncpg/blob/master/asyncpg/connect_utils.py#L464-L465)).

I've added a test for this using `mock.object` to change the `_params` property since I don't see any real-world scenario of this happening without upstream changes.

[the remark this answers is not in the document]
tammy-baylis-swi: Yes, mocks work just fine for this!

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
tammy-baylis-swi: Yes, you are correct 👍

xrmx (approved):

privatwolke:
@xrmx I think something went wrong with the automerge. Should I rebase my branch on the base branch?

PR #3268, opened 2025-02-13
[chore] Replace '-0' and '-1' with semantically meaningful suffixes. Other minor fixes in 'instrumentation-genai' test structure.

michaelsafyan:
# Description

This is a follow up item from feedback on:

 - https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3256/

In particular, the new code proposed here breaks with the existing pattern of adding `-0` and `-1` suffixes in the `tox.ini` file in favor of more intuitive `-old-deps` and `-recent-deps` suffixes, encoding the intended meaning.

It was requested on that PR to send a separate PR to make similar updates to the other cases.

This updates the existing `instrumentation-genai` packages to follow a similar pattern. It likewise updates the naming and location of the corresponding `requirements.txt` files to similarly reflect their meaning. And it adds documentation comments to the `requirements.txt` files that parallel those in that pull request.

## Type of change

Minor internal structuring, build improvement.

# How Has This Been Tested?

Ran `tox` to verify that things still build. In particular, ran the following commands:

  - `tox -l | grep openai | xargs -n 1 tox -e`
  - `tox -l | grep vertexai | xargs -n 1 tox -e`

I verified that both `recent-deps` and `old-deps` variants of each pass for at least one version of Python, verifying that this modification did not induce any breakages in the general build mechanism.

No

# Checklist:

I believe that all checklist items that apply have been followed.


michaelsafyan:
Note: the feedback on that PR that I'm referring to is this:

https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3256#discussion_r1951807052

michaelsafyan:
CC: @aabmass

instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/requirements.latest.txt
```diff
@@ -0,0 +1,52 @@
...
+#   openai-recent-deps: {[testenv]test_deps}
+#   openai-recent-deps: -r {toxinidir}/instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/requirements.latest.txt
+#
+# This provides additional dependencies, namely:
+#
+#   opentelemetry-api
+#   opentelemetry-sdk
+#   opentelemetry-semantic-conventions
+#
+# ... with a "dev" version based on the latest distribution.
```
aabmass, line 33: Can you instead just move this note to https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md#update-supported-instrumentation-package-versions?

This is actually the norm for most of the packages in this repo (they test against recent versions and opentelemetry-python@HEAD) and there's already some discussion in there.

aabmass (approved):
Thanks! I'd like one more review from other maintainers.

codefromthecrypt:
I think in a diff like this best to use two words and be consistent both in tox and also the requirements names.

So, "latest" makes sense, so in both places.

For the floor version I suggest using "floor" or similar and also both places.

Basically, I think that whatever the two words chosen this PR passes when there are 2 not 4 phrases for the two intents, verified that diffs match renames

My 2p

instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/requirements.latest.txt
michaelsafyan: I think it's easier to overlook when the comment is not colocated with the code that it is related to. Putting comments in the affected files rather than centrally, while it can create some redundancy, makes it more obvious.

tox.ini
```diff
@@ -9,14 +9,14 @@ envlist =
     ; for specifying supported Python versions per package.
 
     ; instrumentation-openai
-    py3{8,9,10,11,12,13}-test-instrumentation-openai-v2-{0,1}
-    pypy3-test-instrumentation-openai-v2-{0,1}
+    py3{8,9,10,11,12,13}-test-instrumentation-openai-v2-{old-deps,recent-deps}
```
emdneto, line 12: I would prefer a single word here, like @codefromthecrypt said: {floor, latest} and also something that matches the requirements.txt file name.

At the end, you need to run `tox -e generate-workflows` and commit the changes so things can run at CI

michaelsafyan: Done. Updated to use the same name for both (`oldest` and `latest`). I didn't use `floor`, because to me that implies rounding down.  Also ran the `generate-workflows` command and included in the update.

emdneto: There are some fixes to do for spellcheck, but other than that, LGTM. Maintainers probably need to remove the old checks required from CI. @aabmass would you be able to help with that?

michaelsafyan: Thanks. I wasn't able to get `tox -e spellcheck` to work correctly locally, but I think I fixed the spellcheck issues that were flagged in the GitHub evaluation.

emdneto (approved):

xrmx (approved):
Not a fan of adding copyright headers to requirements files but LGTM

xrmx:
Updating repo setting to have the proper jobs required to pass

PR #3276, opened 2025-02-19
Fix openaiv2 example readme

emdneto:
I was trying the examples and noticed the rst file was rendered weird


codefromthecrypt:
would be nice if we could render before/after like some site-oriented repo PR actions do..

xrmx (approved):

xrmx:
> would be nice if we could render before/after like some site-oriented repo PR actions do..

Another option since these are not built in the documentation we can just write them in markdown

emdneto:
> > would be nice if we could render before/after like some site-oriented repo PR actions do..
> 
> Another option since these are not built in the documentation we can just write them in markdown

+1 for markdown.

PR #3282, opened 2025-02-21
Fix uv setup when using tox

emdneto:
Since we are using `tox-uv`, it is reading the uv configuration under `pyproject.toml` for sources and workspaces (as expected), and this is causing some tests to fail [because uv and tox deps are conflicting](https://github.com/open-telemetry/opentelemetry-python/actions/runs/13449331885/job/37582716831) (e.g., tox is installing opentelemetry-api@CORE_REPO_SHA and uv is telling to install opentelemetry-api@main. I explained a bit about that behavior [here](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3124#discussion_r1944965949)

This change is to make tox ignore whatever is defined in [tool.uv.sources] configuration

Note: I tried `UV_NO_CONFIG` env var but didn't work because workspaces/sources are related to pyproject scope

PR in -core to show it works for cross-reference SHA: https://github.com/open-telemetry/opentelemetry-python/pull/4445

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,6 +60,7 @@ dependencies = [
   "opentelemetry-instrumentation-urllib3[instruments]",
   "opentelemetry-instrumentation-wsgi",
   "opentelemetry-propagator-ot-trace",
+  "opentelemetry-propagator-aws-xray",
   "opentelemetry-util-http",
 ]
 
@@ -126,6 +127,7 @@ opentelemetry-instrumentation-urllib = { workspace = true }
 opentelemetry-instrumentation-urllib3 = { workspace = true }
 opentelemetry-instrumentation-wsgi = { workspace = true }
 opentelemetry-propagator-ot-trace = { workspace = true }
+opentelemetry-propagator-aws-xray = { workspace = true }
 opentelemetry-util-http = { workspace = true }
 
 # https://docs.astral.sh/uv/reference/settings/#workspace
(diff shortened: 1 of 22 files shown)


xrmx (approved):

pyproject.toml
```diff
@@ -126,6 +127,7 @@ opentelemetry-instrumentation-urllib = { workspace = true }
 opentelemetry-instrumentation-urllib3 = { workspace = true }
 opentelemetry-instrumentation-wsgi = { workspace = true }
 opentelemetry-propagator-ot-trace = { workspace = true }
+opentelemetry-propagator-aws-xray = { workspace = true }
```
emdneto, line 130: Solved the issue with aws-xray from the previous PR while at it

aabmass: Could you also add openai-v2 and vertexai while you're at it?

aabmass (approved):
Thank you!

pyproject.toml
emdneto: @aabmass 
```
@emdneto ➜ /workspaces/opentelemetry-python-contrib (fix-uv) $ uv sync
Resolved 230 packages in 945ms
      Built opentelemetry-instrumentation-vertexai @ file:///workspaces/opentelemetry-python-contrib/instrumentation-genai/opentelemetry-instrumentation-vertexai
      Built opentelemetry-instrumentation-openai-v2 @ file:///workspaces/opentelemetry-python-contrib/instrumentation-genai/opentelemetry-instrumentation-openai-v2
      Built google-crc32c==1.6.0
Prepared 30 packages in 13.55s
```

PR #3296, opened 2025-02-24
Add optional traceback as spanattribute for psycopg2 instrumentation

adilbenameur:
# Description

To help debug bad performance, I thought it would be useful to have the traceback of a call to a database. Especially for ORMs like Django's that make a lot of call that might not be wanted in the first place, it could help to find the code responsible for the calls.
The changes are non breaking.

## Type of change

- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

I hadded tests to check that traceback would be indeed provided in the span (and not if the option is set to False).

- test_enable_traceback (line 298 in instrumentation/opentelemetry-instrumentation-dbapi/tests/test_dbapi_integration.py)
- test_disabled_traceback (line 329 in instrumentation/opentelemetry-instrumentation-dbapi/tests/test_dbapi_integration.py)

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

- [ ] Changelogs have been updated (not sure there is one for psycopg2 ?)
- [ ] Documentation has been updated (didn't found the doc to fill but I would happy to to)


xrmx:
I understand the issue but it doesn't feel right to to add something ad-hoc per instrumentation

adilbenameur:
Would it be better if the PR also add support for the other database instrumention libs ? The implementation for psycopg2 was just for my use case but I understand that the libs must be uniform.

PR #3298, opened 2025-02-25
Add support for async and streaming responses in the Google GenAI instrumentation

michaelsafyan:
# Description

This builds on https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3256 to add support for async and streaming variants of the `GenerateContent` operation.

## Change Summary:

 - `generate_content.py`:
    - Fix minor issues in implementation of streaming async (align to underlying library's data model of the result)
    - Removal of guards to short-circuit patching of async, streaming cases
 - tests:
    - Extend existing utilities to enable mocking streaming responses
    - Add testing coverage for streaming, async cases that validate the instrumentation
 - docs:
    - Remove async, streaming support from list of `TODOS` given that this PR checks that off the list

## Type of change

# How Has This Been Tested?

Added unit test coverage.

- [X] No.


instrumentation-genai/opentelemetry-instrumentation-google-genai/TODOS.md
```diff
@@ -4,18 +4,17 @@
 
 Here are some TODO items required to achieve stability for this package:
 
-  1. Add support for streaming interfaces
-  2. Add support for async interfaces
-  3. Add more span-level attributes for request configuration
-  4. Add more span-level attributes for response information
-  5. Verify and correct formatting of events:
+  1. Add more span-level attributes for request configuration
```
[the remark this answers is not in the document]
michaelsafyan: Switched to a bulleted list, because they do not have to land in the specified order (and some can even be done in parallel with each other).

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -646,25 +633,33 @@ async def instrumented_generate_content_stream(
...
+                    helper.process_error(error)
+                    raise
+                finally:
+                    helper.finalize_processing()
+
+        class _GeneratorProvider:
+            def __aiter__(self):
+                return _internal_generator()
+
+        return _GeneratorProvider()
```
aabmass, line 662: Looking at https://googleapis.github.io/python-genai/genai.html#genai.models.AsyncModels.generate_content_stream

I think this code changes the semantics of the wrapped function slightly. Now when the user calls `await client.aio.models.generate_content_stream()`, the `wrapped_func` still has not been called and the span has not started.

Something like this should work but haven't tested it:

```py
async def instrumented_generate_content_stream(...):
  with helper.start_span_as_current_span(...):
    gen = await wrapped_func(...)
 
    async def _internal_generator():
      async for response in gen:
        ...
  
    return _internal_generator()
```

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/common/requests_mocker.py
```diff
@@ -123,6 +124,35 @@ def generate_response_from_dict(args):
...
+        result.encoding = "utf-8"
+        result.headers["transfer-encoding"] = "chunked"
+        contents = []
+        for generator in response_generators:
+            response = generator(args)
+            if response.status_code != 200:
+                continue
+            response_json = response.json()
+            response_json_str = json.dumps(response_json)
+            contents.append(f"data: {response_json_str}")
```
aabmass, line 147: Does the API actually send multiple responses like this or does it stream on a single request?

michaelsafyan: I believe this is how it actually works. You can see the related code in:

[`python-genai/google/genai/_api_client.py`](https://github.com/googleapis/python-genai/blob/33d309c07e9d955854a6ca0a0e16c1c342b7b5d2/google/genai/_api_client.py#L182):

```
     for chunk in self.response_stream.iter_lines():
        if chunk:
          # In streaming mode, the chunk of JSON is prefixed with "data:" which
          # we must strip before parsing.
          if chunk.startswith(b'data: '):
            chunk = chunk[len(b'data: ') :]
          yield json.loads(str(chunk, 'utf-8'))
```

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
michaelsafyan: Good point. It was a bit more complicated than that, but updated based on this feedback to begin eager execution earlier, before the point at which the responses are being consumed.

instrumentation-genai/opentelemetry-instrumentation-google-genai/TODOS.md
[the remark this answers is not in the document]
michaelsafyan: Marking as resolved given the change to unordered.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
michaelsafyan: Marking resolved given the restructuring here to eagerly begin the processing.

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/common/requests_mocker.py
michaelsafyan: Marking as resolved given that this question has been answered.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -647,24 +637,42 @@ async def instrumented_generate_content_stream(
...
+                    helper.process_error(error)
+                    raise
+                finally:
+                    helper.finalize_processing()
+
+        class _GeneratorProvider:
+            def __aiter__(self):
+                return _response_async_generator_wrapper()
+
+        return _GeneratorProvider()
```
aabmass, line 675: I think this is redundant and can just do this

```suggestion
        return _response_async_generator_wrapper()
```

aabmass (approved):

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
michaelsafyan: Done.

Issue #3300, opened 2025-02-25
Starlette instrumentation's `instruments` version is too restrictive


aabmass:
The starlette instrumentation specifies

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/b76119bb8f4be539a1972baa811102e579b1b0e7/instrumentation/opentelemetry-instrumentation-starlette/pyproject.toml#L36-L39

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/b76119bb8f4be539a1972baa811102e579b1b0e7/instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/package.py#L16

This does not mean `>=0.13`, it actually means `starlette>=0.13.0,<0.14.dev0`. Concretely this overly narrow requirement is causing resolution issues with the new UV workspace introduced in #3124 and the google-genai instrumentation PR https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3256.

```
...
      And because opentelemetry-instrumentation-starlette[instruments] depends on starlette>=0.13.0,<0.14.dev0 and opentelemetry-instrumentation-fastapi[instruments] depends on fastapi>=0.58, we can conclude that all of:
...
```

aabmass:
@Kludex @emdneto this is making `tox -e ruff` precommit fail, because of d092c1afe9a21a19aa1d605e32f136a34f216144

emdneto:
> [@Kludex](https://github.com/Kludex) [@emdneto](https://github.com/emdneto) this is making `tox -e ruff` precommit fail, because of [d092c1a](https://github.com/open-telemetry/opentelemetry-python-contrib/commit/d092c1afe9a21a19aa1d605e32f136a34f216144)
Now the `tox -e ruff` doesn't make sense; we can probably change the name to pre-commit instead

aabmass:
> Now the `tox -e ruff` doesn't make sense; we can probably change the name to pre-commit instead

I agree, but I think it might be a little annoying to people with muscle memory for `tox -e ruff`. Is it possible to keep `tox -e ruff` for just running ruff somehow?

emdneto:
@aabmass I think this, or create separated jobs for ruff, ruff-format and uv-lock. 

```
diff --git a/tox.ini b/tox.ini
index c55482e7..f64bce06 100644
--- a/tox.ini
+++ b/tox.ini
@@ -1006,6 +1006,18 @@ commands =
 
[...]

PR #3304, opened 2025-02-26
Loosen `opentelemetry-instrumentation-starlette[instruments]` specifier

aabmass:
# Description

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3300

Ran `uv sync --upgrade` to regenerate the lockfile with latest starlette.

## Type of change

# How Has This Been Tested?

I haven't actually tested this. Hoping @Kludex can confirm if this should be working.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-starlette/pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
 
 [project.optional-dependencies]
 instruments = [
-  "starlette ~= 0.13.0",
+  "starlette ~= 0.13",
```
aabmass, line 38: @Kludex Alternatively, we could just do `starlette >= 0.13.0`. I'm not sure we can reasonable choose an upper bound, wdyt?

Kludex: 0.13 is too old, some security vulnerabilities released since then... I'm not sure the exact version right now (on the phone), but at least 0.19.0 plus.

Kludex: Ah, I misread what you said.

I think we thought it was working... so I guess we can?

xrmx (commented):
I think we should add a test requirements with latest starlette too

aabmass:
> I think we should add a test requirements with latest starlette too

Ya probably, I'll add to this PR. We really need a better way to do this long term I think.

instrumentation/opentelemetry-instrumentation-starlette/pyproject.toml
aabmass: So turns out the instrumentation fails tests on any recent versions of starlette. The newest I could get is `starlette >= 0.13, <0.15` 😞 

Mostly seeing errors with extracting custom headers:
`E   AssertionError: 'http.response.header.custom_test_header_1' not found in mappingproxy({'http.scheme': 'ws', 'http.host': 'testserver', 'net.host.port': 80, 'http.target': '/foobar_web', 'http.url': 'ws://testserver/foobar_web', 'http.server_name': 'testserver', 'http.user_agent': 'testclient', 'net.peer.ip': 'testclient', 'net.peer.port': 50000, 'http.route': '/foobar_web', 'http.status_code': 200})`

aabmass:
So I was able to generate locked dependencies for oldest and newest versions pretty easily with `uv pip compile --resolution lowest-direct | highest` https://docs.astral.sh/uv/concepts/resolution/#resolution-strategy

This works pretty nicely, I think we should adopt this for all packages.

instrumentation/opentelemetry-instrumentation-starlette/pyproject.toml
aabmass: I'd like to merge this PR as is, and I'll create a follow up issue for starlette

xrmx (approved):

instrumentation/opentelemetry-instrumentation-starlette/test-requirements.latest.txt
```diff
@@ -0,0 +1,112 @@
+# This file was autogenerated by uv via the following command:
+#    uv pip compile --python 3.8 --universal --extra instruments instrumentation/opentelemetry-instrumentation-starlette/pyproject.toml -c dev-requirements.txt instrumentation/opentelemetry-instrumentation-starlette/test-requirements.in -o instrumentation/opentelemetry-instrumentation-starlette/test-requirements.latest.txt
+-e file:///usr/local/google/home/aaronabbott/repo/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-asgi
```
xrmx, line 3: Thanks for sharing :)

xrmx (changes requested):

instrumentation/opentelemetry-instrumentation-starlette/test-requirements.latest.txt
aabmass: I still don't know why UV does this but I found an incantation that removes it.

instrumentation/opentelemetry-instrumentation-starlette/pyproject.toml
aabmass: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3317

instrumentation/opentelemetry-instrumentation-starlette/test-requirements.latest.txt
Kludex, line 1: Can you generate this from the pyproject only? 🤔 so you don't need a `test-requirements.in`, but instead you could leverage the `dependency-groups`?

instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/package.py
```diff
@@ -13,6 +13,6 @@
 # limitations under the License.
 
 
-_instruments = ("starlette ~= 0.13.0",)
+_instruments = ("starlette >= 0.13, <0.15",)
```
Kludex, line 16: 0.15 is too old, FYI

xrmx (approved):

instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/package.py
xrmx: We know it's old but it's what the instrumentation is currently able to instrument :sweat_smile:

xrmx:
@aabmass need to update the branch rules settings

instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/package.py
Kludex: That's the same as saying that Starlette doesn't have an instrumentation package. That version is 4 years old.

Kludex: But I was just pointing out.

aabmass: Ya it probably means many people are using a buggy instrumentation. Let's take this discussion into https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3317

instrumentation/opentelemetry-instrumentation-starlette/test-requirements.latest.txt
aabmass: `uv pip compile` does not support specifiying `dependency-groups` https://github.com/astral-sh/uv/issues/8969. I think we would need to use `uv.lock` for that which is complicated by the workspaces setup we did, and idk how to get it working with tox.

I would love to work on this for the whole repo but not sure I have time right now. It may require some feature requests to UV.

aabmass: There's also (presumably) a bug where `uv pip compile` outputs absolute paths if you point it at a workspace pyproject.toml https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3304#discussion_r1975020273

emdneto: I would say I prefer the wording as `lowest`, `highest` instead of latest, oldest. But we can discuss it later for the whole repo

aabmass:
Added 46282d4 so I can submit this since the CI is deadlocked between this PR and https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3322.

emdneto (approved):

PR #3324, opened 2025-03-04
Fix tracing of async cursors for psycopg

### LINKED ISSUE #2486: `opentelemetry-instrumentation-psycopg` claims to work for async queries, but doesn't record time

ibash:
# Description

This copies the traced_execution of AsyncCursorTracer except query_method is awaited within the span.

Fixes #2486 

## Type of change

# How Has This Been Tested?

I ran this locally by making the corresponding changes in my site-packages folder. Then I checked logfire (what we're using for opentelemetry) to confirm that sql queries now had non-zero timings.

Also ran the `tox` command to run tests.

Happy to build and test the built package if you tell me how to build!

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

- [-] Changelogs have been updated
- [-] Unit tests have been added
- [-] Documentation has been updated


ibash:
@federicobond would love a review when you get a chance!

emilingerslev (approved):
I've validated that this indeed solves the problem. Only concern here is minor; The override of `traced_execution` changes the signature from sync to async, effectively also changing the return type. Another option is to extend `CursorTracer` with a `traced_execution_async` method, keeping it all in one class, but supporting both.

But as I said this works ✅

xrmx:
> I've validated that this indeed solves the problem. Only concern here is minor; The override of `traced_execution` changes the signature from sync to async, effectively also changing the return type. Another option is to extend `CursorTracer` with a `traced_execution_async` method, keeping it all in one class, but supporting both.

Thanks for testing. Not sure I follow the return type concern, AFAIU the users of that `_cursor_tracer` are awaiting it so the fact it was not an async function was the issue?

xrmx (changes requested):
I think we are missing tests to avoid regressions in the future

ibash:
@xrmx sorry for the delay here -- added a regression test. Please let me know if there's a better way to assert this behavior.

instrumentation/opentelemetry-instrumentation-psycopg/tests/test_psycopg_integration.py
```diff
@@ -492,3 +496,27 @@ async def test_not_recording_async(self):
             self.assertFalse(mock_span.set_status.called)
 
         PsycopgInstrumentor().uninstrument()
+
+    async def test_tracing_is_async(self):
+        PsycopgInstrumentor().instrument()
+
+        # before this async fix cursor.execute would take 14000 ns, delaying for
```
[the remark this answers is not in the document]
xrmx: Sure, but here we are testing the mocks and not the real code. Would it be possible to test the connection `cursor_factory` we get from `instrument_connection` ?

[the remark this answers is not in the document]
ibash-corpusant: This tests the real code. The test is for the _tracer_ class AsyncCursorTracer. It doesn’t matter what Cursor class it uses.

instrumentation/opentelemetry-instrumentation-psycopg/src/opentelemetry/instrumentation/psycopg/__init__.py
```diff
@@ -381,6 +381,46 @@ def callproc(self, *args: Any, **kwargs: Any):
...
+                        # sqlcomment is only added to executed query
+                        # so db.statement is set before add_sql_comment
+                        self._populate_span(span, cursor, *args)
+                        args = self._update_args_with_added_sql_comment(
+                            args, cursor
+                        )
+                else:
+                    # no sqlcomment anywhere
+                    self._populate_span(span, cursor, *args)
+            return await query_method(*args, **kwargs)
```
ibash, line 421: @xrmx this is what the bugfix is for, the await was missing before

instrumentation/opentelemetry-instrumentation-psycopg/tests/test_psycopg_integration.py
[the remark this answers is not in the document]
ibash: This tests the real code. The test is for the tracer class AsyncCursorTracer. It doesn’t matter what Cursor class the test uses.

jwilm:
Thanks @ibash for this PR and @xrmx for your reviews. I'm eager to see this PR land and published; is there anything I can do to help?

ibash:
@jwilm not ideal but I did publish the patched version to github, in our requirements.txt we use:
```
opentelemetry-instrumentation-fastapi==0.52b1
opentelemetry-instrumentation-psycopg @ https://github.com/ibash/opentelemetry-python-contrib/releases/download/v52b1/opentelemetry_instrumentation_psycopg-0.52b1-py3-none-any.whl
```

ibash:
@xrmx fixed the formatting issue -- any chance this can get reviewed / merged?

xrmx (approved):

instrumentation/opentelemetry-instrumentation-psycopg/src/opentelemetry/instrumentation/psycopg/__init__.py
```diff
@@ -381,6 +381,46 @@ def callproc(self, *args: Any, **kwargs: Any):
     return TracedCursorFactory
 
 
+class AsyncCursorTracer(CursorTracer):
+    async def traced_execution(
+        self,
+        cursor: dbapi.CursorT,
+        query_method: Callable[..., Any],
```
aabmass: Can you update the annotation here to make it more explicit that the result needs to be awaited?
```suggestion
        query_method: Callable[..., Awaitable[Any]],
```

ibash-corpusant: done!

Kludex:
> > I've validated that this indeed solves the problem. Only concern here is minor; The override of `traced_execution` changes the signature from sync to async, effectively also changing the return type. Another option is to extend `CursorTracer` with a `traced_execution_async` method, keeping it all in one class, but supporting both.
> 
> Thanks for testing. Not sure I follow the return type concern, AFAIU the users of that `_cursor_tracer` are awaiting it so the fact it was not an async function was the issue?

When you subclass, you need to maintain the method's signature (Liskov Substitution Principle).

instrumentation/opentelemetry-instrumentation-psycopg/src/opentelemetry/instrumentation/psycopg/__init__.py
```diff
@@ -381,6 +381,46 @@ def callproc(self, *args: Any, **kwargs: Any):
     return TracedCursorFactory
 
 
+class AsyncCursorTracer(CursorTracer):
+    async def traced_execution(
```
Kludex: I suggest (as mentioned in the main thread) to create a new method e.g. `traced_execution_async`.

Kludex (changes requested):

Kludex:
Thanks for using Logfire! ❤️

ibash:
@Kludex @xrmx updated this to have `traced_execution_async` instead

PR #3332, opened 2025-03-05
Fix aiokafka multiple values headers error.

fcfangcc:
# Description

When using the **send_and_wait** method, a **TypeError** occurs:  
**`TypeError: original_send() got multiple values for argument 'headers'`**  

This happens because of the following lines in the OpenTelemetry AIoKafka instrumentation code:  
[[opentelemetry-instrumentation-aiokafka/utils.py#L263-L266](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/81eaea57f919055af56ca94e9bc523e234ca21a7/instrumentation/opentelemetry-instrumentation-aiokafka/src/opentelemetry/instrumentation/aiokafka/utils.py#L263-L266)](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/81eaea57f919055af56ca94e9bc523e234ca21a7/instrumentation/opentelemetry-instrumentation-aiokafka/src/opentelemetry/instrumentation/aiokafka/utils.py#L263-L266).  

If `headers` is present in `args[5]`, these lines modify `kwargs`, causing `headers` to appear in both `args` and `kwargs`, leading to the error.  

Example code triggering the issue(run with otel):  
```python
import asyncio

from aiokafka import AIOKafkaProducer


async def test():
    producer = AIOKafkaProducer(
        bootstrap_servers='bootstrap_servers',
    )
    await producer.start()
    await producer.send_and_wait('topic', value=b'test', headers=[])  # that's ok
    await producer.send_and_wait('topic', value=b'test')  # error here


asyncio.run(test())

```
Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] test_wrap_send_with_headers_as_args

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-aiokafka/tests/test_utils.py
```diff
@@ -120,6 +163,8 @@ async def wrap_send_helper(
         tracer = mock.MagicMock()
         produce_hook = mock.AsyncMock()
         original_send_callback = mock.AsyncMock()
+        original_send_callback.side_effect = original_send
+        original_send_callback.return_value = SEND_RETURN_VALUE
```
xrmx, line 167: Aren't these exclusive?

instrumentation/opentelemetry-instrumentation-aiokafka/src/opentelemetry/instrumentation/aiokafka/utils.py
```diff
@@ -260,6 +269,7 @@ async def _traced_send(
         args: Tuple[Any],
         kwargs: Dict[str, Any],
     ) -> None:
+        args, kwargs = _move_headers_to_kwargs(args, kwargs)
```
xrmx, line 272: Isn't the issue here that `_extract_send_headers` is not seeing headers in args[5] though?

fcfangcc: Since the method being wrapped has 6 parameters, with **header** being the last one, if the header is not in args, the **_move_headers_to_kwargs** method will return the original args and kwargs without affecting the result. 

https://github.com/aio-libs/aiokafka/blob/29b58dbcacc75a62b430ce5243ac684ccf16a7f8/aiokafka/producer/producer.py#L441-L448

I can't think of a better way to extract the **headers** parameter from args and kwargs, and to leave args and kwargs unchanged when it is None in args.

instrumentation/opentelemetry-instrumentation-aiokafka/tests/test_utils.py
fcfangcc: original_send is used to mock the parameter signature of the original method being wrapped. The original mock method was unable to trigger this bug（because of any parameter can by ok）

dimastbk:
Hi. I suggest add smth to tests:

```python
    async def test_send_and_wait(self) -> None:
        AIOKafkaInstrumentor().uninstrument()
        AIOKafkaInstrumentor().instrument(tracer_provider=self.tracer_provider)

        producer = await self.producer_factory()
        add_message_mock: mock.AsyncMock = (
            producer._message_accumulator.add_message
        )
        add_message_mock.side_effect = [mock.AsyncMock()(), mock.AsyncMock()()]

        tracer = self.tracer_provider.get_tracer(__name__)
        with tracer.start_as_current_span("test_span") as span:
            await producer.send_and_wait("topic_1", b"value_1")

        add_message_mock.assert_awaited_with(
            TopicPartition(topic="topic_1", partition=1),
            None,
            b"value_1",
            40.0,
            timestamp_ms=None,
            headers=[("traceparent", mock.ANY)],
        )
        assert (
            add_message_mock.call_args_list[0]
            .kwargs["headers"][0][1]
            .startswith(
                f"00-{format_trace_id(span.get_span_context().trace_id)}-".encode()
            )
        )

        await producer.send_and_wait("topic_2", b"value_2")
        add_message_mock.assert_awaited_with(
            TopicPartition(topic="topic_2", partition=1),
            None,
            b"value_2",
            40.0,
            timestamp_ms=None,
            headers=[("traceparent", mock.ANY)],
        )
```

fcfangcc:
@dimastbk Thank you for your suggestion. This test is more reasonable

dimastbk (approved):

instrumentation/opentelemetry-instrumentation-aiokafka/src/opentelemetry/instrumentation/aiokafka/utils.py
xrmx: I mean, shouldn't we fix `_extract_send_headers` instead of moving things from args to kwargs? Something like the untested below:

```
def _extract_send_headers(args: Tuple[Any], kwargs: Dict[str, Any]):
    """extract headers from `send` method arguments in AIOKafkaProducer class"""
    if len(args) > 5:
        return args[5]
    return kwargs.get("headers", None)
```

dimastbk: Exception raised on line https://github.com/open-telemetry/opentelemetry-python-contrib/blob/16c3d0f6c8051bb115b5ba17a04c9a200ddbe83f/instrumentation/opentelemetry-instrumentation-aiokafka/src/opentelemetry/instrumentation/aiokafka/utils.py#L306

If args is `("topic", None, None, None, None, None)` we set `kwargs` to `{"headers": {...}}`, and run `await func("topic", None, None, None, None, None <- (it's headers too), headers={...})` with an exception.

Another variant of fixing it is to add `_set_send_headers` and use it in `if headers is None` block:
```python
def _set_send_headers(args, kwargs, headers):
    if len(args) > 5:
        args = tuple(list(args[:5]) + [headers])
    else:
        kwargs["headers"] = headers
    return args, kwargs
```

fcfangcc: Maybe mixed *args,**kwargs -> **kwargs is better? 
```python
def transfer_to_kwargs(*args,**kwargs)->dict:
    ...
...
async def _traced_send(...):
    kwargs = transfer_to_kwargs(*args,**kwargs)
    ...
    # return await func(*args, **kwargs) 
    return await func(**kwargs) 
```

fcfangcc: > I mean, shouldn't we fix `_extract_send_headers` instead of moving things from args to kwargs? Something like the untested below:
> 
> ```
> def _extract_send_headers(args: Tuple[Any], kwargs: Dict[str, Any]):
>     """extract headers from `send` method arguments in AIOKafkaProducer class"""
>     if len(args) > 5:
>         return args[5]
>     return kwargs.get("headers")
> ```

Unfortunately, it is not possible to fix this issue by not change **args** or **kwargs**

If **headers** in args like  **("topic", None, None, None, None, None)**, we must change args[5] **headers** to []. 

One option is modify args(tuple type), like @dimastbk. 

Another option is mv args[5] **headers** to kwargs, args change to length 5, like me.

Or use your ** _extract_send_headers** , It is also necessary to change args/kwargs before  line296
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/27d5d93a6ab5034432ec19d4be837f35e7d7afb6/instrumentation/opentelemetry-instrumentation-aiokafka/src/opentelemetry/instrumentation/aiokafka/utils.py#L290-L296

The difference between these ways is where to modify the headers parameter

emdneto: I understand the issue. Even if we return the correct headers value in `_extract_send_headers`, the `return await func(*args, **kwargs)` will be called with headers present in both args and kwargs.

emdneto (approved):

instrumentation/opentelemetry-instrumentation-aiokafka/src/opentelemetry/instrumentation/aiokafka/utils.py
xrmx: Yeah but isn't the duplicated header issue introduced at line 264 (in this branch 274)?

fcfangcc: > Yeah but isn't the duplicated header issue introduced at line 264 (in this branch 274)?

After **args, kwargs = _move_headers_to_kwargs(args, kwargs)** args[5] (headers) remove from args to kwargs(If had).
If headers in kwargs, everthing is not changed.
Then after this line codes changed **headers** from kwargs, **headers** not in args, only in kwargs.

xrmx: > > Yeah but isn't the duplicated header issue introduced at line 264 (in this branch 274)?
> 
> After **args, kwargs = _move_headers_to_kwargs(args, kwargs)** args[5] (headers) remove from args to kwargs(If had). If headers in kwargs, everthing is not changed. Then after this line codes changed **headers** from kwargs, **headers** not in args, only in kwargs.

I'm not doubting the code works it looks like the need to touch args and kwargs for headers is not a necessity but something we inflict ourself :)

xrmx (approved):

PR #3344, opened 2025-03-06
[chore] Address TODO to migrate to VCRpy and remove bespoke RequestMocker code in Google GenAI SDK instrumentation

michaelsafyan:
# Description

Revises the tests to split up into two different categories:

  1. A majority of tests that fully mock out the GenAI SDK (and do not need any request-level mocking).
  2. An end-to-end ("e2e") test that uses VCRpy to test with a non-mocked GenAI SDK to sanity-check the system.

This addresses prior pull request feedback expressing concern regarding bespoke request mocking and deviation from typical pattern elsewhere of using the VCRpy component. This also addresses the original underlying concerns regarding velocity; since most tests do not use VCRpy and directly mock the underlying GenAI SDK, most feature changes can be tested with much simpler mocks of the GenAI SDK outputs.

## Type of change

- Code cleanup

# How Has This Been Tested?

- Recorded with:
 - `tox -e py312-test-instrumentation-google-genai-oldest --vcr-record=all`
 
- Ran the tests with:
 - `tox -e py312-test-instrumentation-google-genai-oldest` 

- No


michaelsafyan:
Friendly ping.... is it possible to get this reviewed?

nirga (approved):

nirga:
@michaelsafyan note the failing CI

michaelsafyan:
Thanks. I had missed those failing checks.

Notes for myself for addressing this:

  1. First fix is in relation to lint (`tox -e lint-instrumentation-google-genai`).
  2. Second is in relation to tests on Python 3.9 (`tox -e py39-test-instrumentation-google-genai-oldest`, `py39-test-instrumentation-google-genai-latest`). It looks like VCRpy reports `content-encoding: gzip` even when it isn't gzipped, and this is causing it to fail in earlier version of Python... I'm guessing more stringent validation compared to later versions.

lmolkova (approved):

michaelsafyan:
@xrmx sorry for bothering you, but since you commented with a rocket emoji earlier and are one of the maintainers, I'm wondering if I might be able to ask you to get this merged?

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/test_e2e.py
```diff
@@ -0,0 +1,507 @@
...
+        gcloud_call_result = subprocess.run(
+            "gcloud config get project",
+            shell=True,
+            capture_output=True,
+            check=True,
+        )
+    except subprocess.CalledProcessError:
+        return None
+    gcloud_output = gcloud_call_result.stdout.decode()
+    return gcloud_output.strip()
```
aabmass, line 70: What are these for? `google.auth.default()` should behave the same as these

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/test_e2e.py
```diff
@@ -0,0 +1,507 @@
...
+    return False
+
+
+def _redact_headers(headers):
+    to_redact = []
+    for header_key in headers:
+        if _should_redact_header(header_key.lower()):
+            to_redact.append(header_key)
+    for header_key in to_redact:
+        headers[header_key] = "<REDACTED>"
```
aabmass, line 124:
```suggestion
    for header_key in headers:
        if _should_redact_header(header_key.lower()):
            headers[header_key] = "<REDACTED>"
```

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/test_e2e.py
```diff
@@ -0,0 +1,507 @@
...
+            and "gzip" not in headers["Content-Encoding"]
+        ):
+            continue
+        response["body"]["string"] = _ensure_gzip_single_response(
+            response["body"]["string"].encode()
+        )
+
+
+def _maybe_ensure_casette_gzip(result):
+    if sys.version_info[0] == 3 and sys.version_info[1] == 9:
```
aabmass, line 281: Is it only broken in 3.9?

aabmass (approved):

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/test_e2e.py
michaelsafyan: It doesn't. `google.auth.default()` gets the project from the credential, which works if you are using a service account. But If you are using `gcloud auth application-default login` rather than `gcloud auth activate-service-account`, then the credential does not have an associated project.

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/test_e2e.py
michaelsafyan: Done. I had previously done it with two loops, because of deleting the keys rather than setting the value to `<REDACTED>` which had created issues with modifying the iterator during iteration.

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/test_e2e.py
michaelsafyan: Yes, that is correct. Version 3.9 seems to be more stringent with validating gzip content. The other versions will return the original content if it does not match the expectations of the gzip format.

PR #3374, opened 2025-03-19
Include more request configuration options into the span attributes for the Google GenAI SDK instrumentation

michaelsafyan:
# Description

Improves the comprehensiveness and uniformity of reporting span-level attributes for Google GenAI SDK operations.

This essentially dumps all of the configuration options into span attributes, with a vendor-specific prefix. The subset of request options that are defined in OTel Semantic Conventions are then renamed to match conventions.

In service of the above, this change also:

 1. Adds a `flatten_dict` util used to generate the span attributes.
 2. Adds tests to more fully cover the set of attributes that are included.

## Type of change

# How Has This Been Tested?

Added more integration tests, covering the relevant changes.

- [X] No.

# Checklist:


michaelsafyan:
Friendly ping. Is it possible to get this reviewed?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -129,21 +131,48 @@ def _determine_genai_system(models_object: Union[Models, AsyncModels]):
...
-        if isinstance(current_context, dict):
-            current_context = current_context.get(path_segment)
-        else:
-            current_context = getattr(current_context, path_segment)
-    return current_context
+        return
+    span_context = span.get_span_context()
+    if not span_context.trace_flags.sampled:
+        return
+    attributes = flatten_dict(
```
xrmx, line 150: Maybe add a comment on why you need to do this?

michaelsafyan: Done. Added comments.

xrmx (approved):

xrmx:
I approved the PR because I think this would be helpful for you to move forward but would be nice to add a google genai specific section to the genai semantic conventions with these attributes you are exporting. Now it's really opaque what we are exporting.

michaelsafyan:
> I approved the PR because I think this would be helpful for you to move forward but would be nice to add a google genai specific section to the genai semantic conventions with these attributes you are exporting. Now it's really opaque and we don't have assertions on what we are actually exporting.

Thanks, @xrmx . Much appreciated. I will follow up on the Semantic Conventions (I know I have a bunch of pending tasks here).

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -129,21 +131,60 @@ def _determine_genai_system(models_object: Union[Models, AsyncModels]):
...
+    if not span_context.trace_flags.sampled:
+        # Avoid potentially costly traversal of config
+        # options if the span will be dropped, anyway.
+        return
+    # Automatically derive attributes from the contents of the
+    # config object. This ensures that all relevant parameters
+    # are captured in the telemetry data (except for those
+    # that are excluded via "exclude_keys").
+    attributes = flatten_dict(
+        _to_dict(config),
```
aabmass, line 157: I saw you're planning to make a semconv PR for some of these parameters. Are you planning to declare `gen_ai.gcp.request` as dynamic or listing out the actual expected request params?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -129,21 +131,60 @@ def _determine_genai_system(models_object: Union[Models, AsyncModels]):
...
-            current_context = getattr(current_context, path_segment)
-    return current_context
+        return
+    span_context = span.get_span_context()
+    if not span_context.trace_flags.sampled:
+        # Avoid potentially costly traversal of config
+        # options if the span will be dropped, anyway.
+        return
+    # Automatically derive attributes from the contents of the
+    # config object. This ensures that all relevant parameters
```
aabmass, line 153: Is it possible that this blows up the attributes size or accidentally includes sensitive data with the dynamic approach? For reference, `http.request.header.<key>` is opt in for this kind of thing https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -129,21 +131,60 @@ def _determine_genai_system(models_object: Union[Models, AsyncModels]):
...
+        ],
+        # Although a custom prefix is used by default, some of the attributes
+        # are captured in common, standard, Semantic Conventions. For the
+        # well-known properties whose values align with Semantic Conventions,
+        # we ensure that the key name matches the standard SemConv name.
+        rename_keys={
+            # TODO: add more entries here as more semantic conventions are
+            # generalized to cover more of the available config options.
+            "gen_ai.gcp.request.temperature": "gen_ai.request.temperature",
+            "gen_ai.gcp.request.top_k": "gen_ai.request.top_k",
```
aabmass, line 177: For the RHS standard semconv keys, we should use the constants from the semconv package

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -317,7 +314,12 @@ def _maybe_update_error_type(self, response: GenerateContentResponse):
     def _maybe_log_system_instruction(
         self, config: Optional[GenerateContentConfigOrDict] = None
     ):
-        system_instruction = _get_config_property(config, "system_instruction")
+        system_instruction = None
```
aabmass, line 317: I thought `_get_config_property()` was doing this for you

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
```diff
@@ -0,0 +1,222 @@
...
+            value.model_dump(),
+            key_prefix=key,
+            exclude_keys=exclude_keys,
+            rename_keys=rename_keys,
+            flatten_functions=flatten_functions,
+        )
+    if _from_json:
+        raise ValueError(
+            f"Cannot flatten value with key {key}; value: {value}"
+        )
```
aabmass, line 115: It looks like this causes the whole call to `flatten_dict()` to fail and no attributes would get added to the span

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
```diff
@@ -0,0 +1,222 @@
...
+    key_prefix: Optional[str] = None,
+    exclude_keys: Optional[Sequence[str]] = None,
+    rename_keys: Optional[Dict[str, str]] = None,
+    flatten_functions: Optional[Dict[str, Callable]] = None,
+):
+    key_prefix = key_prefix or ""
+    if exclude_keys is None:
+        exclude_keys = set()
+    elif isinstance(exclude_keys, list):
+        exclude_keys = set(exclude_keys)
```
aabmass, line 213:
```suggestion
    exclude_keys = set(exclude_keys or [])
```

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
```diff
@@ -0,0 +1,222 @@
...
+        result.update(flattened)
+    return result
+
+
+def flatten_dict(
+    d: Dict[str, Any],
+    key_prefix: Optional[str] = None,
+    exclude_keys: Optional[Sequence[str]] = None,
+    rename_keys: Optional[Dict[str, str]] = None,
+    flatten_functions: Optional[Dict[str, Callable]] = None,
```
aabmass, line 207: Can you add a signature to the callable?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
```diff
@@ -0,0 +1,222 @@
...
+    lst: list[Any],
+    key_prefix: str,
+    exclude_keys: Set[str],
+    rename_keys: Dict[str, str],
+    flatten_functions: Dict[str, Callable],
+) -> FlattenedDict:
+    result = {}
+    result[_concat_key(key_prefix, "length")] = len(lst)
+    for index, value in enumerate(lst):
+        full_key = f"{key_prefix}[{index}]"
```
aabmass, line 190: What are the `foo.length` and `foo[i]` for? I don't see those in the test cases anywhere

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
michaelsafyan: Thanks. I created an issue for defining the Semantic Conventions.

https://github.com/open-telemetry/semantic-conventions/issues/2053

I will get this convention defined and update here before we complete this PR.

I think dynamic is the better way to go, what do you think?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
michaelsafyan: Yes. The `flatten_dict` either succeeds or it does not. I think it is better to fail fully, loudly with a clear error than to partially succeed, making it difficult to debug why certain information is absent.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
michaelsafyan: I'm not sure how to express it in type annotations...

The signature is:


    def flatten_func(
            key: str,
            value: any,
            exclude_keys=Set[str],
            rename_keys=Dict[str, str],
            flatten_functions=Dict[str, flatten_func],
            **kwargs) -> Union[FlattenedDict, FlattenedValue]

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/custom_semconv.py
```diff
@@ -0,0 +1,18 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Prefix to use for LLM model request attributes that are unique GCP
+# (or that have not yet been formally defined in the GenAI/LLM SIG).
+CUSTOM_LLM_REQUEST_PREFIX = "gen_ai.gcp.request"
```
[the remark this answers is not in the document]
michaelsafyan: Current value reflects:

 - https://github.com/open-telemetry/semantic-conventions/pull/2050
 - https://github.com/open-telemetry/semantic-conventions/issues/2053

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
sasha-gitg: The safest approach is to opt-in every field because the config could include sensitive information in newly added fields  in the future and even updating the filter list will still result in previous versions of this opentelemetry integration referencing those fields.

Some headers may be useful for tracking. For example, provisioned throughput: https://cloud.google.com/vertex-ai/generative-ai/docs/use-provisioned-throughput

michaelsafyan: OK. I have updated this to use an allow list for determining which dynamic keys to include.

aabmass (approved):

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
michaelsafyan: Marking resolved.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
michaelsafyan: Thanks for that pointer. I've updated this logic to fail more gracefully and to log an error instead of raising.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
michaelsafyan: Done.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
michaelsafyan: Thanks for the pointer! Defined a protocol and updated type annotations.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
michaelsafyan: Good point. I've updated the AllowList to be capable of expressing wildcards like `foo.bar[*].baz` and `some.prefix.*` so that this is compatible with the allowlist approach.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
michaelsafyan: It was, but its main function was to support the span attributes that are now served with the dictionary flattening. The little bit of extra logic here is simpler than the `_get_config_property` definition, and so it seemed simpler overall to delete it in favor of the small bit of extra logic here.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
michaelsafyan: It should still be possible to align them with this approach, albeit if we treat the Gen AI SDK as the canonical representation of the config (rather than SemConv as the canonical representation of the config).

I think this makes sense given the approach that Gen AI SDK is taking towards unifying the mapping of the different config options.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/allowlist_util.py
```diff
@@ -0,0 +1,62 @@
...
+        if not stripped_entry:
+            continue
+        result.add(stripped_entry)
+    return result
+
+
+class AllowList:
+    def __init__(
+        self,
+        includes: Optional[Union[Set[str], List[str]]] = None,
```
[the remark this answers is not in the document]
michaelsafyan: Applied type annotation simplification.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
michaelsafyan: I think dumping it all to JSON defeats the purpose of flattening and is more likely to run into span attribute size limit issues. Let's leave as-is if you don't mind.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
michaelsafyan: The definition is proposed in PR https://github.com/open-telemetry/semantic-conventions/pull/2125 .

Marking as resolved for now.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
michaelsafyan: Marking resolved.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/custom_semconv.py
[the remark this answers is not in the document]
michaelsafyan: Marking resolved, since this has been updated to reflect the merged Semantic Convention updates (thanks for approving those!)

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
```diff
@@ -115,6 +115,41 @@ def _flatten_with_flatten_func(
...
+    rename_keys: Dict[str, str],
+    flatten_functions: Dict[str, FlattenFunc],
+    _from_json=False) -> FlattenedDict:
+    if _from_json:
+        _logger.debug(
+            "Cannot flatten value with key %s; value: %s", key, value
+        )
+        return {}
+    try:
+        json_string = json.dumps(value)
```
lmolkova, line 132: btw this guy uses ascii by default, try setting some value to non-ascii and enjoy unreadable stuff in console and other places that don't utf8 it at rendering time

michaelsafyan: I don't know that setting `ensure_ascii` to False is the right thing to do here. I believe that would create JSON that is not necessarily correct. And my understanding is that `json.loads` (which processes the string shortly thereafter) will perform the correct unescaping.

To double check this, I added a test to cover the JSON case, including one which uses a Unicode emoji character. 

I think I'd prefer to apply a "test driven development" approach on this one... let's come up with a test case that needs this to be changed. Otherwise, I think this is implemented correctly.

michaelsafyan: Marking resolved.

michaelsafyan:
There is still conversation in SemConv PR 2125 (https://github.com/open-telemetry/semantic-conventions/pull/2125/) related to defining the Semantic Conventions for this change. However, I think it would be good to get this in at this time and then circle back if additional changes are required to minimize risk of merge conflicts / bit rotting.

PR #3381, opened 2025-03-21
Check for None result in gRPC (#3380)

### LINKED ISSUE #3380: GRPC traces called twice when result evaluates to false
### Describe your environment

OS: MacOS
Python version: 3.12.9
Package version: commit `a5474c3b290f5fcd7b5e8f9953ae8b2891c374c0`
google-cloud-pubsub 2.28.0

### What happened?

When instrumenting Google Cloud PubSub, I see a lot of log messages `Calling end() on an ended span.`.

This appears to happen whenever no messages are available after a PubSub call:

```python
...
GrpcInstrumentorClient().instrument()

subscriber = pubsub_v1.SubscriberClient()
subscription_path = subscriber.subscription_path(env.pubsub_project_id, env.subscription_id)
with subscriber:
    while True:
        response = subscriber.pull(
            subscription=subscription_path,
            max_messages=1,
            retry=retry.Retry(timeout=env.timeout),
        )
       ...
```

This appears to be due to https://github.com/open-telemetry/opentelemetry-python-contrib/blob/a5474c3b290f5fcd7b5e8f9953ae8b2891c374c0/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py#L168-L169

which ends the span when `result` evaluates falsy.

But then since there was no exception, program flow continues to https://github.com/open-telemetry/opentelemetry-python-contrib/blob/a5474c3b290f5fcd7b5e8f9953ae8b2891c374c0/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py#L122, where it calls `span.end()`.

My interpretation of the `if not result:` is that it's mainly acting as an indication that there was an exception; at any rate, in my simple use case, spans seem to start and end correctly if I change it to `if result is None:`.

### Steps to Reproduce

```python
...
GrpcInstrumentorClient().instrument()

subscriber = pubsub_v1.SubscriberClient()
subscription_path = subscriber.subscription_path(env.pubsub_project_id, env.subscription_id)
with subscriber:
    while True:
        response = subscriber.pull(
            subscription=subscription_path,
            max_messages=1,
            retry=retry.Retry(timeout=env.timeout),
        )
       ...
```
(This relies on an empty queue. I'm not familiar enough with gRPC to create a fully standalone test at this point)

### Expected Result

Only a single `span.end()` call for each span created.

### Actual Result

`span.end()` is called twice when result contains no messages.

### Additional context

_No response_

Yes

kevmodo:
# Description

This is a fix for #3380 that checks explicitly for a `None` result rather than evaluating truthiness. PubSub results with zero messages evaluate to `False`, causing `span.end()` to be called twice.

Fixes #3380 

## Type of change

# How Has This Been Tested?

- [x] Ran Tox
- [x] Ran locally against an empty PubSub queue and no longer received warning messages, but spans correctly started and ended.
- [x] Ran against queue with messages and behavior was as expected (spans correctly started and ended, as before)

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `opentelemetry-instrumentation-grpc` Check for None result in gRPC
+  ([#3380](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3381))
 - `opentelemetry-instrumentation-redis` Add missing entry in doc string for `def _instrument`
   ([#3247](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3247))
 - `opentelemetry-instrumentation-botocore` sns-extension: Change destination name attribute
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3419](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3419))
 - `opentelemetry-instrumentation` don't print duplicated conflict log error message
   ([#3432](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3432))
+- `opentelemetry-instrumentation-grpc` Check for None result in gRPC
+  ([#3380](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3381))
 
 ## Version 1.32.0/0.53b0 (2025-04-10)
 
@@ -78,8 +80,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- `opentelemetry-instrumentation-grpc` Check for None result in gRPC
-  ([#3380](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3381))
 - `opentelemetry-instrumentation-redis` Add missing entry in doc string for `def _instrument`
   ([#3247](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3247))
 - `opentelemetry-instrumentation-botocore` sns-extension: Change destination name attribute
--- a/instrumentation/opentelemetry-instrumentation-grpc/tests/test_client_interceptor.py
+++ b/instrumentation/opentelemetry-instrumentation-grpc/tests/test_client_interceptor.py
@@ -296,7 +296,7 @@ class TestClientProto(TestBase):
                 ),
                 invoker=invoker,
             )
-            assert span_end_mock.call_count == 1
+            self.assertEqual(span_end_mock.call_count, 1)
 
     def test_client_interceptor_trace_context_propagation(
         self,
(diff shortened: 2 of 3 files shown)


kevmodo:
> Thanks for the PR, could you please write a test for this to avoid regressions in the future?

That was an adventure! Added a test that counts `span.end()` invocations.

CHANGELOG.md
```diff
@@ -43,6 +43,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `opentelemetry-instrumentation-grpc` Check for None result in gRPC
```
xrmx, line 46: This need to be moved to the unreleased version

instrumentation/opentelemetry-instrumentation-grpc/tests/test_client_interceptor.py
```diff
@@ -269,6 +269,47 @@ def test_error_stream_stream(self):
...
+                return original_end(*args, **kwargs)
+
+            return wrapper
+
+        def patched_start_span(*args, **kwargs):
+            span = original_start_span(*args, **kwargs)
+            span.end = counting_span_end(span.end)
+            return span
+
+        tracer.start_span = patched_start_span
```
xrmx, line 295: Have you tried patching the end method of Span instead? something like `with mock.patch.object(Span, "end") as span_end_mock:` and then `self.assertEqual(span_end_mock.call_count, 1)`?

kevmodo: I tried originally and it didn't work, but realized there are two `Span`s and I was patching the wrong one. Much nicer now.

Tested with and without the original fix, and behavior is as expected (without, 2 calls; with, 1 call).

instrumentation/opentelemetry-instrumentation-grpc/tests/test_client_interceptor.py
```diff
@@ -269,6 +272,32 @@ def test_error_stream_stream(self):
...
+            interceptor.intercept_unary(
+                request,
+                {},
+                _UnaryClientInfo(
+                    full_method="/GRPCTestServer/SimpleMethod",
+                    timeout=None,
+                ),
+                invoker=invoker,
+            )
+            assert span_end_mock.call_count == 1
```
xrmx, line 299:
```suggestion
            self.assertEqual(span_end_mock.call_count, 1)
```

And you can drop the no-self-use

xrmx (approved):
With the small nit fixed, LGTM

instrumentation/opentelemetry-instrumentation-grpc/tests/test_client_interceptor.py
```diff
@@ -269,6 +272,32 @@ def test_error_stream_stream(self):
             trace.StatusCode.ERROR,
         )
 
+    def test_client_interceptor_falsy_response(
+        self,
+    ):  # pylint: disable=no-self-use
```
xrmx, line 277:
```suggestion
    ):
```

kevmodo: Of course I use the one test doing things that way as a template...

lzchen (approved):

PR #3408, opened 2025-04-04
asyncio: fix duplicate instrumentation

### LINKED ISSUE #3383: aiokafka instrumentation cause memory leak
### Describe your environment

OS: Ubuntu
Python version: (e.g., Python 3.13.1)
Package version: 0.51b0
Aiokafka[lz4] version: 0.12.0

### What happened?

we have a memory leakage caused by asyncio when using AIOKafkaConsumer in our fastapi app, exactly as documented in aiokafka [documentation](https://github.com/aio-libs/aiokafka/tree/29b58dbcacc75a62b430ce5243ac684ccf16a7f8?tab=readme-ov-file#aiokafkaconsumer).
we do think that using getone()  within a while loop instead of __anext__ solves that issue but we want to follow aiokafka best practices.

### Steps to Reproduce

add the following code as part of the fastapi app startup lifespan:
```python
from aiokafka import AIOKafkaConsumer
import asyncio

async def consume():
    consumer = AIOKafkaConsumer(
        'my_topic', 'my_other_topic',
        bootstrap_servers='localhost:9092',
        group_id="my-group")
    # Get cluster layout and join group `my-group`
    await consumer.start()
    try:
        # Consume messages
        async for msg in consumer:
            print("consumed: ", msg.topic, msg.partition, msg.offset,
                  msg.key, msg.value, msg.timestamp)
    finally:
        # Will leave consumer group; perform autocommit if enabled.
        await consumer.stop()

asyncio.run(consume())

```
then trigger the consume once with a message, and the memory will scale up exponentially.

you can check the heap using guppy3 and tracemalloc. the best way is just to measure the memory utilization of the process.

### Expected Result

stable memory utilization when using aiokafka's best practices.

### Actual Result

exponentially increasing memory utilization.

### Additional context

_No response_

No

bourbonkk:
# Description

This PR addresses duplicate instrumentation of Future objects when using AsyncioInstrumentor multiple times on the same Future. Without proper de-duplication, callbacks can continue to accumulate and cause memory leaks. This change introduces a simple attribute check on each Future, coroutine, or function so that once instrumented, no additional callbacks or spans will be attached.

Fixes #3383 aiokafka instrumentation cause memory leak

## Type of change

# How Has This Been Tested?

- [X] Unit Tests:
  - Added a new test file (test_asyncio_duplicate_instrument.py) which repeatedly calls trace_future on the same Future objects.
  - Verified that the callback count does not grow beyond a small constant (1–2 callbacks).
- [X] Manual Verification:
  - Mocked up a simple scenario mimicking aiokafka.GroupCoordinator usage, confirming that repeated instrumentation does not cause exponential callback growth.

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:


xrmx:
@bourbonkk looks like you commited with multiple accounts and one has not signed the CLA

instrumentation/opentelemetry-instrumentation-asyncio/src/opentelemetry/instrumentation/asyncio/instrumentation_state.py
```diff
@@ -0,0 +1,74 @@
...
+"""
+
+import weakref
+
+# A global dictionary to track whether an object has been instrumented.
+# Keys are weak references to avoid preventing garbage collection.
+_instrumented_tasks = {}
+
+
+def _get_weak_key(obj):
```
bourbonkk, line 29: This code was introduced to safely track whether a coroutine, Future, or function has already been instrumented by the OpenTelemetry asyncio instrumentation layer.

In Python, many objects (such as async_generator_asend or built-in functions) do not support weak references or custom attribute assignment.
Attempting to track these objects directly using WeakKeyDictionary or by setting custom attributes can result in runtime TypeErrors or memory leaks.

To avoid these issues, this helper uses a weak-reference-based tracking approach:

Only objects that support weak references are tracked.

Objects that cannot be weak-referenced are gracefully skipped to prevent instrumentation errors.

A global dictionary is used to store weak references to instrumented objects, allowing for duplicate instrumentation prevention while still being memory-safe and garbage-collection-friendly.

This ensures that coroutines, Futures, and thread functions are only instrumented once, without compromising compatibility or stability across Python versions and object types.

instrumentation/opentelemetry-instrumentation-asyncio/src/opentelemetry/instrumentation/asyncio/instrumentation_state.py
```diff
@@ -0,0 +1,74 @@
...
+    if not _is_instrumented(obj):
+        _mark_instrumented(obj)
+        # instrument the object...
+"""
+
+import weakref
+
+# A global dictionary to track whether an object has been instrumented.
+# Keys are weak references to avoid preventing garbage collection.
+_instrumented_tasks = {}
```
yonathan-wolloch-lendbuzz, line 26: Is there a way to cleanup all weak refs from the dictionary after the gc collects them? it seems like this dictionary will grow indefinitely.

bourbonkk: You're absolutely right to be concerned about memory growth. In the original version where we used a regular dict with weakref.ref(obj) as keys, there is no automatic cleanup — dead weak references would remain in the dictionary unless manually removed, which could cause the dictionary to grow indefinitely over time.

To solve this, we’ve switched to using weakref.WeakKeyDictionary. This built-in data structure is specifically designed to handle this scenario. It automatically removes entries when their keys (i.e., the original objects) are garbage collected. So there is no need for manual cleanup, and the dictionary will not grow indefinitely.

By wrapping all insertions in a try/except TypeError, we also ensure that only objects that support weak references are tracked, preventing runtime errors.

yonathan-wolloch-lendbuzz (approved):
Thank you!

yonathan-wolloch-lendbuzz (approved):

bourbonkk:
@xrmx 
I believe I’ve made all the necessary changes in this PR. I’d appreciate it if you could review and let me know if anything else is required.

instrumentation/opentelemetry-instrumentation-asyncio/src/opentelemetry/instrumentation/asyncio/instrumentation_state.py
```diff
@@ -0,0 +1,60 @@
+"""
```
aabmass, line 1: Please add OTel copyright header. We should have a linter for this, sorry!

instrumentation/opentelemetry-instrumentation-asyncio/src/opentelemetry/instrumentation/asyncio/instrumentation_state.py
```diff
@@ -0,0 +1,60 @@
...
+    Args:
+        obj: A coroutine, function, or Future.
+
+    Returns:
+        True if the object has already been marked as instrumented.
+        False if it has not, or if it does not support weak references.
+    """
+    try:
+        return _instrumented_tasks.get(obj, False)
+    except TypeError:
```
aabmass, line 41: IIUC, the lookup will use the object's `__hash__()` which would not work as expected if the object has a custom impl not based on it's identity and would fail the object is unhashable.

I think the use case here wants to track identity instead

instrumentation/opentelemetry-instrumentation-asyncio/src/opentelemetry/instrumentation/asyncio/instrumentation_state.py
```diff
@@ -0,0 +1,60 @@
...
+"""
+
+import weakref
+
+# A global WeakKeyDictionary to track instrumented objects.
+# Entries are automatically removed when the objects are garbage collected.
+_instrumented_tasks = weakref.WeakKeyDictionary()
+
+
+def _is_instrumented(obj) -> bool:
```
aabmass, line 28: If you can give a more specific type for `obj` please do. If it is unbounded or there are a tons of different types being instrumented, just add an explicit `Any`.

We are planning to make typing **opt-out** for any new files.

CHANGELOG.md
```diff
@@ -31,9 +31,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3113](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3113))
 - `opentelemetry-instrumentation-grpc` Fix error when using gprc versions <= 1.50.0 with unix sockets.
   ([[#3393](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3393)])
+- `opentelemetry-instrumentation-asyncio` Fix duplicate instrumentation.
```
xrmx, line 34: Move to unreleased please

instrumentation/opentelemetry-instrumentation-asyncio/src/opentelemetry/instrumentation/asyncio/instrumentation_state.py
```diff
@@ -0,0 +1,60 @@
...
+        obj: A coroutine, function, or Future.
+
+    Returns:
+        True if the object has already been marked as instrumented.
+        False if it has not, or if it does not support weak references.
+    """
+    try:
+        return _instrumented_tasks.get(obj, False)
+    except TypeError:
+        # The object does not support weak references.
```
aabmass, line 42: What case does this actually happen for?

instrumentation/opentelemetry-instrumentation-asyncio/src/opentelemetry/instrumentation/asyncio/instrumentation_state.py
```diff
@@ -0,0 +1,60 @@
...
+
+    Args:
+        obj: A coroutine, function, or Future.
+
+    Returns:
+        True if the object has already been marked as instrumented.
+        False if it has not, or if it does not support weak references.
+    """
+    try:
+        return _instrumented_tasks.get(obj, False)
```
aabmass, line 40: Since we're just storing bools, is `WeakSet` more appropriate?

aabmass (approved):
I don't want to block this since it's an improvement. Please just address comments

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_duplicate_instrument.py
```diff
@@ -0,0 +1,115 @@
...
+        self.loop.close()
+        asyncio.set_event_loop(None)
+        super().tearDown()
+
+    def test_duplicate_instrumentation_of_futures(self):
+        """
+        If instrumentor.trace_future is called multiple times on the same Future,
+        we should NOT see an unbounded accumulation of callbacks.
+        """
+        coordinator = MockGroupCoordinator()
```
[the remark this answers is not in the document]
bourbonkk: That's right, I'll fix it with a simple test code

instrumentation/opentelemetry-instrumentation-asyncio/src/opentelemetry/instrumentation/asyncio/instrumentation_state.py
bourbonkk: In Python 3.12+, some internal types like `async_generator_asend` raise TypeError when weakref is attempted.

I’ve added comments to the code.

bourbonkk: @aabmass 
No, there’s no memory leak. When a TypeError is raised, we skip instrumentation altogether — the object is neither marked nor wrapped. So it's never stored in _instrumented_tasks, and nothing accumulates.

PR #3424, opened 2025-04-14
infra: Automate SHA procedure during releases

emdneto:
# Description

This PR tries to automate the SHA procedure during releases. The rationale is:

1. For non-PR events (e.g., a push), `CORE_REPO_SHA` defaults to "main".

2. For PRs:
* If there’s a label named `release`, use the PR’s head branch (e.g., opentelemetrybot/prepare-release-1.32.0-0.53b0)

* Else, if there’s a label named "patch-release", use the PR’s base branch (e.g., usually release/v1.32.x-0.53bx)

* Otherwise, use "main".

Since only people with write permission can use labels, we are safe here. The automation is basically telling opentelemetrybot to open the release PRs with labels and use the labels to set the correct environment variable for `CORE_REPO_SHA` or `CORE_CONTRIB_REPO_SHA`.

I tested it locally using [act](https://github.com/nektos/act), with the following event and checked tox is using the correct value for installing packages:

```
{
    "pull_request": {
      "labels": [
        {
          "name": "release"
        }
      ],
    "head": {
      "ref": "release/v1.32.x-0.53bx"
    },
    "base": {
      "ref": "main"
    }
    }
  }
```
and:
```
act -P ubuntu-latest=catthehacker/ubuntu:act-latest pull_request -W .github/workflows/misc_0.yml --container-architecture linux/amd64 -e event.json
```

Testing the behaviour from release workflows is hard, so I'm open to suggestions in terms of how testing it.

### DIFF
--- a/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
+++ b/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
@@ -10,7 +10,14 @@ on:
   pull_request:
 
 env:
-  CORE_REPO_SHA: main
+  # Set the SHA to the branch name if the PR has a label 'release' or 'patch-release' otherwise, set it to 'main'
+  # For PRs you can change the inner fallback ('main')
+  # For pushes you change the outer fallback ('main')
+  CORE_REPO_SHA: ${% raw %}${{ github.event_name == 'pull_request' && (
+       contains(github.event.pull_request.labels.*.name, 'release') && github.event.pull_request.head.ref ||
+       contains(github.event.pull_request.labels.*.name, 'patch-release') && github.event.pull_request.base.ref ||
+       'main'
+    ) || 'main' }}{% endraw %}
   CONTRIB_REPO_SHA: main
   PIP_EXISTS_ACTION: w
 
--- a/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
+++ b/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
@@ -13,7 +13,7 @@ env:
   # Set the SHA to the branch name if the PR has a label 'prepare-release' or 'backport' otherwise, set it to 'main'
   # For PRs you can change the inner fallback ('main')
   # For pushes you change the outer fallback ('main')
-  # This logic below is used during releases and depends on having a equivalent branch name in core.
+  # The logic below is used during releases and depends on having an equivalent branch name in the core repo.
   CORE_REPO_SHA: {% raw %}${{ github.event_name == 'pull_request' && (
        contains(github.event.pull_request.labels.*.name, 'prepare-release') && github.event.pull_request.head.ref ||
        contains(github.event.pull_request.labels.*.name, 'backport') && github.event.pull_request.base.ref ||
--- a/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
+++ b/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
@@ -9,6 +9,10 @@ on:
     - 'release/*'
   pull_request:
 
+concurrency:
+  group: ${% raw %}{{ github.workflow }}-${{ github.head_ref || github.run_id }}{% endraw %}
+  cancel-in-progress: true
+
 env:
   # Set the SHA to the branch name if the PR has a label 'prepare-release' or 'backport' otherwise, set it to 'main'
   # For PRs you can change the inner fallback ('main')
@@ -28,6 +32,7 @@ jobs:
   {{ job_data.name }}:
     name: {{ job_data.ui_name }}
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
       - name: Checkout repo @ SHA - ${% raw %}{{ github.sha }}{% endraw %}
         uses: actions/checkout@v4
--- a/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
+++ b/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
@@ -13,7 +13,7 @@ env:
   # Set the SHA to the branch name if the PR has a label 'release' or 'patch-release' otherwise, set it to 'main'
   # For PRs you can change the inner fallback ('main')
   # For pushes you change the outer fallback ('main')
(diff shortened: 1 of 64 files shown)


.github/workflows/package-prepare-release.yml
```diff
@@ -195,4 +196,5 @@ jobs:
           gh pr create --title "$message" \
                        --body "$body" \
                        --head $branch \
-                       --base main
+                       --base main \
+                       --label release
```
aabmass, line 200: Will this create the label if it doesn't exist in the repo yet?

aabmass (approved):
Thanks! Makes sense to me but would like someone else to double check it

.github/workflows/package-prepare-release.yml
emdneto: Unforunately no. We need to create them

xrmx: Labels created

xrmx:
The manual process is something like this atm:
- run the prepare workflow
- update the sha of the other repo to the same branch (opentelemetrybot/prepare-release...)
- Release
- Update sha reference to the release/vX.Y.Z branch

With automation it should be something like:
- run the prepare worflow
- update the sha of the other repo to the release/vX.Y.Z branch
- add label and have the workflow override the value set in the workflows
- release

.github/workflows/package-prepare-patch-release.yml
```diff
@@ -118,4 +118,5 @@ jobs:
           gh pr create --title "[$GITHUB_REF_NAME] $message" \
                        --body "$message." \
                        --head $branch \
-                       --base $GITHUB_REF_NAME
+                       --base $GITHUB_REF_NAME \
+                       --label release
```
lzchen, line 122: Don't we have a "patch release" label? Not too sure if we need both "release" and "patch release".

.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
```diff
@@ -10,7 +10,14 @@ on:
   pull_request:
 
 env:
-  CORE_REPO_SHA: main
+  # Set the SHA to the branch name if the PR has a label 'release' or 'patch-release' otherwise, set it to 'main'
+  # For PRs you can change the inner fallback ('main')
+  # For pushes you change the outer fallback ('main')
+  CORE_REPO_SHA: {% raw %}${{ github.event_name == 'pull_request' && (
+       contains(github.event.pull_request.labels.*.name, 'release') && github.event.pull_request.head.ref ||
```
lzchen, line 17: Doesn't `github.event.pull_request.head.ref` take the ref of the current pull request, which would be the contrib ref not the core ref? Also, we need the ref of the equivalent core repo PR right?

emdneto: That's right. Let's take the last release as example:

1. [First PR](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3413) = [main](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main) (base.ref) <- [opentelemetrybot/update-version-to-1.33.0.dev-0.54b0.dev](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/opentelemetrybot/update-version-to-1.33.0.dev-0.54b0.dev) (head.ref)
2. [Second PR](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3412) = [release/v1.32.x-0.53bx](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/release/v1.32.x-0.53bx) (base.ref) <- [opentelemetrybot/prepare-release-1.32.0-0.53b0](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/opentelemetrybot/prepare-release-1.32.0-0.53b0) (head.ref)

But the thing is: we are already relying on the branch name to make SHA work and CI pass, right? So, we are just automating it based on labels. For opened PRs with `release` use the head.ref as branch name which will have a correspondent one in -core. At the end, the only necessary action will be a re-run in workflows if the branch doesn't exist in -core (or vice-versa)

.github/workflows/package-prepare-patch-release.yml
emdneto: The patch-release purpose is to be used in backport's PR's, like this:
https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3425. If you want we can use a more appropriate name like "backport" or something else. 

The Patch Release PR itself needs to point to head.ref because there will be an equivalent one in -core, like this https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3426

xrmx: So i think the code is correct but taking a look at this again I agree that the labels are confusing. I think better labels would be `prepare-release` and `backport`. And since the backport workflow will likely fail for changelog conflicts we may start simple and just do the one for the prepare-release/prepare-patch-release pr with the `prepare-release` label?

xrmx: Or alternatively go generic with something like `ci-pull-request-head` and `ci-pull-request-base`

xrmx: I think the workflow is fine it's just that there will be inevitable conflicts for commits updating changelog until we move to generate the changelog from change fragments.

.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
lzchen: I see. This is dependent on the fact that we create an equivalent branch name in core, hence the possible re-run.

lzchen: Could you put a comment stating the above ^? Just incase we change the workflows in the future where the branch names created in core/contrib don't match anymore.

.github/workflows/package-prepare-patch-release.yml
lzchen: I'm okay with getting this PR in for regular releases for now :)

lzchen (approved):

.github/workflows/generate_workflows_lib/src/generate_workflows_lib/lint.yml.j2
emdneto: > I see. This is dependent on the fact that we create an equivalent branch name in core, hence the possible re-run.

Exactly. I'll add a comment and let you know when it's ready to merge.

lzchen (approved):

Issue #3220, opened 2025-01-30
Add `explicit_bucket_boundaries_advisory` to stable semconv duration histograms for HTTP instrumentations


emdneto:
At this moment, we have an issue with all instrumentation that supports duration metrics based on the stable semantic convention.  When using the stable semantic convention opt-in feature, duration histogram data points are recorded in seconds, but histogram buckets are still defined in ms, which leads to a wrong calculation and presentation of percentiles.

The plan is: with the merge of https://github.com/open-telemetry/opentelemetry-python/pull/4361, we can move forward and add the advisory `explicit_bucket_boundaries_advisory` parameter for duration histograms of already migrated HTTP instrumentation that uses the stable semantic convention. 

```
explicit_bucket_boundaries_advisory = [0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10]
```

[HTTP_SERVER_REQUEST_DURATION](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-python-contrib+HTTP_SERVER_REQUEST_DURATION&type=code)
[HTTP_CLIENT_REQUEST_DURATION](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-python-contrib+HTTP_CLIENT_REQUEST_DURATION&type=code)

### Instrumentations

- [x] opentelemetry-instrumentation-flask https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3523
- [x] opentelemetry-instrumentation-django https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3524
- [x] opentelemetry-instrumentation-falcon https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3525
- [x] opentelemetry-instrumentation-wsgi https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3527
- [x] opentelemetry-instrumentation-asgi https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3526
- [x] opentelemetry-instrumentation-urllib https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3519
- [x] opentelemetry-instrumentation-requests https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3464
- [x] opentelemetry-instrumentation-urllib3 https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3518

[...]

PR #3464, opened 2025-05-05
requests: support explicit_bucket_boundaries_advisory in duration metrics

emdneto:
# Description

Refs https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3220

### DIFF
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
@@ -47,6 +47,43 @@ from opentelemetry.semconv.attributes.user_agent_attributes import (
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.trace.status import Status, StatusCode
 
+# Values defined in milliseconds
+_DURATION_HISTOGRAM_OLD_EXPLICIT_BOUNDS = (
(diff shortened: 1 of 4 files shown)


opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -47,6 +47,43 @@
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.trace.status import Status, StatusCode
 
+# Values defined in milliseconds
+DURATION_HISTOGRAM_EXPLICIT_BOUNDS_OLD = (
```
xrmx, line 51: Should these be more like `HTTP_REQUEST_DURATION_EXPLICIT_BOUNDS_OLD` instead to keep the info they are for http request duration? I've dropped the histogram because otherwise it's a bit too long :sweat:

emdneto: Makes sense to me. Will update

emdneto: Done, changed to `HTTP_DURATION_HISTOGRAM_BUCKETS_OLD`

xrmx (approved):

PR #3481, opened 2025-05-09
500Ensure spans complete early on stream closure for Bedrock Streaming APIs

SrdjanLL:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/bedrock.py
```diff
@@ -642,7 +644,8 @@ def stream_done_callback(response):
                     self._converse_on_success(
                         span, response, instrumentor_context, capture_content
                     )
-                    span.end()
+                    if span.is_recording():
```
xrmx, line 647: This is to avoid a double close?

CHANGELOG.md
```diff
@@ -35,7 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3447](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3447))
 - `opentelemetry-instrumentation-botocore` Capture server attributes for botocore API calls
   ([#3448](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3448))
-
+- `opentelemetry-instrumentation-botocore` Ensure spans end on early stream closure for Bedrock Streaming APIs
```
xrmx, line 38: Need to rebase on main, 1.33.0 was released earlier today

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/bedrock.py
```diff
@@ -499,18 +499,20 @@ def _converse_on_success(
...
-        # this path is used by streaming apis, in that case we are already out of the span
-        # context so need to add the span context manually
-        span_ctx = span.get_span_context()
-        event_logger.emit(
-            choice.to_choice_event(
-                trace_id=span_ctx.trace_id,
-                span_id=span_ctx.span_id,
-                trace_flags=span_ctx.trace_flags,
+        # In case of an early stream closure, the result may not contain outputs
+        if "output" in result and "message" in result["output"]:
```
xrmx, line 503: Can we move this check to a small helper named something like `_stream_has_{output,content}` that we can also reuse later?

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/bedrock.py
SrdjanLL: Yeah - guess it would only throw a warning here for double close, but thought it's better to keep the noise down.

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/bedrock.py
SrdjanLL: Good shout - added that in!

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/bedrock.py
xrmx: I'm not sure this is correct though,  `is_recording` is used for checking if the span is sampled and not if it's already closed. So there can be cases where we don't end unsampled spans. I think we may need to do the accounting ourselves.

SrdjanLL: Ahh, good catch. I looked around the other instrumentation examples and couldn't find any where spans are ended only if they haven't been already - do you happen to have an example at hand?

Not finding the examples, made me wonder if we need to be this cautious at all. After reading through the api docs, I found the implementations are free to to either ignore or raise on double closure (see [here](https://opentelemetry-python.readthedocs.io/en/latest/api/trace.span.html#opentelemetry.trace.span.Span.end)) with the default here being a warning log. 
Do you think it's worth dropping this check in the first place and leave the handling to the specific use cases down the stream (i.e if you're trying to close the span twice, maybe something is up with your code)? I can also confirm that double closure doesn't really change the behaviour of the tests, but as extra precaution we can add assertions to the number of finished spans to capture any regressions.

xrmx: I think the easieast way would be to explicitly pass a boolean to `_stream_done_callback` and act accordingly. In the end the check is only to avoid logs spam.

SrdjanLL: I've added that in - for both `_stream_done_callback` and `_stream_error_callback` to be consistent. Let me know what you think.

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/bedrock.py
```diff
@@ -1004,3 +1018,9 @@ def on_error(
                 duration,
                 attributes=metrics_attributes,
             )
+
+    def _stream_has_output(self, response_body: dict[str, Any]):
+        return "output" in response_body
+
+    def _stream_has_content(self, response_body: dict[str, Any]):
+        return "message" in response_body["output"]
```
xrmx, line 1026:
```suggestion
    def _stream_has_output_content(self, response_body: dict[str, Any]):
        return "output" in response_body and "message" in response_body["output"]
```

What do you think on merging the two functions since the caller don't care much on having any intermediate ?

SrdjanLL: Makes sense to me! Should've done it in the first place, but took your previous suggestion too literal 😄

xrmx (approved):
Thanks!

CHANGELOG.md
```diff
@@ -35,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-instrumentation-fastapi`: Drop support for FastAPI versions earlier than `0.92`
   ([#3012](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3012))
 
+
```
xrmx, line 40:
```suggestion
```

Issue #3475, opened 2025-05-07
Stop using SpanAttributes


xrmx:
A lot of code imports semantic conventions attributes from `opentelemetry.semconv.trace.SpanAttributes` instead of the preferred `opentelemetry.semconv.attributes` and `opentelemetry.semconv.incubating`. 

This issue it is only about removing usage of `SpanAttributes` while not changing the output of the telemetry. If some attributes (e.g. in messaging) are not available anymore just keep using the one from `SpanAttributes`.

Here's the list of of packages found grepping `SpanAttributes` on a recent checkout. Some packages may be more easy than others to convert since there may just be a couple of instances to fix.

People willing to help with this, state in the issue on what package you are going to work and you have 1 week to create a PR, after that we'll consider the package available to others.

The PR message should contains `Refs #3475` instead of `Fixes #3475`.

Updated of instrumentation specific tests in tests/opentelemetry-docker-tests should probably go on par with the conversion of the instrumentation package.

- [x] exporter/opentelemetry-exporter-richconsole https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3501
- [x] instrumentation/opentelemetry-instrumentation-aio-pika
- [x] instrumentation/opentelemetry-instrumentation-aiohttp-client https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3502
- [x] instrumentation/opentelemetry-instrumentation-aiohttp-server https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3504
- [x] instrumentation/opentelemetry-instrumentation-aiopg
- [x] instrumentation/opentelemetry-instrumentation-asgi
- [x] instrumentation/opentelemetry-instrumentation-asyncpg https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3505
- [x] instrumentation/opentelemetry-instrumentation-aws-lambda
- [x] instrumentation/opentelemetry-instrumentation-boto
- [x] instrumentation/opentelemetry-instrumentation-boto3sqs
- [x] instrumentation/opentelemetry-instrumentation-botocore
- [x] instrumentation/opentelemetry-instrumentation-cassandra
- [x] instrumentation/opentelemetry-instrumentation-celery
- [x] instrumentation/opentelemetry-instrumentation-confluent-kafka
- [x] instrumentation/opentelemetry-instrumentation-dbapi
- [x] instrumentation/opentelemetry-instrumentation-django
[...]

PR #3492, opened 2025-05-10
refactor: don't use deprecated `SpanAttributes` in confluent  kafka

zhengkezhou1:
# Description

Stop using `SpanAttributes` in kafka.
Part of  #3475 

## Type of change

# How Has This Been Tested?

Exist test

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/utils.py
+++ b/instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/utils.py
@@ -3,9 +3,17 @@ from typing import List, Optional
 
 from opentelemetry import context, propagate
 from opentelemetry.propagators import textmap
+from opentelemetry.semconv._incubating.attributes.messaging_attributes import (
+    MESSAGING_DESTINATION_NAME,
+    MESSAGING_DESTINATION_TEMPORARY,
+    MESSAGING_KAFKA_DESTINATION_PARTITION,
+    MESSAGING_MESSAGE_ID,
+    MESSAGING_OPERATION,
+    MESSAGING_SYSTEM,
+    MessagingOperationTypeValues,
+)
 from opentelemetry.semconv.trace import (
     MessagingDestinationKindValues,
-    MessagingOperationValues,
     SpanAttributes,
 )
 from opentelemetry.trace import Link, SpanKind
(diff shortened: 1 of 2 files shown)


instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/utils.py
```diff
@@ -114,32 +118,25 @@ def _enrich_span(
...
+    operation: Optional[MessagingOperationTypeValues] = None,
 ):
     if not span.is_recording():
         return
 
-    span.set_attribute(SpanAttributes.MESSAGING_SYSTEM, "kafka")
-    span.set_attribute(SpanAttributes.MESSAGING_DESTINATION, topic)
-
+    span.set_attribute(MESSAGING_SYSTEM, "kafka")
+    span.set_attribute(MESSAGING_DESTINATION_NAME, topic)
```
xrmx, line 127: This task is about updating the import, not changing the semantic conventions so the attributes set should have the same value but imported from another path. If the attribute is not available we must keep using the one from SpanAttributes until we move to a newer semantic convention.

zhengkezhou1: Got it! take back deleted attribute value now.

xrmx: Yeah but still this adds changes `messaging.destination` to `messaging.destination.name` which we don't want to do.

zhengkezhou1: OK, revert it.

instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/utils.py
```diff
@@ -114,32 +121,29 @@ def _enrich_span(
...
 ):
     if not span.is_recording():
         return
 
-    span.set_attribute(SpanAttributes.MESSAGING_SYSTEM, "kafka")
+    span.set_attribute(MESSAGING_SYSTEM, "kafka")
     span.set_attribute(SpanAttributes.MESSAGING_DESTINATION, topic)
-
     if partition is not None:
-        span.set_attribute(SpanAttributes.MESSAGING_KAFKA_PARTITION, partition)
```
xrmx, line 126: You changed the attribute here, it's not the only one and I won't add a comment for every one you changed.

PR #3498, opened 2025-05-13
Add tornado WebSocketHandler instrumentation support

### LINKED ISSUE #2761: Broken Tornado Intrumentation for Websocket
### Describe your environment

OS: Ubuntu 22.04
Python version: 3.10.12
Package version: 0.47b0

### What happened?

The `WebSocketHandler.finish` method, called after the WebSocket upgrade, results in lost context at https://github.com/tornadoweb/tornado/blob/1f8b2d78fee5e5a15e686815e0285d4a3c831a2c/tornado/websocket.py#L929

### Steps to Reproduce

```python
from tornado import websocket, web, ioloop

class WebsocketHandler(websocket.WebSocketHandler):
    def open(self):
        print("WebSocket opened")

    def on_message(self, message):
        self.write_message(u"You said: " + message)

    def on_close(self):
        print("WebSocket closed")


app = web.Application([
    (r'/ws', WebsocketHandler),
])

if __name__ == '__main__':
    app.listen(8888)
    ioloop.IOLoop.instance().start()
```

```shell
opentelemetry-instrument \
    --traces_exporter console \
    --metrics_exporter console \
    --logs_exporter console \
    --service_name cmc-front-end \
    python \
    main.py
```

```javascript
const ws = new WebSocket("ws://localhost:8888/ws");
ws.onopen = function() {
   ws.send("Hello, world");
};
ws.onmessage = function (evt) {
    console.log(evt)
};
```

```shell
deno run main.js
```

### Expected Result

Context to flow from the protocol upgrade to the message handling.

### Actual Result

Context is lost after the initial phase is finished.

### Additional context

_No response_

Yes

devmonkey22:
# Description

Add tornado WebSocketHandler instrumentation support.  Previously, the tornado instrumentation would improperly handle tracing and metrics that would end prematurely before the websocket connection was really closed (it would end when the HTTP connection was upgraded).

Fixes #2761

## Type of change

# How Has This Been Tested?

Added `TestTornadoInstrumentation.test_websockethandler` test which makes a websocket connection and ensures the traces are instrumented properly.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
```diff
@@ -394,13 +406,21 @@ def _prepare(
 
 
 def _on_finish(tracer, server_histograms, func, handler, args, kwargs):
-    response = func(*args, **kwargs)
-
-    _record_on_finish_metrics(server_histograms, handler)
+    try:
```
xrmx, line 409: Not sure we should catch user code exception and change code the behavior, or if we do we should re-raise later.

devmonkey22: Its just a finally block to ensure that we finish the trace span and metrics, even if the user code raises an exception.  It doesn't catch, so it shouldn't need to re-raise?

xrmx: Ah, overlooked that. Sorry for the noise!

xrmx (approved):

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
```diff
@@ -405,13 +417,21 @@ def _prepare(
...
+        _finish_span(tracer, handler)
 
-    _finish_span(tracer, handler)
 
-    return response
+def _websockethandler_on_close(
+    tracer, server_histograms, func, handler, args, kwargs
+):
+    try:
+        func()
```
emdneto, line 431: Don't we need to pass args and kwargs here? not sure if it's expected that though

emdneto (changes requested):
Tests are failing. Also, there are some lint fixes to do for ruff

instrumentation/opentelemetry-instrumentation-tornado/tests/test_instrumentation.py
```diff
@@ -454,6 +456,54 @@ def test_handler_on_finish(self):
...
+        close_span, msg_span, req_span = spans
+
+        self.assertEqual(req_span.name, "GET /echo_socket")
+        self.assertEqual(req_span.context.trace_id, msg_span.context.trace_id)
+        self.assertIsNone(req_span.parent)
+        self.assertEqual(req_span.kind, SpanKind.SERVER)
+        self.assertSpanHasAttributes(
+            req_span,
+            {
+                SpanAttributes.HTTP_METHOD: "GET",
```
emdneto: I think we are using the `_incubating` attributes instead of deprecated SpanAttributes.

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
devmonkey22: [WebSocketHandler.on_close](https://github.com/tornadoweb/tornado/blob/a800ee41055c2ed1438275cecadd1f45857f6b2c/tornado/websocket.py#L483) doesn't take any arguments.  I was keeping the patched handler signatures (like `_on_finish`) the same by accepting `args` and `kwargs` though as I believe required by `wrapt`/`_wrap`

instrumentation/opentelemetry-instrumentation-tornado/tests/test_instrumentation.py
devmonkey22: Yep, I see those were updated recently in #3582.  I'll update

emdneto (approved):

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
emdneto: Gotcha

PR #3514, opened 2025-05-16
Update metric support table to include opentelemetry-instrumentation-aiohttp-server

hectorhdzg:


instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/package.py
```diff
@@ -14,3 +14,7 @@
 
 
 _instruments = ("aiohttp ~= 3.0",)
+
+_supports_metrics = True
+
+_semconv_status = "migration"
```
emdneto, line 20:
```suggestion
```

emdneto (approved):

xrmx (approved):

PR #3456, opened 2025-04-29
starlette: Remove maximum version constraint

MattiasDC:
# Description

Remove maximum version constraint on starlette, by fixing unit tests. Fixed some warnings in unit tests.

Problem was that the environment was patched only during the `setUp` execution, and not during the tests themselves. By putting the patch decorator on the class itself, the environment variables are patched during the test class, as is done in another place in the same file.

This PR also includes some small changes to reduce the amount of warnings encountered in the unit tests.
- Get rid of deprecated calls to the `route` and `route_websocket` decorators
- As `StarletteInstrumentor._uninstrument` can clear all instrumented apps, use a `discard` i.s.o. `remove` in the deletor of `_InstrumentedStarlette` to prevent a `KeyError` being raised

Fixes #3317

## Type of change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?
All existing unit tests pass with the newest version of starlette (0.46.2)

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [ ] Followed the style guidelines of this project
- [x] Changelogs have been updated
- [ ] Unit tests have been added
- [ ] Documentation has been updated

PR #3529, opened 2025-05-22
Remove starlette version ceiling

### LINKED ISSUE #3317: Starlette instrumentation tests fail recent versions `starlette >= 0.15.0`
_Originally posted by @aabmass in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3304#discussion_r1972595732_
            
Only really ancient versions of starlette seem to be working with `opentelemetry-insturmentation-starlette`. The tests fail if you increase the version range because of an issue with capturing custom headers. See thread for more context.

anuraaga:
# Description

Removes the starlette version ceiling which is currently set to a very ancient version of starlette. I think this is fairly sane to do since the instrumentation only uses starlette's public API (`add_middleware`) and should be relatively stable.

There was test failure when bumping but it was an issue with the test code itself, not the instrumentation, so I fixed it.

Fixes #3317

## Type of change

# How Has This Been Tested?

- [X] Unit tests of latest and oldest

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/__init__.py
@@ -239,7 +239,7 @@ class StarletteInstrumentor(BaseInstrumentor):
             meter_provider,
             schema_url="https://opentelemetry.io/schemas/1.11.0",
         )
-        if not getattr(app, "is_instrumented_by_opentelemetry", False):
+        if not getattr(app, "_is_instrumented_by_opentelemetry", False):
             app.add_middleware(
                 OpenTelemetryMiddleware,
                 excluded_urls=_excluded_urls,
@@ -251,7 +251,7 @@ class StarletteInstrumentor(BaseInstrumentor):
                 tracer=tracer,
                 meter=meter,
             )
-            app.is_instrumented_by_opentelemetry = True
+            app._is_instrumented_by_opentelemetry = True
 
             # adding apps to set for uninstrumenting
             if app not in _InstrumentedStarlette._instrumented_starlette_apps:
@@ -332,7 +332,7 @@ class _InstrumentedStarlette(applications.Starlette):
         _InstrumentedStarlette._instrumented_starlette_apps.add(self)
 
     def __del__(self):
-        _InstrumentedStarlette._instrumented_starlette_apps.remove(self)
+        _InstrumentedStarlette._instrumented_starlette_apps.discard(self)
 
 
 def _get_route_details(scope: dict[str, Any]) -> str | None:
--- a/instrumentation/opentelemetry-instrumentation-starlette/test-requirements.latest.txt
+++ b/instrumentation/opentelemetry-instrumentation-starlette/test-requirements.latest.txt
@@ -17,31 +17,31 @@
     #   opentelemetry-instrumentation-asgi
     #   opentelemetry-instrumentation-starlette
 anyio==4.5.2 ; python_full_version < '3.9'
-    # via httpx
-anyio==4.8.0 ; python_full_version >= '3.9'
-    # via httpx
+    # via
+    #   httpx
+    #   starlette
+anyio==4.9.0 ; python_full_version >= '3.9'
+    # via
+    #   httpx
+    #   starlette
 asgiref==3.8.1
     # via opentelemetry-instrumentation-asgi
-certifi==2025.1.31
+certifi==2025.4.26
     # via
     #   httpcore
     #   httpx
     #   requests
-charset-normalizer==3.4.1
+charset-normalizer==3.4.2
     # via requests
 colorama==0.4.6 ; sys_platform == 'win32'
     # via pytest
-deprecated==1.2.18
-    # via
-    #   opentelemetry-api
-    #   opentelemetry-semantic-conventions
-exceptiongroup==1.2.2 ; python_full_version < '3.11'
+exceptiongroup==1.3.0 ; python_full_version < '3.11'
     # via
     #   anyio
     #   pytest
-h11==0.14.0
+h11==0.16.0
     # via httpcore
-httpcore==1.0.7
+httpcore==1.0.9
     # via httpx
 httpx==0.28.1
     # via -r instrumentation/opentelemetry-instrumentation-starlette/test-requirements.in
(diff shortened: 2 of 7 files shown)


instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/__init__.py
```diff
@@ -239,7 +239,7 @@ def instrument_app(
             meter_provider,
             schema_url="https://opentelemetry.io/schemas/1.11.0",
         )
-        if not getattr(app, "is_instrumented_by_opentelemetry", False):
+        if not getattr(app, "_is_instrumented_by_opentelemetry", False):
```
anuraaga, line 243: This needs to match the setting at https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3529/files#diff-a4901479b7a21ad4f5d45a7f3e47b2ada3f9d8ae4ae7f9e68c2001e8b8067611L330

I think the problem becomes double telemetry, but presumably no one uses such an old version of starlette to allow the auto-instrumentation to kick in to trigger it.

instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/__init__.py
```diff
@@ -331,9 +332,6 @@ def __init__(self, *args: Any, **kwargs: Any):
         # adding apps to set for uninstrumenting
         _InstrumentedStarlette._instrumented_starlette_apps.add(self)
 
-    def __del__(self):
-        _InstrumentedStarlette._instrumented_starlette_apps.remove(self)
```
anuraaga, line 335: This would fail if the app isn't in the set, which would happen in the `.clear()` call in `StarletteInstrumentation.uninstrument()`. And actually, `__del__` could only ever be called after such a call because the set held a strong reference, preventing `__del__` from being called. I changed to `WeakSet` which should fix it, and we don't need `__del__` since it's automatically cleaned when the object goes away

instrumentation/opentelemetry-instrumentation-starlette/tests/test_starlette_instrumentation.py
```diff
@@ -667,15 +667,17 @@ async def _(websocket: WebSocket) -> None:
...
-    @patch.dict(
-        "os.environ",
-        {
-            OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SANITIZE_FIELDS: ".*my-secret.*",
-            OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST: "Custom-Test-Header-1,Custom-Test-Header-2,Custom-Test-Header-3,Regex-Test-Header-.*,Regex-Invalid-Test-Header-.*,.*my-secret.*",
-            OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_RESPONSE: "Custom-Test-Header-1,Custom-Test-Header-2,Custom-Test-Header-3,my-custom-regex-header-.*,invalid-regex-header-.*,.*my-secret.*",
-        },
-    )
     def setUp(self) -> None:
+        patcher = patch.dict(
```
anuraaga, line 671: Previously patch was decorating `setUp` - it was active only for the duration of `setUp`, not the entire test case. Recent starlette seems to initialize middleware lazily so the patch needs to be active during the test case for it. I switched to this pattern that activates it both for `setUp` (for older eager starlette) and the test cases (for newer lazy starlette). Note, moving the decorator to the class level would loose the former.

But it seems CI is showing failures still that I can't repro locally will check it out more

xrmx:
@anuraaga this is ready to review since you fixed the testcase right?

instrumentation/opentelemetry-instrumentation-starlette/tests/test_starlette_instrumentation.py
```diff
@@ -230,7 +230,8 @@ def test_basic_post_request_metric_success(self):
                         dict(point.attributes), expected_duration_attributes
                     )
                     if metric.name == "http.server.duration":
-                        self.assertAlmostEqual(duration, point.sum, delta=30)
```
anuraaga, line 233: latest had trouble with this assertion because it initializes middleware lazily, so the `duration` ends up being way longer than the actual request's duration. I think in principal, it's only safe to check the request duration is not larger than the test duration rather than comparing the actual values so I changed the assertions. 

I wanted to use `self.assertGreater(point.sum, 0)` but the old ms metric is `round`ed and easily `0` on a fastish machine - not sure if the rounding is actually required for it but that would be a change in the ASGI middleware

anuraaga:
FYI @codefromthecrypt

anuraaga:
> @anuraaga this is ready to review since you fixed the testcase right?

Yup should be ok now

anuraaga:
Oh no, I didn't check existing PRs before making this one and now see @mattiasdc's #3456 is quite similar. I found the tests seem to pass fine for the current baseline version so didn't raise it here which is the main difference. Feel free to reject this PR as a duplicate and get that one in instead though, then I can send another one with just the `WeakSet` and other fixes which should still apply.

instrumentation/opentelemetry-instrumentation-starlette/test-requirements.latest.txt
emdneto, line 1: Funny that running the same command I got different result for this file

emdneto (commented):
@anuraaga Thank you for the fix! To be fair with the already opened PR https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3456 I would suggest that we proceed with #3456 and later you can change your PR with the changes you are proposing here (weakset and the fix for the already instrumented flag)

instrumentation/opentelemetry-instrumentation-starlette/test-requirements.latest.txt
anuraaga: Yeah to get it to update, I removed the file first and ran the command

emdneto:
@anuraaga #3456 was merged.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3012](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3012))
 - `opentelemetry-instrumentation-starlette` Remove max version constraint on starlette
   ([#3456](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3456))
+- `opentelemetry-instrumentation-starlette` Fix memory leak and double middleware
+  ([#3456](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3529))
```
jomcgi, line 23: nit: links to different PR
```suggestion
  ([#3529](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3529))
```

jomcgi (approved):

emdneto (approved):

Issue #3476, opened 2025-05-08
Starlette instrumentation can have ordering issues


anuraaga:
Currently, starlette instrumentation uses a class override pattern which can have ordering issues that cause autoinstrumentation to not apply.

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/__init__.py#L287

In practice, I see most users manually instrument the app because of this, for example here

https://github.com/YunosukeY/otel-and-a2a-sample/blob/main/server.py#L47

I think it should be possible to switch to wrapping the `__init__` method without replacement which can usually solve such ordering issues.

/cc @xrmx was interested in this (I hope :) )

codefromthecrypt:
Basically, we can't magically drop in "opentelemetry-instrumentation-starlette" and have `opentelemetry-instrument` take care of instrumenting A2A due to this glitch in how starlette instrumentation works.

cc @whitlockjc even though the code rag referenced from @YunosukeY used the former samples, still applies to the official SDK that's brewing.

cc @Kludex likely anything else that uses starlette, like MCP or FastA2A, right?

Kludex:
> cc @Kludex likely applies to anything else that uses starlette, like MCP or FastA2A, right?

Yes, it does.

---

I didn't get the "ordering issues" part tho, when does it occur?

xrmx:
> I didn't get the "ordering issues" part tho, when does it occur?

@Kludex the httpx instrumentor had a similar issue where loading the openai library in the openai instrumentation before the httpx instrumentor was loaded made the httpx client substitution not patch the openai subclasses and so no http spans for openai client calls. We solved that doing lazy loading of both openai library in openai instrumentation and lazy patching of the httpx library in the httpx instrumentor (f6b68d0c024cf40d15c08062a18bf70ea73847e6) to avoid top level import of patched libraries.
This will also help making instrumentors loadable when the instrumented library is not available.

anuraaga:
Sorry for the noise, I realized I was misunderstanding things. The main problem was just that the instrumentation targets a very old version of starlette

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-starlette/pyproject.toml#L37

[...]

PR #3530, opened 2025-05-22
starlette: use wrapt for patching instead of class replacement

SrdjanLL:
# Description

Moving starlette instrumentation to patch `Starlette.__init__` method using `wrapt` instead of doing the class replacement to prevent potential ordering issues. 

Left the behaviour of manual instrumentation methods of `StarletteInstrumentor` to work as it did before.

The core issue where this surfaced (https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3476)  has already been closed, but this was still a relevant improvement

## Type of change

# How Has This Been Tested?
- [x] Unit tests

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

- [x] Unit tests have been ~added~ updated

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/__init__.py
@@ -176,10 +176,12 @@ API
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Collection, cast
+from functools import partial
+from typing import TYPE_CHECKING, Any, Collection
 
 from starlette import applications
 from starlette.routing import Match
+from wrapt import wrap_function_wrapper
 
 from opentelemetry.instrumentation.asgi import OpenTelemetryMiddleware
 from opentelemetry.instrumentation.asgi.types import (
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/__init__.py
```diff
@@ -176,10 +176,12 @@ def client_response_hook(span: Span, scope: dict[str, Any], message: dict[str, A
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Collection, cast
+from functools import partial
+from typing import TYPE_CHECKING, Any, Collection
 
 from starlette import applications
 from starlette.routing import Match
```
xrmx, line 183: So if we don't have an ordering issue with other instrumentations then the value of switching to wrapt would be on removing these top level imports and push them at a later time or move them under type checking. Do you see this achievable?

codefromthecrypt:
Would love a fix to the issue, if it is this, so that we don't need to manually do `StarletteInstrumentor.instrument_app(app)`. There are an increasing number of applications using starlette due to the prevalence of JSON-RPC in genai agents, such as Google A2A. Right now, you have to manually instrument the app in order to use "zero code instrumentation" which doesn't make any sense. cc @aabmass

PR #3545, opened 2025-05-27
Allow reraising the root exception if instrumentation fails

### LINKED ISSUE #3550: Allow failing fast for auto-instrumentation problems

If there's an exception in an auto-instrumentation, we currently log the exception and continue. I would like an option to fail faster and immediately re-raise the exception to the caller.

Here's where we currently swallow any instrumentation exceptions:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/7bd0895d01d14d6ee290449d4a2b385b2ae96f03/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py#L121-L135

With this, there's no way to say, fail startup of my service if there's an instrumentation problem.

I'd like to be able to re-raise instrumentation exceptions and fail startup if there's a failure in this case:

```python
from opentelemetry import autoinstrumentation

autoinstrumentation.initialize()
```

_No response_

### Additional Context

_No response_

Yes

phillipuniverse:
# Description

I would rather completely fail startup in my services if instrumentation fails for whatever reason instead of just logging an exception and continuing.

Use case:

```python
from opentelemetry import autoinstrumentation

autoinstrumentation.initialize(swallow_exceptions=False)
```

Fixes #3550 

## Type of change

# How Has This Been Tested?

New unit test

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


tammy-baylis-swi:
Hi thanks for your input and contribution.

Generally speaking, the opt-in does go against the Otel [specification on error handling](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md) and the [mission](https://opentelemetry.io/community/mission/). We avoid disrupting regular service operation if instrumentation fails and, whenever possible, [record errors as telemetry](https://opentelemetry.io/docs/specs/semconv/general/recording-errors/).

Please could you share more details about your use case? e.g. production requirements, user onboarding friction, what docs are being followed

phillipuniverse:
@tammy-baylis-swi thanks for the context. My use case is as follows:

- I centralize my organization's otel support in a common Python library, something like `mycompany-common-python`. We apply various standards on top of stock otel like setting versions, making some instrumentation customizations, etc
- All of our microservices depend on `mycompany-common-python` and transitively pulls in all the otel dependencies
- In each service they call something like:

    ```python
    from mycompany_common_python import otel

    otel.initialize(service_name="myservice")
    ```

- The `mycompany_common_python.otel.initialize()` ends up calling `autoinstrumentation.initialize()` (from my example above)

So when I do something like upgrade the otel version, that process looks like this:

1. Bump otel versions in `mycompany-common-python`
2. Ensure tests pass, etc in `mycompany-common-python`
3. Release new version of `mycompany-common-python`

We then automate the upgrade of `mycompany-common-python` with Dependabot (this isn't a super important detail but I think will help illustrate how I get into a problematic state). So over the next few days. Dependabot goes through and bumps the version of `mycompany-common-python` in each of my services. Eventually that code makes it into production.

Now, sometime later, it turns out maybe there's some sort of bug in the otel auto-instrumentation due to the release, or maybe even in my service, a new version of something like FastAPI or Django is released that causes auto-instrumentation to fail in some way. But I'm not notified in automated tests anywhere since it's only an exception log, not a raised exception. This has happened at various times and in various ways over the years, but it can be eventually traced back to an exception in the auto-instrumentation.

The issue is that this problem is hard to spot from the single error log. I'll end up noticing it when I go to look for some specific span and can't find it - when come to find out the instrumentation failed in some way and nothing was actually being instrumented. But now it's too late because I have already lost out on an important aspect of traceability for whatever problem I'm trying to diagnose.

I personally would much rather fail very loudly whenever there are instrumentation problems and fail a service to start up. Since I have a staging environment/k8s then a pod will fail to start if instrumentation fails which matches the level of importance I put on my services emitting spans and undergoing instrumentation.

----

To make matters worse, if there's an exception in 1 instrumentation it affects all instrumentations after it because of how we re-raise here:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/a164d37a3ca557c6116141ad6f91a9a0701e8249/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py#L104-L106

So 1 instrumentation failure in something like asyncpg could prevent everything loaded after it (fastapi, whatever) from being loaded too. Again, something I really want to know about as loudly as possible.

aabmass (commented):
I think I'm onboard with this option especially since it's opt-in. The spec allows failing fast IRC

phillipuniverse:
@tammy-baylis-swi [done!](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3550)

PR #3567, opened 2025-06-09
fix excluded_urls in instrumentation-asgi

CarlosRCDev:
# Description

If set value "excluded_urls" in OpenTelemetryMiddleware, class fails: 'str' object has no attribute 'url_disabled'.
I'am using Quart, but i think the problem is here:

Fixes # (issue)
excluded_urls with parse_excluded_urls

## Type of change

# How Has This Been Tested?
In instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py

- [ ] test_excluded_urls, fails if return span, with excluded_urls
- [ ] test_no_excluded_urls, fails if not return span, with excluded_urls

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

- [X ] Followed the style guidelines of this project


CarlosRCDev:
> CI is failing. Maybe we should instead check if excluded_urls is already an ExcludeList?

My fault, you are right, the problem:
the class OpenTelemetryMiddleware, when is used directly for example, quart or django, need a string to be convert in ExcludeList class (using parse_excluded_urls function), but the instrumentor like fastapi, instance the class OpenTelemetryMiddleware with excluded_urls as ExcludeList class.

The fix check if excluded_urls isistance of string, if not use the object without changes:
```python
if isinstance(excluded_urls, str):
    excluded_urls = parse_excluded_urls(excluded_urls)
```

emdneto:
> > CI is failing. Maybe we should instead check if excluded_urls is already an ExcludeList?
> 
> My fault, you are right, the problem: the class OpenTelemetryMiddleware, when is used directly for example, quart or django, need a string to be convert in ExcludeList class (using parse_excluded_urls function), but the instrumentor like fastapi, instance the class OpenTelemetryMiddleware with excluded_urls as ExcludeList class.
> 
> The fix check if excluded_urls isistance of string, if not use the object without changes:
> 
> ```python
> if isinstance(excluded_urls, str):
>     excluded_urls = parse_excluded_urls(excluded_urls)
> ```

Right. Can you add a changelog entry?

CarlosRCDev:
> > > CI is failing. Maybe we should instead check if excluded_urls is already an ExcludeList?
> > 
> > 
> > My fault, you are right, the problem: the class OpenTelemetryMiddleware, when is used directly for example, quart or django, need a string to be convert in ExcludeList class (using parse_excluded_urls function), but the instrumentor like fastapi, instance the class OpenTelemetryMiddleware with excluded_urls as ExcludeList class.
> > The fix check if excluded_urls isistance of string, if not use the object without changes:
> > ```python
> > if isinstance(excluded_urls, str):
> >     excluded_urls = parse_excluded_urls(excluded_urls)
> > ```
> 
> Right. Can you add a changelog entry?

Sorry for the wait, I just did it.

emdneto (approved):

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
```diff
@@ -1633,6 +1633,29 @@ async def test_no_metric_for_websockets(self):
...
+        spans = self.memory_exporter.get_finished_spans()
+        self.assertEqual(len(spans), 0)
+
+    async def test_no_excluded_urls(self):
+        self.scope["path"] = "/test_excluded_urls"
+        app = otel_asgi.OpenTelemetryMiddleware(
+            simple_asgi, excluded_urls="test_excluded_urls"
+        )
+        self.seed_app(app)
+        self.scope["path"] = "/test_no_excluded_urls"
```
xrmx, line 1653: I'm a bit confused on why both tests have 0 spans, shouldn't we add a test that asserts that an url not excluded gets the span?

CarlosRCDev: #### In both functions
excluded_urls="test_excluded_urls"

#### This funtion fails if spans has elements
test_excluded_urls, function:
self.scope["path"] = "**/test_excluded_urls**"
self.**assertEqual**(len(spans), 0)

#### This function fails if spans not have elements
test_no_excluded_urls, function:
self.scope["path"] = "**/test_no_excluded_urls**"
self.**assertGreater**(len(spans), 0)

xrmx: Ah I missed that the second one is `Greater` and not `Equal` :sweat: , thanks!

PR #3589, opened 2025-06-19
Rename InMemoryLogExporter to InMemoryLogRecordExporter

hectorhdzg:
# Description

Rename the following classes to maintain consistency with specs and other languages implementations, only used in tests for this repo.
InMemoryLogExporter to InMemoryLogRecordExporter

- [ ] Yes. - Link to PR: https://github.com/open-telemetry/opentelemetry-python/pull/4647
- [ ] No.


instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/common/otel_mocker.py
```diff
@@ -26,7 +26,7 @@
 from opentelemetry.sdk._events import EventLoggerProvider
 from opentelemetry.sdk._logs import LoggerProvider
 from opentelemetry.sdk._logs.export import (
-    InMemoryLogExporter,
+    InMemoryLogRecordExporter,
```
xrmx: I think you should handle the ImportError and alias the old one to the same name as the new one so we can merge this before the other one. Technically there's no constraint the test would run against latest git outside of this CI, so we cannot assume that

xrmx: I was thinking something like the following so we can merge this and tests will be also green on the core PR  with the rename:

```
try:
    from opentelemetry.sdk._logs.export import (
        InMemoryLogRecordExporter
    )
except ImportError:
    from opentelemetry.sdk._logs.export import (
        InMemoryLogExporter as InMemoryLogRecordExporter
    )
```

Said that I commented on the core PR with a proposal to keep the old names deprecated for a release cycle or so and then drop so it's more convenient for people needing to update code?

hectorhdzg:
I removed alias and reverted to original changes; this is not blocking SDK side rename PR anymore so this can be merged when SDK is released, let me know if there is any other feedback.

xrmx:
> I removed alias and reverted to original changes; this is not blocking SDK side rename PR anymore so this can be merged when SDK is released, let me know if there is any other feedback.

While this may be fine for CI, for the genai instrumentations that are not tied to a single opentelemetry version there's no guarantee they are running against latest sdk so we need to handle both or bump the baseline.

From google-genai instrumentation pyproject.toml: 

```
  "opentelemetry-api >=1.31.1, <2",
  "opentelemetry-instrumentation >=0.52b1, <2",
  "opentelemetry-semantic-conventions >=0.52b1, <2"
```

xrmx (approved):

PR #2783, opened 2024-08-07
fix: `fastapi` auto-instrumentation by removing `fastapi-slim` support

smoke:
# Description

`fastapi-slim` support was introduced via https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2683 as a response to `fastapi` maintainers adding a no-dependencies package called `fastapi-slim`.

The changes over `opentelemetry-instrumentation-fastapi` have broken auto-instrumentation via `opentelemetry-instrument` as reproduced here https://github.com/smoke/fastapi-autoinstrumentation-fails
causing `Skipping instrumentation fastapi: DependencyConflict: requested: "fastapi-slim~=0.111.0" but found: "None"` (when used with OTEL_PYTHON_LOG_LEVEL=debug as https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1745 is not yet available)

The `fastapi-slim` itself is discontinued from maintainers as it caused issues with users of it.
Given it causes issues with opentelemetry auto-instrumentation as well, it is removed as most feasible option.

see https://github.com/fastapi/fastapi/pull/11522#issuecomment-2264639417 
see https://fastapi.tiangolo.com/release-notes/#01120

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2683#issuecomment-2271199419

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- see https://github.com/smoke/fastapi-autoinstrumentation-fails

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [x] Followed the style guidelines of this project
- [x] Changelogs have been updated
- [x] Unit tests have been added
- [x] Documentation has been updated

PR #3610, opened 2025-07-02
DRAFT Instruments either

jeremydvoss:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:


jeremydvoss:
This PR's goal is to tackle a longstanding issue from the ground up. I want to make certain that all past stakeholders for the instrumentations I seek to help understand and are okay with it
kafka-python: @rjduffner 
fastapi: @musicinmybrain @smoke @gryevns
psycopg2: @joshowen @emdneto @xrmx

smoke (approved):
I am not subject matter expert, but the idea perfectly makes sense to me and I like it.

musicinmybrain:
Thanks for mentioning me on this. While I still maintain [`python-fastapi` in Fedora](https://src.fedoraproject.org/rpms/python-fastapi), I orphaned [`python-opentelemetry-contrib`](https://src.fedoraproject.org/rpms/python-opentelemetry-contrib) and my other [opentelementry-related packages](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/3LHITWOBP76JNSQICDDX5BHFVFCFS4NA/) some time ago, after I no longer needed them [in connection with `grpc`](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/7ZBAHM7FIB4BXXS7QGOXVETZTQCU5F7J/#7ZBAHM7FIB4BXXS7QGOXVETZTQCU5F7J). Therefore, I no longer have a stake in these instrumentations.

gryevns:
Seems like a sensible approach and I believe this would have resolved the fastapi-slim package issue.

(minor) Maybe `instruments_any` over `instruments_either`.

jeremydvoss:
> `instruments_any`

@gryevns good idea, I'll consider switching the naming convection

rjduffner:
Just starting reading here, but by partially reverting #3202, and not including #3612, doesn't this re-break kafka-python in the auto instrumentation case?

Specifically because kafka-python doesn't have the instruments either functionality yet and only has the instruments functionality which would mean, kafka-python and kafka-python-ng which doesn't ever happen.

scripts/generate_instrumentation_bootstrap.py
```diff
@@ -93,6 +93,14 @@ def main():
                     values=[ast.Str(target_pkg), ast.Str(pkg["requirement"])],
                 )
             )
+        # _instruments_either is an optional field that can be used instead of or in addition to _instruments. While _instruments is a list of dependencies, all of which are expected by the instrumentation, _instruments_either is a list any of which but not all are expected.
```
pmcollins: Just a suggestion but I wonder if `instruments_any` would be more appropriate than `instruments_either` since we're matching on one of multiple rather than one of two.

Update: just noticed this was already suggested above :-)

jeremydvoss:
> Just starting reading here, but by partially reverting #3202, and not including #3612, doesn't this re-break kafka-python in the auto instrumentation case?
> 
> Specifically because kafka-python doesn't have the instruments either functionality yet and only has the instruments functionality which would mean, kafka-python and kafka-python-ng which doesn't ever happen.

Yes, we would want to change the kafka toml accordingly.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/dependencies.py
```diff
@@ -73,4 +136,53 @@ def get_dependency_conflicts(
...
+    required_any: Collection[str] = []
+    found_any: Collection[str] = []
+    for dep in deps_any:
+        if isinstance(dep, Requirement):
+            req = dep
+        else:
+            try:
+                req = Requirement(dep)
+            except InvalidRequirement as exc:
+                logger.warning(
```
emdneto, line 161: Not sure if this will emit a bunch of unintended logs in use cases like the otel-operator. I'll validate that.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/dependencies.py
```diff
@@ -49,8 +80,40 @@ def __str__(self):
...
+
+
 def get_dependency_conflicts(
-    deps: Collection[str | Requirement],
+    deps: Collection[
+        str | Requirement
+    ],  # Dependencies all of which are required
+    deps_any: Collection[
+        str | Requirement
+    ] = None,  # Dependencies any of which are required
```
xrmx:
```suggestion
    ] | None = None,  # Dependencies any of which are required
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -12,22 +12,51 @@
...
+            for dist in distributions()
+            for ep in dist.entry_points
+        }
+
+    def dist_for(self, entry_point: EntryPoint):
+        dist = getattr(entry_point, "dist", None)
+        if dist:
+            return dist
+
+        return self._mapping.get(self._key_for(entry_point))
```
xrmx, line 53: AFAIR we weren't sure this cache was used at all :sweat_smile:

opentelemetry-instrumentation/src/opentelemetry/instrumentation/dependencies.py
jeremydvoss: Not familiar. Could you explain the scenario?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
jeremydvoss: Yeah, I didn't quite understand when it was added. However, since this is part of reverting https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3202 , lets consider removing it separately.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/dependencies.py
jeremydvoss: This is [preexisting](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3610/files#diff-fa513a8f13e9ef85e44311858ced6f35bb55b07f87f19ca27094ff82be08ae34R123). I just duplicated it for this scenario. If it's an issue, let's handle it separately.

instrumentation/opentelemetry-instrumentation-fastapi/pyproject.toml
```diff
@@ -33,8 +33,10 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-instruments = [
+instruments = []
+instruments-any = [
     "fastapi ~= 0.92",
+    "fastapi-slim ~= 0.92",
```
emdneto: I remember we removed that [here](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2783)

CONTRIBUTING.md
```diff
@@ -325,10 +325,14 @@ Below is a checklist of things to be mindful of when implementing a new instrume
...
 - Navigate to the **instrumentation package directory:**
-  - Update **`pyproject.toml`** file by modifying _instruments_ entry in the `[project.optional-dependencies]` section with the new version constraint
-  - Update `_instruments` variable in instrumentation **`package.py`** file with the new version constraint
+  - Update **`pyproject.toml`** file by modifying `instruments` or `instruments-any` entry in the `[project.optional-dependencies]` section with the new version constraint
+  - Update `_instruments` or `_instruments_any` variable in instrumentation **`package.py`** file with the new version constraint
 - At the **root of the project directory**, run `tox -e generate` to regenerate necessary files
 
+Please note that `instruments-any` is an optional field that can be used instead of or in addition to `instruments`. While `instruments` is a list of dependencies, _all_ of which are expected by the instrumentation, `instruments-any` is a list _any_ of which but not all are expected.
+
+<!-- See https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3610 for details on instruments-any -->
```
emdneto, line 345: Do you think we should write an example instead? Like 
```
instruments = [
  "util ~= 1.0"
  "common ~2.0"
]
instruments-any = [
  "foo ~= 3.0"
  "bar ~= 4.0"
]
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -12,22 +12,51 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from functools import cached_property
 from logging import getLogger
 from os import environ
 
-from opentelemetry.instrumentation.dependencies import DependencyConflictError
+from opentelemetry.instrumentation.dependencies import (
+    DependencyConflictError,
```
emdneto, line 20: Do we still need that? It was added before in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3202

emdneto: Also, https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/instrumentor.py#L116

instrumentation/opentelemetry-instrumentation-fastapi/pyproject.toml
jeremydvoss: Yes. This was removed because "instruments" is a list ALL of which are required for instrumentation. With instruments-any, this is finally possible. However, I can remove to leave as a separate PR. It would simplify this PR. What do you think?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
jeremydvoss: Yeah it's not really necessary, but I figured it doesn't hurt to leave it in and it limits the scope of the change. See my comment [here](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3610/files#diff-6ecd6cbdf97812bf55619474db901385d54a6a6931b91b6b2fc5b308b1fb8ab3R118). Can remove if you think that's best

instrumentation/opentelemetry-instrumentation-fastapi/pyproject.toml
jeremydvoss: Now that I think more on this, while the psycopg2 and kafka instrumentations have tests for their alt versions, fastapi does not. So, I will remove the fastapi changes and leave those for whenever you or someone else wants to add fastapi-slim _with_ tests designed specifically for it.

jeremydvoss: Alright, I removed the fastapi changes. The kafka-python and psycopg2 changes are more important because their dependencies wrongfully _include_ second options that need to be retrofitted. But fastapi's change would be something new.

CONTRIBUTING.md
jeremydvoss: How about:
Please note that `instruments-any` is an optional field that can be used instead of or in addition to `instruments`. While `instruments` is a list of dependencies, _all_ of which are expected by the instrumentation, `instruments-any` is a list _any_ of which but not all are expected. For example, the following entry requires both `util` and `common` plus either `foo` or `bar` to be present for the instrumentation to occur:
```
instruments = [
  "util ~= 1.0"
  "common ~= 2.0"
]
instruments-any = [
  "foo ~= 3.0"
  "bar ~= 4.0"
]
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/dependencies.py
jeremydvoss: Done.

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/package.py
```diff
@@ -16,4 +16,5 @@
 _instruments_kafka_python = "kafka-python >= 2.0, < 3.0"
 _instruments_kafka_python_ng = "kafka-python-ng >= 2.0, < 3.0"
 
-_instruments = (_instruments_kafka_python, _instruments_kafka_python_ng)
+_instruments = ()
+_instruments_any = (_instruments_kafka_python, _instruments_kafka_python_ng)
```
rjduffner, line 20: So _instruments is actually used in the kafka-python package here. https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/__init__.py#L126

So you can't really just change it. You can update line 126 above to be _instruments_any instead of _instruments to keep the same functionality in python.

I would check any place that you did this change for the same type of behavior. I would expect to see changes in this and changes in the instrumentation_dependencies function.

CONTRIBUTING.md
emdneto: Yeah. Sounds good to me

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/package.py
emdneto: Good point. For those cases (e.g., psycopg2, kafka-python) I believe we should always return `_instruments_any`

emdneto: But if we return `_instruments_any` it will break the manual-instrumentation path with:
```
ERROR:opentelemetry.instrumentation.instrumentor:DependencyConflict: requested: "kafka-python-ng >= 2.0, < 3.0" but found: "None"
```
because in `_check_dependency_conflicts` function we are checking only for deps and not passing the deps_any as the parameter of the function

emdneto: @jeremydvoss @rjduffner  [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/instrumentor.py#L89) we are not passing deps_any, but it works because in instrumentation that supports multiple packages (e.g., kafka, psycopg) we are checking which distribution is installed and returning the the single `dep`, like [that](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/__init__.py#L116). So, in that case I believe returning _instruments if no distribution is found is the correct approach.

CHANGELOG.md
```diff
@@ -11,6 +11,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Fixed
+
+- `opentelemetry-instrumentation`: Fix dependency conflict detection when instrumented packages are not installed by moving check back to before instrumentors are loaded. Add "instruments-any" feature for instrumentations that target multiple packages.
+  ([#3610](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3610))
+
+### Added
+
+- `opentelemetry-instrumentation-psycopg2` Utilize instruments-any functionality.
+  ([#3610](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3612))
```
```suggestion
  ([#3610](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3610))
```

CHANGELOG.md
```diff
@@ -11,6 +11,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `opentelemetry-instrumentation-kafka-python` Utilize instruments-any functionality.
+  ([#3610](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3612))
```
```suggestion
  ([#3610](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3610))
```

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/package.py
jeremydvoss: I think this should be fine because instrumentation_dependencies actually uses this field as a default and instead, checks for which dependencies to return based on the environment. And I'm keeping _instruments_kafka_python and _instruments_kafka_python_ng. But for the default, we could change it to _instruments_kafka_python instead of instruments. It's somewhat redundant, but that way the dep conflict resolving will block instrumentation correctly. @emdneto @rjduffner am I understanding correctly?

jeremydvoss: or I could keep the current behavior of returning the list of both by returning the _instruments_any list as the default. Either way it's kind of funny style, as is the current status IMO.

emdneto: the last change looks good to me

emdneto: @rjduffner, how do you feel about it?

jeremydvoss: Updated to return _instruments_any with the same value as before. I think it should not be a change on your end but will test

jeremydvoss: Repro from Emidio looks good in all scenarios.

jeremydvoss: Emidio provided a simple repro script for anyone that wants to test:
```python
from opentelemetry.instrumentation.auto_instrumentation import initialize
from os import environ
import logging
logging.basicConfig(level=0)

environ["OTEL_LOGS_EXPORTER"] = "none"
environ["OTEL_TRACES_EXPORTER"] = "none"
environ["OTEL_METRICS_EXPORTER"] = "none"

initialize()
```
try in the following scenarios:
1. kafka-python installed
2. kafka-python-ng installed
3. neither installed

jeremydvoss:
> > `instruments_any`
> 
> @gryevns good idea, I'll consider switching the naming convection

Btw, updated accordingly awhile back

emdneto (approved):

Issue #3434, opened 2025-04-18
DRAFT: Detailed breakdown of dependency conflict check breaking change


jeremydvoss:
Recently, a [breaking change](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3202) was made to how dependency checks work. The change was released in 1.32.0/0.53b0. There were multiple issues with this approach but also multiple benefits. This issue it meant to explain the reasons for the change, the different use cases affected, breakages, and possible solutions.

# Pre-existing dependency conflict logid

"instruments"
"dist"
"requires"

Dependency check was done before instrumentor was instantiated. It is not assumed the instrumented library is installed

## Reason for change: Multi-package instrumentations

kafka, psycopg2, fastapi?

## Change breakdown

Moving dependency check in instrumentor.instrument
Instantiating instrumentor before dependency check

## Use cases

Before explaining the breakages, here are some relevant use cases

### Instrumentation packs

OpenTelemetry clients may include multiple instrumentations automatically. For instance, the azure-monitor-opentelemetry "distro" provides an easy one-line solution to set up OTel providers, exporters, and instrumentations of the most popular libraries. For example, it includes the Flask instrumentation automatically. It is up to dependency conflicts to decide whether that instrumentor should be instantiated and whether the library should be instrumented.

### One-click codeless autoinstrumentation from Cloud providers

Multiple cloud providers, such as Azure provide OpenTelemetry autoinstrumentation as a feature in their UI. This means with a single click, you can enable any and all supported instrumentations. This means the cloud service must install the instrumentations. For both ease of use and to avoid ballooning start-up times, this is down by side loading pre-installed instrumentations. For example, the Flask instrumentation will be instantly present regardless of whether the user has a flask app. It is up to dependency conflicts to decide whether that instrumentor should be instantiated and whether the library should be instrumented.

Note that the fundamental difference between this and other autoinstrumentation scenarios is that instrumented app and autoinstrumentation agent come from 2 different parties: the cloud customer and cloud provider, respectively.
 
## Summarized breakages and fixes

1. Public method get_dist_dependency_conflicts deleted. Fixed in ___
2. Instrumentation requirements are no longer taken from project.toml but rather by abstract Instrumentor.instrumentation_dependencies() method.
[...]

PR #3612, opened 2025-07-03
Utilize instruments-any in fastapi, kafka, and psycopg2

jeremydvoss:
# Description

Relies on https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3610 which new full feature to support "either/or" instrumentation dependency scenarios in manual and autoinstrumentation. Utilizes this feature in fastapi, psycopg2, and kafka.

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3434

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -441,9 +441,9 @@ wheels = [
 name = "boto"
 version = "2.49.0"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/c8/af/54a920ff4255664f5d238b5aebd8eedf7a07c7a5e71e27afcfe840b82f51/boto-2.49.0.tar.gz", hash = "sha256:ea0d3b40a2d852767be77ca343b58a9e3a4b00d9db440efb8da74b4e58025e5a", size = 1478498, upload-time = "2018-07-11T20:58:58.057Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/c8/af/54a920ff4255664f5d238b5aebd8eedf7a07c7a5e71e27afcfe840b82f51/boto-2.49.0.tar.gz", hash = "sha256:ea0d3b40a2d852767be77ca343b58a9e3a4b00d9db440efb8da74b4e58025e5a", size = 1478498 }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/23/10/c0b78c27298029e4454a472a1919bde20cb182dab1662cec7f2ca1dcc523/boto-2.49.0-py2.py3-none-any.whl", hash = "sha256:147758d41ae7240dc989f0039f27da8ca0d53734be0eb869ef16e3adcfa462e8", size = 1359202, upload-time = "2018-07-11T20:58:55.711Z" },
+    { url = "https://files.pythonhosted.org/packages/23/10/c0b78c27298029e4454a472a1919bde20cb182dab1662cec7f2ca1dcc523/boto-2.49.0-py2.py3-none-any.whl", hash = "sha256:147758d41ae7240dc989f0039f27da8ca0d53734be0eb869ef16e3adcfa462e8", size = 1359202 },
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/package.py
```diff
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 
+# TODO: where are these used?
```
rjduffner, line 16: So the python side is how we do the in code dependency checks.

The _instruments* are used by this function here 

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/__init__.py#L108-L126

Which is then loaded here. https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/instrumentor.py#L88

During instrumentation time.

This change doesn't make sense to me as the logic for "either" is already inside the instrumentation_dependencies function. (https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/__init__.py#L108-L126)

rjduffner:
Open question why is this a different PR than #3610? Given #3610 changes this behavior shouldn't this all be rolled into one PR?

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/package.py
jeremydvoss: That get_dependencies method can be overwritten to achieve your desired result, though it's not its intended purpose. The problem is really that overriding it does not solve the scenario where we want to do dependency check before we load the instrumentation. That scenario is fundamental to cloud Auto instrumentation and it was the approach autoinsteumentation used prior to your change. This PR solves the problem from the ground up. Instrumentations no longer need to overwrite methods to provide hacky solutions for narrow scenarios. Now the SDK itself supports either/or scenarios -- whether the check happens before or after loading the instrumentation. In other words, with this change, partial solutions such as those implemented by Kafka and psycopg2 aren't required anymore. However they can still exist. That's why I haven't changed those method overwrites in the pr. This way it shouldn't break existing apps that already adopted that pattern. Let me know if that clears it up. We can sync tomorrow too or discuss in sig.

rjduffner: Sorry, I was mostly reacting to the comment which made it seem like this was still an open question. I do like that we may not need to use the instrumentation_dependencies method anymore but I am worried that its still there at all. Is it possible we could only do the toml method?

I do understand the need for handling more complicated scenarios, (a couple ifs and some ands etc), but in a case like this, the proposed solution in #3610 doesn't work either.

An exact example would be
- If package a, then package b and c are needed
- Or if package d then package e and f are needed

This logic gets super complicated super quickly which is why I went towards doing it in python via the instrumentation_dependencies method but I also totally get that we may not need any more complexity than and's and or's and we can maybe refactor more fully if when we ever need this.

jeremydvoss: Ah, i think I understand. Yeah this method works for all current cases, but I see what you mean about possible "branching" scenarios. This method does not have infinite customizability. I agree that a refactor like that would be a much bigger subject. My focus now is fixing all existing scenarios cleanly without the need for custom overwrites or hacks.

jeremydvoss: (Just a thought for what a long term solution that _could_ support infinite branching might look like: a json-like "instruments" value that includes the logic within and the sdk merely executes.)

jeremydvoss: Missed that you were responding to my inline comment. I understand it but just forgot to remove my comment lol.

emdneto:
Closing. I don't think this is needed anymore.

PR #3651, opened 2025-07-25
Avoid Django's "ALLOWED_HOSTS" check

jourdanrodrigues:
# Description

Django (eventually) calls `HttpRequest.get_host` when calling `HttpRequest.build_absolute_uri`, which we use in the middleware.

In the context I'm at, we have a case for setting up a health check as early as possible in a middleware that needs to bypass that check since Kubernetes hits the containers directly via IP.

I can put the middleware after the health check one but would be good to also instrument it to collect traces if that ever fails.

These changes will bring up the implementation of `build_absolute_uri` without any of the checks.

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
+++ b/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
@@ -194,7 +194,7 @@ class _DjangoMiddleware(MiddlewareMixin):
         # Read more about request.META here:
         # https://docs.djangoproject.com/en/3.0/ref/request-response/#django.http.HttpRequest.META
 
-        if self._excluded_urls.url_disabled(request.build_absolute_uri("?")):
+        if self._url_is_disabled(request):
             return
 
         is_asgi_request = _is_asgi_request(request)
@@ -305,7 +305,7 @@ class _DjangoMiddleware(MiddlewareMixin):
     def process_view(self, request, view_func, *args, **kwargs):
         # Process view is executed before the view function, here we get the
         # route template from request.resolver_match.  It is not set yet in process_request
-        if self._excluded_urls.url_disabled(request.build_absolute_uri("?")):
+        if self._url_is_disabled(request):
             return
 
         if (
@@ -330,7 +330,7 @@ class _DjangoMiddleware(MiddlewareMixin):
                         duration_attrs[HTTP_ROUTE] = route
 
     def process_exception(self, request, exception):
-        if self._excluded_urls.url_disabled(request.build_absolute_uri("?")):
+        if self._url_is_disabled(request):
             return
 
         if self._environ_activation_key in request.META.keys():
@@ -340,7 +340,7 @@ class _DjangoMiddleware(MiddlewareMixin):
     # pylint: disable=too-many-locals
     # pylint: disable=too-many-statements
     def process_response(self, request, response):
-        if self._excluded_urls.url_disabled(request.build_absolute_uri("?")):
+        if self._url_is_disabled(request):
             return response
 
         is_asgi_request = _is_asgi_request(request)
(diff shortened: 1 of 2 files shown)


instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
```diff
@@ -116,6 +116,10 @@ def setUpClass(cls):
     def setUp(self):
         super().setUp()
         setup_test_environment()
+        conf.settings.ALLOWED_HOSTS = [
+            # Django adds "testserver" within "setup_test_environment" so this check doesn't break during tests.
+            "unknown"
+        ]
```
TylerHelmuth, line 122: is there a specific test scenario you can add the asserts the changed conditions are met/not met?

jourdanrodrigues: Would have to get a little creative for this. Will give it a try.

jourdanrodrigues: Using mocks I managed to trigger the behavior within a test!

instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
```diff
@@ -230,6 +234,20 @@ def test_templated_route_get(self):
         self.assertEqual(span.attributes["http.scheme"], "http")
         self.assertEqual(span.attributes["http.status_code"], 200)
 
+    def test_when_middleware_triggers_host_check_then_raises_exception(self):
```
jourdanrodrigues:
```suggestion
    def test_when_middleware_triggers_host_check_then_returns_bad_request(self):
```

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -453,6 +453,15 @@ def process_response(self, request, response):
 
         return response
 
+    def _url_is_disabled(self, request):
+        """
+        Avoid `request.get_host` to bypass Django's ALLOWED_HOST check
```
jourdanrodrigues:
```suggestion
        Avoid `request.get_host` to bypass Django's `ALLOWED_HOSTS` check
```

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -453,6 +453,15 @@ def process_response(self, request, response):
 
         return response
 
+    def _url_is_disabled(self, request):
+        """
+        Avoid `request.get_host` to bypass Django's `ALLOWED_HOSTS` check
+        """
+        url = "{}://{}{}".format(
+            request.scheme, request._get_raw_host(), request.path
+        )
```
TylerHelmuth, line 462: Will this break anyone who was relying on the previous behavior of `request.build_absolute_uri("?")`

jourdanrodrigues: In my experience (not proof of anything, more as an example), host checks are intentional, not an accidental feature of any particular lib (reason why I'm proposing this).

The consequence of these changes is that the host won't be checked anymore, everything else works normally.

I can put these behind an environment variable, Django settings or just the current code under that method abstraction so that it can more easily be overridden, without having to copy over the whole code.

Let me know what you think!

TylerHelmuth: I defer to the python maintainers on how they would want this configurable, if they do. Thanks for the explanation!

TylerHelmuth (approved):

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
pmcollins: It's not clear to me why the previous implementation appended a `?` to the URL being checked, but I'm wondering if we should keep it to minimize changes in behavior. What do you think @jourdanrodrigues?

PR #3652, opened 2025-07-25
Allow `_DjangoMiddleware` to have more override entrypoints

jourdanrodrigues:
# Description

I had a need to dictate how the span creation would happen, specifically because the span I wanted to be the root was at the gunicorn worker level, not Django nor WSGI. Main reason was to include its request log in the trace.

Also, in our context, one span is enough to wrap gunicorn, WSGI and `_DjangoMiddleware` (for billing and practical purposes), so had WSGI and `_DjangoMiddleware` contribute with their attributes to the span created by gunicorn.

I had to override the entirety of `_DjangoMiddleware.process_request` in order to reuse the existing span and to avoid `end_on_exit` that's currently hardcoded to true.

This `_DjangoMiddleware._get_span` will allow achieving this behavior without such wide override.

## Type of change

# How Has This Been Tested?

- [ ] Tests at `instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py`

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
+++ b/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
@@ -13,15 +13,17 @@
 # limitations under the License.
 
 import types
+from contextvars import Token
 from logging import getLogger
 from time import time
 from timeit import default_timer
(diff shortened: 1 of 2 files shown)


tammy-baylis-swi:
Thanks @jourdanrodrigues for contributing! Would it be possible to share example spans from before and after this change?

PR #3657, opened 2025-07-29
Adding new optional `post_injection_hook` to `RequestsInstrumentor`

eddawley:
# Description
The `tracestate` and `baggage` headers may contain sensitive information that is valuable to propagate to trusted systems but could pose concerns when sent to untrusted systems. Currently there is no way to support instrumenting these spans while ensuring these, or other, headers are conditionally propagated. This new option allows callers to define a callable to mutate the post-injection headers based on each request.

## Type of change

# How Has This Been Tested?
`tox -e py312-test-instrumentation-requests`

- [ ] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
@@ -86,6 +86,31 @@ For example,
 
 will exclude requests such as ``https://site/client/123/info`` and ``https://site/xyz/healthcheck``.
 
+Post Injection Hook
+*******************
+To customize headers after the current context is injected, you can provide a function that will be called with
+both the headers and request object.
+
+For example, to avoid propagating `tracestate`/`baggage` to non-allowed hosts while still creating a span:
+
+.. code:: python
+
+    allowed_hosts = frozenset(["my-company.com", "my-cloud-provider.com"])
+
+    def post_injection_hook(headers, request):
+        # Parse the domain from the request URL
+        parsed_url = urlparse(request.url)
+        request_host = parsed_url.hostname
+
+        # Remove headers if the host is NOT in the allowed list
+        if request_host not in allowed_hosts:
+            headers.pop("tracestate", None)
+            headers.pop("baggage", None)
+
+    RequestsInstrumentor().instrument(
+        post_injection_hook=post_injection_hook
+    )
+
 API
 ---
 """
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
```diff
@@ -300,6 +329,9 @@ def get_or_create_headers():
             headers = get_or_create_headers()
             inject(headers)
 
+            if callable(post_injection_hook):
+                post_injection_hook(headers, request)
```
pmcollins, line 333: Would it work for your use case to -- instead of passing in a general-purpose post-injection hook -- pass in an allowlist and inject or not based on that? Seems like the allowlist use case would be common enough to justify baking in explicit support (even at the api/spec level).

robotadam:
If it helps, there have been a few discussions about this in [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification). I had originally requested something like this in https://github.com/open-telemetry/opentelemetry-specification/issues/3799, but the consensus was to focus on baggage, which is in https://github.com/open-telemetry/opentelemetry-specification/issues/3957. But since then there seemed to be some consensus around doing this in instrumentations, like this PR does: https://github.com/open-telemetry/opentelemetry-specification/issues/1633#issuecomment-2703503622

FWIW, this style of hook (or the allowlist pmcollins suggested) would solve my original issue, so I hope some version of this moves forward.

instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py
eddawley: I considered that but I think users will differ in what and when to propagate these headers. For example, this [user](https://github.com/open-telemetry/opentelemetry-specification/issues/1633#issuecomment-2372598895) wants to avoid all headers for one host while only `baggage` for external domains.

I suspect it will be difficult to have a truly one-size fits all. I could even see a case where you base the decision on values in the baggage itself.

edit: also i noticed the request/response hooks started as separate single purpose options that were changed to hooks and I didn't want to repeat that

eddawley:
> If it helps, there have been a few discussions about this in [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification). I had originally requested something like this in [open-telemetry/opentelemetry-specification#3799](https://github.com/open-telemetry/opentelemetry-specification/issues/3799), but the consensus was to focus on baggage, which is in [open-telemetry/opentelemetry-specification#3957](https://github.com/open-telemetry/opentelemetry-specification/issues/3957). But since then there seemed to be some consensus around doing this in instrumentations, like this PR does: [open-telemetry/opentelemetry-specification#1633 (comment)](https://github.com/open-telemetry/opentelemetry-specification/issues/1633#issuecomment-2703503622)
> 
> FWIW, this style of hook (or the allowlist pmcollins suggested) would solve my original issue, so I hope some version of this moves forward.

Thank you so much for these links. I wondered how I could be the first person to consider the security implications of auto instrumentation sending baggage to 3rd parties. I could definitely see this being a broader need of the contrib package. This PR was really just a simple solution to get value as quickly as possible.

bschoenmaeckers:
I would love to see this moving forward.

PR #3664, opened 2025-07-30
Rewrite FastAPI instrumentor middleware stack to be failsafe

### LINKED ISSUE #3637: FastAPI instrumentor stops recording exception event starting v0.55b0
### Describe your environment

OS: MacOS
Python version: (e.g., Python 3.13)
Package version: (e.g., 0.56b0)

### What happened?

Starting 0.55b0, exception event is no longer captured. Screenshot below shows that the span has ended before it gets to record the exception.

Image

### Steps to Reproduce

Below is a script to reproduce the issue. By changing the version from `0.55b0` down to `0.54b1`, we can see the exception event in the span. But for 0.55b0 onward, the span has no event.

```python
# /// script
# dependencies = [
#   "fastapi",
#   "uvicorn",
#   "opentelemetry-sdk",
#   "opentelemetry-instrumentation-fastapi==0.55b0",
# ]
# ///
from fastapi import FastAPI
import uvicorn
from opentelemetry.sdk.trace.export import ConsoleSpanExporter, SimpleSpanProcessor

tracer_provider.add_span_processor(SimpleSpanProcessor(ConsoleSpanExporter()))

app = FastAPI()

@app.get("/")
def root():
    raise RuntimeError()

if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", port=8000)
```

### Expected Result

exception event should be recorded

### Actual Result

empty event list

### Additional context

_No response_

None

### LINKED ISSUE #3642: FastAPI instrumentation: errors in hooks not handled properly
Example script:

```python
from typing import Any

import uvicorn
from fastapi import FastAPI, Request
from fastapi.responses import JSONResponse
from opentelemetry.trace import Span

def server_request_hook(span: Span, scope: dict[str, Any]):
    raise Exception('Hook Exception')


app = FastAPI()
FastAPIInstrumentor.instrument_app(app, server_request_hook=server_request_hook)


@app.exception_handler(Exception)
async def exception_handler(request: Request, exc: Exception):
    return JSONResponse(
        status_code=200,
        content={'message': 'Oops!'},
    )


@app.get('/')
async def index():
    raise Exception('Index Exception')


uvicorn.run(app)
```

Output after visiting http://127.0.0.1:8000/:

```pytb
INFO:     127.0.0.1:55312 - "GET / HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/uvicorn/protocols/http/h11_impl.py", line 403, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/fastapi/applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/starlette/applications.py", line 113, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/starlette/middleware/errors.py", line 186, in __call__
    raise exc
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/starlette/middleware/errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/opentelemetry/instrumentation/asgi/__init__.py", line 754, in __call__
    self.server_request_hook(current_span, scope)
  File "/Users/alex/Library/Application Support/JetBrains/PyCharm2025.1/scratches/scratch_1842.py", line 20, in server_request_hook
    raise Exception('Hook Exception')
Exception: Hook Exception
```

```json
{
    "name": "GET /",
    "context": {
        "trace_id": "0xf2635a292d6a5396c9914909957f3be6",
        "span_id": "0x0760d24d07d2a667",
        "trace_state": "[]"
    },
    "kind": "SpanKind.SERVER",
    "parent_id": null,
    "start_time": "2025-07-16T13:20:32.289554Z",
    "end_time": "2025-07-16T13:20:32.292083Z",
    "status": {
        "status_code": "ERROR",
        "description": "Exception: Hook Exception"
    },
    "attributes": {
        "http.scheme": "http",
        "http.host": "127.0.0.1:8000",
        "net.host.port": 8000,
        "http.flavor": "1.1",
        "http.target": "/",
        "http.url": "http://127.0.0.1:8000/",
        "http.method": "GET",
        "http.server_name": "127.0.0.1:8000",
        "http.user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/138.0.0.0 Safari/537.36",
        "net.peer.ip": "127.0.0.1",
        "net.peer.port": 55312,
        "http.route": "/"
    },
    "events": [
        {
            "name": "exception",
            "timestamp": "2025-07-16T13:20:32.292002Z",
            "attributes": {
                "exception.type": "Exception",
                "exception.message": "Hook Exception",
                "exception.stacktrace": "Traceback (most recent call last):\n  File \"/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/opentelemetry/trace/__init__.py\", line 589, in use_span\n    yield span\n  File \"/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/opentelemetry/instrumentation/asgi/__init__.py\", line 754, in __call__\n    self.server_request_hook(current_span, scope)\n  File \"/Users/alex/Library/Application Support/JetBrains/PyCharm2025.1/scratches/scratch_1842.py\", line 20, in server_request_hook\n    raise Exception('Hook Exception')\nException: Hook Exception\n",
                "exception.escaped": "False"
            }
        }
    ],
    "links": [],
    "resource": {
        "attributes": {
            "telemetry.sdk.version": "1.35.0",
            "service.name": "unknown_service"
        },
        "schema_url": ""
    }
}
```

Problems:

1. Errors in hooks probably shouldn't cause the whole request to fail
2. Even if they should, the `exception_handler` handler isn't used in this case, i.e. the response is 500 instead of 200. This is because of the decision made at the end of https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3012 to not pass user defined error handlers to the outer server middleware. The reasoning was to avoid executing side effects in an error handler twice. On that note, see https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3012/files#r2210430687
3. Also, the 500 status code isn't even recorded on the span.

By contrast, if the server request hook doesn't raise an exception, things work as expected, i.e. error handling inside the route is correct.

outergod:
# Description

Change the way the FastAPI instrumentor deals with the FastAPI middleware stack so that exception handling code doesn't get executed twice, but still has a valid OTEL context available. At the same time, make sure instrumentor hooks failures cannot crash the service itself.

Fixes #3642 
Fixes #3637

## Type of change

# How Has This Been Tested?

Using the MRE in the linked issue, and added unit tests.

- [X] No.

# Checklist:

- [ ] ~~Documentation has been updated~~ (not needed)

### DIFF
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/examples/manual/main.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/examples/manual/main.py
@@ -4,9 +4,12 @@ import os
 from openai import OpenAI
 
 # NOTE: OpenTelemetry Python Logs and Events APIs are in beta
-from opentelemetry import _events, _logs, trace
+from opentelemetry import _events, _logs, metrics, trace
 from opentelemetry.exporter.otlp.proto.grpc._log_exporter import (
     OTLPLogExporter,
+)
+from opentelemetry.exporter.otlp.proto.grpc.metric_exporter import (
+    OTLPMetricExporter,
 )
 from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import (
     OTLPSpanExporter,
@@ -15,6 +18,8 @@ from opentelemetry.instrumentation.openai_v2 import OpenAIInstrumentor
 from opentelemetry.sdk._events import EventLoggerProvider
 from opentelemetry.sdk._logs import LoggerProvider
 from opentelemetry.sdk._logs.export import BatchLogRecordProcessor
+from opentelemetry.sdk.metrics import MeterProvider
+from opentelemetry.sdk.metrics.export import PeriodicExportingMetricReader
 from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.sdk.trace.export import BatchSpanProcessor
 
@@ -30,6 +35,17 @@ _logs.get_logger_provider().add_log_record_processor(
     BatchLogRecordProcessor(OTLPLogExporter())
 )
 _events.set_event_logger_provider(EventLoggerProvider())
+
+# configure metrics
+metrics.set_meter_provider(
+    MeterProvider(
+        metric_readers=[
+            PeriodicExportingMetricReader(
+                OTLPMetricExporter(),
+            ),
+        ]
+    )
+)
 
 # instrument OpenAI
 OpenAIInstrumentor().instrument()
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/examples/manual/requirements.txt
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/examples/manual/requirements.txt
@@ -1,5 +1,5 @@
 openai~=1.57.3
 
-opentelemetry-sdk~=1.30.0
-opentelemetry-exporter-otlp-proto-grpc~=1.30.0
+opentelemetry-sdk~=1.36.0
+opentelemetry-exporter-otlp-proto-grpc~=1.36.0
 opentelemetry-instrumentation-openai-v2~=2.1b0
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/examples/zero-code/requirements.txt
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/examples/zero-code/requirements.txt
@@ -1,6 +1,6 @@
 openai~=1.57.3
 
-opentelemetry-sdk~=1.30.0
-opentelemetry-exporter-otlp-proto-grpc~=1.30.0
-opentelemetry-distro~=0.51b0
+opentelemetry-sdk~=1.36.0
+opentelemetry-exporter-otlp-proto-grpc~=1.36.0
+opentelemetry-distro~=0.57b0
 opentelemetry-instrumentation-openai-v2~=2.1b0


instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -289,22 +288,33 @@ def instrument_app(
...
+            # exceptions from user-provided hooks of tearing through, we wrap
+            # them to return without failure unconditionally.
             def build_middleware_stack(self: Starlette) -> ASGIApp:
+                def failsafe(func):
+                    @functools.wraps(func)
+                    def wrapper(*args, **kwargs):
+                        try:
+                            func(*args, **kwargs)
+                        except Exception:  # pylint: disable=W0718
+                            pass
```
alexmojaki: rather record this exception on the span

outergod: @alexmojaki good point! Addressed in new commit.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -313,23 +323,6 @@ def build_middleware_stack(self: Starlette) -> ASGIApp:
...
-                # Wrap in an outer layer of ServerErrorMiddleware so that any exceptions raised in OpenTelemetryMiddleware
-                # are handled.
-                # This should not happen unless there is a bug in OpenTelemetryMiddleware, but if there is we don't want that
-                # to impact the user's application just because we wrapped the middlewares in this order.
-                if isinstance(
-                    inner_server_error_middleware, ServerErrorMiddleware
-                ):  # usually true
-                    outer_server_error_middleware = ServerErrorMiddleware(
-                        app=otel_middleware,
-                    )
```
alexmojaki, line 326: It's not clear to me why this is being removed or how this relates to the added tests. If you keep this, do the tests fail?

I'm guessing this is meant to fix https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3637 but I don't see how this PR does or how that's being tested.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -289,22 +288,33 @@ def instrument_app(
...
+                return OpenTelemetryMiddleware(
                     inner_server_error_middleware,
                     excluded_urls=excluded_urls,
                     default_span_details=_get_default_span_details,
-                    server_request_hook=server_request_hook,
-                    client_request_hook=client_request_hook,
-                    client_response_hook=client_response_hook,
+                    server_request_hook=failsafe(server_request_hook),
+                    client_request_hook=failsafe(client_request_hook),
+                    client_response_hook=failsafe(client_response_hook),
```
alexmojaki: This wrapping should be in OpenTelemetryMiddleware itself.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
outergod: It's not related to https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3637 but you are correct, this should probably stay as-is. No effect on the tests either way. Addressed in commit.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
outergod: Addressed in commit.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -646,9 +646,21 @@ def __init__(
...
             default_span_details or get_default_span_details
         )
-        self.server_request_hook = server_request_hook
-        self.client_request_hook = client_request_hook
-        self.client_response_hook = client_response_hook
+
+        def failsafe(func):
+            @wraps(func)
+            def wrapper(span: Span, *args, **kwargs):
+                if func is not None:
```
alexmojaki:
```suggestion
                if func is not None:
```

this should be checked outside `wrapper`

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -289,15 +290,16 @@ def instrument_app(
...
 
-            # Instead of using `app.add_middleware` we monkey patch `build_middleware_stack` to insert our middleware
-            # as the outermost middleware.
-            # Otherwise `OpenTelemetryMiddleware` would have unhandled exceptions tearing through it and would not be able
-            # to faithfully record what is returned to the client since it technically cannot know what `ServerErrorMiddleware` is going to do.
-
+            # In order to make traces available at any stage of the request
+            # processing - including exception handling - we wrap ourselves as
+            # the new, outermost middleware. However in order to prevent
+            # exceptions from user-provided hooks of tearing through, we wrap
```
alexmojaki: The wrapping is no longer here. I also don't think this comment is an improvement. It's now less clear than before that this approach ensures the correct http status code.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -339,6 +342,28 @@ def build_middleware_stack(self: Starlette) -> ASGIApp:
...
+                        span.record_exception(exc)
+                        span.set_status(
+                            Status(
+                                status_code=StatusCode.ERROR,
+                                description=f"{type(exc).__name__}: {exc}",
+                            )
+                        )
+                        raise
+
+            app.add_middleware(ExceptionHandlerMiddleware)
```
alexmojaki: I don't actually know where this middleware is going to go, it'd be clearer if this was inserted in  build_middleware_stack.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -1877,3 +1881,154 @@ def test_custom_header_not_present_in_non_recording_span(self):
...
+
+    def tearDown(self) -> None:
+        super().tearDown()
+        with self.disable_logging():
+            otel_fastapi.FastAPIInstrumentor().uninstrument_app(self.app)
+
+    def test_error_handler_context(self):
+        """OTEL tracing contexts must be available during error handler execution"""
+
+        @self.app.exception_handler(Exception)
```
alexmojaki, line 1926: Please test if the exception and http status code are recorded on the span in this case.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -1877,3 +1881,154 @@ def test_custom_header_not_present_in_non_recording_span(self):
...
+                "/foobar",
+            )
+        except Exception:  # pylint: disable=W0718
+            pass
+
+        spans = self.memory_exporter.get_finished_spans()
+
+        self.assertEqual(len(spans), 3)
+        span = spans[2]
+        self.assertEqual(span.status.status_code, StatusCode.ERROR)
```
alexmojaki, line 2022: please also check the http status code on the span.

alexmojaki (changes requested):

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
outergod: That's not possible unless we override Starlette's `build_middleware_stack(self)` completely; its return value is a closure that gets generated from 
```py
        middleware = (
            [Middleware(ServerErrorMiddleware, handler=error_handler, debug=debug)]
            + self.user_middleware
            + [Middleware(ExceptionMiddleware, handlers=exception_handlers, debug=debug)]
        )
```
by invoking each middleware constructor passing in the previously generated middleware instance each. 

I've added a clarifying comment in the next commit.

alexmojaki: it is possible as long as `build_middleware_stack` returns a `ServerErrorMiddleware` or similar object that has an `app` attribute. Then you have whatever layer comes before that, and you insert `ExceptionHandlerMiddleware` between that layer and the outermost one. Same idea as https://github.com/open-telemetry/opentelemetry-python-contrib/issues/795#issuecomment-2396611991

I think what you have now doesn't work if you call `app.add_user_middleware(m)` after instrumenting the app where `m` raises an exception, i.e. it won't be recorded. This also needs testing.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -1877,3 +1881,170 @@ def test_custom_header_not_present_in_non_recording_span(self):
...
+        self.assertEqual(self.request_trace_id, self.error_trace_id)
+
+        spans = self.memory_exporter.get_finished_spans()
+
+        self.assertEqual(len(spans), 1)
+        span = spans[0]
+        self.assertEqual(span.status.status_code, StatusCode.ERROR)
+        self.assertEqual(
+            len(span.events), 2
+        )  # The other span is a `TypeError` from the exception handler not returning a proper response, that's expected
```
alexmojaki: please just fix that in the exception handler

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -339,6 +327,34 @@ def build_middleware_stack(self: Starlette) -> ASGIApp:
                 app,
             )
 
+            # Define an additional middleware for exception handling that gets
+            # added as a regular user middleware.
+            # Normally, `opentelemetry.trace.use_span` covers the recording of
+            # exceptions into the active span, but because of order of
+            # middleware execution, that doesn't work so we have to do it in an
+            # inner middleware.
```
alexmojaki: This explanation is incomplete/incorrect. `use_span` on its own seems like it should work anyway, the problem is not simply the order. It's that the otel middleware ends the span before `use_span` does (or before it can record the exception) as explained in https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3637#issuecomment-3165734793.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
outergod: I've added a check for the exception and the internal status code.

However checking for the HTTP status code wouldn't make much sense here as it would be the one from the installed exception handler; since this one doesn't return a proper response (not part of what's being tested), there's no HTTP status code set on the span however I did add the check below, to `test_exception_span_recording`.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
outergod: Done

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
alexmojaki: checking that the http status code returned by the exception handler is then set on the span here would absolutely be valuable

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
outergod: done

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
outergod: Thank you. I've adapted the explanation.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
outergod: Removed

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
outergod: Done.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
outergod: I'll look into it.

outergod: You are correct about additional user middlewares not getting covered.

outergod: @alexmojaki I think I got it right now. I've added a bunch of comments so the code remains maintainable.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -289,17 +290,78 @@ def instrument_app(
...
 
-            # Instead of using `app.add_middleware` we monkey patch `build_middleware_stack` to insert our middleware
-            # as the outermost middleware.
-            # Otherwise `OpenTelemetryMiddleware` would have unhandled exceptions tearing through it and would not be able
-            # to faithfully record what is returned to the client since it technically cannot know what `ServerErrorMiddleware` is going to do.
-
             def build_middleware_stack(self: Starlette) -> ASGIApp:
-                inner_server_error_middleware: ASGIApp = (  # type: ignore
+                # Define an additional middleware for exception handling that gets
+                # added as a regular user middleware.
```
alexmojaki:
```suggestion
                # Define an additional middleware for exception handling
```

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -1877,3 +1879,207 @@ def test_custom_header_not_present_in_non_recording_span(self):
...
+
+    def test_error_handler_context(self):
+        """OTEL tracing contexts must be available during error handler execution"""
+
+        @self.app.exception_handler(Exception)
+        async def _(*_):
+            self.error_trace_id = (
+                trace.get_current_span().get_span_context().trace_id
+            )
+            return PlainTextResponse("", status_code=500)
```
alexmojaki: please choose a different status_code, e.g. 501

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -1877,3 +1879,207 @@ def test_custom_header_not_present_in_non_recording_span(self):
...
+            self.request_trace_id = (
+                trace.get_current_span().get_span_context().trace_id
+            )
+            raise UnhandledException("Test Exception")
+
+        try:
+            self.client.get(
+                "/foobar",
+            )
+        except Exception:  # pylint: disable=W0718
```
alexmojaki:
```suggestion
        except UnhandledException:
```

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -1877,3 +1879,207 @@ def test_custom_header_not_present_in_non_recording_span(self):
...
+
+class TestTraceableExceptionHandling(TestBase):
+    """Tests to ensure FastAPI exception handlers are only executed once and with a valid context"""
+
+    def setUp(self):
+        super().setUp()
+
+        self.app = fastapi.FastAPI()
+
+        otel_fastapi.FastAPIInstrumentor().instrument_app(self.app)
```
alexmojaki:
```suggestion
        otel_fastapi.FastAPIInstrumentor().instrument_app(self.app, exclude_spans=['receive', 'send'])
```

Then there should only be the spans we care about which should make the test a bit clearer.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -1877,3 +1879,207 @@ def test_custom_header_not_present_in_non_recording_span(self):
...
+        self.assertEqual(len(span.events), 1)
+        event = span.events[0]
+        self.assertEqual(event.name, "exception")
+        assert event.attributes is not None
+        self.assertEqual(
+            event.attributes.get(EXCEPTION_TYPE),
+            f"{__name__}.UnhandledException",
+        )
+
+    def test_error_handler_side_effects(self):
```
alexmojaki: can this be folded into the test above?

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -1877,3 +1879,207 @@ def test_custom_header_not_present_in_non_recording_span(self):
...
+        event = span.events[0]
+        self.assertEqual(event.name, "exception")
+        assert event.attributes is not None
+        self.assertEqual(
+            event.attributes.get(EXCEPTION_TYPE),
+            f"{__name__}.UnhandledException",
+        )
+
+
+class TestFailsafeHooks(TestBase):
```
alexmojaki: i think these tests belong in the asgi tests

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
outergod: done

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
outergod: applied and adapted tests accordingly

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
outergod: done

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
outergod: Done.

alexmojaki (approved):
Looks solid, thank you so much!

outergod:
Thank you for the thorough review and the persistence @alexmojaki!
Do you have any suggestions how to find a maintainer to sponsor the PR? Should I ask on the CNCF Slack?

alexmojaki:
@xrmx @emdneto @codefromthecrypt @lzchen please review? This relates to https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3012 which you reviewed previously. Also cc @Kludex @adriangb

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -289,17 +290,77 @@ def instrument_app(
...
+                    def __init__(self, app):
+                        self.app = app
+
+                    async def __call__(
+                        self, scope: Scope, receive: Receive, send: Send
+                    ) -> None:
+                        try:
+                            await self.app(scope, receive, send)
+                        except Exception as exc:  # pylint: disable=broad-exception-caught
+                            span = get_current_span()
```
emdneto, line 309: We can probably check if span is recording here

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -289,17 +290,77 @@ def instrument_app(
...
                 )
+
+                if not isinstance(
+                    inner_server_error_middleware, ServerErrorMiddleware
+                ):
+                    # Oops, something changed about how Starlette creates middleware stacks
+                    _logger.error(
+                        "Cannot instrument FastAPI as the expected middleware stack has changed"
+                    )
+                    return inner_server_error_middleware
```
emdneto, line 341: Can we test this branch path?

emdneto (commented):
Right. I gave it a try to run the repro and now I can see the recorded exception.

If it helps to others to review, before we had:
ServerErrorMiddleware (outermost) -> OpenTelemetryMiddleware -> ServerErrorMiddleware (innermost)

now we have:
ServerErrorMiddleware (outer -- same as before) -> OpenTelemetryMiddleware -> ServerErrorMiddleware (original handler/debug) -> ExceptionHandlerMiddleware 

But I'm afraid we are not seeing some issues in the current structure of tests for FastAPI. Noticed that while reviewing https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3701

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -289,17 +290,77 @@ def instrument_app(
...
+                    ) -> None:
+                        try:
+                            await self.app(scope, receive, send)
+                        except Exception as exc:  # pylint: disable=broad-exception-caught
+                            span = get_current_span()
+                            span.record_exception(exc)
+                            span.set_status(
+                                Status(
+                                    status_code=StatusCode.ERROR,
+                                    description=f"{type(exc).__name__}: {exc}",
```
emdneto: Not related to your PR I guess, but I can't see the description in the exported span. https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3713

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
outergod: Addressed in commit.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
outergod: Done. I hope the approach is satisfying.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
outergod: What do you suggest I do?

emdneto: My suggestion is we move forward with your fix and address #3713 in a separate PR

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -290,17 +291,78 @@ def instrument_app(
...
+                inner_server_error_middleware: ServerErrorMiddleware = (  # type: ignore
                     self._original_build_middleware_stack()  # type: ignore
                 )
+
+                if not isinstance(
+                    inner_server_error_middleware, ServerErrorMiddleware
+                ):
+                    # Oops, something changed about how Starlette creates middleware stacks
+                    _logger.error(
+                        "Cannot instrument FastAPI as the expected middleware stack has changed"
```
emdneto: Can we assert the message in the test as well?

```suggestion
                        "Skipping FastAPI instrumentation due to unexpected middleware stack: expected %s, got %s", ServerErrorMiddleware.__name__, type(inner_server_error_middleware)
```

emdneto, line 337: Can we assert the message in the test as well?

```suggestion
                        "Skipping FastAPI instrumentation due to unexpected middleware stack: expected %s, got %s", ServerErrorMiddleware.__name__, type(inner_server_error_middleware)
```

emdneto (approved):

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
outergod: Good idea, changed and added test.

Issue #2698, opened 2024-07-11
opentelemetry-instrumentation-asgi: do not set `url.full` attribute for server spans


emdneto:
### Describe your environment

instrumentation-asgi: 0.47b0.dev

### What happened?

asgi instrumentation is emitting `url.full` attribute in server spans

### Steps to Reproduce

Run any example of asgi instrumentation with new semconv opt-in enabled 

```python
import fastapi
import os
import logging
from opentelemetry.instrumentation.fastapi import FastAPIInstrumentor
from opentelemetry import trace
from opentelemetry.sdk.trace import TracerProvider, sampling
from opentelemetry.sdk.resources import Resource
from opentelemetry.sdk.trace.export import (
    BatchSpanProcessor,
    ConsoleSpanExporter,
)
# Specify "http", "http/dup" or ""
os.environ["OTEL_SEMCONV_STABILITY_OPT_IN"] = "http"
logging.basicConfig(level=logging.DEBUG)

provider = TracerProvider()
processor = BatchSpanProcessor(ConsoleSpanExporter())
provider.add_span_processor(processor)
trace.set_tracer_provider(provider)
tracer = trace.get_tracer(__name__)

app = fastapi.FastAPI()

@app.get("/foobar")
async def foobar():
    return {"message": "hello world"}

FastAPIInstrumentor.instrument_app(app)
```

### Expected Result

```json
"attributes": {
        "url.scheme": "http",
        "server.address": "127.0.0.1:8000",
        "server.port": 8000,
        "network.protocol.version": "1.1",
        "url.path": "/foobar",
        "http.request.method": "GET",
        "user_agent.original": "curl/7.68.0",
        "client.address": "127.0.0.1",
        "client.port": 45854,
        "http.route": "/foobar",
        "http.response.status_code": 200
    },
  ```

### Actual Result

Server span Attributes with `url.full` which isn't necessary for server span:
```json
"attributes": {
        "url.scheme": "http",
        "server.address": "127.0.0.1:8000",
        "server.port": 8000,
        "network.protocol.version": "1.1",
        "url.path": "/foobar",
        "url.full": "http://127.0.0.1:8000/foobar",
        "http.request.method": "GET",
        "user_agent.original": "curl/7.68.0",
        "client.address": "127.0.0.1",
        "client.port": 45854,
        "http.route": "/foobar",
        "http.response.status_code": 200
    },
```

### Additional context
[...]

PR #3670, opened 2025-08-01
Fix for FASTAPI unable to record AppService URL (Issue #3654)

### LINKED ISSUE #3654: FastAPI can fail to record App Service URL
### Describe your environment

- **Package Name**: azure-monitor-opentelemetry(-exporter)
- **Package Version**: 
- **Operating System**: Linux App Service
- **Python Version**: 3.11

### What happened?

**Describe the bug**
FastAPI reports IP instead of URL in Application Insights table.

Image

### Steps to Reproduce

**To Reproduce**
1. Deploy Fastapi App Service
2. Query
3. Check App Insights

### Expected Result

**Expected behavior**

**Screenshots**
Image

### Actual Result

Image

### Additional context

_No response_

Yes

rads-1996:
# Description

Fixes #3654 

## Type of change

# How Has This Been Tested?

- [x] Added tests in both asgi and fastapi instrumentation to ensure the host value is fetched.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Version 1.36.0/0.57b0 (2025-07-29)
 
 ### Fixed
-
+- `opentelemetry-instrumentation-asgi` Fixed issue where FastAPI reports IP instead of URL.
+  ([]())
 - `opentelemetry-instrumentation`: Fix dependency conflict detection when instrumented packages are not installed by moving check back to before instrumentors are loaded. Add "instruments-any" feature for instrumentations that target multiple packages.
   ([#3610](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3610))
 - infra(ci): Fix git pull failures in core contrib test
--- a/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
@@ -445,7 +445,20 @@ def get_host_port_url_tuple(scope):
     """Returns (host, port, full_url) tuple."""
     server = scope.get("server") or ["0.0.0.0", 80]
     port = server[1]
+
+    host_header = asgi_getter.get(scope, "host")
+    if host_header:
+        host_value = host_header[0]
+        # Ensure host_value is a string, not bytes
+        if isinstance(host_value, bytes):
+            host_value = host_value.decode("utf-8")
+
+        url_host = host_value + (":" + str(port) if str(port) != "80" else "")
+        
+    else:
+        url_host = server[0] + (":" + str(port) if str(port) != "80" else "")
     server_host = server[0] + (":" + str(port) if str(port) != "80" else "")
+
     # using the scope path is enough, see:
     # - https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope (see: root_path and path)
     # - https://asgi.readthedocs.io/en/latest/specs/www.html#wsgi-compatibility (see: PATH_INFO)
@@ -453,7 +466,7 @@ def get_host_port_url_tuple(scope):
     #       -> that means that the path should contain the root_path already, so prefixing it again is not necessary
     # - https://wsgi.readthedocs.io/en/latest/definitions.html#envvar-PATH_INFO
     full_path = scope.get("path", "")
-    http_url = scope.get("scheme", "http") + "://" + server_host + full_path
+    http_url = scope.get("scheme", "http") + "://" + url_host + full_path
     return server_host, port, http_url
 
 
--- a/instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
+++ b/instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
@@ -1877,3 +1877,259 @@ class TestNonRecordingSpanWithCustomHeaders(TestBase):
         self.assertEqual(200, resp.status_code)
         span_list = self.memory_exporter.get_finished_spans()
         self.assertEqual(len(span_list), 0)
+
+
+class TestFastAPIHostHeaderURL(TestBaseManualFastAPI):
+    """Test suite for Host header URL functionality in FastAPI instrumentation."""
+    
+    def test_host_header_url_construction(self):
+        """Test that URLs use Host header value instead of server IP when available."""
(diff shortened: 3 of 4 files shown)


rads-1996:
Before Change - 

Screenshot 2025-08-01 143500

After Change - 

Screenshot 2025-08-01 143511

CHANGELOG.md
```diff
@@ -14,7 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Version 1.36.0/0.57b0 (2025-07-29)
 
 ### Fixed
-
+- `opentelemetry-instrumentation-asgi` Fixed issue where FastAPI reports IP instead of URL.
```
xrmx: This needs to be moved to unreleased

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -1877,3 +1877,305 @@ def test_custom_header_not_present_in_non_recording_span(self):
...
+                self.assertIsNotNone(server_span)
+
+                actual_url = server_span.attributes[HTTP_URL]
+                self.assertEqual(expected_url, actual_url)
+
+    def test_host_header_bytes_handling(self):
+        """Test that Host header values are properly decoded from bytes."""
+        # This test verifies the fix for bytes vs string handling in our implementation
+        resp = self._client.get(
+            "/foobar", headers={"host": "bytes-test.example.com"}
```
emdneto:
```suggestion
            "/foobar", headers={"host": b"bytes-test.example.com"}
```
am I missing something, or are we supposed to send the value as bytes here?

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -1877,3 +1877,305 @@ def test_custom_header_not_present_in_non_recording_span(self):
...
+        )
+        self.assertEqual(200, resp.status_code)
+
+        spans = self.memory_exporter.get_finished_spans()
+        self.assertEqual(len(spans), 3)
+
+        # Find the server span (the main span, not internal middleware spans)
+        server_span = None
+        for span in spans:
+            if HTTP_URL in span.attributes:
```
emdneto: I think will be better if we check if the span kind is the server? Same for other tests

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -445,15 +445,28 @@ def get_host_port_url_tuple(scope):
     """Returns (host, port, full_url) tuple."""
     server = scope.get("server") or ["0.0.0.0", 80]
     port = server[1]
+
+    host_header = asgi_getter.get(scope, "host")
+    if host_header:
+        host_value = host_header[0]
+        # Ensure host_value is a string, not bytes
+        if isinstance(host_value, bytes):
+            host_value = host_value.decode("utf-8")
```
emdneto: Can we use the already created `_decode_header_item` helper?

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -1877,3 +1877,305 @@ def test_custom_header_not_present_in_non_recording_span(self):
...
+        self.assertIn("url.scheme", server_span.attributes)
+        self.assertIn("server.address", server_span.attributes)
+        self.assertIn("url.path", server_span.attributes)
+        self.assertEqual("https", server_span.attributes.get("url.scheme"))
+        self.assertEqual("/foobar", server_span.attributes.get("url.path"))
+
+        # Current behavior: Host header may not affect server.address in new semantic conventions
+        # This test documents the current behavior rather than enforcing Host header usage
+        server_address = server_span.attributes.get("server.address", "")
+        self.assertIsNotNone(server_address)  # Should have some value
```
emdneto: Can we assert a more deterministic value?

emdneto (commented):
Left some comments. One thought though is we don't even need the HTTP_URL attribute for server spans even for [old semantic convention](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.11.0/specification/trace/semantic_conventions/http.md), but I believe it's another discussion.

>http.url is usually not readily available on the server side but would have to be assembled in a cumbersome and sometimes lossy process from other information

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
Kludex: No, it doesn't need to be bytes.

Kludex: I'm not even sure it accepts bytes? 👀

Kludex:
I think the current implementation is correct. I don't think the instrumentation should trust the `"Host"` header.

rads-1996:
> I think the current implementation is correct. I don't think the instrumentation should trust the `"Host"` header.

But then the IP address appearing in the url field instead of the actual application url address is not very desirable. @Kludex Do you think there is another way to overcome this issue.

jeremydvoss (approved):

jeremydvoss:
Long needed fix. Thanks @rads-1996! Fastapi and asgi should behave like the other instrumentations and use the url when available.

emdneto (approved):
I agree host header is 100% safe, but approving because it looks good to me and is consistent with the WSGI instrumentation. Also, is the way semantic conventions recommends to get the hostname https://opentelemetry.io/docs/specs/semconv/http/http-spans/#setting-serveraddress-and-serverport-attributes. btw, for new semconv we are not recording it https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2698

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -445,15 +445,30 @@ def get_host_port_url_tuple(scope):
...
     port = server[1]
+
+    host_header = asgi_getter.get(scope, "host")
+    if host_header:
+        host_value = host_header[0]
+        # Ensure host_value is a string, not bytes
+        if isinstance(host_value, bytes):
+            host_value = _decode_header_item(
+                host_value
+            )  # use existing function
```
xrmx:
```suggestion
            )
```

Issue #3486, opened 2025-05-09
Tornado instrumentation HTTP_SERVER_DURATION metric is inaccurate and not async-safe


devmonkey22:
### Describe your environment

OS: Ubuntu
Python version: Python 3.10
SDK version: 1.31.1
API version: 0.52b1 (same code in 0.53b0 though)

### What happened?

The tornado instrumenter and how it tracks a request's elapsed time is not async-safe and does not appear to be able to handle multiple concurrent requests without overwriting and using an incorrect value.  This makes the HTTP_SERVER_DURATION metric inaccurate.

### Steps to Reproduce

1. Start one long-running request to a tornado server.
2. Wait a moment, and start a second long-running request to a tornado server.
3. Let the first request finish
4. Let the second request finish.
5. Review the calculated elapsed time of that request.

This effect is much more visible with streaming requests/handlers like an SSE handler or WebSocketHandler (if I ignore #2761) that may stay connected for hours/days.

The `_START_TIME` is stored in [server_histograms](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/9bc77641398cbc582ec24dd4a9defbd21ac276d5/instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py#L382) which is a global dict created by the instrumentor. It stores this in the `_prepare` hook, then reads this back in [_record_on_finish_metrics](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/9bc77641398cbc582ec24dd4a9defbd21ac276d5/instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py#L597).  You can get interleaved start times.

The start time really needs tracked per handler instance (connection).

### Expected Result

```
Request1 Prepare -> StartTime = 10:00:00am
Request2 Prepare -> StartTime = 10:00:05am
Request1 OnFinish -> Reads original StartTime of 10:00:00am, so at 10:00:10am, elapsed time should be 10 seconds.
Request1 OnFinish -> Reads correct original StartTime as 10:00:05am, so at 10:00:10am, elapsed time should be 5 seconds.
```

### Actual Result

```
Request1 Prepare -> StartTime = 10:00:00am
Request2 Prepare -> StartTime = 10:00:05am
[...]

PR #3679, opened 2025-08-05
Fix tornado server (request) duration metric calculation

devmonkey22:
# Description

Fix the tornado instrumenter to track a request's elapsed time in an async-safe way so concurrent requests calculate their own elapsed time for the `HTTP_SERVER_DURATION` metric properly.  This changes the tornado instrumentation to track state (like request start time used to calculate request duration) on the per-request handler instance and not on the more global `server_histogram` object that is shared between requests.

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3486

## Type of change

# How Has This Been Tested?

- Added tornado metrics `test_metrics_concurrent_requests` test

Without the fix, the test run errors with:
image

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


tammy-baylis-swi:
Thank you for this fix. Please could you also add/update the tests?: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/74536f1a92a357c9a25fccba057ce5766d5a8f27/instrumentation/opentelemetry-instrumentation-tornado/tests/test_metrics_instrumentation.py

devmonkey22:
> Thank you for this fix. Please could you also add/update the tests?: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/74536f1a92a357c9a25fccba057ce5766d5a8f27/instrumentation/opentelemetry-instrumentation-tornado/tests/test_metrics_instrumentation.py

@tammy-baylis-swi Can you think of a good way to test this race condition?  I think we'd have to be able start two concurrent fetches, where one hits `prepare`, stays running in a (maybe 1 second sleep), the 2nd request hits (updates start_time), then both requests can complete.  Then we could test that the duration for the 1st request is similar to client duration (ie: not super short like 2nd request should be).  Do you know of an example already?  Otherwise, I was just going to rely on the existing tests to make sure duration is still calculated and similar to client duration (just can't check race condition is fully fixed).

tammy-baylis-swi:
> @tammy-baylis-swi Can you think of a good way to test this race condition? I think we'd have to be able start two concurrent fetches, where one hits `prepare`, stays running in a (maybe 1 second sleep), the 2nd request hits (updates start_time), then both requests can complete. Then we could test that the duration for the 1st request is similar to client duration (ie: not super short like 2nd request should be). Do you know of an example already? Otherwise, I was just going to rely on the existing tests to make sure duration is still calculated and similar to client duration (just can't check race condition is fully fixed).

I'm not actually familiar with Tornado but generally the checks for values resulting from concurrent fetches with predictable times would be helpful, if it's possible. Hmm. Would it make sense to [make_app](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/74536f1a92a357c9a25fccba057ce5766d5a8f27/instrumentation/opentelemetry-instrumentation-tornado/tests/tornado_test_app.py#L124) with any new routes as needed, then also use tornado.httpclient.AsyncHTTPClient to do two concurrent fetches and check resulting metrics in memory?

devmonkey22:
> > @tammy-baylis-swi Can you think of a good way to test this race condition? I think we'd have to be able start two concurrent fetches, where one hits `prepare`, stays running in a (maybe 1 second sleep), the 2nd request hits (updates start_time), then both requests can complete. Then we could test that the duration for the 1st request is similar to client duration (ie: not super short like 2nd request should be). Do you know of an example already? Otherwise, I was just going to rely on the existing tests to make sure duration is still calculated and similar to client duration (just can't check race condition is fully fixed).
> 
> I'm not actually familiar with Tornado but generally the checks for values resulting from concurrent fetches with predictable times would be helpful, if it's possible. Hmm. Would it make sense to [make_app](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/74536f1a92a357c9a25fccba057ce5766d5a8f27/instrumentation/opentelemetry-instrumentation-tornado/tests/tornado_test_app.py#L124) with any new routes as needed, then also use tornado.httpclient.AsyncHTTPClient to do two concurrent fetches and check resulting metrics in memory?

Thanks @tammy-baylis-swi I think I found the test approach.  Just need to iron out CI test failures that didn't happen locally.

devmonkey22:
Should be ready for review again.  Tests are passing.

tammy-baylis-swi (approved):
Nice, thank you for the commented new tests and updated OP. This lgtm; the Maintainers will also have to have a look.

xrmx (approved):

PR #3681, opened 2025-08-05
Fix FastAPI issue with APIRoute subclasses

### LINKED ISSUE #3671: FastAPI Instrumentation Bypasses Middleware for Custom Router
### Describe your environment

OS: Fedora 42
Python: 3.12.11
SDK Version: 1.35.0
API Version: 1.35.0

### What happened?

It appears that if I use anything after splunk-opentelemetry version 2.4.0 to auto-instrument my application,  the middleware for my application is bypassed when using my custom router. When the application is ran without being auto-instrumented, the middleware is not bypassed and everything works as expected.

### Steps to Reproduce

Follow the steps in the README.md for my [minimal issue recreation repository](https://github.com/lannuttia/recreate-otel-issue)

### Expected Result

When the application is instrumented, middleware is not bypassed for requests that use a custom API route.

### Actual Result

When the application is instrumented, middleware is bypassed for requests that use a custom API route

### Additional context

I am unsure of exactly where this issue is coming from but I suspect that it is related to changes made in #3529.

None

lannuttia:
# Description

When an APIRoute subclass would overwrite the matches method with an implementation that depended on non-standard fields existing on the HTTP connection scope, this would cause a failure when the OpenTelemetryMiddleware tried to get the default span details for the incoming request. This has been fixed by using the matches implementation on the Route class for any subclass of Route. This should be sufficient since the only information we are trying to get from that method is the path for the request.

Fixes #3671 

## Type of change

# How Has This Been Tested?

This has been validated with automated tests that are included in this pull request.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3624](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3624))
 - `opentelemetry-instrumentation-dbapi`: fix crash retrieving libpq version when enabling commenter with psycopg
   ([#3796](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3796))
+- `opentelemetry-instrumentation-fastapi`: Fix handling of APIRoute subclasses
+  ([#3681](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3681))
 
 ### Added
 


instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
```diff
@@ -313,6 +341,16 @@ def test_sub_app_fastapi_call(self):
...
         )
 
+
+    def test_custom_api_router(self):
+        """
+        This test is to ensure that custom API routers the OpenTelemetryMiddleware does not cause issues with
+        custom API routers that depend on non-standard fields on the ASGI scope.
+        """
+        resp = self._client.get("/custom-router/success")
+        self.assertEqual(resp.status_code, 200)
```
emdneto: Can you assert spans and it's attributes as well?

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -474,7 +476,11 @@ def _get_route_details(scope):
     route = None
 
     for starlette_route in app.routes:
-        match, _ = starlette_route.matches(scope)
+        match, _ = (
+            Route.matches(starlette_route, scope)
+            if isinstance(starlette_route, Route)
```
emdneto, line 479: This will be true for almost everything no? Maybe check if type(starlette_route).matches is Route.matches

emdneto (commented):
LGTM with CI fix and asserts in the test

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
lannuttia: It's been a while since I wrote this so I forget why I did this specifically. I would have thought that your suggestion would work but when I made that change and ran the test suite, it appears that that change makes several tests fail.

instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
lannuttia: I added some asserts for spans and it's attributes.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
emdneto: Yup. I just tried it, and it doesn't work. Do you see any test we can write to validate that it will go to the else part of the if?

lannuttia: I tried removing the ternary and using just `Route.matches(starlette_route, scope)` and that also caused test failures. I'd assume that whatever those test failures were, were going through the else part. I'll try running them again now and see if I can give a more detailed answer about when it's going through the else.

lannuttia: So I think I tracked it down and it looks like we go through that else when the `startlette_route` is an instance of a starlette `Mount` or `Host`. Mount, Host, and Route all subclass BaseRoute and the routes for the app can be any subclass of BaseRoute.

https://github.com/Kludex/starlette/blob/main/starlette/routing.py#L373

PR #3689, opened 2025-08-13
(Do Not Merge) Add Labeler and instrumentation integration for custom attribute addition to HTTP request duration metrics (Flask, Django, Falcon, WSGI, ASGI)

tammy-baylis-swi:
# Description

⚠️  See this instead: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4288/

Adds new a Python instrumentation Labeler, inspired by Go's `net/http` instrumentation: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/306. It stores KVs added to it in the current request context. The `add` / `add_attributes` can be called by user's instrumented code manually, or by custom distros. The following instrumentors have been updated in this PR so that custom attributes are read from the Labeler in context, then merged into the other attributes (new or old semconv) that are passed to the `record` calls of the request duration histograms:

1. `opentelemetry-instrumentation-flask`
3. `opentelemetry-instrumentation-django`
4. `opentelemetry-instrumentation-falcon`
5. `opentelemetry-instrumentation-wsgi`
6. `opentelemetry-instrumentation-asgi`

Example OTel Collector debug metrics from ASGI OpenTelemetryMiddleware instrumentation, with histograms including custom attributes: https://gist.github.com/tammy-baylis-swi/974ec71d5fb9e88968327b175f2b59a5

Addresses https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3702, https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3703

## Type of change

# How Has This Been Tested?

- [X] New unit tests
- [X] Local installation and manual instrumentation of Flask/Falcon/Django app, as per new doc
- [X] Local installation and manual instrumentation of web.py app (no instrumentation) with WSGI
- [X] Local installation and manual instrumentation of Quart app (no instrumentation) with ASGI

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:


opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+    Returns:
+        Dictionary of custom attributes, or empty dict if no labeler exists
+    """
+    labeler = _labeler_context.get()
+    if labeler is None:
+        return {}
+    return labeler.get_attributes()
+
+
+def enhance_metric_attributes(
```
emdneto: I know the implementation is based on the otel-go one, but I would say "enrich" fits better than enhance, wdyt?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+    base_attributes: Dict[str, Any],
+    include_custom: bool = True,
```
emdneto: if we switch to use the term `enrich` I would suggest replace this with just `enrich` or something similar

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+    def add(self, key: str, value: Union[str, int, float, bool]) -> None:
+        """
+        Add a single attribute to the labeler.
+
+        Args:
+            key: The attribute key
+            value: The attribute value (must be a primitive type)
+        """
+        with self._lock:
+            self._attributes[key] = value
```
emdneto, line 87: I believe we need to skip everything which is not part of the valid primitive types [str, int, float, bool]

emdneto (commented):
This would be nice to have 🙌🏻

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+def get_labeler_attributes() -> Dict[str, Union[str, int, float, bool]]:
+    """
+    Get attributes from the current labeler, if any.
+
+    Returns:
+        Dictionary of custom attributes, or empty dict if no labeler exists
+    """
+    labeler = _labeler_context.get()
+    if labeler is None:
+        return {}
```
emdneto:
```suggestion
        return MappingProxyType({})
```
maybe use MappingProxyType here and in get_attributes return MappingProxyType(self._attributes)

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def add(self, key: str, value: Union[str, int, float, bool]) -> None:
+        """
+        Add a single attribute to the labeler.
+
+        Args:
+            key: The attribute key
+            value: The attribute value (must be a primitive type)
+        """
+        with self._lock:
```
emdneto, line 74: If I remove the locks, test_thread_safety doesn't fail, are we missing asserts?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
```diff
@@ -0,0 +1,168 @@
...
+    enhanced_attributes = base_attributes.copy()
+
+    added_count = 0
+    for key, value in custom_attributes.items():
+        if added_count >= max_custom_attrs:
+            break
+        if key in base_attributes:
+            continue
+
+        if isinstance(value, str) and len(value) > max_attr_value_length:
```
emdneto: can we move this logic and max_custom_attrs to the Labeler level implementation so it can be done at writing time?

tammy-baylis-swi: Good call, changed in [2c64576](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3689/commits/2c64576f13364739df98046889a89193327c940c)

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
tammy-baylis-swi: I did add one new test and updated the existing test in [0bfd192](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3689/commits/0bfd192ed078561c71f719ec1d20dfe2d4cb6ba8), not sure if that helps.

I tried the same (comment out all `with self._lock`) and tried adding other tests, but none of them would fail. I'm not sure if there is a way? The tests should be using real, concurrent threads but without something like writing a C extension I don't think GIL can be released.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
tammy-baylis-swi: Updated in [c86a538](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3689/commits/c86a538ce80cf0832f31ff404adc4b60ba739cc1)

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
tammy-baylis-swi: Yes! Added checks to skip and log warning if invalid in [e10dc80](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3689/commits/e10dc80988d7f366aff02dab915dd034c250efac)

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
tammy-baylis-swi: Sure! Renamed in [068d0c4](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3689/commits/068d0c48069c512248c221b86860bc9388e7388c)

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
tammy-baylis-swi: In [068d0c4](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3689/commits/068d0c48069c512248c221b86860bc9388e7388c) I've changed to `enrich_enabled` to be similar to other bools in the repo (`is_instrumentation_enabled`, `get_future_trace_enabled`, `commenter_enabled`).

tammy-baylis-swi:
@open-telemetry/opentelemetry-python-contrib-approvers ptal, thanks.

tammy-baylis-swi:
> @open-telemetry/opentelemetry-python-contrib-approvers ptal, thanks.

Hi @emdneto could I have another look, please

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
```diff
@@ -0,0 +1,235 @@
+# Copyright The OpenTelemetry Authors
```
emdneto, line 1: can we add the implementation in the typecheck?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -746,6 +800,10 @@ def __call__(
                 duration_attrs_new = _parse_duration_attrs(
                     req_attrs, _StabilityMode.HTTP
                 )
+                # Enhance attributes with any custom labeler attributes
+                duration_attrs_new = enrich_metric_attributes(
```
emdneto, line 804: i'm wondering what we should do here in case user overrides semconv attributes. can we test that?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
```diff
@@ -0,0 +1,235 @@
...
+    Args:
+        base_attributes: The base attributes for the metric
+        enrich_enabled: Whether to include custom labeler attributes
+
+    Returns:
+        Dictionary combining base and custom attributes. If no custom attributes,
+        returns a copy of the original base attributes.
+    """
+    if not enrich_enabled:
+        return base_attributes.copy()
```
emdneto, line 210: since the hot path will be this, can't we just return base_attributes and update the typing to Mapping[str, Any]?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def clear_labeler() -> None:
+    """
+    Clear the Labeler instance from the current request context.
+    """
+    _labeler_context.set(None)
+
+
+def get_labeler_attributes() -> Dict[str, Union[str, int, float, bool]]:
```
emdneto: we probably need to update typing here

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
tammy-baylis-swi: I think with this PR it would be fine, but I think it's less brittle to keep "always copy" moving forward when the instrumentors' parse/record calls are so order-dependent.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
tammy-baylis-swi: Sure! Added in [bac83e2](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3689/commits/bac83e26dffacd132be9e20a701134b387718de4). I had to tweak some Labeler-internal return types to pass.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
tammy-baylis-swi: Done with [bac83e2](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3689/commits/bac83e26dffacd132be9e20a701134b387718de4)

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
tammy-baylis-swi: Yes! Added Labeler unit tests, and WSGI/ASGI unit tests in [12ef547](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3689/commits/12ef547ecf048e8296ad60ac73376352699b09bb)

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+                "feature_enabled": True,
+                "user_segment": "active",
+            }
+        )
+
+        return f"Got user profile for {user_id}"
+
+The labeler also works with auto-instrumentation.
+
+Custom attributes are merged by any instrumentors that use ``enrich_metric_attributes`` before their calls to report individual metrics recording, such as ``Histogram.record``. ``enhance_metrics_attributes`` does not overwrite base attributes that exist at the same keys.
```
emdneto:
```suggestion
Custom attributes are merged by any instrumentors that use ``enrich_metric_attributes`` before their calls to report individual metrics recording, such as ``Histogram.record``. ``enrich_metric_attributes`` does not overwrite base attributes that exist at the same keys.
```

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
emdneto, line 822: should this be enriched as well?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+                        key,
+                        type(value).__name__,
+                    )
+                    continue
+
+                if (
+                    len(self._attributes) >= self._max_custom_attrs
+                    and key not in self._attributes
+                ):
+                    break
```
emdneto:
```suggestion
                    continue
```
Otherwise an already existing key won't be updated. Simple MRE:

```
labeler = Labeler(max_custom_attrs=1)
labeler.add("existing", "old-value")
labeler.add_attributes({"new_key": "should-be-ignored","existing": "new-value"})
assert dict(labeler.get_attributes()) == {"existing": "new-value"}
```

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
```diff
@@ -179,7 +199,23 @@ def setUp(self):
...
+        @self.app.route("/test_labeler")
+        def test_labeler_route():
+            labeler = get_labeler()
+            labeler.add("custom_attr", "test_value")
+            labeler.add_attributes(
+                {"endpoint_type": "test", "feature_flag": True}
+            )
+            return "OK"
+
+        @self.app.route("/no_labeler")
```
emdneto, line 215: Can you also include a call to `/no_labeler` in the test and confirm that there are no attributes in the metrics? I believe we would need a clear_labeler for every request so attributes don't leak.

Something like:
```
    def test_flask_metrics_no_labeler(self):
        self.client.get("/test_labeler")
        self.client.get("/no_labeler")

        ...
        ...

        self.assertNotIn("custom_attr", no_labeler_attrs)

```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_labeler/_internal/__init__.py
tammy-baylis-swi: Addressed in [2efe1f6](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3689/commits/2efe1f6d30b82ab7e749a3c88932febbaf765ece)!

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
tammy-baylis-swi: Originally I wanted to limit this PR to HTTP duration metrics-only since it's already a pretty big diff.

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
tammy-baylis-swi: Good call. There is indeed a leak right now!

`clear_labeler` for every request works for each instrumentor in isolation. It's more complicated when multiple frameworks are instrumented, e.g. Flask and WSGI, where custom attributes are wiped by one before attributes are written by the other. Working on how to deal with that

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
emdneto: That's fine

instrumentation/opentelemetry-instrumentation-flask/tests/test_programmatic.py
tammy-baylis-swi: Hi @emdneto , I've created a new PR where new `Labeler` relies on OTel Context for better custom attributes management and therefore no `clear_labeler` is required to prevent leaks. I'm also dividing this into multiple PRs; new PR only adds Labeler to start. PTAL over there and LMK what you think! 😺 : https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4288/

emdneto:
Thanks for the PR!

Issue #3458, opened 2025-05-02
Improve metrics in Celery instrumentation


rbagd:
### What problem do you want to solve?

Our instrumented Celery applications often are deployed along with Celery Flower to monitor a few metrics on tasks and workers. But those metrics can also be made available through auto-instrumentation. I am sure for those whose usage of Flower is limited having it all in a single instrumentation would be beneficial.

### Describe the solution you'd like

- [x] Add metrics from Celery Flower, i.e. counts for executing and prefetched tasks as well as prefetch and execution duration of a task.
- [x] Rename those metrics according to OTel semantic conventions
- [x] Fix a memory leak in current metrics generation ([this entry](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/e6869cdc75931cbdc0751fcefadf84ece4e10d09/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py#L355) never gets removed)

### Additional Context

~I got some parts working in my local environment but testing has been a challenge. I don't think I ever had all unit tests passing even in clean setup. I have no idea if it's something in my own environment as Celery tests appear to be generally green in PRs.~

### Would you like to implement a fix?

Yes

PR #3690, opened 2025-08-13
Do not keep task id reference indefinitely in Celery instrumenation

rbagd:
# Description

Reference to a task identifier as a string is kept indefinitely by the instrumentation. This just makes sure it gets removed once no longer needed.

Related #3458

## Type of change

# How Has This Been Tested?

- [X] Unit test to ensure reference is not there

- [X] No.

# Checklist:

### DIFF
--- a/instrumentation-genai/opentelemetry-instrumentation-google-genai/examples/manual/main.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-google-genai/examples/manual/main.py
@@ -22,8 +22,7 @@ import os
 
 import google.genai
 
-# NOTE: OpenTelemetry Python Logs and Events APIs are in beta
-from opentelemetry import _events as otel_events
+# NOTE: OpenTelemetry Python Logs API is in beta
 from opentelemetry import _logs as otel_logs
 from opentelemetry import metrics as otel_metrics
 from opentelemetry import trace as otel_trace
@@ -40,7 +39,6 @@ from opentelemetry.instrumentation.google_genai import (
     GoogleGenAiSdkInstrumentor,
 )
 from opentelemetry.instrumentation.requests import RequestsInstrumentor
-from opentelemetry.sdk._events import EventLoggerProvider
 from opentelemetry.sdk._logs import LoggerProvider
 from opentelemetry.sdk._logs.export import BatchLogRecordProcessor
 from opentelemetry.sdk.metrics import MeterProvider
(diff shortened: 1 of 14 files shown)


instrumentation/opentelemetry-instrumentation-celery/tests/test_metrics.py
```diff
@@ -42,6 +42,8 @@ def test_basic_metric(self):
         task_runtime_estimated = (default_timer() - start_time) * 1000
 
         metrics = self.get_metrics()
+        self.assertEqual(CeleryInstrumentor().task_id_to_start_time, {})
```
rads-1996, line 45: Should another test be added where the task_id does not exist and the pop operation returns None instead?

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `opentelemetry-instrumentation-celery`: Fix a memory leak where a reference to a task identifier is kept indefinitely
+  ([#3463](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3463))
```
xrmx:
```suggestion
  ([#3690](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3690))
```

instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
```diff
@@ -362,6 +362,7 @@ def _record_histograms(self, task_id, metric_attributes):
             self.task_id_to_start_time.get(task_id),
             attributes=metric_attributes,
         )
+        self.task_id_to_start_time.pop(task_id, None)
```
xrmx, line 365: What is that we are leaking? On line 355 we are setting this a time object so don't expect to keep alive something on the celery side?

rbagd: `task_id_to_start_time` is a dictionary initialised with `CeleryInstrumentor`. Each time a task finishes the dictionary is augmented with a record of time spent for that task id. In the current state the dictionary just keeps growing even though the task finished long time ago (task identifiers are normally unique). There are no references to any other Python objects that prevent GC, it's just this relatively small issue.

instrumentation/opentelemetry-instrumentation-celery/tests/test_metrics.py
rbagd: Would that test matter actually? We're not testing return value of the `pop` here in any case. I may not see your point. Can you maybe explain the case you have in mind?

PR #3701, opened 2025-08-20
Don't pass bounded `server_request_hook` when using `FastAPIInstrumentor.instrument(server_request_hook=...)`

Kludex:
You can reproduce the issue with the following:

```py
import logfire

logfire.configure()

def server_request_hook(span, scope):
    print('server_request_hook', span, scope)

FastAPIInstrumentor().instrument(server_request_hook=server_request_hook)

from fastapi import FastAPI

app = FastAPI()


@app.get('/')
def read_root():
    return {'message': 'Hello, World!'}

if __name__ == '__main__':
    import uvicorn

    uvicorn.run(app, host='0.0.0.0', port=8000)
```

If you call `localhost:8000` after running with `python main.py`, you'll see an exception. The exception happens because of an issue that was introduced on https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3012 (using `self._server_request_hook` instead of `_InstrumentedFastAPI._server_request_hook`).

This PR simplifies the logic, and fixes the issue above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3719](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3719))
 - `opentelemetry-instrumentation-httpx`: fix missing metric response attributes when tracing is disabled
   ([#3615](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3615))
+- `opentelemetry-instrumentation-fastapi`:  Don't pass bounded server_request_hook when using `FastAPIInstrumentor.instrument()`
+  ([#3701](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3701))
 
 ### Added
 
--- a/instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
+++ b/instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
@@ -1047,7 +1047,7 @@ class TestFastAPIManualInstrumentation(TestBaseManualFastAPI):
         return app
 
 
-class TestFastAPIManualInstrumentationHooks(TestBaseManualFastAPI):
+class TestFastAPIManualInstrumentationHooks(TestBaseFastAPI):
     def _create_app(self):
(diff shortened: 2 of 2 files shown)


Kludex:
cc @xrmx @emdneto :pray:

emdneto (commented):
So, the current `test_hooks` is probably masking the issue since we are passing bounded hooks already. Wdyt of modify the test to something like this or add new one with that:

```python
class TestFastAPIManualInstrumentationHooks(TestBaseManualFastAPI):
    def _create_app(self):
        def server_request_hook(span, scope):
            span.update_name("name from server hook")

        def client_request_hook(receive_span, scope, message):
            receive_span.update_name("name from client hook")
            receive_span.set_attribute("attr-from-request-hook", "set")

        def client_response_hook(send_span, scope, message):
            send_span.update_name("name from response hook")
            send_span.set_attribute("attr-from-response-hook", "value")

        self._instrumentor.instrument(
            server_request_hook=server_request_hook,
            client_request_hook=client_request_hook,
            client_response_hook=client_response_hook,
        )

        app = self._create_fastapi_app()

        return app

    def test_hooks(self):
        self._client.get("/foobar")

        spans = self.sorted_spans(self.memory_exporter.get_finished_spans())
        self.assertEqual(len(spans), 3)

        server_span = spans[2]
        self.assertEqual(server_span.name, "name from server hook")

        response_spans = spans[:2]
        for span in response_spans:
            self.assertEqual(span.name, "name from response hook")
            self.assertSpanHasAttributes(span, {"attr-from-response-hook": "value"})
```

Kludex:
Yeah, that's the case. 👍

Kludex:
I can't join the meeting today, but it will be nice if the line-length of this project changes to 120. We could add a `.git-blame-ignore-revs` file after to ignore that commit.

Most of the project is very lengthy.

Kludex:
How changing a function in a test below changed the result of a test above?

emdneto (commented):
@Kludex mind if I push some changes to your branch? I think I managed to get it working

Kludex:
> @Kludex mind if I push some changes to your branch? I think I managed to get it working

vlw mano, fica a vtd

emdneto:
> > @Kludex mind if I push some changes to your branch? I think I managed to get it working
> 
> vlw mano, fica a vtd

@Kludex da uma olhada ai dps. vou precisar desse fix

Kludex:
Thanks

Issue #3280, opened 2025-02-20
VertexAI tool call may be duplicated in user event


aabmass:
See https://github.com/open-telemetry/semantic-conventions/issues/1883 for discussion. This is a tracking issue for the actual `opentelemetry-instrumentation-vertexai`.

aabmass:
This is only applies to the "old" semconv which will eventually be aged-out. The new behavior with `OTEL_SEMCONV_STABILITY` envvar (https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3709) is implemented and doesn't have the issue. I don't intend to fix this for the old semconv, so closing this out.

Issue #1768, opened 2023-04-20
Support new stable http semantic conventions for http instrumentations - transition plan


lzchen:
Now that the community has decided to adopt ECS [conventions](https://github.com/open-telemetry/semantic-conventions/issues/1012), there is a push for stability for http semantic conventions. Part of this eventual work for languages is the eventual transition for instrumentations to update to the stable semantic conventions (which contain several breaking changes). This [issue](https://github.com/open-telemetry/opentelemetry-specification/issues/3362) discusses the process of how languages will transition their instrumentations. The associated [PR](https://github.com/open-telemetry/opentelemetry-specification/pull/3404) proposes a plan to do so, including bumping instrumentations to major versions. This poses a couple of problems with how we currently version our packages:

1. We release packages in lockstep so these need to be decoupled if we want to update instrumentations separately
2. All instrumentations pin a version of the semantic conventions package -> how to support multiple semantic conventions
3. This only applies to http semantic conventions so we must ensure a clear path forward that doesn't affect transitioning other types of instrumentations.

Opening this issue to discuss how we can tackle this transition and whether we want to chime in to the community if this is not feasible.

tammy-baylis-swi:
Hi there! I'd love to stay in the loop and work on the actual transition if/when the time comes.

So far I've been catching up on related issues. In May, a PR was merged outlining a transition helper variable, `OTEL_SEMCONV_STABILITY_OPT_IN`: https://github.com/open-telemetry/opentelemetry-specification/pull/3443 Setting `OTEL_SEMCONV_STABILITY_OPT_IN=http/dup` before the breaking major version release would "emit both the old and the stable HTTP and networking attributes". This might help with point 2 about supporting multiple semantic conventions.

lzchen:
As discussed in the Python Sig 7/13

We will NOT be implementing the transition plan as defined [here](https://github.com/open-telemetry/opentelemetry-specification/pull/3443). Instead, we will be sticking with our current instrumentation releases (bumping minor versions as appropriate) until we are ready to make a stable release since all of our instrumentations are not yet stable. As for updating to the new semantic conventions, we will be updating to the frozen version v1.21 [this pr] (https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1746) so as to only make a single breaking change. This will be followed by a MINOR version release as usual. Our message to users is that this will be a significant breaking change and we will assist them as needed when issues arise if they want to upgrade.

[...]

PR #3709, opened 2025-08-21
Update vertexai instrumentation to be in-line with the latest semantic conventions

DylanRussell:
# Description

Start making changes to implement the big semantic convention changes made in https://github.com/open-telemetry/semantic-conventions/pull/2179.

Now only a single event (`gen_ai.client.inference.operation.details`) is used to capture Chat History. These changes will be opt-in, users will need to set the environment variable OTEL_SEMCONV_STABILITY_OPT_IN to `gen_ai_latest_experimental` to see them.

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Unit tests

No.. We should wait for a sem conv release (and core repo to pick that up) before releasing this change, but I don't this PR needs to wait for those.

# Checklist:

- [ x] Followed the style guidelines of this project
- [ x] Changelogs have been updated


DylanRussell:
I'm confused by the `OTEL_SEMCONV_STABILITY_OPT_IN` env var. I think the idea is to avoid breaking people,  and make the breaking changes made by https://github.com/open-telemetry/semantic-conventions/pull/2179 opt-in.

Maybe the value of that envvar should just be a version number, and every new version which has breaking changes we put them behind a gate with that version number ?  If a user sets the env var to v3, they get all breaking changes from v1/v2/v3, there will be no way to pick and choose which breaking changes they get.. But making all the permutations of breaking changes opt-in is kind of a pain.. I suppose the flag is repeated so we could let users specify multiple versions in it, to get just the breaking changes they want..

Users will be opted-in to the non-breaking semantic convention changes by default, I think that is fine..

I don't like that new users have to know to set this env var and don't get all the newest stuff by default..

I realize what I've just described doing with an env var, is the same thing that we do with the standard release.. It's kind of like we want 2 releases one with breaking changes and one without..

Would be interested in others thoughts on this..

lmolkova:
Could you please check if https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3280 can be closed now

aabmass:
What do you want to do regarding ordering of this PR and these:
- https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3709
- https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3718

DylanRussell:
Ideally those go in first.  https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3718 may take longer but it's kind of minor and I don't think it should block this one.

This PR is ready for review though, I don't think we need to wait for those to review this one

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -173,6 +174,7 @@ class _StabilityMode(Enum):
     HTTP_DUP = "http/dup"
     DATABASE = "database"
     DATABASE_DUP = "database/dup"
+    GEN_AI_LATEST = "gen_ai_latest_experimental"
```
tammy-baylis-swi: For consistency with the other modes, would it make sense to also have a `GEN_AI_LATEST_DUP`? Or maybe not because this part of the semconv doesn't really have an "old" vs "new" version?

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/patch.py
```diff
@@ -91,14 +95,72 @@ def _extract_params(
...
+        event_logger: EventLogger,
+        capture_content: bool,
+        sem_conv_opt_in_mode: _StabilityMode,
     ) -> None:
         self.tracer = tracer
         self.event_logger = event_logger
         self.capture_content = capture_content
+        self.sem_conv_opt_in_mode = sem_conv_opt_in_mode
+
+    #   Deprecations:
```
tammy-baylis-swi: Could this info and the opt-in be added to the [readme](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation-genai/opentelemetry-instrumentation-vertexai/README.rst)?

DylanRussell: Eventually yes, but it's not ready yet. I also need to make the same changes to the Span that gets written. I'd like for everything to be done before updating the readme

DylanRussell: Planning to make the span changes in a different PR to keep the size of this one reasonable..

tammy-baylis-swi: Sure, that makes sense!

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
DylanRussell: What is the meaning of `DUP` ? The old version is what users get when they don't set the env var

tammy-baylis-swi: `DUP` is to duplicate and write both old and new versions of attributes. It came from the [TC's transition plan](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1768#issuecomment-1760372816) for the HTTP attributes stabilization, and for those Python instrumentors there's "old/new/dup" config to support stepwise opt-in for long-time users ([this epic](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2453)).

That being said I don't know enough about the gen_ai semconv adherence plan! If it's to do this more simple opt-in without duplication because these instrumentors are relatively new, then we don't need `DUP`.

DylanRussell: I see, I don't think that makes sense for this gen ai flag yet.. It's not a new set of attributes that's behind the flag, but a bigger refactoring of the actual events/span that gets written

DylanRussell:
Ok i've updated this PR to use the new gen ai utils that got merged the other day.

DylanRussell:
Need a release of the gen AI utils before this can get merged.. It'd be nice if we also got a release in opentelemetry-python so we can use the new sem conv constants that got added..

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
tammy-baylis-swi: Ok! I agree that DUP is not needed

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/utils.py
```diff
@@ -106,16 +122,20 @@ def get_server_attributes(
...
     generation_config = params.generation_config
     attributes: dict[str, AttributeValue] = {
         GenAIAttributes.GEN_AI_OPERATION_NAME: operation_name.value,
-        GenAIAttributes.GEN_AI_SYSTEM: GenAIAttributes.GenAiSystemValues.VERTEX_AI.value,
         GenAIAttributes.GEN_AI_REQUEST_MODEL: model,
     }
+    if not use_latest_semconvs:
+        attributes[GenAIAttributes.GEN_AI_SYSTEM] = (
+            GenAIAttributes.GenAiSystemValues.VERTEX_AI.value
+        )
```
tammy-baylis-swi, line 138: Re:

> I don't like that new users have to know to set this env var and don't get all the newest stuff by default..
> 
> I realize what I've just described doing with an env var, is the same thing that we do with the standard release.. It's kind of like we want 2 releases one with breaking changes and one without..
> 
> Would be interested in others thoughts on this..

I don't love having to opt-in since what's under `if use_latest_semconvs` below is all additive. But I guess there is a case for it -- this one `GEN_AI_SYSTEM` attribute if not-latest. If it's really important to some users then it makes sense but I'm still quite new around genAI and instrumenting it to know.

DylanRussell: Unfortunately it's not all additive.. This part is additive but the rest of it is a breaking change to existing users

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/__init__.py
```diff
@@ -118,10 +123,30 @@ def _instrument(self, **kwargs: Any):
...
+        elif sem_conv_opt_in_mode == _StabilityMode.GEN_AI_LATEST_EXPERIMENTAL:
+            # Type checker now knows it's the other literal
+            method_wrappers = MethodWrappers(
+                tracer,
+                event_logger,
+                is_content_enabled(sem_conv_opt_in_mode),
+                sem_conv_opt_in_mode,
+            )
+        else:
+            # Impossible to reach here, only 2 opt-in modes exist for GEN_AI.
```
aabmass: FYI https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.assert_never

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/patch.py
```diff
@@ -186,9 +299,31 @@ async def agenerate_content(
...
+                        create_operation_details_event(
+                            params=_extract_params(*args, **kwargs),
+                            response=None,
+                            capture_content=capture_content,
+                            api_endpoint=api_endpoint,
+                        )
+                    )
+                    raise e
+                handle_response(response)
+                return response
```
aabmass: I think it would be good to push this shared boilerplate down into the shared context manager. This would get copied 4 times once you add sync and async streaming https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3331

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/utils.py
```diff
@@ -258,6 +307,101 @@ def request_to_events(
...
+def create_operation_details_event(
+    *,
+    api_endpoint: str,
+    response: prediction_service.GenerateContentResponse
+    | prediction_service_v1beta1.GenerateContentResponse
+    | None,
+    params: GenerateContentParams,
+    capture_content: ContentCapturingMode,
+) -> Event:
+    event = Event(name="gen_ai.client.inference.operation.details")
```
aabmass, line 319: Is there a constant for it generated in `opentelemetry.semconv._incubating`?

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/utils.py
```diff
@@ -258,6 +307,101 @@ def request_to_events(
...
+        **(get_genai_response_attributes(response) if response else {}),
+    }
+    event.attributes = attributes
+    if capture_content in {
+        ContentCapturingMode.NO_CONTENT,
+        ContentCapturingMode.SPAN_ONLY,
+    }:
+        return event
+    if params.system_instruction:
+        attributes["gen_ai.system_instructions"] = [
```
aabmass: there is a constant for this one in `from opentelemetry.semconv._incubating.attributes import gen_ai_attributes`

You'll probably need  to increase the semconv version pinned in pyproject.toml

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/utils.py
```diff
@@ -258,6 +307,101 @@ def request_to_events(
...
+            )
+        )
+    return output_messages
+
+
+def _convert_content_to_message(
+    content: content.Content | content_v1beta1.Content,
+) -> InputMessage:
+    parts: MessagePart = []
+    message = InputMessage(role=content.role, parts=parts)
```
aabmass: nit consider just `return InputMessage(...)` at the end of the function

instrumentation-genai/opentelemetry-instrumentation-vertexai/tests/conftest.py
```diff
@@ -94,13 +115,27 @@ def vertexai_init(vcr: VCR) -> None:
     )
 
 
-@pytest.fixture
+@pytest.fixture(
+    params=[True, False],
+    ids=["newStyleInstrumentation", "oldStyleInstrumentation"],
```
aabmass:
```suggestion
    ids=["experimental_semconv", "stable_semconv"],
```

instrumentation-genai/opentelemetry-instrumentation-vertexai/tests/requirements.oldest.txt
```diff
@@ -66,9 +66,6 @@ yarl==1.15.2
 zipp==3.20.2
 
 # when updating, also update in pyproject.toml
-opentelemetry-api==1.28
-opentelemetry-sdk==1.28
-opentelemetry-semantic-conventions==0.49b0
-opentelemetry-instrumentation==0.49b0
```
aabmass, line 72: Was this needed?

instrumentation-genai/opentelemetry-instrumentation-vertexai/tests/test_chat_completions.py
```diff
@@ -126,44 +143,58 @@ def test_generate_content_without_events(
...
-    }
-    assert choice_log.body == {
-        "finish_reason": "stop",
-        "index": 0,
-        "message": {"role": "model"},
-    }
-
-
-@pytest.mark.vcr
+    if "oldStyleInstrumentation" in request.node.name:
```
aabmass: hmm i kinda like splitting the test cases into two separate test functions with a fixture for `semconv_stable` or `semconv_experimental`, like we have for `instrument_with_content` vs `instrument_no_content`.

I know that would be a lot of work so up to you if it's worth doing.

instrumentation-genai/opentelemetry-instrumentation-vertexai/tests/test_chat_completions.py
```diff
@@ -38,84 +23,116 @@
...
+                    "role": "model",
+                    "parts": ({"type": "text", "content": "This is a test."},),
+                    "finish_reason": "stop",
+                },
+            ),
+        }
 
 
-@pytest.mark.vcr
+@pytest.mark.vcr("cassettes/test_generate_content_without_events.yaml")
```
aabmass: I thought this would be handled by https://github.com/open-telemetry/opentelemetry-python-contrib/blob/ee947cec0c5bddb0b553c95cc6407cda81e26fe7/instrumentation-genai/opentelemetry-instrumentation-vertexai/tests/test_chat_completions.py#L516-L519

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/patch.py
aabmass, line 1: Can you just add a comment somewhere explaining what this is and link[ to the semconv ](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/gen-ai/gen-ai-agent-spans.md?plain=1#L18-L37)?

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/__init__.py
```diff
@@ -118,10 +123,30 @@ def _instrument(self, **kwargs: Any):
             schema_url=Schemas.V1_28_0.value,
             event_logger_provider=event_logger_provider,
         )
-
-        method_wrappers = MethodWrappers(
-            tracer, event_logger, is_content_enabled()
+        sem_conv_opt_in_mode = _OpenTelemetrySemanticConventionStability._get_opentelemetry_stability_opt_in_mode(
```
aabmass, line 134: We need to set the correct `schema_url` (semconv version) in the providers based on what version they opt-in to

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/utils.py
DylanRussell: No this one is missing for some reason

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/utils.py
DylanRussell: Nice yes got it. Thought I already did this .. Fixed now

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/utils.py
DylanRussell: Yep, done

instrumentation-genai/opentelemetry-instrumentation-vertexai/tests/requirements.oldest.txt
DylanRussell: Now that there's been a release I just need to bump them all to the latest release (0.58b0,1.37) in order to get the latest semantic conventions.. All 4 dependencies are intertwined and need to be bumped together..

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/__init__.py
DylanRussell: Ack, set it to 1.36 if they opted into that falg..

instrumentation-genai/opentelemetry-instrumentation-vertexai/tests/requirements.oldest.txt
aabmass: Got it. Better to keep them here and bump, and make sure you increase the one in pyproject.toml too. There's two tox envs, one to test oldest deps and one to test newest.

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/__init__.py
DylanRussell: Ack, added that in

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/patch.py
DylanRussell: Alright I updated the code to put it in the context manager.. I had to put `handle_response` into a `finnally` block to ensure it executes even when the method raises an exception

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/patch.py
DylanRussell: Sure, done. I plan to eventually update the README.md with a better explanation. I want to get the new span added first though before doing that..

instrumentation-genai/opentelemetry-instrumentation-vertexai/tests/requirements.oldest.txt
DylanRussell: Should be all good now

instrumentation-genai/opentelemetry-instrumentation-vertexai/tests/test_chat_completions.py
DylanRussell: Switched to doing it this way

instrumentation-genai/opentelemetry-instrumentation-vertexai/tests/test_chat_completions.py
DylanRussell: When I re-ran the cassette tests w/o this before I was getting lots of duplicates. But when I did it now without this it isn't duplicating them, so not sure what the deal was.. Removed it for now

PR #3734, opened 2025-09-05
botocore: Add support for SNS semantic convention attribute aws.sns.topic.arn

lukeina2z:
# Description
The AWS SNS semantic convention defines the attribute aws.sns.topic.arn: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/aws.md#amazon-sns-attributes

Currently, this attribute is not set in the botocore instrumentation library.

Fixes # (issue)
This PR adds support for it by setting the attribute from both the request and response objects.

## Type of change

# How Has This Been Tested?
New unit tests added and passing.
Verified with:
tox -e spellcheck
tox -e ruff

Backward Compatibility
There is no risk of breaking existing functionality. This change only adds instrumentation for an additional AWS resource without modifying the existing behavior of the auto-instrumentation library.

- [x ] No.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/sns.py
+++ b/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/sns.py
@@ -24,6 +24,10 @@ from opentelemetry.instrumentation.botocore.extensions.types import (
     _AwsSdkCallContext,
     _AwsSdkExtension,
     _BotocoreInstrumentorContext,
+    _BotoResultT,
+)
+from opentelemetry.semconv._incubating.attributes.aws_attributes import (
+    AWS_SNS_TOPIC_ARN,
 )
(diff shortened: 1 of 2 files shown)


instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/sns.py
```diff
@@ -170,3 +177,16 @@ def before_service_call(
     ):
         if self._op:
             self._op.before_service_call(self._call_context, span)
+
+    def on_success(
```
xrmx, line 181: Isn't the attribute already set by extract_attributes? To say that maybe we don't even need to set this callback

xrmx: Ah I see, one is for request and the other for response

PR #3737, opened 2025-09-06
botocore: Add support for AWS Step Functions semantic convention attributes

lukeina2z:
AWS Step Functions defines two semantic convention attributes:

aws.step_functions.activity.arn
aws.step_functions.state_machine.arn

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/aws.md#amazon-step-functions-attributes

Currently, these attributes are not set in the botocore instrumentation library. This PR adds support for them by extracting values from both Request and Response objects.

Tests
Added new unit tests (passing).

Verified with:
tox -e spellcheck
tox -e ruff

Backward Compatibility
This change is backward compatible. It only adds instrumentation for additional AWS resources and does not modify existing behavior in the auto-instrumentation library.

Note
This PR depends on #3736. Since #3736 has not yet been merged, its changes are included here as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3507](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3507))
 - Fix documentation order of sections and headers for Django, Flask, MySQL, mysqlclient, psycopg, psycopg2, pymysql, sqlalchemy instrumentations.
   ([#3719](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3719))
+- `opentelemetry-instrumentation-asgi` Fixed an issue where FastAPI reports IP instead of URL.
+  ([#3670](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3670))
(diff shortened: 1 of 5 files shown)


instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/__init__.py
```diff
@@ -35,6 +35,7 @@ def loader():
     "bedrock-runtime": _lazy_load(".bedrock", "_BedrockRuntimeExtension"),
     "dynamodb": _lazy_load(".dynamodb", "_DynamoDbExtension"),
     "lambda": _lazy_load(".lmbd", "_LambdaExtension"),
+    "stepfunctions": _lazy_load(".sfns", "_StepFunctionsExtension"),
```
yiyuan-he, line 38: nit: Was there a reason we chose to make the naming here plural?

lukeina2z: The string literal comes from the AWS SDK, so we have to stick with it.  https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/stepfunctions.html

xrmx (approved):

PR #3757, opened 2025-09-15
Revise component owners

xrmx:
# Description

Drop from component owners people that have not partecipated in the last year and add an entry for `util/opentelemetry-util-genai`.

While at it sort the entries.


lzchen (approved):

xrmx:
@keith-decker if you are fine to be added as component owner for genai utils could you please approve?

PR #3732, opened 2025-09-04
GenAI Utils | Initial Helpers for Inference

keith-decker:
# Description

The GenAI Utils package will include boilerplate and helpers to standardize instrumentation for Generative AI. 
This PR provides APIs  to minimize the work needed to instrument genai libraries. It provides helpers for starting and stopping LLM invocations, it generates spans aligned with semconvs.  

Metrics and Events to come on future PRs

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [X] Basic unit test for starting and stopping LLM - Spans
- [X] Unit test for parent/child invocations

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [X] Followed the style guidelines of this project
- [X] Changelogs have been updated
- [X] Unit tests have been added
- [X] Documentation has been updated

PR #3768, opened 2025-09-18
GenAI Utils | Inference Type and Span Creation

keith-decker:
# Description

The GenAI Utils package will include boilerplate and helpers to standardize instrumentation for Generative AI. 
This PR provides APIs  to minimize the work needed to instrument genai libraries. It provides helpers for starting and stopping LLM invocations, it generates spans aligned with semconvs.  

Metrics and Events to come on future PRs

## Type of change

# How Has This Been Tested?

- [X] Basic unit test for starting and stopping LLM - Spans
- [X] Unit test for parent/child invocations

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,12 +20,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `opentelemetry-instrumentation`: botocore: Add support for AWS Secrets Manager semantic convention attribute
+- `opentelemetry-instrumentation-botocore`: Add support for AWS Secrets Manager semantic convention attribute
   ([#3765](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3765))
+- `opentelemetry-instrumentation-dbapi`: Add support for `commenter_options` in `trace_integration` function to control SQLCommenter behavior
+  ([#3743](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3743))
 - Add `rstcheck` to pre-commit to stop introducing invalid RST
   ([#3777](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3777))
 - `opentelemetry-exporter-credential-provider-gcp`: create this package which provides support for supplying your machine's Application Default 
   Credentials (https://cloud.google.com/docs/authentication/application-default-credentials) to the OTLP Exporters created automatically by OpenTelemetry Python's auto instrumentation. These credentials authorize OTLP traces to be sent to `telemetry.googleapis.com`. [#3766](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3766).
+- `opentelemetry-instrumentation-psycopg`: Add missing parameter `capture_parameters` to instrumentor.
+  ([#3676](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3676))
+- `opentelemetry-instrumentation-dbapi`: Adds sqlcommenter to documentation.
+  ([#3720](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3720))
 
 ## Version 1.37.0/0.58b0 (2025-09-11)
 
--- a/instrumentation-genai/opentelemetry-instrumentation-google-genai/examples/manual/main.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-google-genai/examples/manual/main.py
@@ -22,8 +22,7 @@ import os
 
 import google.genai
 
-# NOTE: OpenTelemetry Python Logs and Events APIs are in beta
-from opentelemetry import _events as otel_events
+# NOTE: OpenTelemetry Python Logs API is in beta
 from opentelemetry import _logs as otel_logs
 from opentelemetry import metrics as otel_metrics
 from opentelemetry import trace as otel_trace
@@ -40,7 +39,6 @@ from opentelemetry.instrumentation.google_genai import (
     GoogleGenAiSdkInstrumentor,
 )
 from opentelemetry.instrumentation.requests import RequestsInstrumentor
-from opentelemetry.sdk._events import EventLoggerProvider
 from opentelemetry.sdk._logs import LoggerProvider
 from opentelemetry.sdk._logs.export import BatchLogRecordProcessor
 from opentelemetry.sdk.metrics import MeterProvider
@@ -56,12 +54,11 @@ def setup_otel_tracing():
     )
 
 
-def setup_otel_logs_and_events():
+def setup_otel_logs():
     otel_logs.set_logger_provider(LoggerProvider())
     otel_logs.get_logger_provider().add_log_record_processor(
         BatchLogRecordProcessor(OTLPLogExporter())
     )
-    otel_events.set_event_logger_provider(EventLoggerProvider())
 
 
 def setup_otel_metrics():
@@ -77,7 +74,7 @@ def setup_otel_metrics():
 
 def setup_opentelemetry():
     setup_otel_tracing()
-    setup_otel_logs_and_events()
+    setup_otel_logs()
     setup_otel_metrics()
 
 
--- a/instrumentation-genai/opentelemetry-instrumentation-google-genai/pyproject.toml
+++ b/instrumentation-genai/opentelemetry-instrumentation-google-genai/pyproject.toml
@@ -37,7 +37,7 @@ classifiers = [
   "Programming Language :: Python :: 3.12"
 ]
 dependencies = [
-  "opentelemetry-api >=1.37, <2",
+  "opentelemetry-api ~=1.37",
   "opentelemetry-instrumentation >=0.52b1, <2",
   "opentelemetry-semantic-conventions >=0.52b1, <2"
 ]
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/CHANGELOG.md
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Record prompt and completion events regardless of span sampling decision.
   ([#3226](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3226))
+- Migrate off the deprecated events API to use the logs API
+  ([#3625](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3628))
 
 ## Version 2.1b0 (2025-01-18)
 
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/examples/manual/main.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/examples/manual/main.py
@@ -3,8 +3,8 @@ import os
 
 from openai import OpenAI
 
-# NOTE: OpenTelemetry Python Logs and Events APIs are in beta
-from opentelemetry import _events, _logs, metrics, trace
+# NOTE: OpenTelemetry Python Logs API is in beta
+from opentelemetry import _logs, metrics, trace
 from opentelemetry.exporter.otlp.proto.grpc._log_exporter import (
     OTLPLogExporter,
 )
@@ -15,7 +15,6 @@ from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import (
     OTLPSpanExporter,
 )
 from opentelemetry.instrumentation.openai_v2 import OpenAIInstrumentor
-from opentelemetry.sdk._events import EventLoggerProvider
 from opentelemetry.sdk._logs import LoggerProvider
 from opentelemetry.sdk._logs.export import BatchLogRecordProcessor
 from opentelemetry.sdk.metrics import MeterProvider
@@ -29,12 +28,11 @@ trace.get_tracer_provider().add_span_processor(
     BatchSpanProcessor(OTLPSpanExporter())
 )
 
-# configure logging and events
+# configure logging
 _logs.set_logger_provider(LoggerProvider())
 _logs.get_logger_provider().add_log_record_processor(
     BatchLogRecordProcessor(OTLPLogExporter())
 )
-_events.set_event_logger_provider(EventLoggerProvider())
 
 # configure metrics
 metrics.set_meter_provider(
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/pyproject.toml
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/pyproject.toml
@@ -25,9 +25,9 @@ classifiers = [
   "Programming Language :: Python :: 3.13",
 ]
 dependencies = [
-  "opentelemetry-api ~= 1.30",
-  "opentelemetry-instrumentation ~= 0.51b0",
-  "opentelemetry-semantic-conventions ~= 0.51b0"
+  "opentelemetry-api ~= 1.37",
+  "opentelemetry-instrumentation ~= 0.58b0",
+  "opentelemetry-semantic-conventions ~= 0.58b0"
 ]
 
 [project.optional-dependencies]
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/__init__.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/__init__.py
@@ -44,7 +44,7 @@ from typing import Collection
 
 from wrapt import wrap_function_wrapper
 
-from opentelemetry._events import get_event_logger
+from opentelemetry._logs import get_logger
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
 from opentelemetry.instrumentation.openai_v2.package import _instruments
 from opentelemetry.instrumentation.openai_v2.utils import is_content_enabled
@@ -73,12 +73,12 @@ class OpenAIInstrumentor(BaseInstrumentor):
             tracer_provider,
             schema_url=Schemas.V1_28_0.value,
         )
-        event_logger_provider = kwargs.get("event_logger_provider")
-        event_logger = get_event_logger(
+        logger_provider = kwargs.get("logger_provider")
+        logger = get_logger(
             __name__,
             "",
             schema_url=Schemas.V1_28_0.value,
-            event_logger_provider=event_logger_provider,
+            logger_provider=logger_provider,
         )
         meter_provider = kwargs.get("meter_provider")
         self._meter = get_meter(
@@ -94,7 +94,7 @@ class OpenAIInstrumentor(BaseInstrumentor):
             module="openai.resources.chat.completions",
             name="Completions.create",
             wrapper=chat_completions_create(
-                tracer, event_logger, instruments, is_content_enabled()
+                tracer, logger, instruments, is_content_enabled()
             ),
         )
 
@@ -102,7 +102,7 @@ class OpenAIInstrumentor(BaseInstrumentor):
             module="openai.resources.chat.completions",
             name="AsyncCompletions.create",
             wrapper=async_chat_completions_create(
-                tracer, event_logger, instruments, is_content_enabled()
+                tracer, logger, instruments, is_content_enabled()
             ),
         )
 
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/patch.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/patch.py
@@ -18,7 +18,8 @@ from typing import Optional
 
 from openai import Stream
 
-from opentelemetry._events import Event, EventLogger
+from opentelemetry._logs import Logger, LogRecord
+from opentelemetry.context import get_current
 from opentelemetry.semconv._incubating.attributes import (
     gen_ai_attributes as GenAIAttributes,
 )
@@ -26,6 +27,7 @@ from opentelemetry.semconv._incubating.attributes import (
     server_attributes as ServerAttributes,
 )
 from opentelemetry.trace import Span, SpanKind, Tracer
+from opentelemetry.trace.propagation import set_span_in_context
 
 from .instruments import Instruments
 from .utils import (
@@ -40,7 +42,7 @@ from .utils import (
 
 def chat_completions_create(
     tracer: Tracer,
-    event_logger: EventLogger,
+    logger: Logger,
     instruments: Instruments,
     capture_content: bool,
 ):
@@ -57,7 +59,7 @@ def chat_completions_create(
             end_on_exit=False,
         ) as span:
             for message in kwargs.get("messages", []):
-                event_logger.emit(message_to_event(message, capture_content))
+                logger.emit(message_to_event(message, capture_content))
 
             start = default_timer()
             result = None
@@ -65,16 +67,14 @@ def chat_completions_create(
             try:
                 result = wrapped(*args, **kwargs)
                 if is_streaming(kwargs):
-                    return StreamWrapper(
-                        result, span, event_logger, capture_content
-                    )
+                    return StreamWrapper(result, span, logger, capture_content)
 
                 if span.is_recording():
                     _set_response_attributes(
-                        span, result, event_logger, capture_content
+                        span, result, logger, capture_content
                     )
                 for choice in getattr(result, "choices", []):
-                    event_logger.emit(choice_to_event(choice, capture_content))
+                    logger.emit(choice_to_event(choice, capture_content))
 
                 span.end()
                 return result
@@ -98,7 +98,7 @@ def chat_completions_create(
 
 def async_chat_completions_create(
     tracer: Tracer,
-    event_logger: EventLogger,
+    logger: Logger,
     instruments: Instruments,
     capture_content: bool,
 ):
@@ -115,7 +115,7 @@ def async_chat_completions_create(
             end_on_exit=False,
         ) as span:
             for message in kwargs.get("messages", []):
-                event_logger.emit(message_to_event(message, capture_content))
+                logger.emit(message_to_event(message, capture_content))
 
             start = default_timer()
             result = None
@@ -123,16 +123,14 @@ def async_chat_completions_create(
             try:
                 result = await wrapped(*args, **kwargs)
                 if is_streaming(kwargs):
-                    return StreamWrapper(
-                        result, span, event_logger, capture_content
-                    )
+                    return StreamWrapper(result, span, logger, capture_content)
 
                 if span.is_recording():
                     _set_response_attributes(
-                        span, result, event_logger, capture_content
+                        span, result, logger, capture_content
                     )
                 for choice in getattr(result, "choices", []):
-                    event_logger.emit(choice_to_event(choice, capture_content))
+                    logger.emit(choice_to_event(choice, capture_content))
 
                 span.end()
                 return result
@@ -221,7 +219,7 @@ def _record_metrics(
 
 
 def _set_response_attributes(
-    span, result, event_logger: EventLogger, capture_content: bool
+    span, result, logger: Logger, capture_content: bool
 ):
     set_span_attribute(
         span, GenAIAttributes.GEN_AI_RESPONSE_MODEL, result.model
@@ -311,7 +309,7 @@ class StreamWrapper:
         self,
         stream: Stream,
         span: Span,
-        event_logger: EventLogger,
+        logger: Logger,
         capture_content: bool,
     ):
         self.stream = stream
@@ -320,7 +318,7 @@ class StreamWrapper:
         self._span_started = False
         self.capture_content = capture_content
 
-        self.event_logger = event_logger
+        self.logger = logger
         self.setup()
 
     def setup(self):
@@ -396,17 +394,13 @@ class StreamWrapper:
                 event_attributes = {
                     GenAIAttributes.GEN_AI_SYSTEM: GenAIAttributes.GenAiSystemValues.OPENAI.value
                 }
-
-                # this span is not current, so we need to manually set the context on event
-                span_ctx = self.span.get_span_context()
-                self.event_logger.emit(
-                    Event(
-                        name="gen_ai.choice",
+                context = set_span_in_context(self.span, get_current())
+                self.logger.emit(
+                    LogRecord(
+                        event_name="gen_ai.choice",
                         attributes=event_attributes,
                         body=body,
-                        trace_id=span_ctx.trace_id,
-                        span_id=span_ctx.span_id,
-                        trace_flags=span_ctx.trace_flags,
+                        context=context,
                     )
                 )
 
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
@@ -19,7 +19,7 @@ from urllib.parse import urlparse
 from httpx import URL
 from openai import NOT_GIVEN
 
-from opentelemetry._events import Event
+from opentelemetry._logs import LogRecord
 from opentelemetry.semconv._incubating.attributes import (
     gen_ai_attributes as GenAIAttributes,
 )
@@ -123,8 +123,8 @@ def message_to_event(message, capture_content):
         if tool_call_id:
             body["id"] = tool_call_id
 
-    return Event(
-        name=f"gen_ai.{role}.message",
+    return LogRecord(
+        event_name=f"gen_ai.{role}.message",
         attributes=attributes,
         body=body if body else None,
     )
@@ -156,8 +156,8 @@ def choice_to_event(choice, capture_content):
             message["content"] = content
         body["message"] = message
 
-    return Event(
-        name="gen_ai.choice",
+    return LogRecord(
+        event_name="gen_ai.choice",
         attributes=attributes,
         body=body,
     )
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/conftest.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/conftest.py
@@ -11,7 +11,6 @@ from opentelemetry.instrumentation.openai_v2 import OpenAIInstrumentor
 from opentelemetry.instrumentation.openai_v2.utils import (
     OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT,
 )
-from opentelemetry.sdk._events import EventLoggerProvider
 from opentelemetry.sdk._logs import LoggerProvider
 from opentelemetry.sdk._logs.export import (
     InMemoryLogExporter,
@@ -56,13 +55,11 @@ def fixture_tracer_provider(span_exporter):
     return provider
 
 
-@pytest.fixture(scope="function", name="event_logger_provider")
-def fixture_event_logger_provider(log_exporter):
+@pytest.fixture(scope="function", name="logger_provider")
+def fixture_logger_provider(log_exporter):
     provider = LoggerProvider()
     provider.add_log_record_processor(SimpleLogRecordProcessor(log_exporter))
-    event_logger_provider = EventLoggerProvider(provider)
-
-    return event_logger_provider
+    return provider
 
 
 @pytest.fixture(scope="function", name="meter_provider")
@@ -105,9 +102,7 @@ def vcr_config():
 
 
 @pytest.fixture(scope="function")
-def instrument_no_content(
-    tracer_provider, event_logger_provider, meter_provider
-):
+def instrument_no_content(tracer_provider, logger_provider, meter_provider):
     os.environ.update(
         {OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT: "False"}
     )
@@ -115,7 +110,7 @@ def instrument_no_content(
     instrumentor = OpenAIInstrumentor()
     instrumentor.instrument(
         tracer_provider=tracer_provider,
-        event_logger_provider=event_logger_provider,
+        logger_provider=logger_provider,
         meter_provider=meter_provider,
     )
 
@@ -125,16 +120,14 @@ def instrument_no_content(
 
 
 @pytest.fixture(scope="function")
-def instrument_with_content(
-    tracer_provider, event_logger_provider, meter_provider
-):
+def instrument_with_content(tracer_provider, logger_provider, meter_provider):
     os.environ.update(
         {OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT: "True"}
     )
     instrumentor = OpenAIInstrumentor()
     instrumentor.instrument(
         tracer_provider=tracer_provider,
-        event_logger_provider=event_logger_provider,
+        logger_provider=logger_provider,
         meter_provider=meter_provider,
     )
 
@@ -145,7 +138,7 @@ def instrument_with_content(
 
 @pytest.fixture(scope="function")
 def instrument_with_content_unsampled(
-    span_exporter, event_logger_provider, meter_provider
+    span_exporter, logger_provider, meter_provider
 ):
     os.environ.update(
         {OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT: "True"}
@@ -157,7 +150,7 @@ def instrument_with_content_unsampled(
     instrumentor = OpenAIInstrumentor()
     instrumentor.instrument(
         tracer_provider=tracer_provider,
-        event_logger_provider=event_logger_provider,
+        logger_provider=logger_provider,
         meter_provider=meter_provider,
     )
 
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/requirements.oldest.txt
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/requirements.oldest.txt
@@ -26,8 +26,8 @@ pytest-vcr==1.0.2
 pytest-asyncio==0.21.0
 wrapt==1.16.0
 opentelemetry-exporter-otlp-proto-http~=1.30
-opentelemetry-api==1.30  # when updating, also update in pyproject.toml
-opentelemetry-sdk==1.30  # when updating, also update in pyproject.toml
-opentelemetry-semantic-conventions==0.51b0  # when updating, also update in pyproject.toml
+opentelemetry-api==1.37  # when updating, also update in pyproject.toml
+opentelemetry-sdk==1.37  # when updating, also update in pyproject.toml
+opentelemetry-semantic-conventions==0.58b0  # when updating, also update in pyproject.toml
 
 -e instrumentation-genai/opentelemetry-instrumentation-openai-v2
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/test_async_chat_completions.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/test_async_chat_completions.py
@@ -24,9 +24,6 @@ from opentelemetry.semconv._incubating.attributes import (
     error_attributes as ErrorAttributes,
 )
 from opentelemetry.semconv._incubating.attributes import (
-    event_attributes as EventAttributes,
-)
-from opentelemetry.semconv._incubating.attributes import (
     gen_ai_attributes as GenAIAttributes,
 )
 from opentelemetry.semconv._incubating.attributes import (
@@ -797,7 +794,7 @@ async def async_chat_completion_multiple_tools_streaming(
 
 
 def assert_message_in_logs(log, event_name, expected_content, parent_span):
-    assert log.log_record.attributes[EventAttributes.EVENT_NAME] == event_name
+    assert log.log_record.event_name == event_name
     assert (
         log.log_record.attributes[GenAIAttributes.GEN_AI_SYSTEM]
         == GenAIAttributes.GenAiSystemValues.OPENAI.value
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/test_chat_completions.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/test_chat_completions.py
@@ -24,9 +24,6 @@ from opentelemetry.semconv._incubating.attributes import (
     error_attributes as ErrorAttributes,
 )
 from opentelemetry.semconv._incubating.attributes import (
-    event_attributes as EventAttributes,
-)
-from opentelemetry.semconv._incubating.attributes import (
     gen_ai_attributes as GenAIAttributes,
 )
 from opentelemetry.semconv._incubating.attributes import (
@@ -812,7 +809,7 @@ def chat_completion_multiple_tools_streaming(
 
 
 def assert_message_in_logs(log, event_name, expected_content, parent_span):
-    assert log.log_record.attributes[EventAttributes.EVENT_NAME] == event_name
+    assert log.log_record.event_name == event_name
     assert (
         log.log_record.attributes[GenAIAttributes.GEN_AI_SYSTEM]
         == GenAIAttributes.GenAiSystemValues.OPENAI.value
--- a/instrumentation-genai/opentelemetry-instrumentation-vertexai/examples/manual/main.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-vertexai/examples/manual/main.py
@@ -3,7 +3,7 @@ import vertexai
 from vertexai.generative_models import GenerativeModel
 
 # NOTE: OpenTelemetry Python Logs and Events APIs are in beta
-from opentelemetry import _events, _logs, trace
+from opentelemetry import _logs, trace
 from opentelemetry.exporter.otlp.proto.grpc._log_exporter import (
     OTLPLogExporter,
 )
@@ -11,7 +11,6 @@ from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import (
     OTLPSpanExporter,
 )
 from opentelemetry.instrumentation.vertexai import VertexAIInstrumentor
-from opentelemetry.sdk._events import EventLoggerProvider
 from opentelemetry.sdk._logs import LoggerProvider
 from opentelemetry.sdk._logs.export import BatchLogRecordProcessor
 from opentelemetry.sdk.trace import TracerProvider
@@ -28,7 +27,6 @@ _logs.set_logger_provider(LoggerProvider())
 _logs.get_logger_provider().add_log_record_processor(
     BatchLogRecordProcessor(OTLPLogExporter())
 )
-_events.set_event_logger_provider(EventLoggerProvider())
 
 # instrument VertexAI
 VertexAIInstrumentor().instrument()
--- a/instrumentation-genai/opentelemetry-instrumentation-vertexai/pyproject.toml
+++ b/instrumentation-genai/opentelemetry-instrumentation-vertexai/pyproject.toml
@@ -24,7 +24,7 @@ classifiers = [
   "Programming Language :: Python :: 3.12",
 ]
 dependencies = [
-  "opentelemetry-api >= 1.37",
+  "opentelemetry-api ~= 1.37",
   "opentelemetry-instrumentation ~= 0.58b0",
   # TODO https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3786: restrict
   # version after the first release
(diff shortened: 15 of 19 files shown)


keith-decker:
A cleaned up version of: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3732

- Flattened LLMInvocation to not use dictionary keys
- Removed the parent run_id, use OTEL context to manage parent/child relationships on spans
- Return the run_id on creation if one is not provided, in order to specify which span to end on finish

aabmass:
@keith-decker thing I was mentioning in the SIG:
Looks like you already figured out the context manager thing but figured I'd share what I wrote: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/ee947cec0c5bddb0b553c95cc6407cda81e26fe7/instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/patch.py#L101-L107
  
It's used for both [sync](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/ee947cec0c5bddb0b553c95cc6407cda81e26fe7/instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/patch.py#L165-L170) and [async](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/ee947cec0c5bddb0b553c95cc6407cda81e26fe7/instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/patch.py#L189-L194) wrapper, and would work for streaming as well.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,160 @@
...
+        invocation.output_tokens = attributes.pop("output_tokens")
+
+
+class TelemetryHandler:
+    """
+    High-level handler managing GenAI invocation lifecycles and emitting
+    them as spans, metrics, and events.
+    """
+
+    def __init__(self, emitter_type_full: bool = True, **kwargs: Any):
```
DylanRussell: nit: we can leave off emitter type full until the follow up PR that uses it

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,160 @@
...
+    invocation.attributes.
+    """
+    if "provider" in attributes:
+        invocation.provider = attributes.pop("provider")
+    if "response_model_name" in attributes:
+        invocation.response_model_name = attributes.pop("response_model_name")
+    if "response_id" in attributes:
+        invocation.response_id = attributes.pop("response_id")
+    if "input_tokens" in attributes:
+        invocation.input_tokens = attributes.pop("input_tokens")
```
DylanRussell: Why do we expect these strings in attributes ? They aren't part of the semantic conventions: https://github.com/DylanRussell/opentelemetry-python/blob/5c8269c2d33617cca26c63e8627aec89e0d79b17/opentelemetry-semantic-conventions/src/opentelemetry/semconv/_incubating/attributes/gen_ai_attributes.py#L448 ??

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    if not is_experimental_mode() or get_content_capturing_mode() not in (
+        ContentCapturingMode.SPAN_ONLY,
+        ContentCapturingMode.SPAN_AND_EVENT,
+    ):
+        return
+    message_parts: List[Dict[str, Any]] = [
+        asdict(message) for message in input_messages
+    ]
+    if message_parts:
+        span.set_attribute("gen_ai.input.messages", json.dumps(message_parts))
```
DylanRussell: nit: this constant is in that sem conv package

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -0,0 +1,124 @@
...
+
+    generation_parts: List[Dict[str, Any]] = [
+        asdict(generation) for generation in output_messages
+    ]
+    if generation_parts:
+        span.set_attribute(
+            "gen_ai.output.messages", json.dumps(generation_parts)
```
DylanRussell: nit: same thing

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
DylanRussell: Why do we need json.dumps it's already a dictionary ?

DylanRussell: I see the spec is sort of ambiguous on this:

> When the attribute is recorded on events, it MUST be recorded in structured form. When recorded on spans, it MAY be recorded as a JSON string if structured format is not supported and SHOULD be recorded in structured form otherwise.

keith-decker: I just finished testing locally, and these don't get recorded to the span unless they are a string. They will get recorded to an event though. 

I'll leave them as a json string for now

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: Corrected, these were not in semconvs when I started.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: These attributes are referring to `gen_ai.usage.input_tokens` on a span: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-spans.md.

If I am looking at this right, the link for the gen_ai_attribute is used for `gen_ai.token.type` on metrics? https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-metrics.md

Which we are planning on following up this PR with metrics added to the generators.

Any suggestions/thoughts on renaming those keys?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
DylanRussell: Hm that is weird. Don't see SDK dropping anything: https://github.com/open-telemetry/opentelemetry-python/blob/5c8269c2d33617cca26c63e8627aec89e0d79b17/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L845-L854

There's something mentioned on the API about not setting None as an attribute: https://github.com/open-telemetry/opentelemetry-python/blob/5c8269c2d33617cca26c63e8627aec89e0d79b17/opentelemetry-api/src/opentelemetry/trace/span.py#L88-L90 , not sure if that's the issue or not..

DylanRussell: Ok I see spans don't support complex attributes for some reason: https://github.com/open-telemetry/opentelemetry-python/blob/5c8269c2d33617cca26c63e8627aec89e0d79b17/opentelemetry-api/src/opentelemetry/attributes/__init__.py#L49-L53

keith-decker: When I do a debug trace, I have the value as a dictionary until this point: https://github.com/open-telemetry/opentelemetry-python/blob/5c8269c2d33617cca26c63e8627aec89e0d79b17/opentelemetry-api/src/opentelemetry/attributes/__init__.py#L286

Then the cleaned_value becomes None. and it gets dropped.

It looks like a dict is not a valid attribute type here: https://github.com/open-telemetry/opentelemetry-python/blob/5c8269c2d33617cca26c63e8627aec89e0d79b17/opentelemetry-api/src/opentelemetry/attributes/__init__.py#L79

DylanRussell: Ok lets leave it like this then

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -0,0 +1,124 @@
...
+        ContentCapturingMode.SPAN_AND_EVENT,
+    ):
+        return
+    message_parts: List[Dict[str, Any]] = [
+        asdict(message) for message in input_messages
+    ]
+    if message_parts:
+        span.set_attribute("gen_ai.input.messages", json.dumps(message_parts))
+
+    generation_parts: List[Dict[str, Any]] = [
```
DylanRussell: nit: just inline this list comprehension (here and above)  inside `span.set_attribute`?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
DylanRussell: So is the code guessing that `gen_ai.response.model` has been mistakenly added to attributes as `response.model`  ?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -69,3 +80,31 @@ class OutputMessage:
...
+@dataclass
+class LLMInvocation:
+    """
+    Represents a single LLM call invocation.
+    """
+
+    request_model: str
+    context_token: Optional[ContextToken] = None
+    span: Optional[Span] = None
+    parent_run_id: Optional[UUID] = None
```
DylanRussell: i don't see any code setting `parent_run_id` anymore

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,144 @@
...
+            tracer_provider,
+            schema_url=Schemas.V1_36_0.value,
+        )
+
+        self._generator = SpanGenerator(tracer=self._tracer)
+
+    def start_llm(
+        self,
+        request_model: str,
+        prompts: List[InputMessage],
```
DylanRussell: nit: can we standardize the word we use for `input/output_messages` everywhere, maybe just use `input_messages` and `output_messages` ? We have bunch of different terms being used in this PR (prompts, chat_generations etc.).

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: Done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
keith-decker: Thanks for catching that, removed.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: Done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -0,0 +1,121 @@
...
+            json.dumps([asdict(message) for message in input_messages]),
+        )
+    if output_messages:
+        span.set_attribute(
+            GenAI.GEN_AI_OUTPUT_MESSAGES,
+            json.dumps([asdict(message) for message in output_messages]),
+        )
+
+
+def _apply_finish_attributes(span: Span, invocation: LLMInvocation) -> None:
```
DylanRussell, line 109: why underscore these ? Is it b/c they are only for internal (to gen ai util package) use ?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -0,0 +1,121 @@
...
+    Span,
+)
+from opentelemetry.trace.status import Status, StatusCode
+from opentelemetry.util.genai.utils import (
+    ContentCapturingMode,
+    get_content_capturing_mode,
+    is_experimental_mode,
+)
+
+from .types import Error, InputMessage, LLMInvocation, OutputMessage
```
DylanRussell: nit (here and everywhere) - can we use absolute imports everywhere ? https://google.github.io/styleguide/pyguide.html#224-decision -- IDK if we follow any official style guide in this repo, but inside google we follow this guide which discourages relative imports.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: Yep, the generators apply attributes from the invocation to the span. I think the idea is to separate the span creation and management from the instrumentation library.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: Done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: @DylanRussell I've taken this suggestion and removed the whole mapping from a dict to the LLMInvocation. Now we are expecting the Instrumentation library to assemble an LLMInvocation and pass that to the handler. 

Future PRs will handle multiple types of invocations (toolinvocations, agentinvocation, etc) and the instrumentation library will need to assemble the appropriate invocation. 

How does this look?

keith-decker: See https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3768/files#diff-ffef25592f007d6454175a9d35bcd9e0092a401f7ba1f08604bbbb82cfe1fe87R134 for an example

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
DylanRussell: ok, sounds good

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
DylanRussell: > @DylanRussell I've taken this suggestion and removed the whole mapping from a dict to the LLMInvocation. Now we are expecting the Instrumentation library to assemble an LLMInvocation and pass that to the handler.

Yeah that sounds good to me, we should make it clear that the instrumentation library is just expected to fill in the "data" fields and leave span creation stuff to the helper ..

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,111 @@
...
+        self._tracer = get_tracer(
+            __name__,
+            __version__,
+            tracer_provider,
+            schema_url=Schemas.V1_36_0.value,
+        )
+
+        self._generator = SpanGenerator(tracer=self._tracer)
+
+    def start_llm(
```
DylanRussell, line 99: Can we have this be a `ContextManager` like Aaron was saying ?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/generators.py
```diff
@@ -0,0 +1,118 @@
...
+    ):
+        self._tracer: Tracer = tracer or trace.get_tracer(__name__)
+
+        # Store the active span and its context attachment token
+        self._active: Dict[UUID, tuple[Span, ContextToken]] = {}
+
+    def start(self, invocation: LLMInvocation):
+        # Create a span and attach it as current; keep the token to detach later
+        span = self._tracer.start_span(
+            name=f"{GenAI.GenAiOperationNameValues.CHAT.value} {invocation.request_model}",
```
DylanRussell: Maybe raise an error if `invocation.request_model` is not set ?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/generators.py
```diff
@@ -0,0 +1,118 @@
...
+    set_span_in_context,
+)
+from opentelemetry.util.genai.span_utils import (
+    _apply_error_attributes,
+    _apply_finish_attributes,
+)
+from opentelemetry.util.genai.types import Error, LLMInvocation
+
+# Type alias matching the token type expected by opentelemetry.context.detach
+ContextToken: TypeAlias = Token[otel_context.Context]
```
pmcollins: Looks like `Token` is not a generic type?
```suggestion
ContextToken: TypeAlias = Token
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/generators.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+
+    def __init__(
+        self,
+        tracer: Optional[Tracer] = None,
+    ):
+        self._tracer: Tracer = tracer or trace.get_tracer(__name__)
+
+        # Store the active span and its context attachment token
+        self._active: Dict[UUID, tuple[Span, ContextToken]] = {}
```
pmcollins: Maybe this field is vestigial -- is it still being used?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
DylanRussell: I think it's OK to start without one and get this PR merged for now

util/opentelemetry-util-genai/src/opentelemetry/util/genai/generators.py
keith-decker: Done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: @DylanRussell Took a stab at this, let me know what you think

util/opentelemetry-util-genai/src/opentelemetry/util/genai/generators.py
DylanRussell: Sorry I just noticed that the vertex AI instrumentation has a test for model being empty // not being passed, so maybe that is a valid case (it still wants the span to be created, it just uses an empty string like you had before)

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,114 @@
...
+    def __init__(self, **kwargs: Any):
+        tracer_provider = kwargs.get("tracer_provider")
+        self._tracer = get_tracer(
+            __name__,
+            __version__,
+            tracer_provider,
+            schema_url=Schemas.V1_36_0.value,
+        )
+
+        self._generator = SpanGenerator(tracer=self._tracer)
```
pmcollins: Do we want to let folks pass in a generator? Also, related -- looks like `self._tracer` can be local as the value is immediately passed into a new generator. Should the generator handle creating its own tracer/logger/meter?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,114 @@
...
+        Only set data attributes on the invocation object, do not modify the span or context.
+
+        Starts the span on entry. On normal exit, finalizes the invocation and ends the span.
+        If an exception occurs inside the context, marks the span as error, ends it, and
+        re-raises the original exception.
+        """
+        self._generator.start(invocation)
+        try:
+            yield invocation
+        except BaseException as exc:  # noqa: B902 - ensure we capture all exceptions incl. SystemExit, KeyboardInterrupt
```
DylanRussell: I think `except Exception` is better, I see code doing that in various places , `BaseException` catches a few extra things that I'm not sure we'd want to catch: https://stackoverflow.com/questions/44655854/what-is-the-difference-between-except-and-except-baseexception

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
DylanRussell: I think we do want to let all the providers to be passed in here, like it is now. Instrumentations will directly create and use this `TelemetryHandler` object, which means they need to pass the `providers` that get passed into them onto this thing

DylanRussell: Actually I think your right self._tracer and getting the tracer_provider from the kwargs doesn't need to be here, it should be in the SpanGenerator

DylanRussell: We just need to pass kwargs through here

util/opentelemetry-util-genai/src/opentelemetry/util/genai/generators.py
keith-decker: Good catch. It is not used after converting to context, I've removed it.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/generators.py
keith-decker: Turns out we don't even need the type here. it got cleaned up after removing the dead _active class var.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: Moved the building of the tracer to the generator

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: Corrected!

util/opentelemetry-util-genai/src/opentelemetry/util/genai/generators.py
keith-decker: Removed the throwing of an error since it's a required field on the dataclass. it can be set as an empty string to satisfy that use case by Vertex.

DylanRussell: Sounds good

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
DylanRussell: Thanks

DylanRussell:
Looks good overall. I see one type check error:
```
/home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/instrumentation-genai/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/span_manager.py
  /home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/instrumentation-genai/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/span_manager.py:34:5 - error: Type of "children" is partially unknown
    Type of "children" is "list[Unknown]" (reportUnknownVariableType)
```

keith-decker:
> Looks good overall. I see one type check error:
> 
> ```
> /home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/instrumentation-genai/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/span_manager.py
>   /home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/instrumentation-genai/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/span_manager.py:34:5 - error: Type of "children" is partially unknown
>     Type of "children" is "list[Unknown]" (reportUnknownVariableType)
> ```

This is part of the langchain instrumentation that just merged to main. Should we open a separate PR to keep it out of the utils PR?

DylanRussell:
I think `typecheck` runs against everything that's opted into `typecheck`, so this check will block/fail on lots of PRs.

It should be fixed in this PR or a separate one..

util/opentelemetry-util-genai/src/opentelemetry/util/genai/generators.py
```diff
@@ -0,0 +1,117 @@
...
+        raise NotImplementedError
+
+    def finish(self, invocation: LLMInvocation) -> None:
+        raise NotImplementedError
+
+    def error(self, error: Error, invocation: LLMInvocation) -> None:
+        raise NotImplementedError
+
+
+class SpanGenerator(BaseTelemetryGenerator):
```
aabmass: Are you planning to add other implementations?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,129 @@
...
+
+import time
+from contextlib import contextmanager
+from typing import Any, Iterator, Optional
+
+from opentelemetry.util.genai.generators import SpanGenerator
+from opentelemetry.util.genai.types import Error, LLMInvocation
+
+
+class TelemetryHandler:
```
aabmass, line 85: I'm trying to understand the reason for this and BaseTelemetryGenerator. They seem to have nearly the same interface.

util/opentelemetry-util-genai/tests/test_utils.py
```diff
@@ -81,3 +99,200 @@ def test_get_content_capturing_mode_raises_exception_on_invalid_envvar(
...
+        cls.span_exporter = InMemorySpanExporter()
+        tracer_provider = TracerProvider()
+        tracer_provider.add_span_processor(
+            SimpleSpanProcessor(cls.span_exporter)
+        )
+        trace.set_tracer_provider(tracer_provider)
+
+    def setUp(self):
+        self.span_exporter = self.__class__.span_exporter
+        self.span_exporter.clear()
```
aabmass: I would recommend just setting moving everything from `setUpClass` into here. Then you don't have to clear any state between tests

util/opentelemetry-util-genai/tests/test_utils.py
```diff
@@ -81,3 +99,200 @@ def test_get_content_capturing_mode_raises_exception_on_invalid_envvar(
...
+        child_invocation = LLMInvocation(
+            request_model="child-model",
+            input_messages=[message],
+            provider="test-provider",
+        )
+
+        with self.telemetry_handler.llm(parent_invocation):
+            with self.telemetry_handler.llm(child_invocation):
+                # Stop child first by exiting inner context
+                child_invocation.output_messages = [chat_generation]
```
aabmass, line 241: This feels like a bit of an anti pattern to me. I wouldn't have figured out that I need to set the output messages on the invocation object so that the reference can be read when the ctx manager exits.

I think a bit more natural to python/OTel would be to return something from `__enter__()` that can be written to similar to our `use_span()` and many other stdlib ctx managers e.g.

```py
with open(path, "w") as file:
  file.write(bytes)
```

WDYT?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/generators.py
keith-decker: Yes, we have SpanMetricGenerator, SpanMetricEventGenerator planned.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: The telemetry handler will handle which generator to use. It will also be the entry point for start_tool, stop_tool, start_agent, stop_agent and other types of invocations.

util/opentelemetry-util-genai/tests/test_utils.py
keith-decker: Apologies, that test was shorted as much as possible for just testing the parent and child. 

It does actually return the invocation to use. I can update the test file to make this more clear like below:
```python       
with self.telemetry_handler.llm(parent_invocation) as invocation1:
    # Perform things here, calling a model, tool, processing, etc.
    with self.telemetry_handler.llm(child_invocation) as invocation2:
        # Perform things here, calling a model, tool, processing, etc.
        # append output messages before exiting the context
        invocation2.output_messages = [chat_generation]
    # Then stop parent by exiting outer context
    invocation1.output_messages = [chat_generation]
```

keith-decker: Would you like to see the `with ..handler.llm() as invocation:` build a blank invocation? then inside the with block the user would start adding data attributes?

Like so:
```py
        with self.telemetry_handler.llm() as parent_invocation:
            parent_invocation.request_model = "parent-model"
            parent_invocation.input_messages = [message]
            parent_invocation.provider = "test-provider"
            # Perform things here, calling a tool, processing, etc.
            with self.telemetry_handler.llm() as child_invocation:
                child_invocation.request_model = "child-model"
                child_invocation.input_messages = [message]
                child_invocation.provider = "test-provider"
                # Perform things here, calling a tool, processing, etc.
                # Stop child first by exiting inner context
                child_invocation.output_messages = [chat_generation]
            # Then stop parent by exiting outer context
            parent_invocation.output_messages = [chat_generation]
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
DylanRussell: maybe just have the TelemetryHandler own the span / event / metric providers (all 3), and get rid of the `Generator` concept ?

util/opentelemetry-util-genai/tests/test_utils.py
aabmass: Got it thanks, please just update the test

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
aabmass: WDYT @keith-decker? Could we add it in later if needed?

util/opentelemetry-util-genai/tests/test_utils.py
keith-decker: Done!

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: I've removed the generator from this PR while we work on the POC as discussed in a previous conversation

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -0,0 +1,136 @@
...
+    span.set_attribute(
+        GenAI.GEN_AI_OPERATION_NAME, GenAI.GenAiOperationNameValues.CHAT.value
+    )
+    if request_model:
+        span.set_attribute(GenAI.GEN_AI_REQUEST_MODEL, request_model)
+    if provider is not None:
+        # TODO: clean provider name to match GenAiProviderNameValues?
+        span.set_attribute(GenAI.GEN_AI_PROVIDER_NAME, provider)
+
+    finish_reasons: List[str] = []
```
DylanRussell: nit: could be a list comprehension

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -69,3 +79,48 @@ class OutputMessage:
...
+    input_messages: List[InputMessage] = field(
+        default_factory=_new_input_messages
+    )
+    output_messages: List[OutputMessage] = field(
+        default_factory=_new_output_messages
+    )
+    provider: Optional[str] = None
+    response_model_name: Optional[str] = None
+    response_id: Optional[str] = None
+    input_tokens: Optional[AttributeValue] = None
```
DylanRussell: Why are these AttributeValue and not int ? https://github.com/open-telemetry/semantic-conventions/blob/75ea2fcd4a1fe21524860677579ddf9abf9427d8/model/gen-ai/registry.yaml#L180-L185 says they should just be ints ?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -69,3 +79,48 @@ class OutputMessage:
...
+    output_tokens: Optional[AttributeValue] = None
+    attributes: Dict[str, Any] = field(default_factory=_new_str_any_dict)
```
DylanRussell, line 116: is this function necessary can't you just pass `dict` ? Same above with `list`

keith-decker: if I pass a dict, pyright does not like it. 

opentelemetry-python-contrib/util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py:119:5 - error: Type of "attributes" is partially unknown

I ended up having to make that function in order to satisfy pyright.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
keith-decker: corrected.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
keith-decker: @DylanRussell thoughts?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -0,0 +1,134 @@
...
+            json.dumps([asdict(message) for message in input_messages]),
+        )
+    if output_messages:
+        span.set_attribute(
+            GenAI.GEN_AI_OUTPUT_MESSAGES,
+            json.dumps([asdict(message) for message in output_messages]),
+        )
+
+
+def _maybe_set_span_extra_attributes(
```
DylanRussell: Why is this `maybe` ? Can we rename to `_set_span_extra_attributes` ?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -0,0 +1,134 @@
...
+) -> None:
+    """Apply attributes shared by finish() and error() and compute metrics.
+
+    Returns (genai_attributes) for use with metrics.
+    """
+    request_model = invocation.request_model
+    provider = invocation.provider
+    span.update_name(
+        f"{GenAI.GenAiOperationNameValues.CHAT.value} {request_model}"
+    )
```
DylanRussell, line 51: nit: can we add a `.strip()` here to strip whitespace when the request model is empty (vertex test depends on that) ?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -0,0 +1,134 @@
...
+            GenAI.GEN_AI_RESPONSE_FINISH_REASONS, finish_reasons
+        )
+
+    if invocation.response_model_name is not None:
+        span.set_attribute(
+            GenAI.GEN_AI_RESPONSE_MODEL, invocation.response_model_name
+        )
+    if invocation.response_id is not None:
+        span.set_attribute(GenAI.GEN_AI_RESPONSE_ID, invocation.response_id)
+    if isinstance(invocation.input_tokens, (int, float)):
```
DylanRussell: we can get rid of these `isinstance` checks now

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
DylanRussell: Yeah I did a bit of searching but didn't find anything better. It's OK as is I suppose.

DylanRussell:
This looks good to me now. Just added a few comments

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: Done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: Sure, done!

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: Yep, converted them back to the null check.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,180 @@
...
+            __name__,
+            __version__,
+            tracer_provider,
+            schema_url=Schemas.V1_36_0.value,
+        )
+        self._tracer: Tracer = tracer or trace.get_tracer(__name__)
+
+    def start_llm(
+        self,
+        invocation: LLMInvocation,
```
pmcollins, line 101: Another option here would be to use _immutable_ types, e.g. breaking out `LLMInvocation` into an `LLMRequest` and an `LLMResponse`.  It's possible that span and token don't belong in either of these types, but maybe they can be returned in a tuple or a wrapper.

nagkumar91 (approved):

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: Pablo and I explored this in another branch, and decided it felt awkward with the context manager. We'll explore other options as more types come online.

keith-decker:
@aabmass genai-file upload on python 3.9 seems to be flakey. I've seen it fail occasionally, and I don't have the ability to restart the test.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
pmcollins: Thank you Keith -- we definitely hit awkwardness with the context manager, but to me that's a signal. Relying on a mutation-based API has pitfalls: order dependence, late validation, putting the onus on the caller to know what fields to set and when, and fragility in async/streaming cases. I don't want to block this PR and I want folks to be able to benefit from this utility, but I do think this design will be harder to use and maintain over time. If others share these concerns, perhaps they can be addressed later -- WDYT?

util/opentelemetry-util-genai/tests/test_utils.py
aabmass: @keith-decker is this fixed?

aabmass:
Sorry about that tracking here https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3812 (no idea why)

util/opentelemetry-util-genai/tests/test_utils.py
keith-decker: @aabmass if I move this code from the setUpClass to here, tests that run after each other will not pass their span count asserts. 

I was unable to get the clearing of spans to work. Do we have any examples of this?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -69,3 +78,48 @@ class OutputMessage:
...
+    Represents a single LLM call invocation. When creating an LLMInvocation object,
+    only update the data attributes. The span and context_token attributes are
+    set by the TelemetryHandler.
+    """
+
+    request_model: str
+    context_token: Optional[ContextToken] = None
+    span: Optional[Span] = None
+    start_time: float = field(default_factory=time.time)
+    end_time: Optional[float] = None
```
aabmass: Looks like these are only used in tests. Do we need them?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class TelemetryHandler:
+    """
+    High-level handler managing GenAI invocation lifecycles and emitting
+    them as spans, metrics, and events.
+    """
+
+    def __init__(self, **kwargs: Any):
+        tracer_provider = kwargs.get("tracer_provider")
```
aabmass: Any reason not to do this throughout?
```suggestion
    def __init__(self, *, tracer_provider: TracerProvider | None = None):
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,180 @@
...
+        tracer = get_tracer(
+            __name__,
+            __version__,
+            tracer_provider,
+            schema_url=Schemas.V1_36_0.value,
+        )
+        self._tracer: Tracer = tracer or trace.get_tracer(__name__)
```
aabmass: Don't think this is needed, get_tracer() always returns a Tracer. It's also the same function as called on L93

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,180 @@
...
+    # Or, in case of error
+    handler.fail_llm(invocation, Error(type="...", message="..."))
+"""
+
+import time
+from contextlib import contextmanager
+from typing import Any, Iterator, Optional
+
+from opentelemetry import context as otel_context
+from opentelemetry import trace
```
aabmass: nit prefer to either one of `trace.foo` everywhere or import the objects on L71.

util/opentelemetry-util-genai/tests/test_utils.py
keith-decker: @aabmass Disregard, I got it working during the SIG meeting.

Fixed!

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,180 @@
...
+        self.start_llm(invocation)
+        try:
+            yield invocation
+        except Exception as exc:
+            self.fail_llm(invocation, Error(message=str(exc), type=type(exc)))
+            raise
+        self.stop_llm(invocation)
+
+
+def get_telemetry_handler(**kwargs: Any) -> TelemetryHandler:
```
aabmass: Same comment here on kwargs

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,180 @@
...
+            invocation = LLMInvocation(
+                request_model="",
+            )
+        self.start_llm(invocation)
+        try:
+            yield invocation
+        except Exception as exc:
+            self.fail_llm(invocation, Error(message=str(exc), type=type(exc)))
+            raise
+        self.stop_llm(invocation)
```
aabmass, line 163: Is it possible to use the [`use_span()` decorator](https://github.com/open-telemetry/opentelemetry-python/blob/daae23bc7f5514003d1e262730c409875d12d77d/opentelemetry-api/src/opentelemetry/trace/__init__.py#L565-L619) to manage most of the context and exception handling?

It might be hard if you need the imperative attach/detach API of start/stop/fail.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+    request_model = invocation.request_model
+    provider = invocation.provider
+    span.update_name(
+        f"{GenAI.GenAiOperationNameValues.CHAT.value} {request_model}".strip()
+    )
+    span.set_attribute(
+        GenAI.GEN_AI_OPERATION_NAME, GenAI.GenAiOperationNameValues.CHAT.value
+    )
+    if request_model:
```
aabmass: nit can you make this pattern consistent throughout here?

```py
foo = invocation.foo
if foo: ...

# vs 
if invocation.foo: ...
```

i personally prefer the latter

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -0,0 +1,134 @@
...
+) -> None:
+    if not is_experimental_mode() or get_content_capturing_mode() not in (
+        ContentCapturingMode.SPAN_ONLY,
+        ContentCapturingMode.SPAN_AND_EVENT,
+    ):
+        return
+    if input_messages:
+        span.set_attribute(
+            GenAI.GEN_AI_INPUT_MESSAGES,
+            json.dumps([asdict(message) for message in input_messages]),
```
aabmass: Can you use this function which handles encoding bytes and removing whitespace? https://github.com/open-telemetry/opentelemetry-python-contrib/blob/9f393c75f685b7c3920588db70f707e2b870e19c/util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py#L76

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -0,0 +1,134 @@
...
+        span, invocation.input_messages, invocation.output_messages
+    )
+    _set_span_extra_attributes(span, invocation.attributes)
+
+
+def _apply_error_attributes(span: Span, error: Error) -> None:
+    """Apply status and error attributes common to error() paths."""
+    span.set_status(Status(StatusCode.ERROR, error.message))
+    if span.is_recording():
+        span.set_attribute(ErrorAttributes.ERROR_TYPE, error.type.__qualname__)
```
aabmass, line 122: This looks a bit different from what we do in `use_span()` which I believe follows the spec

https://github.com/open-telemetry/opentelemetry-python/blob/daae23bc7f5514003d1e262730c409875d12d77d/opentelemetry-api/src/opentelemetry/trace/__init__.py#L593-L615

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        if invocation is None:
+            invocation = LLMInvocation(
+                request_model="",
+            )
+        self.start_llm(invocation)
+        try:
+            yield invocation
+        except Exception as exc:
+            self.fail_llm(invocation, Error(message=str(exc), type=type(exc)))
```
aabmass, line 161: Could we just pass through the exception instead of having this wrapper class?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -0,0 +1,134 @@
...
+        return
+    if input_messages:
+        span.set_attribute(
+            GenAI.GEN_AI_INPUT_MESSAGES,
+            json.dumps([asdict(message) for message in input_messages]),
+        )
+    if output_messages:
+        span.set_attribute(
+            GenAI.GEN_AI_OUTPUT_MESSAGES,
+            json.dumps([asdict(message) for message in output_messages]),
```
aabmass: same comment on json

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+
+
+def _set_span_extra_attributes(
+    span: Span,
+    attributes: Dict[str, Any],
+) -> None:
+    for key, value in attributes.items():
+        span.set_attribute(key, value)
```
aabmass: There's already a `span.set_attributes()` function to do this https://github.com/open-telemetry/opentelemetry-python/blob/daae23bc7f5514003d1e262730c409875d12d77d/opentelemetry-api/src/opentelemetry/trace/span.py#L81

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -0,0 +1,180 @@
...
+        """Fail an LLM invocation and end its span with error status."""
+        invocation.end_time = time.time()
+        if invocation.context_token is None or invocation.span is None:
+            # TODO: Provide feedback that this invocation was not started
+            return invocation
+
+        _apply_error_attributes(invocation.span, error)
+        # Detach context and end span
+        otel_context.detach(invocation.context_token)
+        invocation.span.end()
```
aabmass, line 138: if this code doesn't get called the span will never get exported. we see `otel_context.detach()` fail in some cases with async generators https://github.com/open-telemetry/opentelemetry-python/issues/2606

anyway my recommendation is to put this in finally (or use the use_span() manager with end_on_exit=True)

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
keith-decker: We don't need them, we rely on the span start and end time

Corrected

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: Updated the entrypoints to remove kwargs

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: it doesn't look like it's needed. Removed since we pass trace_provider through the init now.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: cleaned up imports

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: fixed

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: Corrected

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: yep! no longer using json.dumps here.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: updated to set_status inside is_recording. See comment on the helper function on why it's still here.

I followed through the exception handling in use_span, and it doesn't seem to set the error attributes preset here: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-agent-spans.md

We can remove this helper function if we don't think they are necessary, and fallback to the use_span exception handling.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: See rework of context_manager

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: refactored!

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: Thank you! helper removed, and using the span function now

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: I think this is resolved in the refactor for using use_span, and no longer managing the context manually

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: The whole context manager has been reworked to use native use_span from the sdk. 

this should delegate context management and error handling to the sdk, and still works both in context manager and explicit start_llm and stop_llm calls.

aabmass: Discussed offline

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
aabmass: Discussed offline

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
aabmass: Discussed offline. I don't 100% understand the error handling in semantic conventions if im honest. We should revisit or get Liudmila's help here when back in office

PR #3772, opened 2025-09-22
Update google genai instrumentation to work with latest semantic convention, allow for uploading content.

argaj:
# Description

Accomodation of semantic conventions changes made in [open-telemetry/semantic-conventions#2179](https://github.com/open-telemetry/semantic-conventions/pull/2179).

Also, when logging the completion details, the upload hook is called, so if user has configured `OTEL_INSTRUMENTATION_GENAI_UPLOAD_BASE_PATH` and `OTEL_INSTRUMENTATION_GENAI_UPLOAD_HOOK` env vars, completion details will also be logged as refs (if appropriate capture content env var is set).

## Type of change

# How Has This Been Tested?

Unit tests, manual tests.

No

# Checklist:


instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -33,18 +35,33 @@
...
+from opentelemetry.trace.span import Span
+from opentelemetry.util.genai.types import ContentCapturingMode
+from opentelemetry.util.genai.upload_hook import load_upload_hook
 
 from .allowlist_util import AllowList
 from .custom_semconv import GCP_GENAI_OPERATION_CONFIG
 from .dict_util import flatten_dict
 from .flags import is_content_recording_enabled
+from .message import (
+    InputMessage,
```
DylanRussell: I would import these from the gen ai utils package where they are defined (from opentelemetry.util.genai.types import ...)

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -226,12 +241,42 @@ def _wrapped_config_with_tools(
...
+    return config.system_instruction
+
+
+def _create_completion_details_attributes(
+    input_messages: list[InputMessage],
+    output_messages: list[OutputMessage],
+    system_instruction: Union[InputMessage, None],
+):
+    attributes = {
+        "gen_ai.input.messages": json.dumps(
```
DylanRussell: these strings can be imported as constants (from opentelemetry.semconv._incubating.attributes import (
    gen_ai_attributes as GenAIAttributes,
))

DylanRussell: we only want to json.dumps stuff if it's going into a span. if it goes into an event (log) we shouldn't

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -373,6 +440,56 @@ def _maybe_update_error_type(self, response: GenerateContentResponse):
...
+        ]:
+            span = trace.get_current_span()
+            span.set_attributes(completion_details_attributes)
+        if self._content_recording_enabled in [
+            ContentCapturingMode.EVENT_ONLY,
+            ContentCapturingMode.SPAN_AND_EVENT,
+        ]:
+            attributes.update(completion_details_attributes)
+            event = Event(name="gen_ai.completion.details", attributes=attributes)
+            hook = load_upload_hook()
```
aabmass: I think the instrumentor class should do a DI pattern for this. It probably should accept an optional upload hook as a parameter to `instrument()`, otherwise use `load_upload_hook()`.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -373,6 +440,56 @@ def _maybe_update_error_type(self, response: GenerateContentResponse):
...
+        if self._content_recording_enabled in [
+            ContentCapturingMode.SPAN_ONLY,
+            ContentCapturingMode.SPAN_AND_EVENT,
+        ]:
+            span = trace.get_current_span()
+            span.set_attributes(completion_details_attributes)
+        if self._content_recording_enabled in [
+            ContentCapturingMode.EVENT_ONLY,
+            ContentCapturingMode.SPAN_AND_EVENT,
+        ]:
```
aabmass: @DylanRussell should we still do this if there is an upload hook? One practical concern for GCP logging exporter is that if you include the JSON in the log directly and it exceeds the size limit (256kB) the whole log gets dropped. Since the upload hook works on the same log, that would mean everything gets dropped.

Alternatively, maybe we could emit two separate logs to avoid this problem

DylanRussell: The user can both configure whether they want the upload and whether they want the content in the log right ? So maybe lets just leave it to the user and recommend one way or the other ? Probably should discuss it more internally..

aabmass: Ah thanks I think you're right [from semconv](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/gen-ai/gen-ai-spans.md#uploading-content-to-external-storage)

> The hook SHOULD operate independently of the opt-in flags that control capturing of
`gen_ai.system_instructions`, `gen_ai.input.messages`, and `gen_ai.output.messages`.

I kind of read this the opposite way at first but it makes sense.

aabmass: So the code should always call the hook

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/instrumentor.py
```diff
@@ -58,8 +59,10 @@ def _instrument(self, **kwargs: Any):
             event_logger_provider=event_logger_provider,
             meter_provider=meter_provider,
         )
+        completion_hook = kwargs.get("completion_hook") or load_completion_hook()
```
DylanRussell: is this used at all ? you should call it at the end with the span, event etc. etc. https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3799 -- check this PR out if u want to take a look at how I did it (and tested it)

argaj: @aabmass suggested to inject the completion_hook in instrumentor, this is later passed to InstrumentorHelper, which calls completion_hook in _maybe_log_completion_details.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
argaj: So, IIUC, the hook operates independently, is called always (potentially a noop), potentially stamps the attributes with refs, and if span/event logging is enabled, that then is logged, yes?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/instrumentor.py
DylanRussell: Ack I see it now. LGTM

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
aabmass: @DylanRussell can you share the code you did since you had a good look at the specs?

I think we would always create the event and span as well and pass them to the hook. But just not add the content directly to them depending on the environment variable.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/message.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class Role(str, Enum):
+    SYSTEM = "system"
+    USER = "user"
+    ASSISTANT = "assistant"
+    TOOL = "tool"
+
+
+class Base64JsonEncoder(json.JSONEncoder):
```
DylanRussell: this can be imported from the gen ai utils module now

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/flags.py
```diff
@@ -12,12 +12,25 @@
...
 # limitations under the License.
 
-import os
+from os import environ
+from typing import Union
 
-_CONTENT_RECORDING_ENV_VAR = (
-    "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT"
+from opentelemetry.instrumentation._semconv import _StabilityMode
+from opentelemetry.util.genai.environment_variables import (
```
aabmass, line 19: `opentelemetry-util-genai` needs to be added as a dep in pyproject.toml

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/message.py
```diff
@@ -0,0 +1,166 @@
...
+
+def to_output_messages(
+    *,
+    candidates: list[genai_types.Candidate],
+) -> list[OutputMessage]:
+    def content_to_output_message(
+        candidate: genai_types.Candidate,
+    ) -> OutputMessage | None:
+        if not candidate.content:
+            return None
```
DylanRussell, line 77: I'd just return an empty list in this case. the upload hook always expects a list, and uses an empty list to denote no upload.

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_streaming[logcontent-gemini-2.5-flash-vertexaiapi-async].yaml
```diff
@@ -0,0 +1,123 @@
+interactions:
+- request:
+    body: |-
+      {
```
Comment (DylanRussell) [inline instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_streaming[logcontent-gemini-2.5-flash-vertexaiapi-async].yaml]: Mind adding a test for the completion hook somewhere ? https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation-genai/opentelemetry-instrumentation-vertexai/tests/test_function_calling_experimental.py#L262 -- this is how I did it for vertex AI

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/requirements.oldest.txt
```diff
@@ -20,12 +20,13 @@ pytest-asyncio==0.21.0
...
-google-genai==1.0.0
-opentelemetry-api==1.31.1
-opentelemetry-sdk==1.31.1
-opentelemetry-semantic-conventions==0.52b1
-opentelemetry-instrumentation==0.52b1
+google-genai==1.32.0
+opentelemetry-api==1.37.0
+opentelemetry-sdk==1.37.0
+opentelemetry-semantic-conventions==0.58b0
+opentelemetry-instrumentation==0.58b0
```
DylanRussell, line 27: Make sure to update to these versions in pyproject.toml too

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/utils/test_tool_call_wrapper.py
```diff
@@ -278,3 +284,48 @@ def somefunction(arg=None):
...
+
+                    span = self.otel.get_span_named(
+                        "execute_tool somefunction"
+                    )
+
+                    if mode in [
+                        ContentCapturingMode.NO_CONTENT,
+                        ContentCapturingMode.EVENT_ONLY,
+                    ]:
+                        self.assertNotIn(
```
DylanRussell, line 322: Can we possibly do an assert equal test the entire content of inputs/outputs/system instruction in one of these tests ?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/message.py
argaj: This is just internal func. It is used below to return actual list, which filters Nones, so if there no output messages, it returns an empty list.

DylanRussell: Ah I see. Ok this LGTM Then

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -232,23 +282,71 @@ def _wrapped_config_with_tools(
...
+        ],
+    }
+    if system_instructions:
+        attributes[gen_ai_attributes.GEN_AI_SYSTEM_INSTRUCTIONS] = [
+            dataclasses.asdict(sys_instr) for sys_instr in system_instructions
+        ]
+
+    if as_str:
+        return {
+            k: json.dumps(v, cls=_GenAiJsonEncoder)
```
aabmass: can you use this function instead? it will remove extra whitespace too https://github.com/open-telemetry/opentelemetry-python-contrib/blob/a4adc93c8ec04e4172bf4014f908cbf2662140fb/util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py#L76

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -373,6 +490,68 @@ def _maybe_update_error_type(self, response: GenerateContentResponse):
...
+                )
+            )
+            span.set_attributes(completion_details_attributes)
+            span.set_attributes(attributes)
+        if self._content_recording_enabled in [
+            ContentCapturingMode.EVENT_ONLY,
+            ContentCapturingMode.SPAN_AND_EVENT,
+        ]:
+            completion_details_attributes = (
+                _create_completion_details_attributes(
```
aabmass: nit iiuc this may duplicate the work already done on L529. Can you check if it's already happened here?

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/test_e2e.py
```diff
@@ -389,7 +390,13 @@ def fixture_gcloud_api_key(gemini_api_key):
 @pytest.fixture(name="nonvertex_client_factory")
 def fixture_nonvertex_client_factory(gemini_api_key):
     def _factory():
-        return google.genai.Client(api_key=gemini_api_key, vertexai=False)
+        return google.genai.Client(
+            api_key=gemini_api_key,
+            vertexai=False,
+            http_options=types.HttpOptions(
+                headers={"accept-encoding": "identity"}
```
aabmass, line 446: can you add a comment explaining this is to prevent compression?

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_streaming[excludecontent-gemini-2.5-flash-vertexaiapi-sync].yaml
```diff
@@ -0,0 +1,109 @@
...
+      - google-genai-sdk/1.32.0 gl-python/3.11.9
+      x-goog-api-client:
+      - <REDACTED>
+      x-goog-user-project:
+      - <REDACTED>
+    method: POST
+    uri: https://test-location-aiplatform.googleapis.com/v1beta1/projects/test-project/locations/test-location/publishers/google/models/gemini-2.5-flash:streamGenerateContent?alt=sse
+  response:
+    body:
+      string: "data: {\"candidates\": [{\"content\": {\"role\": \"model\",\"parts\"\
```
Comment (aabmass) [inline instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_streaming[excludecontent-gemini-2.5-flash-vertexaiapi-sync].yaml]: Any idea why this one is formatted like this vs the other RPC recordings?

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_streaming[excludecontent-gemini-2.5-flash-vertexaiapi-async].yaml
```diff
@@ -0,0 +1,102 @@
...
+      - google-genai-sdk/1.32.0 gl-python/3.11.9
+      x-goog-api-client:
+      - <REDACTED>
+      x-goog-user-project:
+      - <REDACTED>
+    method: POST
+    uri: https://test-location-aiplatform.googleapis.com/v1beta1/projects/test-project/locations/test-location/publishers/google/models/gemini-2.5-flash:streamGenerateContent?alt=sse
+  response:
+    body:
+      string: "data: {\"candidates\": [{\"content\": {\"role\": \"model\",\"parts\"\
```
Comment (aabmass) [inline instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_streaming[excludecontent-gemini-2.5-flash-vertexaiapi-async].yaml]: Same question, why is the YAML not nice like the other recordings?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -55,6 +55,23 @@ class Text:
...
+
+
+@dataclass
+class FileDataPart:
+    mime_type: str
+    uri: str
+    type: Literal["file_data"] = "file_data"
+
+    class Config:
+        extra = "allow"
```
aabmass: @DylanRussell how did you handle this for the Vertex instrumentation?

Maybe just put these two definitions in the opentelemetry-instrumentation-google-genai package for now since we're still working out the spec here (https://github.com/open-telemetry/semantic-conventions/pull/2754)

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
argaj: Done, but won't this require util-genai package version bump? I see this func is not in the currently released version (0.1b0)

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
argaj: Yeah, indeed duplicates, refactored.

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_streaming[excludecontent-gemini-2.5-flash-vertexaiapi-sync].yaml
Comment (argaj) [reply instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_streaming[excludecontent-gemini-2.5-flash-vertexaiapi-sync].yaml]: Dylan updated these, so I don't know :/ @DylanRussell?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
argaj: Moved.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
aabmass: Ya I can do a release today before merging

DylanRussell: Maybe for now don't pin a version and use it. https://github.com/open-telemetry/opentelemetry-python-contrib/blob/9f393c75f685b7c3920588db70f707e2b870e19c/instrumentation-genai/opentelemetry-instrumentation-vertexai/pyproject.toml#L29-L31 -- for vertexAI we don't have a version pinned currently

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_streaming[excludecontent-gemini-2.5-flash-vertexaiapi-sync].yaml
Comment (DylanRussell) [reply instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_streaming[excludecontent-gemini-2.5-flash-vertexaiapi-sync].yaml]: Not sure.. will take a look

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
argaj: Done, I linked it to the same issue, will add a comment there that it's also the case for google-genai instrumentation.

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_streaming[excludecontent-gemini-2.5-flash-vertexaiapi-sync].yaml
Comment (DylanRussell) [reply instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_streaming[excludecontent-gemini-2.5-flash-vertexaiapi-sync].yaml]: ok, so it's only like this for the streaming tests. It's like this in the current streaming cassettes at head too.

It's because the response is sent using Server-Sent Events (SSE) format, which is a stream of separate data messages. `json.loads` fails here: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/9f393c75f685b7c3920588db70f707e2b870e19c/instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/test_e2e.py#L207-L208 and we just return the raw string..

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/utils/test_tool_call_wrapper.py
argaj: Isn't that what happens in nonstreaming_base.py test_new_semconv_record_completion_as_log?
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/05bed52c7775463226c026efcc5d4734850f94e0/instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/nonstreaming_base.py#L298

DylanRussell: Ok true. LGTM

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_streaming[logcontent-gemini-2.5-flash-vertexaiapi-async].yaml
Comment (argaj) [reply instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_streaming[logcontent-gemini-2.5-flash-vertexaiapi-async].yaml]: Added, PTAL.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/message.py
```diff
@@ -0,0 +1,173 @@
...
+class Role(str, Enum):
+    SYSTEM = "system"
+    USER = "user"
+    ASSISTANT = "assistant"
+    TOOL = "tool"
+
+
+@dataclass
+class BlobPart:
+    data: bytes
```
DylanRussell, line 44: Are there finalized semantic conventions for BlobPart/FiledataPart yet ? These shoud be part of gen AI utils if so..

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_upload_hook_non_streaming[logcontent-gemini-2.5-flash-vertexaiapi-sync-enable_completion_hook-experimental].yaml
```diff
@@ -0,0 +1,102 @@
+interactions:
```
Comment (DylanRussell) [inline instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_upload_hook_non_streaming[logcontent-gemini-2.5-flash-vertexaiapi-sync-enable_completion_hook-experimental].yaml:1]: I don't think we need a whole other set of cassette's with the completion hook bool set... We probably just need 1 test/cassette for it. The completion hook doesn't change or affect the request/cassette.

Comment (DylanRussell) [reply instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_upload_hook_non_streaming[logcontent-gemini-2.5-flash-vertexaiapi-sync-enable_completion_hook-experimental].yaml:1]: Ok I suppose it's OK as is.  I see you just have the 4 variants of sync/async streaming

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/message.py
argaj: Originally I put it in the utils, but Aaron said that it's not yet finalized, and should be put here for the time being.

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_upload_hook_non_streaming[logcontent-gemini-2.5-flash-vertexaiapi-sync-enable_completion_hook-experimental].yaml
Comment (argaj) [reply instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/cassettes/test_upload_hook_non_streaming[logcontent-gemini-2.5-flash-vertexaiapi-sync-enable_completion_hook-experimental].yaml:1]: Yep, it's sync/async and log/excludecontent

PR #3796, opened 2025-09-30
opentelemetry-instrumentation-dbapi: fix libpq version detection

### LINKED ISSUE #3793: Psycopg instrumentation errors when retrieving libpq build version
### Describe your environment

OS: Debian Trixie (13)
Python version: 3.13
Package version: 0.58b0

### What happened?

I was trying to hunt down the source of a mysterious error on my server startup.
I finally narrowed it down to `opentelemetry-instrumentation-psycopg`, with error:

```
  File "/opt/python/lib/python3.13/site-packages/opentelemetry/instrumentation/dbapi/__init__.py", line 317, in __init__
    self.commenter_data = self.calculate_commenter_data()
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/opt/python/lib/python3.13/site-packages/opentelemetry/instrumentation/dbapi/__init__.py", line 356, in calculate_commenter_data
    libpq_version = self.connect_module.pq.__build_version__
                    ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: type object 'AsyncConnection' has no attribute 'pq'
```

The relevant part of code is:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/4ae8662ab21f0a2bf799c37ad8449312e489d7f2/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py#L352-L357

I'm not sure if either:
- psycopg (3) doesn't correctly implement the `pg` attribute of the AsyncConnector class, **or**
- `.pg` isn't valid in the Python [DBAPI 2.0 spec](https://peps.python.org/pep-0249)

psycopg2 previously had [__libpq_version__](https://www.psycopg.org/docs/module.html#psycopg2.__libpq_version__) to access on the connection object, so this worked fine.

psycopg has a [few equivalent attributes](https://www.psycopg.org/psycopg3/docs/api/pq.html#module-content) on the top level module, instead of the connection object.

### Steps to Reproduce

As part of FastAPI web server:

database.py
```python
from typing import Union
import logging
import psycopg


log = logging.getLogger(__name__)


def create_connection(db: Union[str, psycopg.Connection]) -> psycopg.Connection:
    """Get db connection from existing psycopg connection, or URL string.

    Args:
        db (str, psycopg.Connection):
            string or existing db connection.
            If `db` is a string, a new connection is generated.
            If `db` is a psycopg connection, the connection is reused.

    Returns:
        conn: DBAPI connection object to generate cursors from.
    """
    if isinstance(db, psycopg.Connection):
        conn = db
    elif isinstance(db, str):
        conn = psycopg.connect(db)
    else:
        msg = "The `db` variable is not a valid string or psycopg connection."
        log.error(msg)
        raise ValueError(msg)

    return conn
```

main.py
```python
@asynccontextmanager
async def lifespan(
    app: FastAPI,  # dead: disable
) -> AsyncIterator[None]:
    """FastAPI startup/shutdown event."""
    log.debug("Starting up FastAPI server")

    # Create a pooled db connection and make available in lifespan state
    # https://asgi.readthedocs.io/en/latest/specs/lifespan.html#lifespan-state
    # NOTE to use within a request (this is wrapped in database.py already):
    # from typing import cast
    # db_pool = cast(AsyncConnectionPool, request.state.db_pool)
    # async with db_pool.connection() as conn:
    db_pool = get_db_connection_pool()
    log.debug("Opening database connection pool for server")
    await db_pool.open()
    log.debug("Database pool opened, attaching to lifespan")

    yield {"db_pool": db_pool}

    # Shutdown events
    log.debug("Shutting down FastAPI server")
    await db_pool.close()
```

### Expected Result

- The instrumentation should either succeed connecting to the database driver, else fail gracefully.
- There should be some error handling here at the very least, and perhaps a warning log.

### Actual Result

- Currently the error is not caught, so it halts server startup and causes an infinite loop.

### Additional context

_No response_

None

### Tip

xrmx:
# Description

This fixes a crash in the detection of libpq version in the psycopg module when we wrap a module that is not the one containing the pq module that has the version. In that case we just don't report the libpq version instead of crashing.

References:
- psycopg2 https://www.psycopg.org/docs/module.html#psycopg2.__libpq_version__
- psycopg https://www.psycopg.org/psycopg3/docs/api/pq.html#module-content

Fixes #3793

## Type of change

# How Has This Been Tested?

- [x] tox

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
@@ -350,11 +350,25 @@ class DatabaseApiIntegration:
         }
 
         if self.database_system == "postgresql":
-            if hasattr(self.connect_module, "__libpq_version__"):
-                libpq_version = self.connect_module.__libpq_version__
-            else:
-                libpq_version = self.connect_module.pq.__build_version__
-            commenter_data.update({"libpq_version": libpq_version})
+            libpq_version = None
+            # psycopg
+            if hasattr(self.connect_module, "pq"):
+                try:
+                    libpq_version = self.connect_module.pq.version()
+                except Exception:
+                    pass
+
+            # psycopg2
+            if libpq_version is None:
+                # this the libpq version the client has been built against
+                libpq_version = getattr(
+                    self.connect_module, "__libpq_version__", None
+                )
+
+            # psycopg also instrument modules that are not the root one, in that case you
+            # won't get the libpq_version
+            if libpq_version is not None:
+                commenter_data.update({"libpq_version": libpq_version})
         elif self.database_system == "mysql":
             mysqlc_version = ""
             if db_driver == "MySQLdb":
--- a/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
@@ -355,7 +355,7 @@ class DatabaseApiIntegration:
             if hasattr(self.connect_module, "pq"):
                 try:
                     libpq_version = self.connect_module.pq.version()
-                except Exception:
+                except Exception:  # pylint: disable=broad-exception-raised
                     pass
 
             # psycopg2
--- a/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
@@ -355,7 +355,7 @@ class DatabaseApiIntegration:
             if hasattr(self.connect_module, "pq"):
                 try:
                     libpq_version = self.connect_module.pq.version()
-                except Exception:  # pylint: disable=broad-exception-raised
+                except Exception:  # pylint: disable=broad-exception-caught
                     pass
 
             # psycopg2
--- a/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
@@ -352,11 +352,10 @@ class DatabaseApiIntegration:
         if self.database_system == "postgresql":
             libpq_version = None
             # psycopg
-            if hasattr(self.connect_module, "pq"):
-                try:
-                    libpq_version = self.connect_module.pq.version()
-                except Exception:  # pylint: disable=broad-exception-caught
-                    pass
+            try:
+                libpq_version = self.connect_module.pq.version()
+            except AttributeError:
+                pass
 
             # psycopg2
             if libpq_version is None:
@@ -365,7 +364,7 @@ class DatabaseApiIntegration:
                     self.connect_module, "__libpq_version__", None
                 )
 
-            # psycopg also instrument modules that are not the root one, in that case you
+            # we instrument psycopg modules that are not the root one, in that case you
             # won't get the libpq_version
             if libpq_version is not None:
                 commenter_data.update({"libpq_version": libpq_version})
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,12 +13,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `opentelemetry-instrumentation-dbapi`: fix crash retrieving libpq version when enabling commenter with psycopg
+  ([#3796](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3796))
+
 ### Added
 - `opentelemetry-instrumentation`: botocore: Add support for AWS Secrets Manager semantic convention attribute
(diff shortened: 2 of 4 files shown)


instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -350,11 +350,25 @@ def calculate_commenter_data(self) -> dict[str, Any]:
         }
 
         if self.database_system == "postgresql":
-            if hasattr(self.connect_module, "__libpq_version__"):
-                libpq_version = self.connect_module.__libpq_version__
-            else:
-                libpq_version = self.connect_module.pq.__build_version__
-            commenter_data.update({"libpq_version": libpq_version})
+            libpq_version = None
+            # psycopg
```
aabmass, line 354: Do we run any tests for pscopg 1? Like requirements-oldest we have in other packages.

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -350,11 +350,25 @@ def calculate_commenter_data(self) -> dict[str, Any]:
...
+            if hasattr(self.connect_module, "pq"):
+                try:
+                    libpq_version = self.connect_module.pq.version()
+                except Exception:  # pylint: disable=broad-exception-caught
+                    pass
```
aabmass: nit would this work?
```suggestion
            try:
                libpq_version = self.connect_module.pq.version()
            except Exception:  # pylint: disable=broad-exception-caught
                pass
```

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -350,11 +350,25 @@ def calculate_commenter_data(self) -> dict[str, Any]:
...
-                libpq_version = self.connect_module.__libpq_version__
-            else:
-                libpq_version = self.connect_module.pq.__build_version__
-            commenter_data.update({"libpq_version": libpq_version})
+            libpq_version = None
+            # psycopg
+            if hasattr(self.connect_module, "pq"):
+                try:
+                    libpq_version = self.connect_module.pq.version()
+                except Exception:  # pylint: disable=broad-exception-caught
```
aabmass: Could we use a more specific exception?

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
emdneto: I don't think so

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
emdneto: @xrmx, in which scenario can this happen? like alpine-based environments?

emdneto:
```
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "opentelemetry-instrumentation-dbapi",
#     "psycopg",
# ]
# ///
import psycopg
from opentelemetry.instrumentation.dbapi import DatabaseApiIntegration
dbi = DatabaseApiIntegration("test", "postgresql", enable_commenter=True, connect_module=psycopg.AsyncConnection)
```
Was able to reproduce this against main (which is failing) and against this branch (which is working fine). Also tested in Alpine container with the -binary version of psycopg.

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
xrmx: Sure, the `hasattr` was to discriminate between psycopg and psycopg2 and psycopg root module and another one.

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
xrmx: This is broad because ancient libpq version (< 9.1, ~10 years ago) do not have that version api and I have no idea what would raise, I can reduce that to AttributeError and ignore the ancient libpq case.

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
xrmx: No we have only tests against mocks and that's why this got un-noticed. I should add something to the docker tests.

xrmx: Added docker tests

xrmx: Added the tests

PR #3802, opened 2025-10-01
Update completion upload hook to make input / output / system instructions optional

DylanRussell:
# Description

It's possible inputs/ouputs/system_instruction are empty for a given request, so making these 3 params optional and obviously will not do a file upload when they are..

## Type of change

# How Has This Been Tested?

- Unit tests

- [ ] Yes. - Link to PR: 
- [x ] No.

# Checklist:

### DIFF
--- a/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
+++ b/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
@@ -60,9 +60,9 @@ _logger = logging.getLogger(__name__)
 
 @dataclass
 class Completion:
-    inputs: list[types.InputMessage]
-    outputs: list[types.OutputMessage]
-    system_instruction: list[types.MessagePart]
+    inputs: list[types.InputMessage] | None
+    outputs: list[types.OutputMessage] | None
+    system_instruction: list[types.MessagePart] | None
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
```diff
@@ -203,17 +203,19 @@ def _do_upload(
     def on_completion(
         self,
         *,
-        inputs: list[types.InputMessage],
-        outputs: list[types.OutputMessage],
-        system_instruction: list[types.MessagePart],
+        inputs: list[types.InputMessage] | None = None,
+        outputs: list[types.OutputMessage] | None = None,
+        system_instruction: list[types.MessagePart] | None = None,
```
aabmass: - is it useful to distinguish between None and empty list? IMO better to avoid additional edge cases and just use an empty list throughout
- This is getting called generically through the `CompletionHook` protocol from the other file. So the instrumentations would still need to comply with that signature

util/opentelemetry-util-genai/tests/test_upload.py
```diff
@@ -162,6 +162,21 @@ def test_upload_then_shutdown(self):
...
+            system_instruction=FAKE_SYSTEM_INSTRUCTION,
+        )
+        # all items should be consumed
+        self.hook.shutdown()
+
+        self.assertEqual(
+            self.mock_fs.open.call_count,
+            1,
+            "should have uploaded 1 file",
+        )
```
aabmass, line 180: can you add an assertion that the empty ones don't create a ref on the span or log?

aabmass (approved):

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
DylanRussell: Ok that works for me, I'll just add a comment on the class that an empty list will not be uploaded

util/opentelemetry-util-genai/tests/test_upload.py
DylanRussell: Yeah, added

PR #3814, opened 2025-10-07
Use an sha 256 hash of the system instructions as it's upload filename

DylanRussell:
# Description

Use an md5 hash of the system instructions as it's upload filename, and check if the file exists before re-uploading it..

In many (most?) cases system instructions will be text and is unlikely to change, this prevents uploading the same system instructions twice.

## Type of change

- [ x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Unit tests

- [ ] Yes. - Link to PR: 
- [x ] No.

# Checklist:

- [x ] Unit tests have been added
- [x ] Documentation has been updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3624](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3624))
 - `opentelemetry-instrumentation-dbapi`: fix crash retrieving libpq version when enabling commenter with psycopg
   ([#3796](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3796))
+- `opentelemetry-instrumentation-fastapi`: Fix handling of APIRoute subclasses
+  ([#3681](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3681))
 
 ### Added
 
--- a/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
@@ -191,7 +191,7 @@ from weakref import WeakSet as _WeakSet
 import fastapi
 from starlette.applications import Starlette
 from starlette.middleware.errors import ServerErrorMiddleware
-from starlette.routing import Match
+from starlette.routing import Match, Route
 from starlette.types import ASGIApp, Receive, Scope, Send
 
 from opentelemetry.instrumentation._semconv import (
@@ -474,7 +474,11 @@ def _get_route_details(scope):
     route = None
 
     for starlette_route in app.routes:
-        match, _ = starlette_route.matches(scope)
+        match, _ = (
+            Route.matches(starlette_route, scope)
+            if isinstance(starlette_route, Route)
+            else starlette_route.matches(scope)
+        )
         if match == Match.FULL:
             try:
                 route = starlette_route.path
--- a/instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
+++ b/instrumentation/opentelemetry-instrumentation-fastapi/tests/test_fastapi_instrumentation.py
@@ -20,14 +20,17 @@ import unittest
 import weakref as _weakref
 from contextlib import ExitStack
 from timeit import default_timer
-from typing import Any, cast
+from typing import Any, Final, cast
 from unittest.mock import Mock, call, patch
 
 import fastapi
 import pytest
 from fastapi.middleware.httpsredirect import HTTPSRedirectMiddleware
 from fastapi.responses import JSONResponse, PlainTextResponse
+from fastapi.routing import APIRoute
 from fastapi.testclient import TestClient
+from starlette.routing import Match
+from starlette.types import Receive, Scope, Send
 
 import opentelemetry.instrumentation.fastapi as otel_fastapi
 from opentelemetry import trace
@@ -131,6 +134,23 @@ _recommended_attrs_both["http.server.active_requests"].extend(
 )
 
 
+class CustomMiddleware:
+    def __init__(self, app: fastapi.FastAPI) -> None:
+        self.app = app
+
+    async def __call__(
+        self, scope: Scope, receive: Receive, send: Send
+    ) -> None:
+        scope["nonstandard_field"] = "here"
+        await self.app(scope, receive, send)
+
+
+class CustomRoute(APIRoute):
+    def matches(self, scope: Scope) -> tuple[Match, Scope]:
+        assert "nonstandard_field" in scope
+        return super().matches(scope)
+
+
 class TestBaseFastAPI(TestBase):
     def _create_app(self):
         app = self._create_fastapi_app()
@@ -191,6 +211,7 @@ class TestBaseFastAPI(TestBase):
         self._instrumentor = otel_fastapi.FastAPIInstrumentor()
         self._app = self._create_app()
         self._app.add_middleware(HTTPSRedirectMiddleware)
+        self._app.add_middleware(CustomMiddleware)
         self._client = TestClient(self._app, base_url="https://testserver:443")
         # run the lifespan, initialize the middleware stack
         # this is more in-line with what happens in a real application when the server starts up
@@ -210,6 +231,7 @@ class TestBaseFastAPI(TestBase):
     def _create_fastapi_app():
         app = fastapi.FastAPI()
         sub_app = fastapi.FastAPI()
+        custom_router = fastapi.APIRouter(route_class=CustomRoute)
 
         @sub_app.get("/home")
         async def _():
@@ -234,6 +256,12 @@ class TestBaseFastAPI(TestBase):
         @app.get("/error")
         async def _():
             raise UnhandledException("This is an unhandled exception")
+
+        @custom_router.get("/success")
+        async def _():
+            return None
+
+        app.include_router(custom_router, prefix="/custom-router")
 
         app.mount("/sub", app=sub_app)
         app.host("testserver2", sub_app)
@@ -312,6 +340,28 @@ class TestBaseManualFastAPI(TestBaseFastAPI):
             "https://testserver/sub/home",
             span.attributes[HTTP_URL],
         )
+
+    def test_custom_api_router(self):
+        """
+        This test is to ensure that custom API routers the OpenTelemetryMiddleware does not cause issues with
+        custom API routers that depend on non-standard fields on the ASGI scope.
+        """
+        resp: Final = self._client.get("/custom-router/success")
+        spans: Final = self.memory_exporter.get_finished_spans()
+        spans_with_http_attributes = [
+            span
+            for span in spans
+            if (HTTP_URL in span.attributes or HTTP_TARGET in span.attributes)
+        ]
+        self.assertEqual(200, resp.status_code)
+        for span in spans_with_http_attributes:
+            self.assertEqual(
+                "/custom-router/success", span.attributes[HTTP_TARGET]
+            )
+            self.assertEqual(
+                "https://testserver/custom-router/success",
+                span.attributes[HTTP_URL],
+            )
 
     def test_host_fastapi_call(self):
         client = TestClient(self._app, base_url="https://testserver2:443")
@@ -1017,6 +1067,7 @@ class TestFastAPIManualInstrumentation(TestBaseManualFastAPI):
     def _create_fastapi_app():
         app = fastapi.FastAPI()
         sub_app = fastapi.FastAPI()
+        custom_router = fastapi.APIRouter(route_class=CustomRoute)
 
         @sub_app.get("/home")
         async def _():
@@ -1041,6 +1092,12 @@ class TestFastAPIManualInstrumentation(TestBaseManualFastAPI):
         @app.get("/error")
         async def _():
             raise UnhandledException("This is an unhandled exception")
+
+        @custom_router.get("/success")
+        async def _():
+            return None
+
+        app.include_router(custom_router, prefix="/custom-router")
 
         app.mount("/sub", app=sub_app)
 


util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
```diff
@@ -166,13 +178,17 @@ def _calculate_ref_path(self) -> CompletionRefs:
...
+                f"{system_instruction_hash or uuid_str}_system_instruction.{self._format}",
             ),
         )
 
     def _do_upload(
         self, path: str, json_encodeable: Callable[[], JsonEncodeable]
     ) -> None:
+        # FileSystem class has this method. Only check for system instructions as that's the only where the filename is a hash.
+        # https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.exists
+        if "_system_instruction" in path and self._fs.exists(path):  # pyright: ignore[reportUnknownMemberType]
```
aabmass: Can you update the type definition to capture this? It will resolve the pyright errors https://github.com/open-telemetry/opentelemetry-python-contrib/blob/e3cb05396bf4f9ba23a5cb208b9b1e31e02d86f2/typings/fsspec/__init__.pyi#L26

aabmass: Do you know if doing this with GCS 

1. Is relatively fast? My guess is it's about the same as doing a small upload
2. Caches the result? If not, we could stick this behind a small `@lru_cache` to avoid network calls

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
```diff
@@ -152,10 +153,21 @@ def done(future: Future[None]) -> None:
...
+        system_instruction_hash = None
+        # Use an md5 hash of the system instructions as a filename, when system instructions are text.
+        if all(isinstance(x, types.Text) for x in system_instruction):
+            md5_hash = hashlib.md5()
+            md5_hash.update(
+                "\n".join(x.content for x in system_instruction).encode(  # pyright: ignore[reportUnknownMemberType, reportAttributeAccessIssue, reportUnknownArgumentType]
+                    "utf-8"
+                )
+            )
+            system_instruction_hash = md5_hash.hexdigest()
```
aabmass: How does this compare with using python's `hash()` on the concatenated string?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
```diff
@@ -152,10 +153,21 @@ def done(future: Future[None]) -> None:
...
+    def _calculate_ref_path(
+        self, system_instruction: list[types.MessagePart]
+    ) -> CompletionRefs:
         # TODO: experimental with using the trace_id and span_id, or fetching
         # gen_ai.response.id from the active span.
-
+        system_instruction_hash = None
+        # Use an md5 hash of the system instructions as a filename, when system instructions are text.
+        if all(isinstance(x, types.Text) for x in system_instruction):
+            md5_hash = hashlib.md5()
```
xrmx: Please use something that is available in FIPS environments, e.g. sha256

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
DylanRussell: done thanks

DylanRussell: The implementation seems like it might cache https://github.com/fsspec/gcsfs/blob/main/gcsfs/core.py#L997, but when I turned on audit logs for GCS and called it repeatedly I see a `get` call to GCS is made every time.

Yeah a LRU cache sounds good

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
DylanRussell: Apparently `hash()` is randomly salted (per-python-process) so it's not good to use for this.

DylanRussell: Although that can be disabled by setting an env variable:  PYTHONHASHSEED=0

DylanRussell: I'm thinking we should just use CRC32, which is a checksum. WE don't really need a hash for this.. just a checksum

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
DylanRussell: I'm thinking we should just use CRC32, which is a checksum. WE don't really need a hash for this.. just a checksum

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
aabmass: Oh nice didn't know that. As long as there is reasonably low chance of collisions, SGTM

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
aabmass: Would md5 still work in fips environment with [`usedforsecurity=False`](https://docs.python.org/3/library/hashlib.html#hashlib.md5:~:text=Changed%20in%20version%203.9%3A%20All%20hashlib%20constructors%20take%20a%20keyword%2Donly%20argument%20usedforsecurity%20with%20default%20value%20True.%20A%20false%20value%20allows%20the%20use%20of%20insecure%20and%20blocked%20hashing%20algorithms%20in%20restricted%20environments.)?

If the perf difference is negligible, sha256 seems reasonable too.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
aabmass: > Since the algorithm is designed for use as a checksum algorithm, it is not suitable for use as a general hash algorithm.

https://docs.python.org/3/library/zlib.html#zlib.crc32 not sure what that means regarding collisions.

DylanRussell: https://stackoverflow.com/questions/10953958/can-crc32-be-used-as-a-hash-function 
https://stackoverflow.com/questions/14210298/probability-of-collision-when-using-a-32-bit-hash
https://stackoverflow.com/questions/2694740/can-one-construct-a-good-hash-function-using-crc32c-as-a-base

It seems pretty good for our needs. I'm guessing it says that because it's not "secure" and the hashes are not evenly distributed, but I don't think we care about either of those.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
DylanRussell: ok using timeit sha 256 was the fastest: .0006 seconds for 1000 runs of a 600 character string hash, so I've switched to using sha256 I also set `usedforsecurity=False`

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
DylanRussell: Hmm LRU cache doesn't work because it's caching `False` return values, and I can't seem to modify it

DylanRussell: Just used an ordereddict to handle the LRU stuff

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
```diff
@@ -97,10 +107,13 @@ def __init__(
         base_path: str,
         max_size: int = 20,
         upload_format: Format | None = None,
+        lru_cache_max_size: int = 1024,
     ) -> None:
         self._max_size = max_size
         self._fs, base_path = fsspec.url_to_fs(base_path)
         self._base_path = self._fs.unstrip_protocol(base_path)
+        self.lru_dict: OrderedDict[str, bool] = OrderedDict()
+        self.lru_cache_max_size = lru_cache_max_size
```
aabmass, line 116: Can we use the built in `functools.lru_cache` decorator for the caching?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
```diff
@@ -194,6 +242,11 @@ def _do_upload(
                 gen_ai_json_dump(message, file)
                 file.write("\n")
 
+        if contents_hashed_to_filename:
+            self.lru_dict[path] = True
+            if len(self.lru_dict) > self.lru_cache_max_size:
+                self.lru_dict.popitem(last=False)
```
aabmass, line 248: Can you use the builtin `@functools.lru_cache` decorator to handle the caching?

util/opentelemetry-util-genai/tests/test_upload.py
```diff
@@ -121,13 +122,14 @@ def setUp(self):
...
+        self.mock_fs.exists.return_value = False
 
         self.hook = UploadCompletionHook(
-            base_path=BASE_PATH,
-            max_size=MAXSIZE,
+            base_path=BASE_PATH, max_size=MAXSIZE, lru_cache_max_size=5
         )
 
     def tearDown(self) -> None:
+        self.mock_fs.reset_mock()
```
aabmass: Was this needed?

util/opentelemetry-util-genai/tests/test_upload.py
```diff
@@ -157,13 +159,87 @@ def test_upload_then_shutdown(self):
...
+        system_instructions = [
+            types.Text(content="You are a helpful assistant."),
+            types.Text(content="You will do your best."),
+        ]
+        expected_hash = hashlib.sha256(
+            "\n".join(text.content for text in system_instructions).encode(
+                "utf-8"
+            ),
+            usedforsecurity=False,
+        ).hexdigest()
```
aabmass: nit can you hardcode it

util/opentelemetry-util-genai/tests/test_upload.py
```diff
@@ -157,13 +159,87 @@ def test_upload_then_shutdown(self):
...
+        self.hook.shutdown()
+
+        self.assertEqual(
+            self.mock_fs.open.call_count,
+            1,
+            "should have uploaded 1 file",
+        )
+        assert record.attributes is not None
+        self.assertEqual(
+            record.attributes["gen_ai.system_instructions_ref"].split("/")[-1],
```
aabmass: could hardcode the `expected_hash` since it should always be the same.
```suggestion
            record.attributes["gen_ai.system_instructions_ref"],
            f"memory://{expected_hash}_system_instruction.json",
```

util/opentelemetry-util-genai/tests/test_upload.py
```diff
@@ -157,13 +159,87 @@ def test_upload_then_shutdown(self):
...
+        record = LogRecord()
+        self.hook.on_completion(
+            inputs=[],
+            outputs=[],
+            system_instruction=system_instructions,
+            log_record=record,
+        )
+        # Wait a bit for file upload to finish..
+        time.sleep(0.5)
+        self.mock_fs.exists.return_value = True
```
aabmass: imo this would be better as a separate test case in `TestUploadCompletionHookIntegration`  (which doesn't mock) where you create the file in BASE_PATH before the upload to test

instrumentation-genai/opentelemetry-instrumentation-google-genai/CHANGELOG.md
aabmass, line 1: please revert

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
DylanRussell: No because we don't want to cache `False` values

util/opentelemetry-util-genai/tests/test_upload.py
DylanRussell: No.. can;t remember why I added that. Removed

util/opentelemetry-util-genai/tests/test_upload.py
DylanRussell: true, done

util/opentelemetry-util-genai/tests/test_upload.py
DylanRussell: done

DylanRussell: The entire file name is ThreadSafeMagicMock/mock.unstrip_protocol()/140162515070480/75357d685f238b6afd7738be9786fdafde641eb6ca9a3be7471939715a68a4de_system_instruction.json that's why I did .split("/")[-1]

util/opentelemetry-util-genai/tests/test_upload.py
DylanRussell: OK updated it to be like that

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
aabmass: Got it. The builtin thing would be nice to avoid some complexity. Would just caching the write call achieve the same result?

```py
# Same as _do_upload but with a cache and existence check before
@lru_cache(maxsize=lru_cache_max_size)
def _do_upload_hashed(
  path: str,
  json_encodeable: Callable[[], JsonEncodeable],
):
  if exists(path):
    return
  with _fs.open(path, "w"):
    ...
```

then choose to call either `_do_upload()` or `_do_upload_hashed()`?

util/opentelemetry-util-genai/CHANGELOG.md
```diff
@@ -15,8 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3795](#3795))
 - Make inputs / outputs / system instructions optional params to `on_completion`,
   ([https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3802](#3802)).
-  - `opentelemetry-instrumentation-google-genai`: migrate off the deprecated events API to use the logs API
-  ([#3625](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3624))
+  - Use an crc32 checksum of the system instructions as it's upload filename, and check
```
aabmass: needs update

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
DylanRussell: That's a good idea but `JsonEncodeable` is not hashable

DylanRussell: I think I could get it working with cache tools: https://stackoverflow.com/questions/30730983/make-lru-cache-ignore-some-of-the-function-arguments -- we'd need to add that package here.. sound good ?

aabmass: Nah let's just go ahead with your hand rolled code to avoid new deps, sorry for the back and forth

util/opentelemetry-util-genai/tests/test_upload.py
```diff
@@ -157,13 +157,46 @@ def test_upload_then_shutdown(self):
...
+        record = LogRecord()
+        self.hook.on_completion(
+            inputs=[],
+            outputs=[],
+            system_instruction=FAKE_SYSTEM_INSTRUCTION,
+            log_record=record,
+        )
+        # Wait a bit for file upload to finish..
+        time.sleep(0.5)
+        assert record.attributes is not None
```
aabmass: Can you use the self.assert* methods just to keep it consistent

util/opentelemetry-util-genai/CHANGELOG.md
DylanRussell: done thanks

util/opentelemetry-util-genai/tests/test_upload.py
DylanRussell: yeah good idea, done

PR #3817, opened 2025-10-07
Implement OpenAI Agents span processing

nagkumar91:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

### DIFF
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/__init__.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/__init__.py
@@ -12,34 +12,95 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Barebones OpenAI Agents instrumentation package.
+"""OpenAI Agents instrumentation for OpenTelemetry."""
 
-This branch provides only the minimal package skeleton:
-- Instrumentor class stub
-- Version module
-- Packaging metadata/entry point
-"""
+from __future__ import annotations
 
+import importlib
+import os
 from typing import Collection
 
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
+from opentelemetry.semconv._incubating.attributes import (
+    gen_ai_attributes as GenAI,
+)
+from opentelemetry.semconv.schemas import Schemas
+from opentelemetry.trace import get_tracer
 
 from .package import _instruments
+from .span_processor import _OpenAIAgentsSpanProcessor
 from .version import __version__  # noqa: F401
 
-__all__ = [
-    "OpenAIAgentsInstrumentor",
-]
+__all__ = ["OpenAIAgentsInstrumentor"]
+
+
+def _load_tracing_module():
+    return importlib.import_module("agents.tracing")
+
+
+def _resolve_system(value: str | None) -> str:
+    if not value:
+        return GenAI.GenAiSystemValues.OPENAI.value
+
+    normalized = value.strip().lower()
+    for member in GenAI.GenAiSystemValues:
+        if normalized == member.value:
+            return member.value
+        if normalized == member.name.lower():
+            return member.value
+    return value
+
+
+def _get_registered_processors(provider) -> list:
+    multi = getattr(provider, "_multi_processor", None)
+    processors = getattr(multi, "_processors", ())
+    return list(processors)
 
 
 class OpenAIAgentsInstrumentor(BaseInstrumentor):
-    """Minimal instrumentor stub (no-op)."""
+    """Instrumentation that bridges OpenAI Agents tracing to OpenTelemetry spans."""
 
-    def _instrument(self, **kwargs) -> None:  # pragma: no cover - stub
-        return
+    def __init__(self) -> None:
+        super().__init__()
+        self._processor: _OpenAIAgentsSpanProcessor | None = None
 
-    def _uninstrument(self, **kwargs) -> None:  # pragma: no cover - stub
-        return
+    def _instrument(self, **kwargs) -> None:
+        if self._processor is not None:
+            return
+
+        tracer_provider = kwargs.get("tracer_provider")
+        tracer = get_tracer(
+            __name__,
+            "",
+            tracer_provider,
+            schema_url=Schemas.V1_28_0.value,
+        )
+
+        system_override = kwargs.get("system") or os.getenv(
+            "OTEL_INSTRUMENTATION_OPENAI_AGENTS_SYSTEM"
+        )
+        system = _resolve_system(system_override)
+
+        processor = _OpenAIAgentsSpanProcessor(tracer=tracer, system=system)
+
+        tracing = _load_tracing_module()
+        provider = tracing.get_trace_provider()
+        existing = _get_registered_processors(provider)
+        provider.set_processors([*existing, processor])
+        self._processor = processor
+
+    def _uninstrument(self, **kwargs) -> None:
+        if self._processor is None:
+            return
+
+        tracing = _load_tracing_module()
+        provider = tracing.get_trace_provider()
+        current = _get_registered_processors(provider)
+        filtered = [proc for proc in current if proc is not self._processor]
+        provider.set_processors(filtered)
+
+        self._processor.shutdown()
+        self._processor = None
 
     def instrumentation_dependencies(self) -> Collection[str]:
         return _instruments
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
@@ -1,0 +1,422 @@
+from __future__ import annotations
+
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
+from datetime import datetime
+from threading import RLock
+from typing import Any
+from urllib.parse import urlparse
+
+try:  # pragma: no cover - used when OpenAI Agents is available
+    from agents.tracing.processor_interface import TracingProcessor
+    from agents.tracing.spans import Span as AgentsSpan
+    from agents.tracing.traces import Trace as AgentsTrace
+except ImportError:  # pragma: no cover - fallback for tests
+
+    class TracingProcessor:  # type: ignore[misc]
+        pass
+
+    AgentsSpan = Any  # type: ignore[assignment]
+    AgentsTrace = Any  # type: ignore[assignment]
+from opentelemetry.semconv._incubating.attributes import (
+    gen_ai_attributes as GenAI,
+)
+from opentelemetry.semconv._incubating.attributes import (
+    server_attributes as ServerAttributes,
+)
+from opentelemetry.trace import Span, SpanKind, Tracer, set_span_in_context
+from opentelemetry.trace.status import Status, StatusCode
+
+
+def _parse_iso8601(timestamp: str | None) -> float | None:
+    """Return POSIX timestamp (seconds) for ISO8601 string."""
+
+    if not timestamp:
+        return None
+
+    try:
+        if timestamp.endswith("Z"):
+            timestamp = timestamp[:-1] + "+00:00"
+        dt = datetime.fromisoformat(timestamp)
+    except ValueError:
+        return None
+
+    return dt.timestamp()
+
+
+def _extract_server_attributes(
+    config: Mapping[str, Any] | None,
+) -> dict[str, Any]:
+    if not config:
+        return {}
+
+    base_url = config.get("base_url")
+    if not isinstance(base_url, str):
+        return {}
+
+    try:
+        parsed = urlparse(base_url)
+    except ValueError:
+        return {}
+
+    attributes: dict[str, Any] = {}
+    if parsed.hostname:
+        attributes[ServerAttributes.SERVER_ADDRESS] = parsed.hostname
+    if parsed.port:
+        attributes[ServerAttributes.SERVER_PORT] = parsed.port
+
+    return attributes
+
+
+def _looks_like_chat(messages: Sequence[Mapping[str, Any]] | None) -> bool:
+    if not messages:
+        return False
+    for message in messages:
+        if isinstance(message, Mapping) and message.get("role"):
+            return True
+    return False
+
+
+def _collect_finish_reasons(choices: Sequence[Any] | None) -> list[str]:
+    reasons: list[str] = []
+    if not choices:
+        return reasons
+
+    for choice in choices:
+        if isinstance(choice, Mapping):
+            reason = choice.get("finish_reason") or choice.get("stop_reason")
+            if reason:
+                reasons.append(str(reason))
+            continue
+
+        finish_reason = getattr(choice, "finish_reason", None)
+        if finish_reason:
+            reasons.append(str(finish_reason))
+
+    return reasons
+
+
+def _clean_stop_sequences(value: Any) -> Sequence[str] | None:
+    if value is None:
+        return None
+    if isinstance(value, str):
+        return [value]
+    if isinstance(value, Sequence):
+        cleaned: list[str] = []
+        for item in value:
+            if item is None:
+                continue
+            cleaned.append(str(item))
+        return cleaned if cleaned else None
+    return None
+
+
+@dataclass
+class _SpanContext:
+    span: Span
+    kind: SpanKind
+
+
+class _OpenAIAgentsSpanProcessor(TracingProcessor):
+    """Convert OpenAI Agents traces into OpenTelemetry spans."""
+
+    def __init__(self, tracer: Tracer, system: str) -> None:
+        self._tracer = tracer
+        self._system = system
+        self._root_spans: dict[str, Span] = {}
+        self._spans: dict[str, _SpanContext] = {}
+        self._lock = RLock()
+
+    def _operation_name(self, span_data: Any) -> str:
+        span_type = getattr(span_data, "type", None)
+        if span_type == "generation":
+            if _looks_like_chat(getattr(span_data, "input", None)):
+                return GenAI.GenAiOperationNameValues.CHAT.value
+            return GenAI.GenAiOperationNameValues.TEXT_COMPLETION.value
+        if span_type == "agent":
+            return GenAI.GenAiOperationNameValues.INVOKE_AGENT.value
+        if span_type == "function":
+            return GenAI.GenAiOperationNameValues.EXECUTE_TOOL.value
+        if span_type == "response":
+            return GenAI.GenAiOperationNameValues.CHAT.value
+        return span_type or "operation"
+
+    def _span_kind(self, span_data: Any) -> SpanKind:
+        span_type = getattr(span_data, "type", None)
+        if span_type in {"generation", "response", "speech", "transcription"}:
+            return SpanKind.CLIENT
+        return SpanKind.INTERNAL
+
+    def _span_name(self, operation: str, attributes: Mapping[str, Any]) -> str:
+        model = attributes.get(GenAI.GEN_AI_REQUEST_MODEL)
+        agent_name = attributes.get(GenAI.GEN_AI_AGENT_NAME)
+        tool_name = attributes.get(GenAI.GEN_AI_TOOL_NAME)
+
+        if operation in (
+            GenAI.GenAiOperationNameValues.CHAT.value,
+            GenAI.GenAiOperationNameValues.TEXT_COMPLETION.value,
+            GenAI.GenAiOperationNameValues.GENERATE_CONTENT.value,
+            GenAI.GenAiOperationNameValues.EMBEDDINGS.value,
+        ):
+            return f"{operation} {model}" if model else operation
+        if operation == GenAI.GenAiOperationNameValues.INVOKE_AGENT.value:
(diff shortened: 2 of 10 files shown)


instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -0,0 +1,422 @@
...
+    def __init__(self, tracer: Tracer, system: str) -> None:
+        self._tracer = tracer
+        self._system = system
+        self._root_spans: dict[str, Span] = {}
+        self._spans: dict[str, _SpanContext] = {}
+        self._lock = RLock()
+
+    def _operation_name(self, span_data: Any) -> str:
+        span_type = getattr(span_data, "type", None)
+        if span_type == "generation":
```
rads-1996: are these span_types always going to be in lowercase?

instrumentation-genai/opentelemetry-instrumentation-openai-agents/tests/stubs/agents/tracing/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+
+@dataclass
+class FunctionSpanData:
+    name: str | None = None
+    input: Any = None
+    output: Any = None
+
+    @property
+    def type(self) -> str:
+        return "function"
```
rads-1996: these strings like function, generation, etc. are being used in multiple places, can these be made constants instead of hardcoding the values.

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/__init__.py
```diff
@@ -12,34 +12,95 @@
...
+        )
+
+        system_override = kwargs.get("system") or os.getenv(
+            "OTEL_INSTRUMENTATION_OPENAI_AGENTS_SYSTEM"
+        )
+        system = _resolve_system(system_override)
+
+        processor = _OpenAIAgentsSpanProcessor(tracer=tracer, system=system)
+
+        tracing = _load_tracing_module()
```
lzchen: Any reason why we are loading the module during init of the instrumentor and not at the beginning of runtime/

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/__init__.py
```diff
@@ -12,34 +12,95 @@
...
+    normalized = value.strip().lower()
+    for member in GenAI.GenAiSystemValues:
+        if normalized == member.value:
+            return member.value
+        if normalized == member.name.lower():
+            return member.value
+    return value
+
+
+def _get_registered_processors(provider) -> list:
```
lzchen: Nit: Include type annotations for provider and provide docstring on what the fields look like.

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/__init__.py
```diff
@@ -12,34 +12,95 @@
...
+        tracer_provider = kwargs.get("tracer_provider")
+        tracer = get_tracer(
+            __name__,
+            "",
+            tracer_provider,
+            schema_url=Schemas.V1_28_0.value,
+        )
+
+        system_override = kwargs.get("system") or os.getenv(
+            "OTEL_INSTRUMENTATION_OPENAI_AGENTS_SYSTEM"
```
lzchen: Is there a generic env var name we can use? Does this apply to multiple libraries and not just openai? If so, consider adding it [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/util/opentelemetry-util-genai/src/opentelemetry/util/genai/environment_variables.py)

instrumentation-genai/opentelemetry-instrumentation-openai-agents/examples/zero-code/.env.example
```diff
@@ -0,0 +1,14 @@
...
+# OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
+# OTEL_EXPORTER_OTLP_PROTOCOL=grpc
+
+OTEL_SERVICE_NAME=opentelemetry-python-openai-agents-zero-code
+
+# Enable auto-instrumentation for logs if desired
+OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED=true
+
+# Optionally override the agent name reported on spans
+# OTEL_GENAI_AGENT_NAME=Travel Concierge
```
rads-1996, line 14: If a user chooses to override the agent name and populates the `OTEL_GEAI_AGENT_NAME`, where is the logic for using that, currently I can see that the `GenAI.GEN_AI_AGENT_NAME` is used for populating the agent_name? Or am I missing something here?

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -0,0 +1,422 @@
...
+            "speech",
+            "transcription",
+            "mcp_tools",
+        }:
+            return self._attributes_from_generic(span_data)
+        return self._base_attributes()
+
+    def on_trace_start(self, trace: AgentsTrace) -> None:
+        attributes = self._base_attributes()
+        start_time = _parse_iso8601(getattr(trace, "started_at", None))
```
lzchen: I'm not sure if `AgentsTrace` is expected to have a `started_at` property so the time should just be `now()` by default. https://github.com/openai/openai-agents-python/blob/9078e29c0c4134d1b850dcaf936a4ef8975d6fcb/src/agents/tracing/traces.py#L13

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -0,0 +1,422 @@
...
+        if operation == GenAI.GenAiOperationNameValues.INVOKE_AGENT.value:
+            return f"{operation} {agent_name}" if agent_name else operation
+        if operation == GenAI.GenAiOperationNameValues.EXECUTE_TOOL.value:
+            return f"{operation} {tool_name}" if tool_name else operation
+        if operation == GenAI.GenAiOperationNameValues.CREATE_AGENT.value:
+            return f"{operation} {agent_name}" if agent_name else operation
+        return operation
+
+    def _base_attributes(self) -> dict[str, Any]:
+        return {GenAI.GEN_AI_SYSTEM: self._system}
```
lzchen: `gen.ai.system` is [deprecated](https://github.com/open-telemetry/semantic-conventions/blob/ae200362aa62d7c976b59e9695e8b789b3cb174d/docs/registry/attributes/gen-ai.md#deprecated-genai-attributes)

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -0,0 +1,422 @@
...
+        }:
+            return self._attributes_from_generic(span_data)
+        return self._base_attributes()
+
+    def on_trace_start(self, trace: AgentsTrace) -> None:
+        attributes = self._base_attributes()
+        start_time = _parse_iso8601(getattr(trace, "started_at", None))
+
+        with self._lock:
+            span = self._tracer.start_span(
```
lzchen, line 427: Does it make sense to have a span created for trace start? It's not really clearly defined in the semantic conventions what this span should look like or where in the agent process does `on_trace_start` get called. Might be something to discuss in the SIG.

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -0,0 +1,422 @@
...
+        tool_name = attributes.get(GenAI.GEN_AI_TOOL_NAME)
+
+        if operation in (
+            GenAI.GenAiOperationNameValues.CHAT.value,
+            GenAI.GenAiOperationNameValues.TEXT_COMPLETION.value,
+            GenAI.GenAiOperationNameValues.GENERATE_CONTENT.value,
+            GenAI.GenAiOperationNameValues.EMBEDDINGS.value,
+        ):
+            return f"{operation} {model}" if model else operation
+        if operation == GenAI.GenAiOperationNameValues.INVOKE_AGENT.value:
```
lzchen, line 215: I don't think these are following the [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/ae200362aa62d7c976b59e9695e8b789b3cb174d/docs/gen-ai/gen-ai-agent-spans.md#create-agent-span).

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -0,0 +1,422 @@
...
+    def on_trace_end(self, trace: AgentsTrace) -> None:
+        end_time = _parse_iso8601(getattr(trace, "ended_at", None))
+
+        with self._lock:
+            span = self._root_spans.pop(trace.trace_id, None)
+
+        if span:
+            span.end(end_time=end_time)
+
+    def on_span_start(self, span: AgentsSpan[Any]) -> None:
```
lzchen, line 444: Just curious, when is `on_span_start` called? On ANY agent operation, including creation?

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -0,0 +1,422 @@
...
+            if _looks_like_chat(getattr(span_data, "input", None)):
+                return GenAI.GenAiOperationNameValues.CHAT.value
+            return GenAI.GenAiOperationNameValues.TEXT_COMPLETION.value
+        if span_type == "agent":
+            return GenAI.GenAiOperationNameValues.INVOKE_AGENT.value
+        if span_type == "function":
+            return GenAI.GenAiOperationNameValues.EXECUTE_TOOL.value
+        if span_type == "response":
+            return GenAI.GenAiOperationNameValues.CHAT.value
+        return span_type or "operation"
```
lzchen, line 191: Think we are missing a case for [create](https://github.com/open-telemetry/semantic-conventions/blob/ae200362aa62d7c976b59e9695e8b789b3cb174d/docs/gen-ai/gen-ai-agent-spans.md#spans).
[AgentSpanData](https://github.com/openai/openai-agents-python/blob/9078e29c0c4134d1b850dcaf936a4ef8975d6fcb/src/agents/tracing/span_data.py#L28).

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -0,0 +1,422 @@
...
+
+    def _span_kind(self, span_data: Any) -> SpanKind:
+        span_type = getattr(span_data, "type", None)
+        if span_type in {"generation", "response", "speech", "transcription"}:
```
lzchen: Why these types in particular? What about types like "function"?

nagkumar91: The OpenAI Agents SDK emits span_data.type values of `"generation", "response", "speech", and "transcription"` when it is actually calling the OpenAI model endpoints. Per the GenAI
semantic conventions (span.gen_ai.inference.client and the OpenAI-specific extension), those operations SHOULD be CLIENT span

It also produces types such as `"agent", "function", "guardrail", "handoff", "speech_group", and "mcp_tools"`. These represent orchestration steps or tool executions that
happen inside the agent runtime, so they remain INTERNAL. In particular, "function" corresponds to the execute_tool convention, which the spec explicitly marks as INTERNAL.

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
nagkumar91: `on_span_start` is invoked by the OpenAI Agents SDK every time its own `Span.start()` runs, because our `_OpenAIAgentsSpanProcessor` is registered with the SDK’s tracing provider. The hook
fires for every agent-emitted span like - generations, responses, tool/function calls, guardrail steps, invoke-agent spans, and any agent-creation spans if the SDK emits them.

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
nagkumar91: `on_trace_start` fires in the OpenAI Agents SDK whenever it starts a new agent workflow `Trace.start()` 
we translate that event into a root span so every downstream generation/response/tool span
has a common parent. We currently tag that root as `SpanKind.SERVER` because, in the Agent runtime model, the trace represents the server side orchestrating a request; this mirrors how other server
frameworks build a tree rooted at a SERVER span. The GenAI spec doesn’t yet spell out guidance for this top-level agent span, so we can bring it to the SIG for confirmation, but today the root span
gives us the correct hierarchy and timing for the whole workflow.

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
nagkumar91: I think it defaults to `time_ns()` somewhere down the funnel, couldn't find the right source, so making the change to be more explicit.

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/__init__.py
nagkumar91: to avoid errors at import if the openai agents SDK is not installed. We can move it around if it makes more sense.

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/__init__.py
nagkumar91: done!

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/__init__.py
nagkumar91: Removed it.

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
nagkumar91: Replaced the deprecated `gen_ai.system` usage with the recommended `gen_ai.provider.name` attribute when building base span attributes

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
nagkumar91: Updated `_span_name` to append the request model for inference operations, falling back to the response model when that’s all we have, so names follow the `{gen_ai.operation.name} {gen_ai.request.model}` convention.

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
nagkumar91: updated.

instrumentation-genai/opentelemetry-instrumentation-openai-agents/examples/zero-code/.env.example
nagkumar91: Great call-out! Thanks for checking. I’ve now wired `OTEL_GENAI_AGENT_NAME` (and the optional agent_name argument) straight into the instrumentor: when we build `_OpenAIAgentsSpanProcessor`, we pass along the trimmed override, and the processor prefers that
value whenever it fills `gen_ai.agent.name` or names create/invoke-agent spans. I added coverage in `test_agent_name_override_applied_to_agent_spans` so we’re alerted if the override ever stops flowing through.

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
nagkumar91: yes

instrumentation-genai/opentelemetry-instrumentation-openai-agents/tests/stubs/agents/tracing/__init__.py
nagkumar91: updated! Thanks

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/__init__.py
lzchen: Don't think there is a need for that. All other instrumentations follow the convention that the underlying library is assumed to be installed. You can move the import statements to top level.

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/__init__.py
```diff
@@ -12,34 +12,120 @@
...
+__all__ = ["OpenAIAgentsInstrumentor"]
+
+
+def _load_tracing_module():
+    return importlib.import_module("agents.tracing")
+
+
+def _resolve_system(value: str | None) -> str:
+    if not value:
+        return GenAI.GenAiSystemValues.OPENAI.value
```
lzchen: I think `gen_ai.provider.name` MUST be `openai` for all openai spans. https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/openai.md#spans

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -0,0 +1,513 @@
...
+        SPAN_TYPE_GENERATION,
+        SPAN_TYPE_RESPONSE,
+        SPAN_TYPE_SPEECH,
+        SPAN_TYPE_TRANSCRIPTION,
+        SPAN_TYPE_AGENT,
+        SPAN_TYPE_AGENT_CREATION,
+    }
+)
+
+_GEN_AI_PROVIDER_NAME = "gen_ai.provider.name"
```
lzchen: Why not use the sem conv symbol?

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
nagkumar91: handling this in - https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3825

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/__init__.py
nagkumar91: changed!

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/__init__.py
nagkumar91: Fixed

instrumentation-genai/opentelemetry-instrumentation-openai-agents/src/opentelemetry/instrumentation/openai_agents/span_processor.py
nagkumar91: fixed

PR #3829, opened 2025-10-10
Add ArangoDB Instrumentation

mdgilene:
# Description

Adds instrumentation for the `python-arango` library

## Type of change

# How Has This Been Tested?

Originally implemented this instrumentation for a project at work. Unit tests verify ArangoDB client executions are intercepted and relevant request/response information is attached to the span.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


mdgilene:
First time contributing here so looking for feedback.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
```diff
@@ -0,0 +1,149 @@
...
+
+        else:
+            query = str(request.data)
+
+        attributes = {
+            db_attributes.DB_SYSTEM_NAME: "arangodb",
+            db_attributes.DB_NAMESPACE: instance.db_name,
+            db_attributes.DB_OPERATION_NAME: request.endpoint,
+            db_attributes.DB_QUERY_TEXT: textwrap.dedent(query.strip("\n")),
+        }
```
thompson-tomo: As per https://opentelemetry.io/docs/specs/semconv/database/database-spans/ `server.address`/`server.port` should also be included.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
```diff
@@ -0,0 +1,149 @@
...
+                bind_vars = request.data.get("bindVars")
+                options = request.data.get("options")
+
+            else:
+                query = str(request.data)
+
+        else:
+            query = str(request.data)
+
+        attributes = {
```
thompson-tomo: Is there any case the table/collection could be extracted from the request and set?

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
```diff
@@ -0,0 +1,149 @@
...
+            db_attributes.DB_NAMESPACE: instance.db_name,
+            db_attributes.DB_OPERATION_NAME: request.endpoint,
+            db_attributes.DB_QUERY_TEXT: textwrap.dedent(query.strip("\n")),
+        }
+
+        if bind_vars:
+            for key, value in bind_vars.items():
+                attributes[f"db.query.parameter.{key}"] = json.dumps(value)
+
+        attributes["db.query.options"] = json.dumps(options)
```
thompson-tomo: This feels too generic for me, I would recommend identifying if there is any options which would be important and defining them as `arrangodb.*`.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
```diff
@@ -0,0 +1,149 @@
...
+
+        attributes["db.execution.cached"] = response.body.get("cached", False)
+
+        if (count := response.body.get("count")) is not None:
+            attributes["db.execution.count"] = count
+
+        if extra := response.body.get("extra"):
+            stats = extra.get("stats")
+            for key, value in stats.items():
+                attributes["db.execution.stats." + key] = value
```
thompson-tomo: Would avoid this as it is too generic and I'd remove it. I would suggest populating the key stats into metrics.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
```diff
@@ -0,0 +1,149 @@
...
+
+            warnings = extra.get("warnings")
+            attributes["db.execution.warnings"] = json.dumps(warnings)
```
thompson-tomo: This feels more like data that would be on an event.  I would leave off as no where else do we log warnings on the span.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
mdgilene: Potentially? It would likely only be a guess at best. ArangoDB queries use a "SQL-ish" language called AQL. AQL queries can operate on one or more collections at a time. I'm not sure how this is handled in other database implementations, but open to suggestions.

thompson-tomo: So it would only be set if the query is known to only be operating on 1 table/collection. Looking at the http api there is api's for collections & documents would those be traced? If so you should be able to get collection from url.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
mdgilene: Added server address and port info

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
mdgilene: I went through and explicitly defined which query options to pull into the attributes. There are more than what I have defined but these seemed to me to be the "most useful"

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
mdgilene: I just removed the stats part of this. I haven't really worked with metrics a whole lot so I will leave that as a future enhancement if needed and just focus on the traces.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
mdgilene: Moved these warnings to span events

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
mdgilene: Yes those should be traced. However it looks like those APIs take the collection name as part of the URL path. So would likely need to try to parse it out

`_api/collection/<collection-name>` or `_api/document/<collection-name>`

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+        attributes: dict[str, Any] = {
+            db_attributes.DB_RESPONSE_STATUS_CODE: response.status_code
+        }
+
+        if "cached" in response.body:
+            attributes["arangodb.response.cached"] = response.body.get(
+                "cached"
+            )
+        if "count" in response.body:
+            attributes["arangodb.response.count"] = response.body.get("count")
```
thompson-tomo: What does count represent? Number of items?

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+        if "hasMore" in response.body:
+            attributes["arangodb.response.hasMore"] = response.body.get(
+                "hasMore"
+            )
```
thompson-tomo: I assume that has more is related to pagination? If so I don't see much benefit of it.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+            if "allowRetry" in options:
+                attributes[arangodb_attributes.ALLOW_RETRY] = options.get(
+                    "allowRetry"
+                )
+            if "cache" in options:
+                attributes[arangodb_attributes.CACHE] = options.get("cache")
+            if "failOnWarning" in options:
+                attributes[arangodb_attributes.FAIL_ON_WARNING] = options.get(
+                    "failOnWarning"
+                )
```
thompson-tomo: This shouldn't be needed as the trace should convey this ie trace errored with xyz reason and trace has warning.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+            if "fullCount" in options:
+                attributes[arangodb_attributes.FULL_COUNT] = options.get(
+                    "fullCount"
+                )
+            if "maxRuntime" in options:
+                attributes[arangodb_attributes.MAX_RUNTIME] = options.get(
+                    "maxRuntime"
+                )
```
thompson-tomo: Is this max execution time of query? If so would leave out as the error reason should convey when the limit has been hit.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+            if "cache" in options:
+                attributes[arangodb_attributes.CACHE] = options.get("cache")
+            if "failOnWarning" in options:
+                attributes[arangodb_attributes.FAIL_ON_WARNING] = options.get(
+                    "failOnWarning"
+                )
+            if "fullCount" in options:
+                attributes[arangodb_attributes.FULL_COUNT] = options.get(
+                    "fullCount"
+                )
```
thompson-tomo: Leave this out as relates to stats

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+
+        if options and isinstance(options, dict):
+            if "allowDirtyReads" in options:
+                attributes[arangodb_attributes.ALLOW_DIRTY_READS] = (
+                    options.get("allowDirtyReads")
+                )
+            if "allowRetry" in options:
+                attributes[arangodb_attributes.ALLOW_RETRY] = options.get(
+                    "allowRetry"
+                )
```
thompson-tomo: I would leave this out as the usage is limited to batch reads & network failures.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+            if "cache" in options:
+                attributes[arangodb_attributes.CACHE] = options.get("cache")
```
thompson-tomo, line 144: Rather than using `true`/`false` as the value, I would use `skip` for `false` & `use` for `true`.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/arangodb_attributes.py
```diff
@@ -0,0 +1,15 @@
+ALLOW_DIRTY_READS = "arangodb.options.allowDirtyReads"
+
+ALLOW_RETRY = "arangodb.options.allowRetry"
+
+CACHE = "arangodb.options.cache"
```
thompson-tomo: Suggest naming this `arangodb.query.cache`.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/arangodb_attributes.py
```diff
@@ -0,0 +1,15 @@
...
+
+FAIL_ON_WARNING = "arangodb.options.failOnWarning"
+
+FULL_COUNT = "arangodb.options.fullCount"
+
+MAX_RUNTIME = "arangodb.options.maxRuntime"
+
+STREAM = "arangodb.options.stream"
```
thompson-tomo: Suggest naming this `arangodb.query.stream`.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+            if "maxRuntime" in options:
+                attributes[arangodb_attributes.MAX_RUNTIME] = options.get(
+                    "maxRuntime"
+                )
+            if "stream" in options:
+                attributes[arangodb_attributes.STREAM] = options.get("stream")
+            if "usePlanCache" in options:
+                attributes[arangodb_attributes.USE_PLAN_CACHE] = options.get(
+                    "usePlanCache"
+                )
```
thompson-tomo: I can't see this option in the js sdk, suggest removing it.

thompson-tomo (commented):
Looks better but I feel we need to be careful about how many attributes we are adding to ensure that they all bring value as each attribute has a cost associated with adding them.

mdgilene:
Sorry it's taken so long for me to look at this again. The last month has been pretty busy.

> Looks better but I feel we need to be careful about how many attributes we are adding to ensure that they all bring value as each attribute has a cost associated with adding them.

Removed a lot of the superfluous attributes.

instrumentation/opentelemetry-instrumentation-arangodb/src/opentelemetry/instrumentation/arangodb/__init__.py
mdgilene: Per the ArangoDB Docs

> The total number of result documents available (only available if the query was executed with the count attribute set).

https://docs.arango.ai/arangodb/3.12/develop/http-api/queries/aql-queries/#read-the-next-batch-from-a-cursor_res_200_count

PR #3841, opened 2025-10-13
feature: hydrate AsyncPG span attributes at creation

### LINKED ISSUE #3643: Support filtering AsyncPG spans

When using the AsyncPG instrumentation, spans are created with no attributes and the attributes are added immediately after by using span.set_attribute. This means samplers can't rely on request attributes which are known during span creation, and in practice have no way to sample any span.

The attribute hydration function doesn't incur a meaningful performance penalty and can be used during span creation instead of within the span itself. This is also more correct for timing the actual request time.

There is no way to filter spans on specific tables or actions without relying on other logic such as some contextvar and code changes.

### Additional Context

_No response_

None

forafox:
# Description

This change enables filtering of AsyncPG spans by moving attribute hydration from inside the span context to before span creation. Previously, spans were created with no attributes and attributes were added after span creation using `span.set_attribute()`, which prevented samplers from filtering based on request attributes.

The fix moves the `_hydrate_span_from_args()` call outside the span context and passes the hydrated attributes directly to `start_as_current_span()` using the `attributes` parameter. This ensures that database attributes (system, name, user, statement, network info) are available during span creation for sampling decisions.

Fixes #3643

## Type of change

# How Has This Been Tested?

The changes were tested by:

- Running existing AsyncPG instrumentation tests to ensure no regressions
- Verifying that attributes are now available during span creation by examining the modified code structure
- Confirming that the `attributes` parameter is properly passed to `start_as_current_span()` in both `_do_execute` and `_do_cursor_execute` methods

The implementation maintains backward compatibility and does not change the external API.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `opentelemetry-instrumentation-asyncpg`: Hydrate span attributes before creation so samplers can filter on database details
-  ([#3643](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3643))
+  ([#3841](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3841))
 - `opentelemetry-instrumentation-django`: Fix exemplars generation for `http.server.(request.)duration`
   ([#3945](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3945))
 - `opentelemetry-util-http`, `opentelemetry-instrumentation-requests`, `opentelemetry-instrumentation-wsgi`, `opentelemetry-instrumentation-asgi`: normalize byte-valued user-agent headers before detecting synthetic sources so attributes are recorded reliably.


xrmx (commented):
Which attributes do you need for sampling decisions? If the tests don't fail it looks like we don't have tests for not recording :)
And we need a changelog entry.

forafox:
> Which attributes do you need for sampling decisions? If the tests don't fail it looks like we don't have tests for not recording :) And we need a changelog entry.

The sampler now receives DB system/name/user, DB statement (and parameters when capture_parameters is on), and peer host/port/transport because we hydrate attributes before start_as_current_span. Added tests covering execute and cursor paths with non-recording spans to ensure those attributes are present for sampling decisions, and added a changelog entry

tammy-baylis-swi:
Thanks for contributing! Please also have a look at the [docker-tests](https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/20009539228/job/57465409663?pr=3841), and do a `tox -e precommit` on local to address any formatting issues.

forafox:
> Thanks for contributing! Please also have a look at the [docker-tests](https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/20009539228/job/57465409663?pr=3841), and do a `tox -e precommit` on local to address any formatting issues.

@tammy-baylis-swi 

Hi! Thanks for the feedback — I’ve addressed the formatting issues and updated the docker tests as requested.
Everything should be fixed now

tests/opentelemetry-docker-tests/tests/asyncpg/test_asyncpg_functional.py
```diff
@@ -425,3 +427,126 @@ def test_instrumented_fetch_method_broken_asyncpg(self, *_, **__):
...
+
+    def is_recording(self):
+        return self._recording
+
+    def set_status(self, status):
+        self.set_status_calls += 1
+        self.status = status
+
+
+class _CapturingTracer:
```
tammy-baylis-swi: This could potentially be replaced with a real Tracer instance that uses [InMemorySpanExporter](https://github.com/open-telemetry/opentelemetry-python/blob/ea7da67c406ae382b8d1a677d3ac23937a81cb94/opentelemetry-sdk/src/opentelemetry/sdk/trace/export/in_memory_span_exporter.py#L22).

tests/opentelemetry-docker-tests/tests/asyncpg/test_asyncpg_functional.py
```diff
@@ -425,3 +427,126 @@ def test_instrumented_fetch_method_broken_asyncpg(self, *_, **__):
         self.assertIs(StatusCode.UNSET, spans[0].status.status_code)
         self.assertEqual(spans[0].name, "postgresql")
         self.assertEqual(spans[0].attributes[SpanAttributes.DB_STATEMENT], "")
+
+
+class _NonRecordingSpan:
```
tammy-baylis-swi: Similarly there is a [NonRecordingSpan](https://github.com/open-telemetry/opentelemetry-python/blob/ea7da67c406ae382b8d1a677d3ac23937a81cb94/opentelemetry-api/src/opentelemetry/trace/span.py#L513), maybe use directly or subclass for this test?

forafox: > Similarly there is a [NonRecordingSpan](https://github.com/open-telemetry/opentelemetry-python/blob/ea7da67c406ae382b8d1a677d3ac23937a81cb94/opentelemetry-api/src/opentelemetry/trace/span.py#L513), maybe use directly or subclass for this test?

Dropped the bespoke non-recording span; the tests now use the SDK tracer pipeline (no custom span subclass needed)

tests/opentelemetry-docker-tests/tests/asyncpg/test_asyncpg_functional.py
forafox: > This could potentially be replaced with a real Tracer instance that uses [InMemorySpanExporter](https://github.com/open-telemetry/opentelemetry-python/blob/ea7da67c406ae382b8d1a677d3ac23937a81cb94/opentelemetry-sdk/src/opentelemetry/sdk/trace/export/in_memory_span_exporter.py#L22).

Replaced the custom tracer with a real SDK tracer + SimpleSpanProcessor + InMemorySpanExporter, so the sampling-attribute checks now validate actual exported spans

CHANGELOG.md
```diff
@@ -73,6 +73,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3936](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3936))
 - `opentelemetry-instrumentation-aiohttp-client`: Update instrumentor to respect suppressing http instrumentation
   ([#3957](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3957))
+- `opentelemetry-instrumentation-asyncpg`: Hydrate span attributes before creation so samplers can filter on database details
```
tammy-baylis-swi: Nit: Please move this up under `## Unreleased`

CHANGELOG.md
```diff
@@ -73,7 +78,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3936](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3936))
 - `opentelemetry-instrumentation-aiohttp-client`: Update instrumentor to respect suppressing http instrumentation
   ([#3957](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3957))
-
 ## Version 1.38.0/0.59b0 (2025-10-16)
```
xrmx, line 86:
```suggestion

## Version 1.38.0/0.59b0 (2025-10-16)
```

PR #3848, opened 2025-10-15
django: handle cancelled requests more cleanly

rtpg:
# Description

Django's ASGI handler [cancels tasks that have seen a disconnect](https://github.com/django/django/blob/02eed4f37879b2077496f86bb1378a076b981233/django/core/handlers/asgi.py#L204). This manifests itself in requests potentially opening an OTel span through the Django middleware but not having `process_response` run in the existing mode, meaning the `use_span` ([this one](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/ebe5bdf465b9cf893e0b0db72820a89a38ad1601/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py#L285)) closure + context `detach` call isn't run.

This results in dreaded "this token was created in a different context" errors, which happen because:

- a task gets cancelled before the `use_span` activation gets `__exit__`'d in `get_response`
- the `use_span` generator instead gets garbage collected after the fact
- the garbage collection runs the `finally` clause inside the context management code of `use_span`
- this causes a call to `detach` many miles away from the request context

The current Django Middleware design dates back to supporting "old style" middleware, making this problem hard to solve cleanly. Given Django 1.x is now 5 years out of support,  I think it's reasonable to drop Django 1.x support to make context management here clearer, through using the new-style `__call__` middleware along with a `try/finally` block.

Fixes # (issue)

This PR does some cleanup to the middleware to make sure that our context and span cleanup always happen, including when there's an `asyncio.CancelledError` during process handling. But this PR also drops support for Django 1.x in the process, by expecting this middleware to be called via `__call__`

(I'll handle doc changes etc if I can get approval on dropping Django 1.x support)

## Type of change

# How Has This Been Tested?

- [x] I added a test into the suite that catches this change
- [x] Testing "live"

In order to see this "in the wild", I added the following sort of thing to a random Django view:

```
print("sleeping")
sleep(5)
print("done")
```

I then ran this app with `uvicorn`, and made requests. When I saw "sleeping" I would close the browser tabs, cancelling the requests. This would lead to the context error appearing.

- [X] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -77,30 +77,8 @@
 except ImportError:
     from django.urls import Resolver404, resolve
 
-DJANGO_2_0 = django_version >= (2, 0)
```
xrmx, line 80: Take a look pyproject.toml  `instruments` and `opentelemetry.instrumentation.django.package`, when you change supported versions you should also change them

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -357,74 +370,56 @@ def process_response(self, request, response):
         )
         request_start_time = request.META.pop(self._environ_timer_key, None)
 
-        if activation and span:
```
rtpg, line 360: in the case where this method is run, there's definitely an `activation` and `span`, so we don't need this `if` block anymore

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
rtpg: OK I went in and removed references to `DJANGO_2_0`, updated the `pyproject.toml` `instruments` and `package` files.

rtpg:
Alright I handled the Django 1.x deprecations across the package, and went ahead and added changelog entries.

If you want for me to squash the commits I can, but it looks like maintainers are the ones managing commit creation based on the `git log`

rtpg:
Adding a comment here that we've been running this patch successfully for a good amount of time now.

zmievsa-semasoftware:
This would help us significantly. Any expectation when this could be merged?

zmievsa-semasoftware:
@xrmx Could you assist here?

xrmx:
My suggestion would be to extract the changes to drop django < 2 support in another PR that would be easier to review and merge so this one will become simpler to review.

splatonov:
@xrmx hi! I think this PR was closed accidentally

zmievsa:
@splatonov I don't think it was.

We merged the django PR separately and now we will have to recreate this PR but with just the ASGI changes

splatonov:
Oh, okay

PR #3852, opened 2025-10-15
Updating workflows and RELEASING.md to allow openai-agents-v2 individual release

lzchen:


RELEASING.md
aabmass, line 85: one more spot its needed

aabmass, line 1: Also needed in https://github.com/open-telemetry/opentelemetry-python-contrib/blob/168f3902ef5c5a5557092de1cc75489ff8ac8e7f/eachdist.ini#L49-L50

PR #3862, opened 2025-10-15
GenAI Utils | Add more SemConv Attributes

keith-decker:
# Description

Add missing SemConv attributes to spans for LLMInvocation calls, clean up some repeated code in the unit tests for genai utils.

## Type of change

# How Has This Been Tested?

Updated unit tests to check for additional attributes

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3904](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3904))
 - build: bump ruff to 0.14.1
   ([#3842](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3842))
+- `opentelemetry-instrumentation-redis`: Add default span name for pipeline operations
+  ([#3941](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3941))
+- `opentelemetry-instrumentation-pymongo`: Fix invalid mongodb collection attribute type
+  ([#3942](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3942))
 
 ## Version 1.38.0/0.59b0 (2025-10-16)
 
--- a/instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py
@@ -138,7 +138,7 @@ class CommandTracer(monitoring.CommandListener):
         command_name = event.command_name
         span_name = f"{event.database_name}.{command_name}"
         statement = self._get_statement_by_command_name(command_name, event)
-        collection = event.command.get(event.command_name)
+        collection = _get_command_collection_name(event)
 
         try:
             span = self._tracer.start_span(span_name, kind=SpanKind.CLIENT)
@@ -224,6 +224,13 @@ class CommandTracer(monitoring.CommandListener):
         if command and self.capture_statement:
             statement += " " + str(command)
         return statement
+
+
+def _get_command_collection_name(event: CommandEvent) -> str | None:
+    collection_name = event.command.get(event.command_name)
+    if not collection_name or not isinstance(collection_name, str):
+        return None
+    return collection_name
 
 
 def _get_span_dict_key(
--- a/instrumentation/opentelemetry-instrumentation-pymongo/tests/test_pymongo.py
+++ b/instrumentation/opentelemetry-instrumentation-pymongo/tests/test_pymongo.py
@@ -278,6 +278,43 @@ class TestPymongo(TestBase):
             span.attributes[SpanAttributes.DB_STATEMENT], "aggregate"
         )
 
+    def test_collection_name_attribute(self):
+        scenarios = [
+            (
+                {
+                    "command_name": "find",
+                    "find": "test_collection",
+                },
+                "test_collection",
+            ),
+            ({"command_name": "find"}, None),
+            ({"command_name": "find", "find": b"invalid"}, None),
+        ]
+        for command_attrs, expected in scenarios:
+            with self.subTest(command_attrs=command_attrs, expected=expected):
+                mock_event = MockEvent(command_attrs)
+
+                command_tracer = CommandTracer(
+                    self.tracer, capture_statement=True
+                )
+                command_tracer.started(event=mock_event)
+                command_tracer.succeeded(event=mock_event)
+
+                spans_list = self.memory_exporter.get_finished_spans()
+
+                self.assertEqual(len(spans_list), 1)
+                span = spans_list[0]
+
+                self.assertEqual(
+                    span.attributes[SpanAttributes.DB_STATEMENT], "find"
+                )
+
+                self.assertEqual(
+                    span.attributes.get(SpanAttributes.DB_MONGODB_COLLECTION),
+                    expected,
+                )
+                self.memory_exporter.clear()
+
 
 class MockCommand:
     def __init__(self, command_attrs):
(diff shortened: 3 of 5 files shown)


util/opentelemetry-util-genai/tests/test_utils.py
```diff
@@ -123,72 +197,71 @@ def tearDown(self):
...
-        assert input_messages[0].get("role") == "Human"
-        assert output_messages[0].get("role") == "AI"
-        assert output_messages[0].get("finish_reason") == "stop"
-        assert (
-            output_messages[0].get("parts")[0].get("content") == "hello back"
+        span_attrs = _get_span_attributes(span)
+        _assert_span_attributes(
+            span_attrs,
+            {
+                "gen_ai.operation.name": "chat",
```
zhirafovod: here, and everywhere else in the code - `GenAI.GEN_AI_OPERATION_NAME`, `GenAI.GEN_AI_SYSTEM`

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -93,7 +93,7 @@ def __init__(self, tracer_provider: TracerProvider | None = None):
             __name__,
             __version__,
             tracer_provider,
-            schema_url=Schemas.V1_36_0.value,
+            schema_url=Schemas.V1_37_0.value,
```
zhirafovod, line 96: Please add a test to validate the attribute is present in the telemetry. Document what is it set in the telemetry

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from dataclasses import asdict
-from typing import List
+from typing import Any, Dict, List, Optional
```
aabmass: nit prefer to use `dict[k, v]`, and `Foo | None` with `from __future__ import annotations`.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -112,6 +93,8 @@ def _apply_finish_attributes(span: Span, invocation: LLMInvocation) -> None:
     _maybe_set_span_messages(
         span, invocation.input_messages, invocation.output_messages
     )
+    _apply_request_attributes(span, invocation)
```
lmolkova, line 98: not blocking for this PR, but we should eventually allow to set attributes at start time. I noticed we don't properly mark which attributes should be provided at the start time, fixing it in the https://github.com/open-telemetry/semantic-conventions/pull/2994

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -122,7 +105,68 @@ def _apply_error_attributes(span: Span, error: Error) -> None:
...
+    if invocation.response_finish_reasons is not None:
+        finish_reasons = invocation.response_finish_reasons
+    elif invocation.output_messages:
+        finish_reasons = [
+            message.finish_reason for message in invocation.output_messages
+        ]
+    else:
+        finish_reasons = None
+
+    if finish_reasons:
```
lmolkova, line 153: I believe if finish_reasons is not available, we should set `error` - https://github.com/open-telemetry/semantic-conventions/issues/2919#issuecomment-3423908668

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -111,9 +111,17 @@ class LLMInvocation:
     provider: Optional[str] = None
     response_model_name: Optional[str] = None
     response_id: Optional[str] = None
+    response_finish_reasons: Optional[List[str]] = None
```
lmolkova: nit: is there a need for `response_`?

```suggestion
    finish_reasons: Optional[List[str]] = None
```

keith-decker: done!

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
keith-decker: Done!

util/opentelemetry-util-genai/tests/test_utils.py
keith-decker: Done!

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -122,7 +105,68 @@ def _apply_error_attributes(span: Span, error: Error) -> None:
...
+
+def _apply_response_attributes(span: Span, invocation: LLMInvocation) -> None:
+    """Attach GenAI response semantic convention attributes to the span."""
+    attributes: Dict[str, Any] = {}
+
+    finish_reasons: Optional[List[str]]
+    if invocation.finish_reasons is not None:
+        finish_reasons = invocation.finish_reasons
+    elif invocation.output_messages:
+        finish_reasons = [
```
DylanRussell, line 145: wonder if this should be a sorted set instead of a list ? or at least converted to a set and then converted back to a sorted list to get rid of duplicates

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
DylanRussell: hmm what would you set error to in that case ? I feel like instrumentations should set error explicitly (or call into a util that sets it explicitly) rather than try to infer it like that

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: Finish reasons are sorted and unique now. some unit tests for this added as well.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
keith-decker: cleaned up!

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -76,18 +77,18 @@ class InputMessage:
...
 
 
-def _new_output_messages() -> List[OutputMessage]:
+def _new_output_messages() -> list[OutputMessage]:
     return []
 
 
-def _new_str_any_dict() -> Dict[str, Any]:
+def _new_str_any_dict() -> dict[str, Any]:
     return {}
```
DylanRussell, line 94: So my reading of https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/ is that "recommended" attributes SHOULD be included by default.. So for the attributes that correspond to these fields that are "recommended" I think we should be defaulting these to  empty strings and 0 values instead of None (which means they will not be added).. Thoughts @aabmass @lmolkova  ?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
aabmass: I think Liudmila means literally setting the reason to error `finish_reason = FinishReasons.ERROR`. Not talking about the `error` attribute itself

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
lmolkova: @DylanRussell you're right that they should be provided by default, but only when they are available, we don't need to include empty values - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/attribute-requirement-level.md#recommended

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
lmolkova: offline discussion with @keith-decker - it should probably be an instrumentation responsibility to provide error here and utils should not make assumptions. resolving here

PR #3891, opened 2025-10-22
GenAI Utils | Add metrics to LLMInvocations

keith-decker:
# Description

Add duration and token metrics to LLMInvocation traces.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Unit tests show metrics exporting.

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
+++ b/util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
@@ -11,7 +11,10 @@ from opentelemetry.semconv._incubating.attributes import (
     gen_ai_attributes as GenAI,
 )
 from opentelemetry.trace import Span, set_span_in_context
-from opentelemetry.util.genai.instruments import Instruments
+from opentelemetry.util.genai.instruments import (
+    create_duration_histogram,
+    create_token_histogram,
+)
 from opentelemetry.util.genai.types import LLMInvocation
 from opentelemetry.util.types import AttributeValue
 
@@ -20,11 +23,8 @@ class InvocationMetricsRecorder:
     """Records duration and token usage histograms for GenAI invocations."""
 
     def __init__(self, meter: Meter):
-        instruments = Instruments(meter)
-        self._duration_histogram: Histogram = (
-            instruments.operation_duration_histogram
-        )
-        self._token_histogram: Histogram = instruments.token_usage_histogram
+        self._duration_histogram: Histogram = create_duration_histogram(meter)
+        self._token_histogram: Histogram = create_token_histogram(meter)
 
     def record(
         self,
--- a/util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py
+++ b/util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py
@@ -36,17 +36,19 @@ _GEN_AI_CLIENT_TOKEN_USAGE_BUCKETS = [
 ]
 
 
-class Instruments:
-    def __init__(self, meter: Meter):
-        self.operation_duration_histogram: Histogram = meter.create_histogram(
-            name=gen_ai_metrics.GEN_AI_CLIENT_OPERATION_DURATION,
-            description="Duration of GenAI client operation",
-            unit="s",
-            explicit_bucket_boundaries_advisory=_GEN_AI_CLIENT_OPERATION_DURATION_BUCKETS,
-        )
-        self.token_usage_histogram: Histogram = meter.create_histogram(
-            name=gen_ai_metrics.GEN_AI_CLIENT_TOKEN_USAGE,
-            description="Number of input and output tokens used by GenAI clients",
-            unit="{token}",
-            explicit_bucket_boundaries_advisory=_GEN_AI_CLIENT_TOKEN_USAGE_BUCKETS,
-        )
+def create_duration_histogram(meter: Meter) -> Histogram:
+    return meter.create_histogram(
+        name=gen_ai_metrics.GEN_AI_CLIENT_OPERATION_DURATION,
+        description="Duration of GenAI client operation",
+        unit="s",
+        explicit_bucket_boundaries_advisory=_GEN_AI_CLIENT_OPERATION_DURATION_BUCKETS,
+    )
+
+
+def create_token_histogram(meter: Meter) -> Histogram:
+    return meter.create_histogram(
+        name=gen_ai_metrics.GEN_AI_CLIENT_TOKEN_USAGE,
+        description="Number of input and output tokens used by GenAI clients",
+        unit="{token}",
+        explicit_bucket_boundaries_advisory=_GEN_AI_CLIENT_TOKEN_USAGE_BUCKETS,
+    )


util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -88,13 +91,41 @@ class TelemetryHandler:
...
             __name__,
             __version__,
             tracer_provider,
             schema_url=Schemas.V1_37_0.value,
         )
+        self._metrics_recorder: InvocationMetricsRecorder | None = None
+        try:
+            meter = get_meter(__name__, meter_provider=meter_provider)
+            self._metrics_recorder = InvocationMetricsRecorder(meter)
+        except Exception:  # pragma: no cover - defensive fallback  # pylint: disable=broad-exception-caught
```
aabmass: Why would this fail though, the OTel API should never raise exceptions

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -88,13 +91,41 @@ class TelemetryHandler:
...
+    ) -> None:
+        if self._metrics_recorder is None or span is None:
+            return
+        try:
+            self._metrics_recorder.record(
+                span,
+                invocation,
+                error_type=error_type,
+            )
+        except Exception:  # pragma: no cover - defensive fallback  # pylint: disable=broad-exception-caught
```
aabmass: same comment

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -0,0 +1,125 @@
...
+        invocation: LLMInvocation,
+        *,
+        error_type: Optional[str] = None,
+    ) -> None:
+        """Record duration and token metrics for an invocation if possible."""
+        if span is None:
+            return
+
+        tokens: list[tuple[int, str]] = []
+        if isinstance(invocation.input_tokens, int):
```
aabmass: can you instead just check if not None, IMO it's much more idiomatic

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -0,0 +1,125 @@
...
+        if isinstance(invocation.output_tokens, int):
+            tokens.append(
+                (
+                    invocation.output_tokens,
+                    GenAI.GenAiTokenTypeValues.COMPLETION.value,
+                )
+            )
+
+        if not tokens:
+            return
```
aabmass: Is there a reason not to record duration in this case?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -0,0 +1,125 @@
...
+        # Calculate duration from span timing
+        duration_seconds = _calculate_duration_seconds(span)
+
+        span_context = set_span_in_context(span)
+        if error_type:
+            attributes["error.type"] = error_type
+
+        if (
+            duration_seconds is not None
+            and isinstance(duration_seconds, Number)
```
aabmass, line 81: Is this check needed, the type should already be float

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -0,0 +1,125 @@
...
+            and duration_seconds >= 0
+        ):
+            duration_attributes: Dict[str, AttributeValue] = dict(attributes)
+            self._duration_histogram.record(
+                float(duration_seconds),
```
aabmass:
```suggestion
                duration_seconds,
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -0,0 +1,125 @@
...
+                attributes=duration_attributes,
+                context=span_context,
+            )
+
+        for token in tokens:
```
aabmass: nit i think this would be more readable
```suggestion
        for token_count, token_type in tokens:
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -0,0 +1,125 @@
...
+            token_attributes: Dict[str, AttributeValue] = dict(attributes)
+            token_attributes[GenAI.GEN_AI_TOKEN_TYPE] = token[1]
+            self._token_histogram.record(
+                token[0],
+                attributes=token_attributes,
+                context=span_context,
+            )
```
aabmass: nit imo this is a bit easier to understand rather than mutable state

```suggestion
            self._token_histogram.record(
                token[0],
                attributes=attributes | {GenAI.GEN_AI_TOKEN_TYPE: token[1]}
                context=span_context,
            )
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -0,0 +1,125 @@
...
+        span: Optional[Span],
+        invocation: LLMInvocation,
+        *,
+        error_type: Optional[str] = None,
+    ) -> None:
+        """Record duration and token metrics for an invocation if possible."""
+        if span is None:
+            return
+
+        tokens: list[tuple[int, str]] = []
```
aabmass: can you use a more descriptive variable name, I thought this was holding the actual tokens at first.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -0,0 +1,125 @@
...
+                (
+                    invocation.input_tokens,
+                    GenAI.GenAiTokenTypeValues.INPUT.value,
+                )
+            )
+        if isinstance(invocation.output_tokens, int):
+            tokens.append(
+                (
+                    invocation.output_tokens,
+                    GenAI.GenAiTokenTypeValues.COMPLETION.value,
```
aabmass: Was COMPLETION intentional? It is deprecated and should use OUTPUT instead  https://github.com/open-telemetry/semantic-conventions/blob/v1.38.0/docs/registry/attributes/gen-ai.md#deprecated-genai-attributes

util/opentelemetry-util-genai/tests/test_handler_metrics.py
```diff
@@ -0,0 +1,160 @@
...
+                getattr(resource_metric, "scope_metrics", []) or []
+            ):
+                for metric in getattr(scope_metric, "metrics", []) or []:
+                    points = list(
+                        getattr(metric.data, "data_points", []) or []
+                    )
+                    if points:
+                        metrics_by_name.setdefault(metric.name, []).extend(
+                            points
+                        )
```
aabmass: Can you do it without `getattr()` since it breaks editor LSP integration/typing. You can do `data.resource_metrics or []` to get the same behavior

aabmass: I actually noticed the type annotations for InMemoryMetricReader are broken, will send a PR. But the comment above still applies

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -0,0 +1,125 @@
...
+from opentelemetry.util.types import AttributeValue
+
+_NS_PER_SECOND = 1_000_000_000
+
+
+def _now_ns() -> int:
+    return time.time_ns()
+
+
+def _get_span_start_time_ns(span: Optional[Span]) -> Optional[int]:
```
aabmass: We generally measure the durations separately with [`default_timer()`](https://docs.python.org/3/library/timeit.html#timeit.default_timer) which is monotonic clock and more accurate when absolute timestamps are not needed. https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-python-contrib%20default_timer&type=code

It would also work even if the Tracing SDK is not set up and start_time can't be read.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
keith-decker: Good catch, duration should still be recorded even if tokens didn't return

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
keith-decker: no, just a bit of legacy copy paste from when the semconvs were still prompt and completion

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
keith-decker: Well, that led me down a rabbit hole of monotonic clocks for duration versus wall clocks for times. Interesting stuff.

I used to calculate the duration based on the span start. That used a wall clock based on the tracer setting it and metrics should use the monotonic clock, I had to add a property and set it on start_llm.

Take a look at these changes and let me know if you think they are too much.

util/opentelemetry-util-genai/tests/test_handler_metrics.py
aabmass: https://github.com/open-telemetry/opentelemetry-python/pull/4820 fyi

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -0,0 +1,136 @@
...
+        if isinstance(start_s, (int, float)):
+            elapsed_s = max(timeit.default_timer() - float(start_s), 0.0)
+            return elapsed_s
+
+    # Fall back to span start_time (wall clock epoch ns)
+    start_time_ns = _get_span_start_time_ns(span)
+    if start_time_ns is None:
+        return None
+    elapsed_ns = max(time.time_ns() - start_time_ns, 0)
+    return elapsed_ns / _NS_PER_SECOND
```
aabmass: Can remove this and `_get_span_start_time_ns` altogether? The caller should have called `start_llm()` which would set the clock first

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -0,0 +1,136 @@
...
+        return None
+    for attr in ("start_time", "_start_time"):
+        value = getattr(span, attr, None)
+        if isinstance(value, int):
+            return value
+    return None
+
+
+def _calculate_duration_seconds(
+    span: Optional[Span], invocation: Optional[LLMInvocation] = None
```
aabmass: FYI I pushed a commit to clean this up a bit by making the typing more restrictive so the code handle fewer defensive cases

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -123,6 +123,10 @@ class LLMInvocation:
     max_tokens: int | None = None
     stop_sequences: list[str] | None = None
     seed: int | None = None
+    # Monotonic start time in seconds (from timeit.default_timer) used
+    # for duration calculations to avoid mixing clock sources. This is
+    # populated by the TelemetryHandler when starting an invocation.
```
aabmass: Should this be a docstring?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -0,0 +1,130 @@
...
+
+class InvocationMetricsRecorder:
+    """Records duration and token usage histograms for GenAI invocations."""
+
+    def __init__(self, meter: Meter):
+        instruments = Instruments(meter)
+        self._duration_histogram: Histogram = (
+            instruments.operation_duration_histogram
+        )
+        self._token_histogram: Histogram = instruments.token_usage_histogram
```
aabmass: consider making this two functions instead of class
```py
self._duration_histogram = create_operation_histogram(meter)
self._token_histogram = create_token_histogram(meter)
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -0,0 +1,110 @@
...
+)
+from opentelemetry.trace import Span, set_span_in_context
+from opentelemetry.util.genai.instruments import Instruments
+from opentelemetry.util.genai.types import LLMInvocation
+from opentelemetry.util.types import AttributeValue
+
+_NS_PER_SECOND = 1_000_000_000
+
+
+def _get_span_start_time_ns(span: Optional[Span]) -> Optional[int]:
```
aabmass: I think this can be deleted now

PR #3894, opened 2025-10-24
Fix ImportError with slash-delimited Lambda handler paths

### LINKED ISSUE #1465: Slash-delimited handler paths result in ImportError within AwsLambdaInstrumentor.instrument
**Describe your environment** 

**Runtime:** Python 3.9, amd64
**Dependencies:** Only those that come with the OpenTelemetry layer for AWS Lambda (v1.14.0)
**Cloud Provider**: AWS (specifically, AWS Lambda)

This issue does still appear to exist in `main`, but I have not had a chance to verify. The issue centers around the format of the handler string (it breaks if it is slash-delimited) and the instrumenter currently loads `ORIG_HANDLER` without modification, ultimately leading to attempting to patch an invalid import path (e.g., `package1/package2/module.target`).
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/80d0b89b8b9dbdfa1379f7cbbbcb2cd47ff0649a/instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py#L372-L377

**Steps to reproduce**

I created https://github.com/myles2007/OpenTelemIssue-PythonLambdaLayer-v1.14.0 which defines a sample application which reproduces the error, includes CloudFormation for the infrastructure, and [commands to deploy the example](https://github.com/myles2007/OpenTelemIssue-PythonLambdaLayer-v1.14.0#reproducing-the-issue). From there, you only need to invoke the Lambda function named `SlashPathWithLayerBroken`.

If you would prefer to recreate manually, the basic reproduction steps are:

 1. Create a Python application with a structure like: 
     ```
     package1 /
         package2/
             module.py
                 handler (function)
    ```
 2. Deploy the application using AWS Lambda and specify the handler as: `package1/package2/module.handler`
     > Without instrumentation applied, this works.
 3. Enable the region-appropriate Python Lambda Layer for OpenTelemetry:
     -  For `us-east-1`: `arn:aws:lambda:us-east-1:901920570463:layer:aws-otel-python-amd64-ver-1-14-0:1`
     - Also set `AWS_LAMBDA_EXEC_WRAPPER` to "/opt/otel-instrument"
 4. Invoke the Lambda function and receive the following error:
      ```json
      {
            "errorMessage": "Unable to import module 'otel_wrapper': No module named 'package1/package2/module'",
            "errorType": "Runtime.ImportModuleError",
            "requestId": "b8fefc59-1377-4cf3-aec6-ba7ea8cdfa84",
            "stackTrace": []
     }
     ```

As it does without instrumentation, the invocation should succeed without issue even when slash-delimited paths are used.

Invocation with the AWS Lambda auto instrumentation in enabled results in an import error at runtime.

**Additional context**
Lambda function handlers can be specified in at least the two following ways for Python Lambda functions:

 - `package1.package2.module.handler`
 - `package1/package2/module.handler`

Either of these forms is valid and executes `module.handler` as expected when the function is invoked. However, the OpenTelemetry Lambda layer does not properly handle the second form. If it is enabled, invocation results in the following error response when slash-based paths are used:

```json
{
  "errorMessage": "Unable to import module 'otel_wrapper': No module named 'package1/package2/module'",
  "errorType": "Runtime.ImportModuleError",
  "requestId": "b8fefc59-1377-4cf3-aec6-ba7ea8cdfa84",
  "stackTrace": []
}
```

The [otel_wrapper.py](https://github.com/open-telemetry/opentelemetry-lambda/blob/f48ed2cff0c18722639d6062b93b02c64fc9cec4/python/src/otel/otel_sdk/otel_wrapper.py) script handles both forms by modifying the `ORIG_HANDLER` path (see [here](https://github.com/open-telemetry/opentelemetry-lambda/blob/f48ed2cff0c18722639d6062b93b02c64fc9cec4/python/src/otel/otel_sdk/otel_wrapper.py#L43)), but [AwsLambdaInstrumentor().instrument()](https://github.com/open-telemetry/opentelemetry-lambda/blob/f48ed2cff0c18722639d6062b93b02c64fc9cec4/python/src/otel/otel_sdk/otel_wrapper.py#L52) does not and instrumenting the function fails with the error above.

I believe the error originates within the call to `wrapt.wrap_function_wrapper` [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/b6b269064c5e9bf1304dcdbe6686cca2dd87eb42/instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py#L347) when patching occurs because the slash-based path format is not a valid Python import path.

**Thoughts on Resolving the Issue**

Below are a few thoughts on how this issue might be resolved, including a simple workaround. I'm happy to contribute a change, but would like guidance from maintainers on the preferred approach, whether one of mine or something different.

_Workaround_
This issue can easily be worked around by specifying the handler location using dot-based paths. Nothing about project structure needs to change.

_Modify ORIG_HANDLER in `otel_wrapper.py`_
Modifying the `ORIG_HANDLER` environment variable within `otel_wrapper.py` before instrumentation occurs by replacing it with the modified path that is computed for use within itself resolves the issue.

While this works, it does mutate the `ORIG_HANDLER` variable in a way which makes it no longer truly the original handler path. Side effects outside of the intended scope are definitely possible, but it might also help prevent similar issues sharing ths root cause.

This is the fix I implemented when debugging this issue. It does work and requires only minimal modifications. This change would be made in https://github.com/open-telemetry/opentelemetry-lambda, not this repository.

_Modify `AwsLambdaInstrumentor.instrument`_
The `AwsLambdaInstrumentor.instrument` method could be modified to locally compute the path to the `ORIG_HANDLER` as `otel_wrapper.py` does (i.e., by replacing `/` with `.`). This would ensure that the path is valid for import, preventing the error.

This fix duplicates the path modification logic, but it reduces the potential of unexpected side effects by keeping the scope small and targeted.

----

Please let me know if anything is unclear or if more information is needed somewhere. Thanks!

DivMode:
# Description

AWS Lambda accepts handler paths in both formats:
- Slash-delimited: `python/functions/api.handler`
- Dot-delimited: `python.functions.api.handler`

The instrumentation was failing when slash-delimited paths were used because it attempted to import invalid Python module paths containing forward slashes. This fix normalizes the handler path by converting slashes to dots before splitting the module and function names.

Fixes #1465

## Type of change

# How Has This Been Tested?

- [x] Added unit test `test_slash_delimited_handler_path()` that verifies both slash-delimited and dot-delimited handler paths work correctly
- [x] Tested with AWS Lambda function using slash-delimited handler (`python/functions/api.handler`) in production environment
- [x] Verified OpenTelemetry instrumentation layer successfully imports and instruments the handler
- [x] Previously failed with `ImportError: No module named 'python/functions/api'`, now works correctly
- [x] Tested in production with SST framework which uses slash-delimited paths by default
- [x] All existing tests pass with the change

- [x] No.

# Checklist:

- [x] Documentation has been updated (not applicable - internal bug fix, no API changes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added 
+
+- `opentelemetry-instrumentation-aiohttp-client`: add support for url exclusions via `OTEL_PYTHON_EXCLUDED_URLS` / `OTEL_PYTHON_AIOHTTP_CLIENT_EXCLUDED_URLS`
+  ([#3850](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3850))
+- `opentelemetry-instrumentation-httpx`: add support for url exclusions via `OTEL_PYTHON_EXCLUDED_URLS` / `OTEL_PYTHON_HTTPX_EXCLUDED_URLS`
+  ([#3837](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3837))
+- `opentelemetry-instrumentation-flask`: improve readthedocs for sqlcommenter configuration.
+  ([#3883](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3883))
+
 ### Fixed
 
+- `opentelemetry-instrumentation-botocore`: Handle dict input in _decode_tool_use for Bedrock streaming
+  ([#3875](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3875))
 - `opentelemetry-instrumentation-aws-lambda`: Fix ImportError with slash-delimited handler paths
   ([#3894](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3894))
 
--- a/instrumentation-genai/opentelemetry-instrumentation-langchain/examples/manual/README.rst
+++ b/instrumentation-genai/opentelemetry-instrumentation-langchain/examples/manual/README.rst
@@ -1,4 +1,4 @@
-OpenTelemetry Langcahin Instrumentation Example
+OpenTelemetry Langchain Instrumentation Example
 ===============================================
 
 This is an example of how to instrument Langchain when configuring OpenTelemetry SDK and instrumentations manually.
@@ -8,14 +8,14 @@ Traces include details such as the span name and other attributes.
 
 Note: `.env <.env>`_ file configures additional environment variables:
 - :code:`OTEL_LOGS_EXPORTER=otlp` to specify exporter type.
-- :code:`OPENAI_API_KEY` open AI key for accessing the OpenAI API.
+- :code:`OPENAI_API_KEY` key for accessing the OpenAI API.
 - :code:`OTEL_EXPORTER_OTLP_ENDPOINT` to specify the endpoint for exporting traces (default is http://localhost:4317).
 
 Setup
 -----
 
 Minimally, update the `.env <.env>`_ file with your :code:`OPENAI_API_KEY`.
-An OTLP compatible endpoint should be listening for traces http://localhost:4317.
+An OTLP compatible endpoint should be listening for traces at http://localhost:4317.
 If not, update :code:`OTEL_EXPORTER_OTLP_ENDPOINT` as well.
 
 Next, set up a virtual environment like this:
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- `opentelemetry-instrumentation-aws-lambda`: Fix ImportError with slash-delimited handler paths
+  ([#1465](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1465))
```
tammy-baylis-swi: Nit: we link the PRs in changelog:


```suggestion
  ([#3894](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3894))
```

emdneto (approved):

PR #3896, opened 2025-10-25
Use canonical name for aiohttp request span name

krnr:
# Description

Current implementation doesn't use [canonical](https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.AbstractResource.canonical) attribute of a `Resource`, which leads to the situation when generated spans all have unique names. It creates high cardinality of spans. Other frameworks that use path params create spans with "canonical" paths in names (e.g. fastapi [[1]](https://github.com/open-telemetry/opentelemetry-python-contrib/commit/c05e1798b7463fdadab6cef208f2c3eb6fb501b6#diff-f69ad8214c98cac3eb1283e4b94dedc5d25842518368e6ec114c6f77cdc6364fR72) )

## Type of change

# How Has This Been Tested?

add a new unit test: 

- [x] `test_url_params_instrumentation`

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-aiohttp-server/tests/test_aiohttp_server_integration.py
```diff
@@ -139,6 +148,53 @@ async def test_status_code_instrumentation(
...
+    client = await aiohttp_client(server)
+    for path in example_paths:
+        await client.get(path)
+
+    assert len(memory_exporter.get_finished_spans()) == 2
+
+    for request_path, span in zip(
+        example_paths, memory_exporter.get_finished_spans()
+    ):
+        assert url == span.name
```
bisgaard-itis: 👍🏻

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
```diff
@@ -123,6 +123,10 @@ def get_default_span_details(request: web.Request) -> Tuple[str, dict]:
         a tuple of the span name, and any attributes to attach to the span.
     """
     span_name = request.path.strip() or f"HTTP {request.method}"
+    if request.match_info and request.match_info.route.resource:
```
bisgaard-itis: Is `match_info.route` guaranteed to be not `None` when `match_info` is not `None`?

bisgaard-itis (approved):
Looks good. Thanks a lot for the effort! 👍🏻

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
krnr: thank you for the good point. i think this can be the proof: https://github.com/aio-libs/aiohttp/blob/v3.13.1/aiohttp/web_urldispatcher.py#L271

(in fact, even `match_info` can't be `None`: https://github.com/aio-libs/aiohttp/blob/v3.13.1/aiohttp/web_request.py#L890 but it may raise issues with type hints)

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
```diff
@@ -123,6 +123,10 @@ def get_default_span_details(request: web.Request) -> Tuple[str, dict]:
...
+        resource = request.match_info.route.resource
+        if resource.canonical:
+            span_name = resource.canonical
```
tammy-baylis-swi: Thanks for working on this! It will definitely be better to have lower cardinality span names.

Could this be adjusted to make sure the new functionality adheres to the [semconv for http server span names](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name)? This currently sets span_name as `{target}` but it should be `{method} {target}` like L125.

krnr: sure thing! thanks for pointing that out.

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
tammy-baylis-swi: I'm not super familiar with aiohttp, but it could we be extra careful and wrap this with a `try-except AttributeError`? So it's like `_get_view_func` below.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- `opentelemetry-instrumentation-aiohttp-server`: Use `canonical` attribute of the `Resource` as a span name.
```
tammy-baylis-swi: NIt: we add PR urls to the changelog entries:


```suggestion
- `opentelemetry-instrumentation-aiohttp-server`: Use `canonical` attribute of the `Resource` as a span name.
  ([#3896](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3896))
```

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
krnr: i am pretty sure that the attributes always exist: https://github.com/aio-libs/aiohttp/blob/v3.5.0/aiohttp/web_request.py#L704 https://github.com/aio-libs/aiohttp/blob/v3.5.0/aiohttp/web_urldispatcher.py#L161 (at least since v3.5.0)

but we definitely can add this "extra-safe" layer. i am just not sure yet what should the name be in case of AttributeError - the default one from `request.path` or some sentinel because "we-found-smth-unexpected"... having thought about it for a while i decided that the first option is better. let me know if you have any other ideas.

krnr: in fact, there's another possibility of "high cardinality abuse": https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3896/files#diff-682c3e961f134b947f243e707c7a75856ea5640e0dfb8f32dad7ffd1b3fa8c9eR145

every non existent name is a span name by itself. so, you can generate an endless amount of 404 and each of them will be a different span. this logic should be (should it?) fixed in a separate PR, because i can't decide right on the spot what shall we do with 404 span names. should they all be `"GET <Not Found>"` if there's no `match_info` (i think so, and only get the path from the span attributes), or some users rely on the existing behavior and collect an "inventory" of such calls using span names? i don't know.

tammy-baylis-swi: Good questions, hmm. [This part](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name) of the semconv says

> HTTP span names SHOULD be `{method} {target}` if there is a (low-cardinality) target available. If there is no (low-cardinality) `{target}` available, HTTP span names SHOULD be `{method}`.

I'm guessing cardinality of `request.path.strip()` can depend on the app but does have the potential to be high. What do you think of something like below, where we prefer target (`path`) as `canonical` > `None`? If we stop relying `request.path.strip()` entirely, maybe this would address the endless-404s concern, though it might be a small breaking change to some users:

```python
    try:
        resource = request.match_info.route.resource
        if resource:
            path = resource.canonical
    except AttributeError:
        path = ""

    if path:
        return f"{request.method} {path}"
    else:
        return f"{request.method}"
```

krnr: yes, i think it will be better. i just modified it a bit to silent a type checker.

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
```diff
@@ -238,14 +238,22 @@ def _parse_active_request_count_attrs(req_attrs):
 
 
 def get_default_span_name(request: web.Request) -> str:
-    """Default implementation for get_default_span_details
+    """Returns the span name.
     Args:
         request: the request object itself.
     Returns:
-        The span name.
+        The canonical name of a resource if possible or just request path.
```
tammy-baylis-swi: Nit: Needs an update


```suggestion
        The span name as "{method} {canonical_name}" of a resource if possible or just "{method}".
```

tammy-baylis-swi (approved):
Thank you @krnr for this. Lgtm though I want some other Approvers to have a look at this change.

If you wanted to bring this up more synchronously, please feel welcome to join the Python SIG meeting on Zoom on Thursdays -- [Calendar with links here](https://calendar.google.com/calendar/embed?src=c_2bf73e3b6b530da4babd444e72b76a6ad893a5c3f43cf40467abc7a9a897f977%40group.calendar.google.com). Else this PR should still be visible on the [PR digest board](https://github.com/orgs/open-telemetry/projects/88/views/1).

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
tammy-baylis-swi: Hi again @krnr , sorry for a bit of scope creep but I think it's important since this will be a breaking change (though a needed one!)

I cc'd you on a very similar [PR for the ASGI instrumentor](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2650#discussion_r2511304381). The change for span name cardinality+semconv will need to be behind an opt-in flag for users. Would it be reasonable to implement something similar here, or in a separate PR? cc @aabmass

krnr: (now it's my turn for being sorry to be late) well, as they say "better be safe than sorry". i agree with a flag idea, however, it's beyond my mind to find a good name for that. OTEL_PYTHON_AIOHTTP_SERVER_CANONICAL_PATH_SPAN_NAME ? too verbose?

tammy-baylis-swi: There is a shared util at [opentelemetry.instrumentation._semconv](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/bd3c1f2aa2e58490e45ba9c95e140112e9016c56/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py#L162C34-L185), which reads `OTEL_SEMCONV_STABILITY_OPT_IN` for us. This might require a middleware update to accept something like `sem_conv_opt_in_mode`, like what the [AsgiInstrumentor does](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/bd3c1f2aa2e58490e45ba9c95e140112e9016c56/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L588-L592).

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
```
emdneto: I think it should be in the fixed section.

emdneto (approved):

xrmx (approved):

xrmx (approved):

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
krnr: i apologize again for a late response @tammy-baylis-swi . today i checked the semconv module and it seems i misunderstood the concept a bit. the semconv is about the _names of attributes_ (of smth - span, metric, whatever). not about the _values_ of the attributes. and here we're talking about the _value_ of `span.name`, "should it be 'fooooooooooo' or 'foo...'" and if _that_ must be decided by a feature-flag, that is out of the scope of semconv feature flags in my opinion. 

moreover, introducing semconv leads to a lot of changes unrelated to the topic of this PR. i'm happy to continue working on opt-in, but in a separate branch and a separate PR. i've already made a draft commit ( https://github.com/open-telemetry/opentelemetry-python-contrib/commit/cc882b44f82b4827d6fc80b535144068f6a2d3c1 ) in a separate branch to illustrate how much is being changed and - more importantly - how many new discussion points emerge. which in turn will draw the attention from the span name even further.

tammy-baylis-swi: Hi @krnr no problem at all! I had some chat with others last week and this PR you've already done will be fine as it is. ✅  We've approved and we'll merge it for an upcoming release.

I'd say don't worry about the opt-in for span name, i.e. don't do [what I asked here](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3896#discussion_r2520052839). It is indeed confusing. The semconv does have Must/Should for [span name](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name). You're also right in that the Python implementation so far is focused on the names of attributes.

If you do want to put in a separate PR at some point for names of attributes, indeed it would be a lot of change. You're welcome to work on it if you're interested! Otherwise I'd say this much-needed fix is done.

tammy-baylis-swi (approved):

tammy-baylis-swi:
Hi @xrmx I've just resolved some conflicts. Any other suggestions?

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
```diff
@@ -252,14 +252,22 @@ def _parse_duration_attrs(
...
         request: the request object itself.
     Returns:
-        The span name.
+        The span name as "{method} {canonical_name}" of a resource if possible or just "{method}".
     """
-    span_name = request.path.strip() or f"HTTP {request.method}"
-    return span_name
+    try:
+        resource = request.match_info.route.resource
+        assert resource
```
xrmx: Do we really need to add this assert? If this is None resource.canonical will give an AttributeError as well.

xrmx (commented):
precommit is failing

PR #3905, opened 2025-10-29
Fix a few bugs in `opentelemetry-instrumentation-google-genai` instrumentation package

### LINKED ISSUE #3596: BaseModel.model_dump() called on class instead of instance in opentelemetry-instrumentation-google-genai
### Describe your environment

_No response_

### What happened?

When enabling the GoogleGenAiSdkInstrumentor from the opentelemetry-instrumentation-google-genai package (v0.2b0), an exception is raised during the execution of generate_content() using the google-genai SDK (v1.21.1) with a Pydantic response schema (pydantic v2.11.7).

The root cause appears to be that the instrumentation incorrectly calls .model_dump() on a Pydantic class rather than an instance, resulting in a TypeError.

### Steps to Reproduce

1.	Set up Gemini client using google-generativeai and enable OpenTelemetry instrumentation:
```
from opentelemetry.instrumentation.google_generativeai import GoogleGenerativeAiInstrumentor

GoogleGenerativeAiInstrumentor().instrument()
```
2. Use a structured response_schema with GenerateContentConfig, like a Pydantic model.
```
generation_config_with_thinking = GenerateContentConfig(
                    temperature=self.temperature,
                    response_mime_type="application/json",
                    response_schema=SomePydanticModel,
            
                )
```
3.	Execute an async generate_content() call via:

```
response = await self.client.aio.models.generate_content(
    model=self.model,
    contents=content,
    config=generation_config_with_thinking,
)
```
4. You’ll see the following error:
`TypeError: BaseModel.model_dump() missing 1 required positional argument: 'self'`

### Expected Result

The instrumentation should safely handle all supported response schemas without interfering with the normal execution of generate_content(). Specifically, it should validate and inspect Pydantic models without causing runtime exceptions, ensuring full compatibility with structured schema-based responses.

### Actual Result

When calling generate_content() with a Pydantic schema and OpenTelemetry instrumentation enabled, the application fails with the following error:
`TypeError: BaseModel.model_dump() missing 1 required positional argument: 'self'`

### Additional context

_No response_

None

DylanRussell:
# Description

- Fixes an issue where the instrumentation would crash when a `pydantic.BaseModel` class was passed as the response schema to `GenerateContentConfig`, the code expected a class instance for some reason (see #3596).
- Ensures the same Sem Conv attributes are added to the Log and the Span. The gen AI semantic conventions are basically the same for [log events](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-events.md) and [spans](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-spans.md). Adds the `error.type` sem conv attribute when there's an exception raised.
- Moves the `_maybe_log_completion_details(` function call into the `finally` block for the 4 instrumentation variants (sync/async streaming/non). This ensures we always write the event and call the completion hook etc. even when there's an exception thrown.
- Ensure the new `gen_ai.client.inference.operation.details` log event is written and completion upload hook are called when `OTEL_SEMCONV_STABILITY_OPT_IN=gen_ai_latest_experimental`.
- Some code cleanup / refactoring.

Fixes #3596

## Type of change

# How Has This Been Tested?

Unit tests

- [ ] Yes. - Link to PR: 
- [ x] No.

# Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -198,7 +198,6 @@ pythonVersion = "3.9"
 reportPrivateUsage = false  # Ignore private attributes added by instrumentation packages.
 # Add progressively instrumentation packages here.
 include = [
-  "instrumentation-genai/opentelemetry-instrumentation-google-genai",
   "instrumentation/opentelemetry-instrumentation-aiokafka",
   "instrumentation/opentelemetry-instrumentation-asyncclick",
   "instrumentation/opentelemetry-instrumentation-threading",


instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
aabmass: Can we just make this function typed since we know we're receiving `GenerateContentConfigOrDict`? It should be simple enough to make this code type check without and "prove" that model_dump() can be called.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -162,30 +162,24 @@ def _to_dict(value: object):
     if isinstance(value, dict):
         return value
     if hasattr(value, "model_dump"):
-        return value.model_dump()
+        try:
+            return value.model_dump()
+        except TypeError:
+            return {"ModelName": str(value)}
+
```
aabmass, line 169: Can we just make this function typed since we know we're receiving `GenerateContentConfigOrDict`? I think it's "provable" that we don't need this hasattr check and that the call is correct

One issue is I don't think this file is opted in to pyright 🙁  but doesn't change the implementation

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -212,37 +206,21 @@ def _add_request_options_to_span(
...
-    if config.seed:
-        attributes[gen_ai_attributes.GEN_AI_REQUEST_SEED] = config.seed
-    if config.candidate_count:
-        attributes[gen_ai_attributes.GEN_AI_REQUEST_CHOICE_COUNT] = (
-            config.candidate_count
-        )
-    if config.response_mime_type:
-        if config.response_mime_type == "text/plain":
+    response_mime_type = config.get("response_mime_type")
+    if response_mime_type and is_experimental_mode:
```
aabmass, line 210: Will this make the GEN_AI_OUTPUT_TYPE go away for people not using experimental semconv?

Looking at the test seems like none of the "old" semconv changes

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -212,37 +206,21 @@ def _add_request_options_to_span(
...
             attributes[gen_ai_attributes.GEN_AI_OUTPUT_TYPE] = "text"
-        elif config.response_mime_type == "application/json":
+        elif response_mime_type == "application/json":
             attributes[gen_ai_attributes.GEN_AI_OUTPUT_TYPE] = "json"
         else:
             attributes[gen_ai_attributes.GEN_AI_OUTPUT_TYPE] = (
-                config.response_mime_type
+                response_mime_type
             )
+    for key in list(attributes.keys()):
```
aabmass, line 219:
```suggestion
    for key in attributes:
```

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -162,30 +162,24 @@ def _to_dict(value: object):
...
+            return value.model_dump()
+        except TypeError:
+            return {"ModelName": str(value)}
+
     return json.loads(json.dumps(value))
 
 
-def _add_request_options_to_span(
-    span: Span,
+def create_request_attributes(
```
aabmass: can you leave it private?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -488,11 +440,15 @@ def _maybe_update_error_type(self, response: GenerateContentResponse):
 
     def _maybe_log_completion_details(
         self,
+        request_attributes: dict[str, Any],
+        final_attributes: dict[str, Any],
+        is_experimental_mode: bool,
```
aabmass: I think you can just use `self.sem_conv_opt_in_mode`

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -748,19 +704,32 @@ def instrumented_generate_content(
...
             model, "google.genai.Models.generate_content"
-        ):
-            helper.add_request_options_to_span(config)
+        ) as span:
+            span.set_attributes(request_attributes)
             if helper.sem_conv_opt_in_mode == _StabilityMode.DEFAULT:
                 helper.process_request(contents, config)
+                span.set_attribute(
+                    gen_ai_attributes.GEN_AI_SYSTEM, helper._genai_system
+                )
```
aabmass: I'm surprised helper doesn't set this itself, that would be simpler imo. Feel free to disregard not too familiar with this code

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/base.py
```diff
@@ -94,7 +94,10 @@ def configure_valid_response(self, **kwargs):
         response = create_response(**kwargs)
         self._responses.append(response)
 
-    def _create_and_install_mocks(self):
+    def configure_exception(self, e, **kwargs):
+        self._create_and_install_mocks(e)
+
+    def _create_and_install_mocks_with(self):
```
aabmass: doesn't look like this function is used

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/base.py
```diff
@@ -114,17 +126,23 @@ def _default_impl(*args, **kwargs):
             self._response_index += 1
             return result
 
-        mock.side_effect = _default_impl
+        if not e:
+            mock.side_effect = _default_impl
+        else:
+            mock.side_effect = e
```
aabmass: nit feel free to ignore
```suggestion
        mock.side_effect = e or _default_impl
```

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/nonstreaming_base.py
```diff
@@ -92,6 +99,41 @@ def test_generated_span_has_minimal_genai_attributes(self):
...
+            },
+        )
+        with patched_environ:
+            _OpenTelemetrySemanticConventionStability._initialized = False
+            _OpenTelemetrySemanticConventionStability._initialize()
+            try:
+                self.generate_content(
+                    model="gemini-2.0-flash", contents="Does this work?"
+                )
+            except ValueError:
```
aabmass: https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

instrumentation-genai/opentelemetry-instrumentation-google-genai/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Fixes bugs
```
aabmass: needs update

aabmass (approved):

instrumentation-genai/opentelemetry-instrumentation-google-genai/CHANGELOG.md
DylanRussell: done

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
DylanRussell: yep, done

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
DylanRussell: I don't think we know that -- `_to_dict` is called in a bunch of places in this file and it's not just that type being passed in.. Didn't know pyright was missing.. I opened up https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3933 to hopefully address that at some point. Opting in generated 1874 errors lol.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
DylanRussell: That causes a size changed during iteration error because I delete from the dict below

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
DylanRussell: In both this and the vertex AI instrumentation we put the request seed, output type and request choice count labels behind the experimental flag. But looking back these were added prior to 1.36 (in 1.30/31) so we should've added them to both.. But I didn't remove the label from the old instrumentation, it didn't exist there.. So probably we should backfill these labels to the old instrumentations (in a separate PR) ? What is the plan for all new semantic conventions going forward -- do we add them to just the "experimental" one ?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
DylanRussell: True, done

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/base.py
DylanRussell: ADded thanks

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/nonstreaming_base.py
DylanRussell: nice, done

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/base.py
DylanRussell: True.. Removed it

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
DylanRussell: Yeah good point. moved it into the helper

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
aabmass: > So probably we should backfill these labels to the old instrumentations (in a separate PR) ?

SGTM or at least open an issue to track.

> What is the plan for all new semantic conventions going forward -- do we add them to just the "experimental" one ?

Yes if the schema version is past 1.36 it should go in the new one

But can you confirm, does this PR remove anything from the "old" telemetry at all?

DylanRussell: Ok opened https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3934 for now.

No this doesn't remove anything from the old telemetry.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
aabmass: Ah thanks didn't realize that

DylanRussell:
Yes fixed now

PR #3929, opened 2025-11-09
Add exception recording to httpx instrumentation spans

axel584:
# Description

This PR enhances the httpx instrumentation by recording exceptions directly in spans using span.record_exception(exception).
Previously, only span.set_status(StatusCode.ERROR) was set, which made it harder to inspect the actual error details in traces.

Fixes # (issue)

## Type of change

- [X] No.


xrmx:
Are you sure the common code in the api isn't already taking care of adding it? If not, please add a test that is not passing with the current code.

PR #2942, opened 2024-10-30
Add PyMySQL instrumentor support for sqlcommenting

tammy-baylis-swi:
# Description

Updates PyMySQL instrumentor to support sqlcommenting.

Depends on # [2897](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2897)

Fixes # [2902](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2902)

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

Added unit tests. Used local installations of opentelemetry-instrumentation-dbapi (in #https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2897) and upstream dependencies with this updated downstream instrumentor on a Flask app that use PyMySQL to query MySQL with general logs enabled to check sqlcomments.

Example general log entry after this update, which now includes sqlcomment:

`2024-10-30T19:33:46.830661Z	   25 Query	SELECT * FROM city WHERE id = '1818' /*db_driver='pymysql%3A1.4.6',dbapi_level='2.0',dbapi_threadsafety=1,driver_paramstyle='pyformat',mysql_client_version='1.4.6',traceparent='00-f8688332bc5467bfa8fe1a9753d7271a-f7add29c81e79082-01'*/`

Example corresponding exported span, whose span ID matches sqlcomment traceparent span ID:

```
InstrumentationScope opentelemetry.instrumentation.pymysql 0.49b0.dev
Span #0
    Trace ID       : f8688332bc5467bfa8fe1a9753d7271a
    Parent ID      : c125e4709cfe1361
    ID             : f7add29c81e79082
    Name           : SELECT
    Kind           : Client
    Start time     : 2024-10-30 19:33:46.829773588 +0000 UTC
    End time       : 2024-10-30 19:33:46.833294379 +0000 UTC
    Status code    : Unset
    Status message :
Attributes:
     -> db.system: Str(mysql)
     -> db.name: Str(world)
     -> db.statement: Str(SELECT * FROM city WHERE id = %s)
     -> db.user: Str(world)
     -> net.peer.name: Str(mysql-world-db)
     -> net.peer.port: Int(3306)
```

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
[...]

PR #3931, opened 2025-11-11
fix(google-genai): prevent TypeError when handling Pydantic model class

### LINKED ISSUE #3596: BaseModel.model_dump() called on class instead of instance in opentelemetry-instrumentation-google-genai

arloc:
# Description

This PR fixes a bug in the `opentelemetry-instrumentation-google-genai` package where passing an uninstantiated Pydantic model class (instead of an instance) to dict conversion functions would cause a `TypeError`:

```
BaseModel.model_dump() missing 1 required positional argument: 'self'
```

The issue occurred in three functions that attempt to convert objects to dictionaries:
- `_to_dict()` in `generate_content.py`
- `_to_otel_value()` in `tool_call_wrapper.py`
- `_flatten_dict()` in `dict_util.py`

These functions check `hasattr(value, "model_dump")` which returns `True` for both Pydantic classes and instances, but `model_dump()` is an instance method that cannot be called on the class itself.

**Solution:**
- Added `isinstance(value, type)` check to distinguish between classes and instances before calling `model_dump()`
- When a Pydantic class is detected, the code now calls `model_json_schema()` (a classmethod) to get the schema representation
- Added comprehensive test coverage for both class and instance scenarios

Fixes #3596 

## Type of change

# How Has This Been Tested?

Added unit tests to verify the fix:

1. **test_flatten_with_pydantic_class_not_instance** in `tests/utils/test_dict_util.py`
   - Verifies that passing a Pydantic class (not instance) doesn't raise the `model_dump()` error
   - Confirms the class is properly serialized using `model_json_schema()`

2. **test_handles_pydantic_class_not_instance** in `tests/utils/test_tool_call_wrapper.py`
   - Tests the same scenario for the tool call wrapper functionality
   - Ensures proper handling of Pydantic model classes in tool call contexts

- [x] No.

# Checklist:

- [ ] Documentation has been updated (not needed - internal implementation fix)

### DIFF
--- a/instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/dict_util.py
@@ -188,9 +188,17 @@ def _flatten_compound_value(
             rename_keys=rename_keys,
             flatten_functions=flatten_functions,
         )
-    if hasattr(value, "model_dump"):
+    if hasattr(value, "model_dump") and not isinstance(value, type):
         return _flatten_dict(
             value.model_dump(),
+            key_prefix=key,
+            exclude_keys=exclude_keys,
+            rename_keys=rename_keys,
+            flatten_functions=flatten_functions,
+        )
+    if hasattr(value, "model_json_schema"):
+        return _flatten_dict(
+            value.model_json_schema(),
             key_prefix=key,
             exclude_keys=exclude_keys,
             rename_keys=rename_keys,
--- a/instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
@@ -161,8 +161,10 @@ def _determine_genai_system(models_object: Union[Models, AsyncModels]):
 def _to_dict(value: object):
     if isinstance(value, dict):
         return value
-    if hasattr(value, "model_dump"):
+    if hasattr(value, "model_dump") and not isinstance(value, type):
         return value.model_dump()
+    if hasattr(value, "model_json_schema"):
+        return value.model_json_schema()
     return json.loads(json.dumps(value))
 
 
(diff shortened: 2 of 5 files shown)


aabmass:
@DylanRussell this covers some of the same things as https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3905, but also includes the JSON schema. WDYT?

DylanRussell:
I just put `str(Class)` when I came across this. This also works, but it's a lot more information.. Do we want the whole json schema each time ?

aabmass:
We have a semconv PR for tool schemas https://github.com/open-telemetry/semantic-conventions/pull/2942. I think it's generally useful but maybe best to hold off until we have conventions. @arloc was capturing the schema in particular important to you or mostly just avoiding this exception?

aabmass:
We merged #2942, but I'll leave this open until we decide on including the json schema inline.

PR #3957, opened 2025-11-18
opentelemetry-instrumentation-aiohttp-client: update instrumentor to respect http instrumentation suppression

### LINKED ISSUE #3953: opentelemetry-instrumentation-aiohttp-client: instrumentor does not respect "suppress_http_instrumentation"
### Describe your environment

OS: Ubuntu
Python version: 3.13.3
Package version: 0.60b0.dev

### What happened?

The aiohttp client instrumentation library currently does not support suppressing instrumentation with `suppress_http_instrumentation()` which leads to unexpected behavior when using this functionality. For example, the following code produces a `Span` even though none should be created: 

```
async with aiohttp.ClientSession() as session:
    with suppress_http_instrumentation():
        async with session.get('http://localhost:8080/success') as response:
            _ = await response.text()
```
My guess is that this library simply never got updated when the `suppress_http_instrumentation()` API was introduced. The fix should be as trivial as replacing the call to `is_instrumentation_enabled()` to `is_http_instrumentation_enabled()` in the library [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-aiohttp-client/src/opentelemetry/instrumentation/aiohttp_client/__init__.py#L328)
```
async def on_request_start(
    unused_session: aiohttp.ClientSession,
    trace_config_ctx: types.SimpleNamespace,
    params: aiohttp.TraceRequestStartParams,
):
    if (
        not is_instrumentation_enabled()
        or trace_config_ctx.excluded_urls.url_disabled(str(params.url))
    ):
        trace_config_ctx.span = None
        return
    ...
```

### Expected Result

No `Span`s are created when using `suppress_http_instrumentation()`

### Actual Result

`Span`s are created when using `suppress_http_instrumentation()`

Yes

### Tip

herin049:
# Description

Fixes issue where the `opentelemetry-instrumentation-aiohttp-client` library does not respect the `suppress_http_instrumentation()` context manager.

Fixes #3953

## Type of change

# How Has This Been Tested?

Test cases have been updated to test suppression using `suppress_http_instrumentation()` in addition to `suppress_instrumentation()`

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3941](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3941))
 - `opentelemetry-instrumentation-pymongo`: Fix invalid mongodb collection attribute type
   ([#3942](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3942))
+- `opentelemetry-instrumentation-aiohttp-client`: Update instrumentor to respect suppressing http instrumentation
+  ([#3957](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3957))
 
 ## Version 1.38.0/0.59b0 (2025-10-16)
 


instrumentation/opentelemetry-instrumentation-aiohttp-client/src/opentelemetry/instrumentation/aiohttp_client/__init__.py
```diff
@@ -485,9 +485,6 @@ def _instrument(
 
     # pylint:disable=unused-argument
     def instrumented_init(wrapped, instance, args, kwargs):
-        if not is_instrumentation_enabled():
```
herin049, line 488: I opted to remove this check since I don't get its purpose. In addition, it could lead to unexpected behavior if the user creates a `ClientSession` object with instrumentation disabled and then subsequently re-enables instrumentation.

xrmx: I guess the purpose was an early exit to avoid doing some work like calling the trace_config callbacks.

herin049: That makes sense, although I still think this check probably doesn't belong here because of the reasoning above. In my view this function is a part of the core monkey patching logic, so as long as the library is instrumented it should be executed regardless. If the user needs more control they can always explicitly pass the Otel trace config into the client session for sessions they wish to selectively instrument. Let me know if you disagree or have a different opinion.

emdneto (approved):

PR #3961, opened 2025-11-20
feat: add extension point to add custom attributes to `generate_content` spans in google genai instrumentor.

RKest:
# Description

Introduces ability to add additional attributes to `generate_content {model.name}` spans to Google GenAI instrumentation.

Motivation behind this change is to add (non-semconv) `gcp.vertex.agent.event_id` attribute, to `generate_content {model.name}` spans, in order to correlate events with spans in [google/adk-python](https://github.com/google/adk-python).

Broader motivation is:
- Currently in ADK we trace inference in both `generate_content` and legacy `call_llm` spans.
- We want to migrate fully to `generate_content` spans and deprecate `call_llm` spans.
- However `call_llm` spans contain `gcp.vertex.agent.event_id` attribute, and we still want to correlate events with spans.

**Note:** Event in ADK is not synonymous with event in OTEL. In ADK, Event == LLM Request/Response or Tool Call/Result, in OTEL IIUC Event == Log.

## Type of change

# How Has This Been Tested?

Tested using:
- Automated unit tests.
- Using `opentelemetry-instrumentor-google-genai` off my fork in ADK, and successfully adding custom attributes to target span
image

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -80,6 +84,7 @@
 # Constant used for the value of 'gen_ai.operation.name".
 _GENERATE_CONTENT_OP_NAME = "generate_content"
 
+GENERATE_CONTENT_EXTRA_ATTRIBUTES = contextvars.ContextVar[Mapping[str, AttributeValue]]("GENERATE_CONTENT_EXTRA_ATTRIBUTES", default={})
```
aabmass: What do you think of using [OTel context api](https://opentelemetry-python.readthedocs.io/en/stable/api/context.html) instead of contextvars directly? It uses contextvars under the hood but it would then also work with code that does explicit context passing

```py
def foo(context: Context):
  with tracer.start_as_current_span("foo", context=context):
    ...
```

(Admittedly it's a little clunkier/indirect)

aabmass: I think contextvars are usually named with snake case since the values aren't constants. Also, consider exposing it in `__init__.py` explicitly to be clear it's in the public API.

aabmass (approved):
LGTM, it could probably be generalized more to other GenAI instrumentations if this proves useful. We can also use this to set the `gen_ai.conversation.id` from ADK 🙂

aabmass:
@RKest any blockers here?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
RKest: Good call, Done

PR #3964, opened 2025-11-21
opentelemetry-instrumentation-dbapi: instrument commit and rollback

bmwalters:
# Description

Adds instrumentation for database `commit()` and `rollback()` transaction operations across all DB-API instrumentations.

## Type of change

## Changes

- Added `commit()` and `rollback()` span instrumentation to `dbapi` module
- Added `enable_transaction_spans` configuration flag (default: `True`)
- Created `AsyncTracedConnectionProxy` for async connections (psycopg)
- Plumbed configuration through all dependent instrumentors: pymysql, mysql, mysqlclient, psycopg, psycopg2, sqlite3, pymssql

# How Has This Been Tested?

- Added unit tests for sync commit/rollback in `test_dbapi_integration.py`
- Added unit tests for sync and async commit/rollback in `test_psycopg_integration.py`
- Added functional tests for pymysql in `test_pymysql_functional.py`
- [x] All existing tests pass

- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-instrumentation-psycopg2`: Add parameter `capture_parameters` to instrumentor.
   ([#4212](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4212))
 
+- `opentelemetry-instrumentation-dbapi`: Add instrumentation for `commit()` and `rollback()` transaction operations
+  ([#3964](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3964))
+- `opentelemetry-instrumentation-dbapi`: Add `enable_transaction_spans` configuration flag to control transaction span creation (default: `True`)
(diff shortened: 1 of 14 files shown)


bmwalters:
Not stale.

tammy-baylis-swi:
Thanks for this feature! I'm ok with `enable_transaction_spans` being True by default, though would be good to get others' thoughts.

Could you please add to OP an example of an exported trace with the commit or rollback span included? e.g. using otel collector debug exporter.

What happens when user also sets `enable_commenter=True`? And `enable_attribute_commenter=True` for more visibility. Is sqlcommenting attempted for commit/rollback spans, or are they excluded. Would be helpful if unit test coverage was added for this scenario.

bmwalters:
> Could you please add to OP an example of an exported trace with the commit or rollback span included? e.g. using otel collector debug exporter.

Done in follow-up PR: #4519 

> What happens when user also sets `enable_commenter=True`?

Comments aren't added. `commit()` constructs the query in the driver (which could be C code). If we wanted that behavior, we'd have to reimplement commit as a hardcoded `COMMIT` query.

Issue #3962, opened 2025-11-21
Add http.request.header.<key> attributes to HTTP client spans


herin049:
### What problem do you want to solve?

The OpenTelemetry semantic conventions list `http.request.header.<key>` as a stable `Opt-In` attribute for HTTP `Span`s [[link](https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/#http-response-header)] In particular, the semantic conventions mention that 

> `http.request.header.<key>`: Instrumentations SHOULD require an explicit configuration of which headers are to be captured. Including all request headers can be a security risk - explicit configuration helps avoid leaking sensitive information.

### Describe the solution you'd like

Similar to HTTP server instrumentation, we can implement these changes by introducing two new environment variables: 

- `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_CLIENT_REQUEST`
- `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_CLIENT_RESPONSE`

These are analogous to the currently existing `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST` and `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_RESPONSE` environment variables. We can reuse the existing `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SANITIZE_FIELDS` environment variable for determining which headers should be sanitized, although it might also be valid to consider introducing client and server specific header sanitation environment variables that would apply individually to client and server spans. 

### Would you like to implement a fix?

Yes

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

xrmx:
I think we did all the http client instrumentations, closing.

PR #3969, opened 2025-11-25
opentelemetry-util-http: add http client header attribute env vars

herin049:
# Description

This PR adds constants for two new environment variables: 

- `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_CLIENT_REQUEST`
- `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_CLIENT_RESPONSE`

to the `opentelemetry-util-http` library. These changes are required in order to add support for capturing certain HTTP headers for HTTP client requests & responses as described in #3962. 

## Type of change

# How Has This Been Tested?

N/A

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `opentelemetry-instrumentation-django`: improve docs for response_hook with examples of providing attributes from middlewares
   ([#3923](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3923))
 - Update for Log SDK breaking changes. Rename InMemoryLogExporter to InMemoryLogRecordExporter in several tests
-  ([#3850](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3589))
+  ([#3589](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3589))
+- opentelemetry-instrumentation: allow to skip all instrumentations loading with a wildcard
+  ([#3967](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3967))
 - `opentelemetry-util-http`: add http client header attribute env vars
   ([#3969](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3969))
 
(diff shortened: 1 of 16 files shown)


tammy-baylis-swi:
I think this is fair given the existing `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_*` as you mentioned in the issue, and it's confined to the contrib/instrumentation.

I appreciate a small PR but it would be good to see how these will be supported with security mindfulness. Wdyt?

herin049:
@tammy-baylis-swi I created a separate PR since I'm assuming the team wants to stick with one changelog entry per PR, but I'm happy to group the implementation for client libraries with the PR to add these constants to the HTTP utility library. 

With regards to security mindfulness, the implementations for the client libraries will by default capture no headers at all, unless the `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_CLIENT_*` environment variables are set. So the default behavior is not changing and is Opt-in only. This is analogous to how this behavior is handled for server headers as well. 

As far as the implementation goes, it will utilize the `get_custom_headers()` and `sanitize_header_values()` in the `SanitizeValue` class. For reference you can look at the `asgi` instrumentation library, in particular [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L426)

tammy-baylis-swi:
> As far as the implementation goes, it will utilize the `get_custom_headers()` and `sanitize_header_values()` in the `SanitizeValue` class. For reference you can look at the `asgi` instrumentation library, in particular [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L426)

That answers my question 👍  Sounds like a plan!

xrmx:
> @tammy-baylis-swi I created a separate PR since I'm assuming the team wants to stick with one changelog entry per PR, but I'm happy to group the implementation for client libraries with the PR to add these constants to the HTTP utility library.

I would have added the constants with the first instrumentation PR so we can see how this is used and if there would be a followup PR at all. You then won't need a changelog entry just for adding a couple of headers but only for the first instrumentation implementation.

PR #3974, opened 2025-11-26
Fix RecursionError by preventing duplicate instrumentation in urllib3

shrxya07:
### Summary
This PR fixes RecursionError reported in issue [#4532](https://github.com/open-telemetry/opentelemetry-python/issues/4532).

### Fix
A `_instrumented` guard flag was added to urllib3 instrumentation to ensure
that `_instrument()` cannot run multiple times. The absence of this guard
caused recursive instrumentation and eventually RecursionError.

### Details
The change:
- Adds `_instrumented = False` at the class level
- Early-returns from `_instrument()` if instrumentation already happened

### Result
This prevents infinite recursion and ensures stable instrumentation behavior.

Fixes: [#4532](https://github.com/open-telemetry/opentelemetry-python/issues/4532)


xrmx:
@shrxya07 thanks for the PR, how do you reproduce it? Is it manual calling the instrumentor more than one time or in a different way? Please add a test, an entry in changelog and setup pre-commit so CI will be happy

dust63:
Can I help to finalize this merge request ?

PR #3986, opened 2025-11-30
Add retry/backoff support to Prometheus Remote Write exporter

### LINKED ISSUE #3985: Add configurable retries with backoff to Prometheus Remote Write exporter

Remote Write pushes are currently fail-fast: transient 429/5xx/network blips drop metric batches without retry or backpressure. That causes silent data loss during throttling or short outages.

Add a configurable retry policy to the exporter: max attempts, exponential backoff with jitter, and a retryable status list (e.g., 429/408/5xx plus connection/timeouts). Log retry decisions and final failures. Defaults can be modest (e.g., 3 retries, small backoff) with max_retries=0 to disable.

Accept drops on error (current behavior): loses data during transient issues.
Rely on external queues/proxies: adds operational overhead and still needs exporter-side retries.
Push retry logic to remote endpoints: not always possible; exporters should be polite and back off themselves.

### Additional Context

_No response_

Yes

### Tip

towseef41:
# Description
Add configurable retries with exponential backoff/jitter to the Prometheus Remote Write exporter so transient 429/408/5xx and connection/timeouts don’t drop metrics silently. Updated README with the new retry knobs.

Fixes #3985

## Type of change

# How Has This Been Tested?
- [x] tox -e py311-test-exporter-prometheus-remote-write
- [x] tox -e lint-exporter-prometheus-remote-write

- [ ] Yes. - Link to PR:
- [x] No.

# Checklist:


herin049:
Can we not just use the builtin `urllib3` retry functionality (see https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry and https://requests.readthedocs.io/en/latest/user/advanced/#example-automatic-retries)

I find this approach preferable to rolling our own backoff-retry loop.

towseef41:
> Can we not just use the builtin `urllib3` retry functionality (see https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry and https://requests.readthedocs.io/en/latest/user/advanced/#example-automatic-retries)
> 
> I find this approach preferable to rolling our own backoff-retry loop.

Thanks for the suggestion. I’m good to move to a requests.Session + HTTPAdapter using urllib3.Retry (POST allowed), mapping our existing knobs, and I’ll add a tiny Retry subclass only to keep jitter/backoff cap. I’ll drop the manual loop and update tests.

For context, I initially considered a custom loop to keep full control over jitter/backoff cap and explicit logging and avoid relying on adapter/session setup, but I agree urllib3.Retry is battle-tested and clearer.

herin049:
> > Can we not just use the builtin `urllib3` retry functionality (see https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry and https://requests.readthedocs.io/en/latest/user/advanced/#example-automatic-retries)
> > I find this approach preferable to rolling our own backoff-retry loop.
> 
> Thanks for the suggestion. I’m good to move to a requests.Session + HTTPAdapter using urllib3.Retry (POST allowed), mapping our existing knobs, and I’ll add a tiny Retry subclass only to keep jitter/backoff cap. I’ll drop the manual loop and update tests.
> 
> For context, I initially considered a custom loop to keep full control over jitter/backoff cap and explicit logging and avoid relying on adapter/session setup, but I agree urllib3.Retry is battle-tested and clearer.

Sounds good, we can get the opinion of other members as I might be in the minority here. 

On a related note, I'm not sure if sub-classing `urllib3.Retry` is necessary if you just want to bound the backoff delay, you can simply set `backoff_jitter` and `backoff_max` to something reasonable.

towseef41:
> > > Can we not just use the builtin `urllib3` retry functionality (see https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry and https://requests.readthedocs.io/en/latest/user/advanced/#example-automatic-retries)
> > > I find this approach preferable to rolling our own backoff-retry loop.
> > 
> > 
> > Thanks for the suggestion. I’m good to move to a requests.Session + HTTPAdapter using urllib3.Retry (POST allowed), mapping our existing knobs, and I’ll add a tiny Retry subclass only to keep jitter/backoff cap. I’ll drop the manual loop and update tests.
> > For context, I initially considered a custom loop to keep full control over jitter/backoff cap and explicit logging and avoid relying on adapter/session setup, but I agree urllib3.Retry is battle-tested and clearer.
> 
> Sounds good, we can get the opinion of other members as I might be in the minority here.
> 
> On a related note, I'm not sure if sub-classing `urllib3.Retry` is necessary if you just want to bound the backoff delay, you can simply set `backoff_jitter` and `backoff_max` to something reasonable.

Makes sense. I’ll switch to urllib3.Retry and avoid subclassing if possible: use a requests.Session + HTTPAdapter with Retry(total=..., backoff_factor=..., backoff_max=..., status_forcelist=..., allowed_methods={"POST"}). The requests-bundled urllib3 we have doesn’t expose backoff_jitter, so if we want jitter I’ll add the smallest possible override; otherwise I’ll stick to base Retry with a sensible backoff_max.

xrmx:
The OTLP exports implement retries manually but AFAICS don't expose any tunable (e.g. `_export` in `exporter/opentelemetry-exporter-otlp-proto-grpc/src/opentelemetry/exporter/otlp/proto/grpc/exporter.py` from opentelemetry-python repo). If you don't expose this mechanism I think it's fine to reuse http libraries code for that.

towseef41:
> The OTLP exports implement retries manually but AFAICS don't expose any tunable (e.g. `_export` in `exporter/opentelemetry-exporter-otlp-proto-grpc/src/opentelemetry/exporter/otlp/proto/grpc/exporter.py` from opentelemetry-python repo). If you don't expose this mechanism I think it's fine to reuse http libraries code for that.

@xrmx 
Thanks for the pointers. I’m planning to reuse the HTTP stack’s retries instead of a custom loop: `requests.Session` + `HTTPAdapter` with `urllib3.Retry` (POST allowed, small status_forcelist). I kept a few knobs exposed (max retries, backoff factor/cap, status list) so users can tune if needed, but otherwise it follows the built-in behavior. If you’d rather keep it non-tunable (closer to OTLP) and just lean on `Retry` defaults, I can pare that back.

exporter/opentelemetry-exporter-prometheus-remote-write/src/opentelemetry/exporter/prometheus_remote_write/__init__.py
```diff
@@ -55,6 +58,27 @@
...
+        self.backoff_max = backoff_max
+        self.jitter_ratio = jitter_ratio
+
+    def get_backoff_time(self) -> float:
+        backoff = super().get_backoff_time()
+        backoff = min(backoff, self.backoff_max)
+        if self.jitter_ratio:
+            jitter = backoff * self.jitter_ratio
+            backoff += random.uniform(-jitter, jitter)
+        return max(backoff, 0.0)
```
herin049:
```suggestion
        if self.jitter_ratio:
            backoff += backoff * self.jitter_ratio * random.uniform(-1, 1)
        return min(max(backoff, 0.0), self.backoff_max)
```

herin049 (changes requested):

herin049 (approved):

xrmx:
I think this is missing a changelog entry

xrmx:
@towseef41 Please fix precommit

PR #3994, opened 2025-12-03
opentelemetry-util-genai: add support for emitting inference events and enrich message types

### LINKED ISSUE #3065: Library for uploading blobs as part of instrumentation #GenAI #MultiModal

Cirilla-zmh:
# Description

- Add support for emitting inference events.
- Add `gen_ai.system_instructions` attributes for inference span or events.
- Refactor `span_utils.py` for more brief implementation.
- Add `Reasoning`, `Blob`, `File`, `Uri` message types.

Fixes #3065

## Type of change

# How Has This Been Tested?

- [X] Add unit tests

- [x] No.

# Checklist:


Cirilla-zmh:
Typecheck failure is not related to this PR, so it may be ignored. cc @aabmass 

```plaintext
typecheck: commands[0]> pyright
/home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
  /home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py:175:17 - error: No overloads for "join" match the provided arguments (reportCallIssue)
  /home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py:175:27 - error: Argument of type "Generator[str | Unknown | bytes | Any, None, None]" cannot be assigned to parameter "iterable" of type "Iterable[str]" in function "join"
    "Generator[str | Unknown | bytes | Any, None, None]" is not assignable to "Iterable[str]"
      Type parameter "_T_co@Iterable" is covariant, but "str | Unknown | bytes | Any" is not a subtype of "str"
        Type "str | Unknown | bytes | Any" is not assignable to type "str"
          "bytes" is not assignable to "str" (reportArgumentType)
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -41,63 +45,162 @@
...
+    logger: Logger | None,
+    span: Span,
+    invocation: LLMInvocation,
+    error: Error | None = None,
+) -> None:
+    """Emit a gen_ai.client.inference.operation.details event to the logger.
+
+    This function creates a LogRecord event following the semantic convention
+    for gen_ai.client.inference.operation.details as specified in the GenAI
+    event semantic conventions.
```
DylanRussell, line 160: nit: can we link to the sem conv that defines this event

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -41,63 +45,162 @@
...
+    """
+    if not is_experimental_mode() or get_content_capturing_mode() not in (
+        ContentCapturingMode.EVENT_ONLY,
+        ContentCapturingMode.SPAN_AND_EVENT,
+    ):
+        return
```
DylanRussell: this isn't right, this flag is for putting content attributes on the event, it shouldn't block the event being written altogether

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -41,63 +45,162 @@
...
+
+    # Build event attributes by reusing the attribute getter functions
+    attributes: dict[str, Any] = {}
+    attributes.update(_get_llm_common_attributes(invocation))
+    attributes.update(_get_llm_request_attributes(invocation))
+    attributes.update(_get_llm_response_attributes(invocation))
+    attributes.update(
+        _get_llm_messages_attributes_for_event(
+            invocation.input_messages,
```
DylanRussell, line 175: this could be one line I think ?

`attributes.update(get_dict | get_dict | get_dict...)`

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -53,18 +53,53 @@ class ToolCallResponse:
...
-
-
 @dataclass()
 class Text:
     content: str
     type: Literal["text"] = "text"
 
 
-MessagePart = Union[Text, ToolCall, ToolCallResponse, Any]
+@dataclass()
```
DylanRussell, line 83: is it possible to link to the sem convs where  these types are defined ?

util/opentelemetry-util-genai/tests/test_utils.py
```diff
@@ -28,6 +28,23 @@
...
+    )
+except ImportError:
+    # Fallback to old name for compatibility with older SDK versions
+    from opentelemetry.sdk._logs.export import (
+        InMemoryLogExporter as InMemoryLogRecordExporter,
+    )
+    from opentelemetry.sdk._logs.export import (
+        SimpleLogRecordProcessor,
+    )
+from opentelemetry.sdk._logs import LoggerProvider
```
DylanRussell: should we bother with this or instead just update pyproject to require the latest version ?

JWinermaSplunk:
I think changing some of the apply attributes methods to output a dictionary instead of attaching the attributes for events sake is ok, though making the methods llm specific will lead to a lot of overlapping code overtime if more genai types are implemented in the future. More so applies to methods like _apply_common_span_attributes, _apply_finish_attributes, etc, which could take different genai type invocations.

Cirilla-zmh:
> More so applies to methods like _apply_common_span_attributes, _apply_finish_attributes, etc, which could take different genai type invocations.

@JWinermaSplunk Thanks for your suggestions! We do need to classify the attributes for common, inference, agent and so on to reuse them as much as possible. However, I don't want to involve many features in this PR because that means we'd better refactor the design of `LLMInovcation` to a layered model with `CommonInvocation`, `MessagesInocation` and so on.

Could you please create another issue to mark this proposal? I believe @keith-decker and @aabmass would also be interested in this.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
Cirilla-zmh: @DylanRussell 

You have a point, we might need an extra config flag to control inference event emission. But I have to ask: if the inference events don't log the chat history and just keep other attributes, how valuable are they really, since we already have inference spans?

IMO, recording the chat history is one of the main purposes of inference events.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
Cirilla-zmh: Please check https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/non-normative/models.ipynb

Do you think if I should add links in the comments?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
DylanRussell: today in Google AI instrumentations we always write the events.. They still have some value I think, that's why the flag was introduced not to just enable/disable the entire event, but just the sensitive content on the event..

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
DylanRussell: IMO yes, it'd be nice if we could link to an actual definition for each type. If they are just defined in that notebook I suppose 1 link is fine too

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
Cirilla-zmh: I'd still prefer using the `update()` API, for two reasons.

- Firstly, given the length of the methods, readability doesn't get much better with the `|` operator.
- Secondly—and this is a bit of a selfish reason—some of our customers are still on Python 3.8, and we want to ensure fewer conflicts for our `genai-util` extensions.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
Cirilla-zmh: Added a config named `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT` to control if instrumentations should emit events. False for default.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
Cirilla-zmh: @DylanRussell Hey! I just set this question resolved. If you have further concern, feel free to reopen it. :)

keith-decker:
This PR looks good to me. 

I agree with @JWinermaSplunk about the duplication of attributes. Once we add the next type (Tool, Agent, etc.) it'll be a great time to refactor those invocations to be more generic.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
lmolkova: Discussion from GenAI SIG on Jan 6th and Jan 12th
- `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT` is false by default
- if `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT: event` or `span_and_event`, then record events and content on events
- if `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT` is true, but `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT = span`, then emit events without content
- if `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT` is explicitly false (not default), probably don't emit events at all

----
then users can opt-into this with just one env var 
`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT = event | span_and_event`

Cirilla-zmh: @lmolkova @DylanRussell @keith-decker 
By default, message content will not be captured.
Set the environment variable `OTEL_SEMCONV_STABILITY_OPT_IN` to `gen_ai_latest_experimental` to enable experimental features.
Set the environment variable `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` to one of:
- `NO_CONTENT`: Do not capture message content (default).
- `SPAN_ONLY`: Capture message content in spans only.
- `EVENT_ONLY`: Capture message content in events only.
- `SPAN_AND_EVENT`: Capture message content in both spans and events.

To control event emission, you can optionally set `OTEL_INSTRUMENTATION_GENAI_EMIT_EVENT` to `true` or `false` (case-insensitive).
This variable controls whether to emit `gen_ai.client.inference.operation.details` events.
If not explicitly set, the default value is automatically determined by `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`:
- When `NO_CONTENT` or `SPAN_ONLY` is set: defaults to `false`
- When `EVENT_ONLY` or `SPAN_AND_EVENT` is set: defaults to `true`
If explicitly set, the user's value takes precedence over the default.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -27,87 +29,183 @@
...
-    Returns (genai_attributes) for use with metrics.
+    Returns a dictionary of attributes.
     """
-    span.update_name(
-        f"{GenAI.GenAiOperationNameValues.CHAT.value} {invocation.request_model}".strip()
-    )
-    span.set_attribute(
-        GenAI.GEN_AI_OPERATION_NAME, GenAI.GenAiOperationNameValues.CHAT.value
-    )
+    attributes: dict[str, Any] = {}
```
DylanRussell: nit: this could be a for loop or even a dictionary comprehension

DylanRussell (approved):

DylanRussell:
LGTM. Thanks

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
Cirilla-zmh: Now it's in the dictionary comprehension pattern.

Cirilla-zmh:
@xrmx Hello, could you please help merge this PR? I believe it's all set.

Cirilla-zmh:
@aabmass @lzchen @xrmx 

Hello, I believe this PR has been thoroughly reviewed by the relevant parties. How should I proceed with merging it? Also, is there anything else I need to do beforehand?

We have some further changes we would like to submit, which depends this PR.

aabmass:
Hey @Cirilla-zmh sorry for the delay. If you fix the conflicts, I've marked it auto merge.

Cirilla-zmh:
> Hey @Cirilla-zmh sorry for the delay. If you fix the conflicts, I've marked it auto merge.

Hey @aabmass , I have fixed the conflicts. But there're several tests are failed, which is not related to this PR. Could you please ignore these failing checks? Or should we fix this issue first in #4180? cc @xrmx

PR #4043, opened 2025-12-19
Fix broken database semantic convention links

sam-mosleh:
Fixed some broken links


JWinermaSplunk (approved):
Probably doesn't need a CHANGELOG entry.

PR #4069, opened 2025-12-30
GenAI utils: Add server address, port and additional metric attributes to LLM invocation

lmolkova:
Adding `server.address` and `server.port` ([recommended attributes](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-spans.md#inference)) and ability to provide additional attributes for metrics (needed for [OpenAI](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/openai.md#metrics))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3980](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3980))
 - `opentelemetry-instrumentation`: add database stability attribute setters in `_semconv` utilities
   ([#4108](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4108))
+- `opentelemetry-instrumentation-tornado`: Implement new semantic convention opt-in migration
+  ([#3993](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3993))
+- `opentelemetry-instrumentation-pyramid` Implement new semantic convention opt-in migration
+  ([#3982](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3982))
 
 ### Fixed
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -122,6 +122,12 @@ Some packages may require additional system-wide dependencies to be installed. F
 
 For `docs` building, you may need to install `mysql-client` and other required dependencies as necessary. Ensure the Python version used in your local setup matches the version used in the [CI](./.github/workflows/) to maintain compatibility when building the documentation.
 
+If you are using `tox-uv` for tests and have issues with resolving OpenTelemetry dependencies try:
+
+```sh
+uv sync --refresh
+```
+
 ### Benchmarks
 
 Some packages have benchmark tests. To run them, run `tox -f benchmark`. Benchmark tests use `pytest-benchmark` and they output a table with results to the console.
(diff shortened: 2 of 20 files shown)


util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -26,6 +31,15 @@ def __init__(self, meter: Meter):
         self._duration_histogram: Histogram = create_duration_histogram(meter)
         self._token_histogram: Histogram = create_token_histogram(meter)
 
+    @staticmethod
+    def _calculate_duration_seconds(invocation: LLMInvocation) -> Optional[float]:
```
lmolkova: extracted to separate method to fix  too-many-branches lint

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -64,23 +78,21 @@ def record(
...
+            attributes.update(invocation.metric_attributes)
 
         span_context = set_span_in_context(span)
         if error_type:
-            attributes["error.type"] = error_type
+            attributes[ErrorAttributes.ERROR_TYPE] = error_type
 
-        if (
-            duration_seconds is not None
-            and isinstance(duration_seconds, Number)
```
lmolkova, line 82: it can't really be anything else (if not None)
and there is no reason not to record 0 duration - why would we drop points for it? Assuming it can happen, we'd be skewing counts and everything else

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
xrmx: Don't stress much about pylint warnings, most of them are really arbitrary

util/opentelemetry-util-genai/src/opentelemetry/util/genai/metrics.py
```diff
@@ -3,13 +3,18 @@
...
 import timeit
-from numbers import Number
 from typing import Dict, Optional
 
 from opentelemetry.metrics import Histogram, Meter
 from opentelemetry.semconv._incubating.attributes import (
     gen_ai_attributes as GenAI,
 )
+from opentelemetry.semconv.attributes import (
+    error_attributes as ErrorAttributes,
```
xrmx: WDYT of keeping the original module name instead? `ErrorAttributes` looks more like a class than a module

JWinermaSplunk (approved):
Looks good to me! I will admit the first docstring in types.py is a little confusing i.e. saying `metric_attributes` won't be set on metrics, though I understand it. Also, do we not intend to set `server_address` or `server_port` for spans or events based on the second docstring? 

Edit: SIG, yes, no address or port for spans, only metrics

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -116,13 +116,26 @@ class LLMInvocation:
     input_tokens: int | None = None
     output_tokens: int | None = None
     attributes: dict[str, Any] = field(default_factory=_new_str_any_dict)
+    """
```
wrisa, line 119: Is this comment for below attributes and the below comment for `metric_attributes ` ?

lmolkova: yeah, it's a common practice in python to write docstrings under the property.

It probably would be less confusing if all properties had it, and hope we will eventually get there :)

util/opentelemetry-util-genai/tests/test_utils.py
```diff
@@ -106,7 +107,13 @@ def _assert_span_attributes(
     span_attrs: Mapping[str, Any], expected_values: Mapping[str, Any]
 ) -> None:
     for key, value in expected_values.items():
-        assert span_attrs.get(key) == value
+        if value == "*":
```
xrmx: Not sure it's better but you can avoid adding this special case by using something like the following as expected value that is similar to mock.ANY in spirit:

```
class MockNotNone:
    def __eq__(self, other):
        return other is not None

NOT_NONE = MockNotNone()
```

lmolkova: cool, let me try this

PR #4082, opened 2026-01-05
update opentelemetry-instrumentation to wrapt v2

MeggyCal:
# Description

Wrapt 2 introduced some changes in object wrappers (this [blog](https://grahamdumpleton.me/posts/2025/10/detecting-object-wrappers/) summarizes it nicely).

This change at least allows tests to pass with wrapt 2.

## Type of change

# How Has This Been Tested?

I ran the testsuite of opentelemetry-instrumentation.

- [x] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#4006](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4006))
 - `opentelemetry-instrumentation-flask`: Add support for 3.1+ streaming responses
   ([#3938](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3938))
+- `opentelemetry-instrumentation`: Add support for wrapt 2
+  ([#4082](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4082))
 - `opentelemetry-instrumentation-aiohttp-server`: Support passing `TracerProvider` when instrumenting.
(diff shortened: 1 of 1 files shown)


JWinermaSplunk:
I am unsure about the necessity of this PR unless there are plans to update the version of wrapt, do we have any plans to use the new functionality from the major version? In turn, I am not sure allowing all versions of wrapt, i.e. `"wrapt >= 1.0.0"` would be the best idea. I think updating the major version of the dependency would classify the PR as a Breaking Change and not a New Feature. The PR also may require a changelog entry.

xrmx:
I'm not worried on permitting a newer wrapt version but I'm worried that:
- we are not testing with a newer wrapt
- we have instrumentations that make use of the object proxy that we have not changed to allow newer wrapt and we should to them at the same time to avoid incompatibilities between different packages

blarghmatey:
Having Wrapt hard locked to <2 is a blocking bug for me to be able to use OpenTelemetry at all in one of my critical applications. That application is using Wrapt 2 and that means I can't even install this package to be able to instrument the application. The application in question is https://github.com/openedx/openedx-platform

MeggyCal:
Hi, I agree it would be nice to test this PR with wrapt 2 in several usecases. Sadly, I have almost no experience with automatic testing on GitHub, could someone please do it for me?

This PR still preserves compatibility with wrapt 1, so I don't see why it would be breaking, unless there are more problems with wrapt 2. I agree changelog entry would be nice to have, I will try to add it.

xrmx:
Closing as superseded by https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4203

PR #4090, opened 2026-01-07
feat: add `commenter_for_all_spans` option to enable SQLCommenter for non-recording spans

Elyahou:
# Description

Add `commenter_for_all_spans` option to SQLAlchemy and DBAPI instrumentations, enabling SQLCommenter to add SQL comments (including `traceparent`) for not sampled spans.

**Motivation:** By default, SQLCommenter only adds comments when `span.is_recording()` is `True`. This means not sampled spans (e.g., when using a sampling rate < 100%) don't get SQL comments. However, some users want `traceparent` propagated in SQL comments for all requests regardless of sampling decision, enabling database-side correlation with distributed traces.

**Changes:**
- Added `commenter_for_all_spans` parameter (default: `False`) to both SQLAlchemy and DBAPI instrumentations
- When enabled, SQL comments are added even for non-recording but valid spans
- Span attributes are still only set when `span.is_recording()` is `True` (existing behavior preserved)

## Type of change

# How Has This Been Tested?

Added unit tests for both instrumentations:


instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -136,6 +136,31 @@
 | ``opentelemetry_values``  | OpenTelemetry context as traceparent at time of query.    | ``traceparent='00-03afa25236b8cd948fa853d67038ac79-405ff022e8247c46-01'`` |
 +---------------------------+-----------------------------------------------------------+---------------------------------------------------------------------------+
 
+SQLComment for non-recording spans
+**********************************
+By default, sqlcommenter only adds comments to recording (sampled) spans.
```
tammy-baylis-swi: Hi again @Elyahou , thanks for this!

The overall approach to changing the instrumentors looks reasonable. The application of concepts needs a bit of fine-tuning though and apologies if my Slack message was confusing. 

"Sampled" is whether span data is exported or not, i.e. the `01` vs `00` traceflags. iiuc we want something like `commenter_for_nonsampled_spans`.

"Recording" is whether span data is kept through instrumentation and propagation and corresponding metrics are recorded, whether sampled or not. If not recorded, telemetry is dropped completely.

So I think if `commenter_for_nonsampled_spans` is True, there would be sqlcomment appended for both `01` and `00`. But no sqlcomment appended if not recording because there won't be a valid trace id anyway.

Elyahou: Hi! Thanks for the feedback. I investigated this thoroughly and want to clarify a few things:

**Non-recording spans DO have valid trace IDs**

I tested this directly:

```python
from opentelemetry.sdk.trace.sampling import ALWAYS_OFF

provider = TracerProvider(sampler=ALWAYS_OFF)
span = provider.get_tracer(__name__).start_span('test')

print(f'is_recording: {span.is_recording()}')
print(f'trace_id: {span.get_span_context().trace_id:032x}')
print(f'span_id: {span.get_span_context().span_id:016x}')
print(f'is_valid: {span.get_span_context().is_valid}')
```

Output:
| Property | Value | Status |
|----------|-------|--------|
| is_recording | False | |
| trace_id | 173482935693c85a55f29d2678655838 | ✅ Valid |
| span_id | ecd82b622a62fbdb | ✅ Valid |
| is_valid | True | ✅ |

Non-recording spans maintain valid trace context for propagation, even though they're not exported.

**The original code used `is_recording()`, not `trace_flags.sampled`**

The existing logic before my change:

```python
if span.is_recording():
    if self.enable_commenter:
        # add SQL comment
```

There was no logic checking `trace_flags.sampled`.

With standard samplers like `TraceIdRatioBasedSampler`:
- Sampled → `RECORD_AND_SAMPLE` → `is_recording()=True`
- Unsampled → `DROP` → `is_recording()=False`

**Naming proposition**

I propose renaming `commenter_for_nonrecording_spans` → `commenter_for_all_spans`

This makes it clearer that it enables SQL comments regardless of the sampling decision. What do you think?

tammy-baylis-swi: Thanks for that! I agree with proposal to rename the flag as `commenter_for_all_spans`.

Regarding your test: yes, the standard samplers AlwaysOff and TraceIdRatioBased do propagate valid trace IDs even if decision is `DROP` via [NonRecordingSpan](https://github.com/open-telemetry/opentelemetry-python/blob/784442fe1a58c199e7ef572d5e1c1c6555a9e075/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L1201).

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -726,24 +765,34 @@ def traced_execution(
...
-                        # so db.statement is set before add_sql_comment
+                        # sqlcomment only in executed query, not in db.statement
                         self._populate_span(span, cursor, *args)
                         args = self._update_args_with_added_sql_comment(
                             args, cursor
                         )
                 else:
-                    # no sqlcomment anywhere
-                    self._populate_span(span, cursor, *args)
+                    # Non-recording span - just add comment for context propagation
```
tammy-baylis-swi: If span is non-recording, the span could potentially have [invalid span context](https://github.com/open-telemetry/opentelemetry-python/blob/784442fe1a58c199e7ef572d5e1c1c6555a9e075/opentelemetry-api/src/opentelemetry/trace/span.py#L577-L586). I think it would be good to add a check like the [TraceContextTextMapPropagator](https://github.com/open-telemetry/opentelemetry-python/blob/784442fe1a58c199e7ef572d5e1c1c6555a9e075/opentelemetry-api/src/opentelemetry/trace/propagation/tracecontext.py#L101-L102) does for HTTP propagation. Wdyt?

Elyahou: Good catch, added

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -726,24 +765,39 @@ def traced_execution(
...
         ) as span:
-            if span.is_recording():
-                if args and self._commenter_enabled:
+            should_comment = (
+                args
+                and self._commenter_enabled
+                and (span.is_recording() or self._commenter_for_all_spans)
+            )
+
+            if should_comment:
```
tammy-baylis-swi: Is there a way to make these nested if-else blocks (here and in sqlalchemy) more concise? If the original code gets chopped up a bit then it's fine as long as the tests continue to pass 🙂

tammy-baylis-swi:
This is looking good and I really appreciate the docs updates and comments so far. Just left one minor suggestion.

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
Elyahou: I refactored it a little bit, let me know if it is better

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -726,24 +789,13 @@ def traced_execution(
...
-                else:
-                    # no sqlcomment anywhere
-                    self._populate_span(span, cursor, *args)
+            can_add_comment = self._can_add_comment(span, args)
+
+            if can_add_comment:
+                args = self._apply_commenter(span, cursor, args)
+            elif span.is_recording():
+                # No sqlcomment, but still populate span for recording spans
+                self._populate_span(span, cursor, *args)
```
tammy-baylis-swi: We might be able to make this even shorter with something like this, so we wouldn't need `_apply_commenter` nor its duplicate `_populate_span` call:


```suggestion
            if can_add_comment:
                commented_args = self._update_args_with_added_sql_comment(args, cursor)
                attr_args = (
                    commented_args if self._enable_attribute_commenter else args
                )
                args = commented_args
            else:
                attr_args = args

            if span.is_recording():
                self._populate_span(span, cursor, *attr_args)
```

Elyahou: Done !

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `opentelemetry-instrumentation-sqlalchemy`, `opentelemetry-instrumentation-dbapi`: Add `commenter_for_all_spans` option to enable SQLCommenter for all spans regardless of sampling decision
```
tammy-baylis-swi: One last nit: please move this to bottom of `Added` list.

tammy-baylis-swi (approved):
Thanks @Elyahou this lgtm!

I've approved; the Maintainers of this repo also need to have a look.

CHANGELOG.md
Elyahou: Done

tammy-baylis-swi:
Hi @Elyahou sorry the approvals have been slow; we've had a ton of contributions this year! Please could you sort out the current merge conflicts.

Pinging other @open-telemetry/opentelemetry-python-contrib-approvers to have a look at this one.

Elyahou:
@tammy-baylis-swi Done

xrmx (commented):
I'm not sure I understand the use case, what are you going to correlate if the span is dropped by the sampler?

tammy-baylis-swi:
> I'm not sure I understand the use case, what are you going to correlate if the span is dropped by the sampler?

It's to maintain context propagation even if signals aren't exported. We chatted a bit about it in https://cloud-native.slack.com/archives/C01PD4HUVBL/p1767789497270419

xrmx:
> > I'm not sure I understand the use case, what are you going to correlate if the span is dropped by the sampler?
> 
> It's to maintain context propagation even if signals aren't exported. We chatted a bit about it in https://cloud-native.slack.com/archives/C01PD4HUVBL/p1767789497270419

I'm still failing to see how this is related to the sql commenter though. I guess the issue is that for some reason we lose the context for non-recording span?

tammy-baylis-swi:
> > > I'm not sure I understand the use case, what are you going to correlate if the span is dropped by the sampler?
> > 
> > 
> > It's to maintain context propagation even if signals aren't exported. We chatted a bit about it in https://cloud-native.slack.com/archives/C01PD4HUVBL/p1767789497270419
> 
> I'm still failing to see how this is related to the sql commenter though. I guess the issue is that for some reason we lose the context for non-recording span?

Yes, right now the context is not propagated via sqlcommenter at all if not sampled (`00`). But some users might want context/trace_id to still be written in a sqlcomment, i.e. propagate regardless of sampling decision. This would also better match behaviour of w3c propagation of context by HTTP headers.

Elyahou:
> I'm not sure I understand the use case, what are you going to correlate if the span is dropped by the sampler?

Our practical use case is to correlate between a sql slow query log and its trace id in the application logs.

Sampling decision should only affect the traces exports, but not propagation.

aabmass:
@Elyahou can you fix the merge conflicts?

emdneto:
Thanks for the PR!

lzchen (approved):

PR #4094, opened 2026-01-08
fix(flask): align http.server.active_requests metric with semconv helper

### LINKED ISSUE #4093: flask: different http.server.active_requests metrics description and unit
Depending on how the flask app is instrumented two different counter are initialized for `http.server.active_requests metrics`.

```
            active_requests_counter = meter.create_up_down_counter(
                name=MetricInstruments.HTTP_SERVER_ACTIVE_REQUESTS,
                unit="{request}",
                description="Number of active HTTP server requests.",
            )
```

vs

```
        active_requests_counter = meter.create_up_down_counter(
            name=MetricInstruments.HTTP_SERVER_ACTIVE_REQUESTS,
            unit="requests",
            description="measures the number of concurrent HTTP requests that are currently in-flight",
        )

```

The first one matches the 1.21.0 spec that we are targeting for stable, but the last one is the one used in auto-instrumentation and should be fixed. BTW other instrumentation are using the helper `create_http_server_active_requests` from the semconv package.

Spotted by weaver registry live-check.

rite7sh:
## Description

This PR fixes inconsistent initialization of the `http.server.active_requests`
metric in Flask instrumentation by aligning it with the semantic conventions
helper used by other instrumentations.

Depending on how Flask is instrumented, different units and descriptions were
being used for the same metric. This change ensures consistent behavior by
using the semconv helper everywhere.

Fixes #4093

## Type of change

## How Has This Been Tested?

- `python -m ruff check instrumentation/opentelemetry-instrumentation-flask`
- Existing Flask instrumentation tests
- Verified no lint or formatting issues after changes

- [x] No.

## Checklist:

- [ ] Changelogs have been updated (not required for this fix)
- [ ] Unit tests have been added (existing coverage is sufficient)
- [ ] Documentation has been updated (no user-facing changes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Add stale PR GitHub Action
+  ([#4220](https://github.com/open-telemetry/opentelemetry-python/pull/4220))
+
 ### Added
 
 - Add Python 3.14 support
@@ -62,6 +65,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `opentelemetry-instrumentation-flask`: Align `http.server.active_requests` initialization with semantic convention helpers to ensure consistent names, units, and descriptions.
   ([#4094](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4094))
+- `opentelemetry-instrumentation-asyncio`: Fix environment variables not appearing in Read the Docs documentation
+  ([#4256](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4256))
 - `opentelemetry-instrumentation-mysql`: Refactor MySQL integration test mocks to use concrete DBAPI connection attributes, reducing noisy attribute type warnings.
   ([#4116](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4116))
 - `opentelemetry-instrumentation-cassandra`: Use `_instruments_any` instead of `_instruments` for driver dependencies so that having either `cassandra-driver` or `scylla-driver` installed is sufficient
@@ -116,6 +121,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#4078](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4171))
 - `opentelemetry-instrumentation-aiohttp-server`: fix HTTP error inconsistencies
   ([#4175](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4175))
+- `opentelemetry-docker-tests` Fix docker-tests assumption by Postgres-Sqlalchemy case about scope of metrics
+  ([#4258](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4258))
+- `opentelemetry-instrumentation-threading`: fix AttributeError when Thread is run without starting
+  ([#4246](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4246))
 
 ### Breaking changes
 
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,35 @@
+name: Mark stale PRs
+
+on:
+  schedule:
+    - cron: "12 3 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  stale:
+    permissions:
+      contents: read
+      issues: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@b5d41d4e1d5dceea10e7104786b73624c18a190f # v10.2.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          days-before-pr-stale: 14
+          days-before-pr-close: 14
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
+          stale-pr-message: >
+            This PR has been automatically marked as stale because it has not had
+            any activity for 14 days. It will be closed if no further activity
+            occurs within 14 days of this comment.
+
+            If you're still working on this, please add a comment or push new commits.
+          close-pr-message: >
+            This PR has been closed due to inactivity. Please reopen if you would
+            like to continue working on it.
+          exempt-pr-labels: "hold,WIP,blocked-by-spec,do not merge"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,6 +31,7 @@ Please also read the [OpenTelemetry Contributor Guide](https://github.com/open-t
     - [How to Receive Comments](#how-to-receive-comments)
     - [How to Get PRs Reviewed](#how-to-get-prs-reviewed)
     - [How to Get PRs Merged](#how-to-get-prs-merged)
+    - [Stale PRs](#stale-prs)
   - [Design Choices](#design-choices)
     - [Focus on Capabilities, Not Structure Compliance](#focus-on-capabilities-not-structure-compliance)
   - [Running Tests Locally](#running-tests-locally)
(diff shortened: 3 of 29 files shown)


instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -680,7 +682,7 @@ def __init__(self, *args, **kwargs):
         duration_histogram_old = None
         if _report_old(_InstrumentedFlask._sem_conv_opt_in_mode):
             duration_histogram_old = meter.create_histogram(
-                name=MetricInstruments.HTTP_SERVER_DURATION,
+                name="http.server.duration",
```
JWinermaSplunk: Are we sure we want the names here and line #815 to be hardcoded and not a constant?

JWinermaSplunk (commented):
Left a small comment, but it looks simple enough if the helper functions suffice. Also, this change probably requires a changelog entry, otherwise LGTM.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
rite7sh: oh i am sorry,  I can switch this to use a constant for consistency if one is available. I’ll check the existing semconv metrics helpers and update accordingly so we’re not introducing a new hardcoded string.

rite7sh:
@JWinermaSplunk  I’ve pushed a small follow-up commit that replaces the hardcoded legacy metric name with a constant, keeping it consistent with the existing semantic convention helpers.
Please let me know if this approach looks good to you, and whether you’d like me to add a changelog entry as well, given that this doesn’t introduce any behavioral changes.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
JWinermaSplunk: I am probably missing something but why are we changing the name value from `MetricInstruments.HTTP_SERVER_DURATION` at all?

JWinermaSplunk:
A changelog entry never hurts. It is simple enough and even the current PR name would be fine as an entry, but it isn't up to me in the end :)

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
rite7sh: @JWinermaSplunk sorry if that wasn’t clear from the diff.
The intent here isn’t to change the metric name or semantics. The legacy metric remains http.server.duration.
This follow-up was only to avoid introducing a new hardcoded string and instead make that legacy name explicit via a local constant, similar to how other legacy/compat paths are handled.

If you’d prefer, I’m also happy to keep using MetricInstruments.HTTP_SERVER_DURATION directly here I mainly wanted to avoid adding another literal string while keeping behavior unchanged.
hopefully i was able to explain the intent :)

rite7sh:
@JWinermaSplunk you got a point, i surely will add a changelog update.
thanks for the suggestion.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
JWinermaSplunk: Yes, let's go ahead and keep using `MetricInstruments.HTTP_SERVER_DURATION` in the meantime. I would be happy to approve the PR after that :)

rite7sh: @JWinermaSplunk I’ve rolled back to using MetricInstruments.HTTP_SERVER_DURATION as suggested and updated the PR accordingly.
Everything should be good to go now, but please feel free to ping me if you’d like any further tweaks before approval.
Thanks for these suggestions btw :P

JWinermaSplunk (approved):

rite7sh:
@xrmx Thanks for the clarification, I agree this is a small behavior change limited to unit and description, and I’m glad it aligns both paths on the same semconv, I’ve updated the branch and am happy to tweak anything if needed. :)

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -287,6 +287,9 @@ def response_hook(span: Span, status: str, response_headers: List):
     HTTP_ROUTE,
     HTTP_TARGET,
 )
+from opentelemetry.semconv._incubating.metrics.http.server import (
```
emdneto:
```suggestion
from opentelemetry.semconv._incubating.metrics.http_metrics import (
```

CHANGELOG.md
```diff
@@ -38,6 +38,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `opentelemetry-instrumentation-flask`: Align HTTP server metrics initialization with semantic convention helpers to ensure consistent names, units, and descriptions.
```
```suggestion
- `opentelemetry-instrumentation-flask`: Align `http.server.active_requests` initialization with semantic convention helpers to ensure consistent names, units, and descriptions.
```

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -692,11 +694,7 @@ def __init__(self, *args, **kwargs):
                 description="Duration of HTTP server requests.",
                 explicit_bucket_boundaries_advisory=HTTP_DURATION_HISTOGRAM_BUCKETS_NEW,
             )
-        active_requests_counter = meter.create_up_down_counter(
-            name=MetricInstruments.HTTP_SERVER_ACTIVE_REQUESTS,
-            unit="requests",
-            description="measures the number of concurrent HTTP requests that are currently in-flight",
-        )
+        active_requests_counter = create_http_server_active_requests(meter)
```
xrmx: Can we do something like

```
if newsemconv:
   active_requests_counter = create_http_server_active_requests(meter)
else:
   active_requests_counter = meter.create_up_down_counter(
            name=MetricInstruments.HTTP_SERVER_ACTIVE_REQUESTS,
            unit="requests",
            description="measures the number of concurrent HTTP requests that are currently in-flight",
        )
```

in both hunks so that the metrics matches the semconv?

rite7sh: @xrmx I’ve updated both paths to use the explicit semconv split as suggested. Let me know if this looks good now, thanks for suggestion.

instrumentation/opentelemetry-instrumentation-flask/src/opentelemetry/instrumentation/flask/__init__.py
```diff
@@ -686,17 +688,13 @@ def __init__(self, *args, **kwargs):
             )
         duration_histogram_new = None
         if _report_new(_InstrumentedFlask._sem_conv_opt_in_mode):
-            duration_histogram_new = meter.create_histogram(
```
xrmx, line 689: This is another metric that wasn't supposed to be removed

rite7sh: @xmrx I’ve restored the duration histogram and completed the explicit semconv split for the active requests metric in both initialization paths.,would appreciate another look, thanks!

rite7sh: hey! an updated review on this would be great if possible so that i can work upon any changes that are required.

xrmx (approved):

PR #4102, opened 2026-01-14
django: handle cancelled requests more cleanly

zmievsa:
# Description

This is a duplicate of: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3848 but without the django version removal.

## Type of change

# How Has This Been Tested?

Only the newly created unit test for now. Can test more later

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -427,21 +455,6 @@ def process_response(self, request, response):
                     max(duration_s, 0),
                     duration_attrs_new,
                 )
-        self._active_request_counter.add(-1, active_requests_count_attrs)
```
zmievsa, line 376: counter decrement, activation.exit, and detach(token)
are now handled in call finally block to ensure cleanup
even when requests are cancelled

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -394,9 +425,6 @@ def process_response(self, request, response):
             if propagator:
                 propagator.inject(response)
 
-            # record any exceptions raised while processing the request
```
zmievsa, line 347: exception handling moved to call finally block

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -335,11 +369,8 @@ def process_response(self, request, response):
         if not _is_asgi_supported and is_asgi_request:
             return response
 
-        activation = request.META.pop(self._environ_activation_key, None)
+        activation = request.META.get(self._environ_activation_key)
         span = request.META.pop(self._environ_span_key, None)
-        active_requests_count_attrs = request.META.pop(
```
zmievsa: active_requests_count_attrs is popped in call finally block

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -271,24 +316,13 @@ def process_request(self, request):
...
         activation = use_span(span, end_on_exit=True)
-        activation.__enter__()  # pylint: disable=E1101
         request_start_time = default_timer()
         request.META[self._environ_timer_key] = request_start_time
         request.META[self._environ_activation_key] = activation
         request.META[self._environ_span_key] = span
         if token:
             request.META[self._environ_token] = token
 
-        if _DjangoMiddleware._otel_request_hook:
```
zmievsa, line 251: request_hook is called in call after activation.enter()

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -271,24 +316,13 @@ def process_request(self, request):
                 span.set_attribute(key, value)
 
         activation = use_span(span, end_on_exit=True)
-        activation.__enter__()  # pylint: disable=E1101
```
zmievsa: activation.enter() is called in call to ensure
proper cleanup via finally block even on request cancellation

zmievsa:
[Lint 0 / instrumentation-google-genai (pull_request)](https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/21012087605/job/60409235661?pr=4102)

This one broke because of [this](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4101)

zmievsa:
@xrmx Pinging you here since you were involved in the original PR.
@rtpg Pinging you since you are the author of the original fix. Thank you very much for it btw, I tried to keep it as close to your version as possible

xrmx:
Please drop the google genai changes

zmievsa:
@xrmx done!

zmievsa:
I would appreciate someone testing it on their setup if possible. Seems to be fine on mine

freakaton:
@xrmx Is it possible to move on with that bug? I recently stucked with that problem also.

emdneto:
Thanks for the PR!

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -149,9 +149,54 @@ def __init__(self, get_response):
         self.get_response = get_response
 
     def __call__(self, request):
+        if self._excluded_urls.url_disabled(request.build_absolute_uri("?")):
```
tammy-baylis-swi: Thanks for this PR.

This `excluded` check probably isn't needed after resolving current merge conflicts.

tammy-baylis-swi:
Please also see [this comment](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4102#issuecomment-4435602329) about the new way we update the changelog.

zmievsa:
@tammy-baylis-swi does this look good?

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -130,9 +130,47 @@ def __init__(self, get_response):
...
+        is_asgi_request = _is_asgi_request(request)
+        if not _is_asgi_supported and is_asgi_request:
+            return self.get_response(request)
+
         self.process_request(request)
-        response = self.get_response(request)
-        return self.process_response(request, response)
+        activation = request.META.get(self._environ_activation_key)
+        if activation is None:
+            return self.get_response(request)
```
tammy-baylis-swi, line 142: Thanks for revisions!

I would help to add a comment [related to OP](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3848) about why this is no longer `process_response`:

> This manifests itself in requests potentially opening an OTel span through the Django middleware but not having process_response run in the existing mode, meaning the use_span ([this one](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/ebe5bdf465b9cf893e0b0db72820a89a38ad1601/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py#L285)) closure + context detach call isn't run.

IIUC, maybe this:

```suggestion
            # This is to make sure that context and span cleanup always happen
            # if Django's ASGI handler cancels tasks that have seen a disconnect
            return self.get_response(request)
```

zmievsa: Sounds good, accepted.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -299,9 +328,8 @@ def process_response(self, request, response):
         if not _is_asgi_supported and is_asgi_request:
             return response
 
-        activation = request.META.pop(self._environ_activation_key, None)
+        activation = request.META.get(self._environ_activation_key)
```
tammy-baylis-swi: Is it necessary to change some of these `pop` to `get`?

zmievsa: Nope, pretty sure it is irrelevant here. Let me change it back. The get in `__call__` needs to remain because process_view and process_exception inspect the activation key while get_response is running. Seems to me like this one was the only useless one.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
```diff
@@ -130,9 +130,49 @@ def __init__(self, get_response):
...
+                    exception,
+                    getattr(exception, "__traceback__", None),
+                )
+            else:
+                activation.__exit__(None, None, None)
+
+            # Detach context token
+            token = request.META.pop(self._environ_token, None)
+            if token is not None:
+                detach(token)
```
tammy-baylis-swi: Should we keep the original, slightly more cautious way that this was done before? Or did that come with issues?

```suggestion
            if request.META.get(self._environ_token, None) is not None:
                detach(request.META.get(self._environ_token))
                request.META.pop(self._environ_token)
```

tammy-baylis-swi (approved):
Thanks for the revisits, I think this looks good. Maintainers will also look

Issue #2679, opened 2024-07-08
opentelemetry-instrumentation-sqlalchemy: semantic convention stability migration


emdneto:
### What problem do you want to solve?

a

### Describe the solution you'd like

b

### Would you like to implement a fix?

None

hyfj44255:
I would like to work on this and create pr soon

hyfj44255:
Hi I would like to work on this, please assign this to me

emdneto:
Hey @hyfj44255 are you still willing to work on this?

hyfj44255:
Sorry for the delay, I'm just a beginner seems a little hard for me.I'll skip it. Hope some senior volunteer can take it soon

PR #4110, opened 2026-01-21
DB and HTTP semantic convention stability migration for SQLAlchemy

tammy-baylis-swi:
# Description

Add SQLAlchemy instrumentor support for when `OTEL_SEMCONV_STABILITY_OPT_IN` includes

* "database" or "database/dup"
* "http" or "http/dup"

This changes the `SQLComment in span attribute` feature if opted in. Inclusion of sqlcomment on query span would be on `db.statement` and/or `db.query.text` attribute if default, "database", or "database/dup".

To make the changes cleaner: adds sqlalchemy helper `_get_db_name_from_cursor` to pull directly instead of having a condition to check attributes taken originally from cursor (see [this conversation](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110#discussion_r2719062417)).

For reusability by other instrumentors later: this PR adds schema url helpers to `_semconv.py` for choosing the latest appropriate version when an instrumentor uses multiple signal types (http, db, genai).

Adds `_set_db_operation` to `_semconv.py`. This adds SQLAlchemy instrumentor setting of `db.operation` and/or `db.operation.name` attributes -- this was only done for span name before.

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2679

## Type of change

# How Has This Been Tested?

- [x] Added unit tests

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#4305](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4305))
 - Don't import module in unwrap if not already imported
   ([#4321](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4321))
+- `opentelemetry-instrumentation-logging`: Map Python `CRITICAL` log level to OTel `FATAL` severity text and `WARNING` to `WARN`
+  ([#4365](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4365))
 - `opentelemetry-instrumentation-logging`: Add recursion guard in LoggingHandler.emit to prevent deadlock
   ([#4302](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4302))
 - `opentelemetry-instrumentation-grpc`: Fix bidirectional streaming RPCs raising `AttributeError: 'generator' object has no attribute 'add_done_callback'`
--- a/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/handler.py
+++ b/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/handler.py
@@ -188,10 +188,14 @@ class LoggingHandler(logging.Handler):
             else:
                 body = record.getMessage()
 
-        # related to https://github.com/open-telemetry/opentelemetry-python/issues/3548
-        # Severity Text = WARN as defined in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#displaying-severity.
-        level_name = (
-            "WARN" if record.levelname == "WARNING" else record.levelname
+        # Map Python log level names to OTel severity text as defined in
+        # https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#displaying-severity
+        _python_to_otel_severity_text = {
+            "WARNING": "WARN",
+            "CRITICAL": "FATAL",
+        }
+        level_name = _python_to_otel_severity_text.get(
+            record.levelname, record.levelname
         )
 
         return LogRecord(
--- a/instrumentation/opentelemetry-instrumentation-logging/tests/test_handler.py
+++ b/instrumentation/opentelemetry-instrumentation-logging/tests/test_handler.py
@@ -357,7 +357,7 @@ class TestLoggingHandler(unittest.TestCase):
                     record.log_record.body,
                     "Critical message within span",
                 )
-                self.assertEqual(record.log_record.severity_text, "CRITICAL")
+                self.assertEqual(record.log_record.severity_text, "FATAL")
                 self.assertEqual(
                     record.log_record.severity_number,
                     SeverityNumber.FATAL,
@@ -390,7 +390,7 @@ class TestLoggingHandler(unittest.TestCase):
             self.assertEqual(
                 record.log_record.body, "Critical message within span"
             )
-            self.assertEqual(record.log_record.severity_text, "CRITICAL")
+            self.assertEqual(record.log_record.severity_text, "FATAL")
             self.assertEqual(
                 record.log_record.severity_number, SeverityNumber.FATAL
             )


instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -289,11 +314,18 @@ def _before_cur_exec(
...
+        )
         if not found:
-            attrs = _get_attributes_from_cursor(self.vendor, cursor, attrs)
+            attrs = _get_attributes_from_cursor(
+                self.vendor, cursor, attrs, self._sem_conv_opt_in_mode
+            )
+
+        # Extract db_name for operation name
+        # Prefer old semconv (db.name) over new semconv (db.namespace) for backwards compatibility
+        db_name = attrs.get(DB_NAME) or attrs.get(DB_NAMESPACE) or ""
```
tammy-baylis-swi: This was a small design choice; lmk if should be another way.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -143,6 +161,12 @@ def __init__(
         commenter_options=None,
         enable_attribute_commenter=False,
     ):
+        # Initialize semantic conventions opt-in if needed
+        _OpenTelemetrySemanticConventionStability._initialize()
+        self._sem_conv_opt_in_mode = _OpenTelemetrySemanticConventionStability._get_opentelemetry_stability_opt_in_mode(
```
herin049: Shouldn't we really be extracting both the HTTP and DB semconv opt in modes since a user might enable HTTP migration but not DB migration?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -119,13 +127,23 @@ def _wrap_connect_internal(func, module, args, kwargs):
         if not is_instrumentation_enabled():
             return func(*args, **kwargs)
 
+        # Initialize semantic conventions opt-in if needed
+        _OpenTelemetrySemanticConventionStability._initialize()
+        sem_conv_opt_in_mode = _OpenTelemetrySemanticConventionStability._get_opentelemetry_stability_opt_in_mode(
```
herin049: Same comment as below.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
tammy-baylis-swi: Yikes, thank you! Added extraction and usage of both in [653a1e3](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110/commits/653a1e340a2c042e9d486b85c417dc728ce64822). Also updated PR title and description.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
tammy-baylis-swi: Addressed in [653a1e3](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110/commits/653a1e340a2c042e9d486b85c417dc728ce64822)

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
```diff
@@ -181,12 +186,18 @@ def _instrument(self, **kwargs):
...
+            _OpenTelemetryStabilitySignalType.DATABASE,
+        )
+
         tracer_provider = kwargs.get("tracer_provider")
         tracer = get_tracer(
             __name__,
             __version__,
             tracer_provider,
-            schema_url="https://opentelemetry.io/schemas/1.11.0",
+            schema_url=_get_schema_url(sem_conv_opt_in_mode),
```
herin049: Not sure if we need to worry about this scenario, and if it is even possible to define a schema URL for it, but what is the expected behavior if only of of DB or HTTP is opt in?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
herin049: It seems like the `db_name` should be the same regardless, right? We should be able to simply pull the DB name from the cursor, since pulling it back from the attributes seems like a very roundabout way to do this.

My preference would be to create a separate function to pull the DB name from the cursor. This can be used in `_get_attributes_from_cursor` and here for the Span name.

herin049 (approved):

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
tammy-baylis-swi: Good point. Initial expected behaviour would have only been DB opt-in, but now both HTTP and DB are (and should be) considered in [653a1e3](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110/commits/653a1e340a2c042e9d486b85c417dc728ce64822). Right now, DB opt-in dictates whether schema url is default or new. This is ok if mode is `http,database` or dup or both default, but a mix (`http,default` or `default,database`) will mix network and db attributes from the same instrumentor.

But I _think_ this is ok. If users/vendors are knowledgeable enough to be migrating before the breaking change to opt-in-by-default, then it can be dup for both anyway. THoughts?

This doesn't fix things but it's more consistent now in [b9ef17a](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110/commits/b9ef17ad59ddc9d9a6d936a56a16cc8e2caeeda9) for doing the same for `meter`.

herin049: My concern is more that once we start emitting stable HTTP attributes by default, the schema URL should be using `1.21` (since this is the version HTTP became made stable), now if DB opt is is enabled then we should be using `1.25`. If we only check the DB opt in, then it could fallback on `1.11`. 

Of course, this is assuming we're using version `1.25` of DB semconv and not `1.21`. Do we know which one we plan on targeting?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
tammy-baylis-swi: Added `_get_db_name_from_cursor` in [66e0dbd](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110/commits/66e0dbd693332aa7036e61eabc558069e0e522b3). It looks a little odd for a few reasons:

1. From existing `_normalize_vendor` and `_get_attributes_from_cursor` methods, it seems this instrumentor only "supports" postgresql and sqlite connection. So I followed this to keep scope to semconv migration.
2. The new helper gets postgres cursor connection the same way `_get_attributes_from_cursor` does.
3. We use `conn.engine.url.database` if sqlite, instead of from-cursor, else we'd have to create a new cursor ([execute](https://docs.python.org/3/library/sqlite3.html)) and do [PRAGMA database_list](https://www.sqlite.org/pragma.html) and that's not ideal.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
tammy-baylis-swi: Yes I think [1.11](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1977) (old hardcode), [1.21](https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.21.0), and [1.25](https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.25.0) make sense. How about this:

| opt-in | semconv |
|-------|-----------|
| default |  1.11.0 |
| http,default | 1.21.0 |
| http/dup,default | 1.21.0 |
| *,database| 1.25.0|
| *,database/dup| 1.25.0|

"default" case would not exist anymore if stable HTTP is new default and no option to opt out (iirc from last sig). DB would dictate highest and HTTP takes lower precedence.

herin049: Yes, exactly what I had in mind! I understand that this adds some scope, but figured we should do this correctly. Perhaps, we can add either a function directly in `_OpenTelemetrySemanticConventionStability` to get the schema URL to be used or expose a way to return the Semcov stability mapping in `_OpenTelemetrySemanticConventionStability` (e.g. with a function `_get_stability_signal_type_mapping`), and make a backwards compatible change to `_get_schema_url` to have the ability to accept this mapping and return the highest semconv version found. I also agree that we should simply use the highest associated semconv version based on the opted in semconv types. For example, if we add `rpc` that uses semconv 1.38, if this is set then it will take precedence over everything else.

tammy-baylis-swi: As a step in that direction, in [5477ff8](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110/commits/5477ff84e6f3c29a6e58ba9bfb06ed53afa33981) I've added a separate `_get_schema_url_for_signal_types` to choose the highest version if default, `http`, `database`, or `gen_ai`. In that commit, SQLAlchemy instrumentor is the only one that uses it while all other instrumentors are using the classic `_get_schema_url`.

Lmk what you think!

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -19,6 +19,8 @@
 from enum import Enum
 from typing import Container, Mapping, MutableMapping
 
+from packaging import version as package_version
```
tammy-baylis-swi, line 23: `packaging` is already a dep of `opentelemetry-instrumentation` so might as well use its built-in equality checks.

opentelemetry-instrumentation/tests/test_semconv.py
```diff
@@ -188,6 +191,133 @@ def test_stability_mode_dup_precedence(self):
         )
 
 
+class TestOpenTelemetrySemConvSchemaUrl(TestCase):
+    @stability_mode("")
```
tammy-baylis-swi, line 198: These decorators are needed for clean env at each test

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
tammy-baylis-swi: > it seems this instrumentor only "supports" postgresql and sqlite connection

I was wrong! Fixed in [903e1f8](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110/commits/903e1f874fa527188d0ff8894ec848b2555a9806) onwards so that `docker-tests` still pass. Cursor connection is checked for PostgreSQL, MySQL, MSSQL variants. I've split into multiple helpers for linting branch/return rules.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -634,8 +639,57 @@ def _set_status(
...
+
+def _get_schema_version_for_opt_in_mode(
+    signal_type: _OpenTelemetryStabilitySignalType,
+    mode: _StabilityMode,
+) -> str:
+    """Get the schema version for a specific signal type and opt-in mode."""
+    if mode == _StabilityMode.DEFAULT:
+        return _LEGACY_SCHEMA_VERSION
+
+    signal_versions = {
```
MikeGoldsmith, line 675: Could we import / use _semconv.py and use it's version strings instead of these custom strings? eg `Schemas.V1_25_0.value`

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -22,26 +22,81 @@
...
 
 
+def _get_db_name_from_cursor(vendor, cursor):
+    """Return DB name from cursor for PostgreSQL, MySQL, or MSSQL -- else None"""
+    if not vendor:
+        return None
+
+    vendor = vendor.lower()
+    db_name = None
+    if "postgres" in vendor:
```
MikeGoldsmith, line 55: We're missing `sqlite` which is needed in `_get_attributes_from_cursor`

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -278,9 +359,10 @@ def _get_commenter_data(self, conn) -> dict:
 
     def _set_db_client_span_attributes(self, span, statement, attrs) -> None:
         """Uses statement and attrs to set attributes of provided Otel span"""
-        span.set_attribute(DB_STATEMENT, statement)
-        span.set_attribute(DB_SYSTEM, self.vendor)
-        for key, value in attrs.items():
+        span_attrs = dict(attrs)
```
MikeGoldsmith, line 374: Should we also set `db.operation.name` here too?

opentelemetry-instrumentation/tests/test_semconv.py
```diff
@@ -188,6 +191,133 @@ def test_stability_mode_dup_precedence(self):
...
+    def test_get_schema_version_for_opt_in_mode_database_stable(self):
+        version = _get_schema_version_for_opt_in_mode(
+            _OpenTelemetryStabilitySignalType.DATABASE, _StabilityMode.DATABASE
+        )
+        self.assertEqual(version, "1.25.0")
+
+    @stability_mode("")
+    def test_get_schema_version_for_opt_in_mode_gen_ai_stable(self):
+        version = _get_schema_version_for_opt_in_mode(
+            _OpenTelemetryStabilitySignalType.GEN_AI, _StabilityMode.HTTP
```
MikeGoldsmith: Should this should be GEN_AI stability?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
tammy-baylis-swi: We can use `Schemas.V1_21_0.value` and they're [full urls](https://github.com/open-telemetry/opentelemetry-python/blob/80bf95517a48a14c3876271d9eab16c8c79f890a/opentelemetry-semantic-conventions/src/opentelemetry/semconv/schemas.py#L19). So in [c74d7ca](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110/commits/c74d7ca7973427dacacb9b5a4ae982c047cb4690) I added that with a urlparse call, as I couldn't find any utils in `opentelemetry-instrumentation` nor `semantic-conventions` to extract the `1.XX.X` part

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
tammy-baylis-swi: Did a cleanup in [e15fe2a](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110/commits/e15fe2a671e4009ec8bf2920e6f63cbe5a1cb783) so the new helpers are now `_get_db_name_from_cursor_or_conn` and `_get_attributes_from_cursor_or_conn`. This is hopefully less muddled than previous `if "sqlite"` checks.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
tammy-baylis-swi: Added in [b4ab9a4](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110/commits/b4ab9a4679af6bc79fa116b35af1f40943043c2a) as a toggle between `db.operation` (old) and/or `db.operation.name` (new) attributes -- this was only being done for span name before. Updated PR description as well.

EDIT: The existing `_set_db_client_span_attributes` helper whose args I changed isn't used by any other instrumentors nor by core repo.

opentelemetry-instrumentation/tests/test_semconv.py
tammy-baylis-swi: Yes, `GEN_AI_LATEST_EXPERIMENTAL`! Fixed in [30b15b3](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110/commits/30b15b33f75fc306b88744c553e86bae103c3f36)

tammy-baylis-swi:
@open-telemetry/python-maintainers May I get a review please?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
```diff
@@ -181,20 +186,32 @@ def _instrument(self, **kwargs):
...
+            ]
+        )
+
         tracer_provider = kwargs.get("tracer_provider")
         tracer = get_tracer(
             __name__,
             __version__,
             tracer_provider,
-            schema_url="https://opentelemetry.io/schemas/1.11.0",
+            schema_url=schema_url,
```
xrmx, line 206: I'm not sure I've seen tests in the sqlalchemy instrumentation assert the schema when testing old / stable semconv

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -22,26 +22,91 @@
...
 from opentelemetry.semconv._incubating.attributes.net_attributes import (
-    NET_PEER_NAME,
-    NET_PEER_PORT,
     NET_TRANSPORT,
     NetTransportValues,
 )
 from opentelemetry.trace.status import Status, StatusCode
 
 
+def _get_db_name_from_cursor_or_conn(vendor, conn, cursor):
```
xrmx, line 48: Question: why have we started looking also in the connection?

tammy-baylis-swi: It's more robust to deal with the different DBs that SQLAlchemy can connect to, namely sqlite where connection is more reliable than cursor. It's also a general update from the roundabout way mentioned in [this thread](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110#discussion_r2780349850). It's a bit extra for a semconv update but might as well since we're changing attributes values anyway.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
tammy-baylis-swi: Added in [04de712](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110/commits/04de7125a2d371ee8113f8f7bca3265d16d5afe6)

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
xrmx: Thanks for the pointer.

PR #4112, opened 2026-01-21
opentelemetry-instrumentation-logging: don't add out of spec attributes by default

### LINKED ISSUE #3687: Unnecessary/noncompliant attributes added via logging instrumentation
Running the app in https://opentelemetry.io/docs/languages/python/getting-started/ with

```
export OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED=true
opentelemetry-instrument \
    --traces_exporter console \
    --metrics_exporter console \
    --logs_exporter console \
    --service_name dice-server \
    flask run -p 8080
```

results in logs like:

```
{
    "body": "Anonymous player is rolling the dice: 1",
    "severity_number": 13,
    "severity_text": "WARN",
    "attributes": {
        "otelSpanID": "57a328e3a762f95e",
        "otelTraceID": "a794f5ff6f076ca3e8744dd222d27760",
        "otelTraceSampled": true,
        "otelServiceName": "dice-server",
        "code.file.path": "/tmp/tmp.ZHMcsMeRYL/otel-getting-started/app.py",
        "code.function.name": "roll_dice",
        "code.line.number": 17
    },
    "dropped_attributes": 0,
    "timestamp": "2025-07-01T15:07:59.544312Z",
    "observed_timestamp": "2025-07-01T15:07:59.544731Z",
    "trace_id": "0xa794f5ff6f076ca3e8744dd222d27760",
    "span_id": "0x57a328e3a762f95e",
    "trace_flags": 1,
    "resource": {
        "attributes": {
            "telemetry.sdk.version": "1.34.1",
            "service.name": "dice-server",
            "telemetry.auto.version": "0.55b1"
        },
        "schema_url": ""
    }
}
```

You can see the `trace_id`, `span_id` , `trace_flags` and `resource::attributes::service.name` are also present as log-level attributes with the names `otelSpanID`, `otelTraceID` , `otelTraceSampled` and `otelServiceName`  respectively. This is mentioned here: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/b1c2c7941b9f5aa4762c648b0e3644de80c1e0a4/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/constants.py#L25-L30

I spoke with @xrmx on Slack about this (see https://cloud-native.slack.com/archives/C01PD4HUVBL/p1751382766109069) and this seems to be historical artifact that could be removed. Since this is reported by opentelemetry-demo and it's on our official docs, it has some weight as an 'official' thing when it seems like it is not, so I would suggest removing it.

xrmx:
# Description

Add span context attributes to logging module LogRecord only if the instrumentation has been configured to do so via set_logging_format argument or OTEL_PYTHON_LOG_CORRELATION environment variable.

Fixes #3687

## Type of change

# How Has This Been Tested?

- [x] tox

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-logging/tests/test_logging.py
+++ b/instrumentation/opentelemetry-instrumentation-logging/tests/test_logging.py
@@ -47,7 +47,9 @@ class TestLoggingInstrumentorProxyTracerProvider(TestBase):
 
     def setUp(self):
         super().setUp()
-        LoggingInstrumentor().instrument(tracer_provider=FakeTracerProvider())
+        LoggingInstrumentor().instrument(
+            tracer_provider=FakeTracerProvider(), set_logging_format=True
+        )
 
     def tearDown(self):
         super().tearDown()
@@ -94,7 +96,21 @@ class TestLoggingInstrumentor(TestBase):
             self.assertEqual(record.otelTraceSampled, trace_sampled)
             self.assertEqual(record.otelServiceName, "unknown_service")
 
-    def test_trace_context_injection(self):
+    @mock.patch.dict("os.environ", {"OTEL_PYTHON_LOG_CORRELATION": "true"})
+    def test_trace_context_injection_with_log_correlation_from_env_var(self):
+        LoggingInstrumentor().uninstrument()
+        LoggingInstrumentor().instrument()
+        with self.tracer.start_as_current_span("s1") as span:
+            span_id = format(span.get_span_context().span_id, "016x")
+            trace_id = format(span.get_span_context().trace_id, "032x")
+            trace_sampled = span.get_span_context().trace_flags.sampled
+            self.assert_trace_context_injected(
+                span_id, trace_id, trace_sampled
+            )
+
+    def test_trace_context_injection_with_log_correlation_instrument_arg(self):
+        LoggingInstrumentor().uninstrument()
+        LoggingInstrumentor().instrument(set_logging_format=True)
         with self.tracer.start_as_current_span("s1") as span:
             span_id = format(span.get_span_context().span_id, "016x")
             trace_id = format(span.get_span_context().trace_id, "032x")
@@ -104,6 +120,8 @@ class TestLoggingInstrumentor(TestBase):
             )
 
     def test_trace_context_injection_without_span(self):
+        LoggingInstrumentor().uninstrument()
+        LoggingInstrumentor().instrument(set_logging_format=True)
         self.assert_trace_context_injected("0", "0", False)
 
     @mock.patch("logging.basicConfig")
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry-instrumentation-logging/tests/test_logging.py
```diff
@@ -158,15 +174,35 @@ def test_custom_format_and_level_api(self, basic_config_mock):  # pylint: disabl
...
+                record = self.caplog.records[0]
+                self.assertFalse(hasattr(record, "otelSpanID"))
+                self.assertFalse(hasattr(record, "otelTraceID"))
+                self.assertFalse(hasattr(record, "otelServiceName"))
+                self.assertFalse(hasattr(record, "otelTraceSampled"))
+                self.assertEqual(
+                    record.custom_user_attribute_from_log_hook, "some-value"
+                )
+
+    def test_log_hook_with_set_logging_format(self):
```
tammy-baylis-swi, line 207: If log hook is used and `set_logging_format` is not set but env var is, would `otel*` be in the record?

instrumentation/opentelemetry-instrumentation-logging/tests/test_logging.py
```diff
@@ -94,16 +96,34 @@ def assert_trace_context_injected(self, span_id, trace_id, trace_sampled):
...
+        with self.tracer.start_as_current_span("s1") as span:
+            span_ctx = span.get_span_context()
+            span_id = format(span_ctx.span_id, "016x")
+            trace_id = format(span_ctx.trace_id, "032x")
+            trace_sampled = span_ctx.trace_flags.sampled
+            self.assert_trace_context_injected(
+                span_id, trace_id, trace_sampled
+            )
+
+    def test_trace_context_injection_with_log_correlation_instrument_arg(self):
```
tammy-baylis-swi, line 112: Maybe we could have a 3rd test like this without hooks, but with neither `set_logging_format` nor env var

instrumentation/opentelemetry-instrumentation-logging/tests/test_logging.py
xrmx: Yes, the path to set the variable used to decide if adding the records is shared between env var and instrumentor parameter.

instrumentation/opentelemetry-instrumentation-logging/tests/test_logging.py
xrmx: Good catch, forgot to add it :sweat_smile: Thanks!

JWinermaSplunk (approved):
LGTM!

tammy-baylis-swi (approved):
Thanks!

emdneto (approved):

takeda:
@xrmx after this change I'm unable to get integration working.

I added `%(otelSpanID)s %(otelTraceID)s %(otelServiceName)s %(otelTraceSampled)s` to the log format, but those entries are not being logged.

Looking at the changes I see the following code:

```py
                    if set_logging_format:
```                     
So looks like  `set_logging_format` is required to get those attributes.

But enabling this option also runs `basicConfig()`:

```py
        if set_logging_format:
            log_format = kwargs.get(
                "logging_format", environ.get(OTEL_PYTHON_LOG_FORMAT, None)
            )
            log_format = log_format or DEFAULT_LOGGING_FORMAT

            log_level = kwargs.get(
                "log_level", LEVELS.get(environ.get(OTEL_PYTHON_LOG_LEVEL))
            )
            log_level = log_level or logging.INFO

            logging.basicConfig(format=log_format, level=log_level)
```

Which I don't want to do as I am configuring logging when the application starts. Can the integration always inject the log attributes? I don't see a scenario where the user wouldn't want that.

It even contradicts the documentation: https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/logging/logging.html#enable-trace-context-injection  and https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/logging/logging.html#manually-calling-logging-basicconfig which says that what I'm trying to do should exactly like that and it implies that those attributes are always injected.

LukasBoersma:
@xrmx any update on the problem that @takeda mentioned? I too need the attributes injected, but am relying on a complex logging setup that the `basicConfig()` call would break. Had to pin our dependency to an older version for now.

takeda:
@LukasBoersma I suspect that perhaps we need to open a brand new issue to get attention

Issue #4111, opened 2026-01-21
Give more information to samplers for http server instrumentations


xrmx:
Or pass the request attributes at span creation instead of adding them later. Some http instrumentation make it hard to do sampling because the sampler does not have enough information (e.g. user-agent, url path) in the attributes.

These are status as of be56431caa05be1cd8ccd610ea13f7584f995acd (there are some in flight PRs for adding stable semconv but they are not changing this):

- [x] aiohttp-server, needs fixes: attributes added after span creation
- [x] asgi, has user_agent and url.path
- [x] django, has user_agent and url.path
- [x] falcon, needs fixes: attributes added after span creation
- [x] fastapi, see asgi
- [x] flask, has user and url.path
- [x] pyramid, needs fixes: attributes added after span creation
- [x] starlette, see asgi
- [x] tornado, needs fixes: attributes added after span creation
- [x] wsgi, has user_agent and url.path

Most of the time this would be matter of moving the call that gather the attributes for the request before the span creation.

PR #4118, opened 2026-01-22
aiohttp-server: pass request attributes at span creation

xrmx:
# Description

Pass request attributes at span creation so it's possible to do sampling decision based on that.

Refs #4111

## Type of change

# How Has This Been Tested?

- [x] tox

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,18 +29,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3819](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3819))
 - `opentelemetry-instrumentation-system-metrics`: Add support for the `OTEL_PYTHON_SYSTEM_METRICS_EXCLUDED_METRICS` environment variable
   ([#3959](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3959))
-- `opentelemetry-instrumentation-httpx`: add ability to capture custom headers 
+- `opentelemetry-instrumentation-httpx`: add ability to capture custom headers
   ([#4047](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4047))
 - `opentelemetry-instrumentation-urllib3`: add ability to capture custom headers
   ([#4050](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4050))
 - `opentelemetry-instrumentation-urllib`: add ability to capture custom headers
   ([#4051](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4051))
+- `opentelemetry-instrumentation-confluent-kafka`: Increase confluent-kafka upper bound to support newer versions (2.13.0)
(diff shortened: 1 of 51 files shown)


tammy-baylis-swi (approved):
Makes sense. [These pass](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/533ece69046bd8d31be864cfbcb17391ad5eeb81/instrumentation/opentelemetry-instrumentation-aiohttp-server/tests/test_aiohttp_server_integration.py#L473-L484). Could maybe add "assert called with" for the tracer but I think that's more brittle than useful.

xrmx:
> Makes sense. [These pass](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/533ece69046bd8d31be864cfbcb17391ad5eeb81/instrumentation/opentelemetry-instrumentation-aiohttp-server/tests/test_aiohttp_server_integration.py#L473-L484). Could maybe add "assert called with" for the tracer but I think that's more brittle than useful.

If we want to keep these tested for the sampler use case there could be tests with a sampler that is asserting whatever attribute it expects but for me the fact that it's not breaking anything is enough :)

tammy-baylis-swi:
> If we want to keep these tested for the sampler use case there could be tests with a sampler that is asserting whatever attribute it expects but for me the fact that it's not breaking anything is enough :)

Works for me! 👍

instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
```diff
@@ -468,13 +468,12 @@ async def _middleware(request, handler):
             span_name,
             context=extract(request, getter=getter),
             kind=trace.SpanKind.SERVER,
+            attributes=request_attrs,
         ) as span:
             if span.is_recording():
-                request_headers_attributes = (
+                span.set_attributes(
                     collect_request_headers_attributes(request)
```
herin049, line 475: What's the reasoning behind not moving this to before the span is created?

xrmx: Less work in case of unsampled span, we can move that as well if there's a need to do sampling based on the captured request http headers.

herin049: Makes sense, although I would imagine that sampling based on request headers might be a pretty common use case (e.g. sampling if certain headers are present or if a given header matches a given value). Let me know what you think, also fine with keep this as is if the consensus is that sampling based on header attributes is an uncommon use case.

xrmx: AFAIK nobody asked for that yet :) We can revise this later and take care of keeping all the instrumentations behave the same wrt the attributes they provide to samplers.

PR #4142, opened 2026-01-26
opentelementry-instrumentation-google-genai: add gen_ai.tool.definitions to experimental semconv

wikaaaaa:
# Description

Change adds [gen_ai.tool.definitions](https://github.com/open-telemetry/semantic-conventions/blob/v1.39.0/docs/registry/attributes/gen-ai.md) from otel v1.39 to `gen_ai.client.inference.operation.details` when experimental semconv is enabled. It's done similarly to how [gen_ai.system_instructions](https://github.com/open-telemetry/semantic-conventions/blob/v1.39.0/docs/registry/attributes/gen-ai.md) are added. 

## Type of change

# How Has This Been Tested?

- Tested manually by running adk web on top of the added changes and verified the event contains the proper tool.definitions field.
- Added the corresponding assertions to relevant test cases in [opentelemetry-python-contrib/instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/nonstreaming_base.py](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content), tests pass.

No.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
@@ -306,6 +306,9 @@ def failing_hook(msg):
     return hook
 
 
+SCOPE = "opentelemetry.instrumentation.asgi"
+
+
 # pylint: disable=too-many-public-methods
 class TestAsgiApplication(AsyncAsgiTestBase):
     def setUp(self):
@@ -340,16 +343,7 @@ class TestAsgiApplication(AsyncAsgiTestBase):
     def _assert_exemplars_present(
         self, metric_names: set[str], context: str = ""
     ):
-        metrics_list = self.memory_metrics_reader.get_metrics_data()
-        print(metrics_list)
-        metrics = []
-        for resource_metric in (
-            getattr(metrics_list, "resource_metrics", []) or []
-        ):
-            for scope_metric in (
-                getattr(resource_metric, "scope_metrics", []) or []
-            ):
-                metrics.extend(getattr(scope_metric, "metrics", []) or [])
+        metrics = self.get_sorted_metrics(SCOPE)
 
         found = {name: 0 for name in metric_names}
         for metric in metrics:
@@ -421,7 +415,7 @@ class TestAsgiApplication(AsyncAsgiTestBase):
             self.assertIsNone(exc_info)
 
         # Check spans
-        span_list = self.memory_exporter.get_finished_spans()
+        span_list = self.get_finished_spans()
         expected_old = [
             {
                 "name": "GET / http receive",
@@ -557,7 +551,7 @@ class TestAsgiApplication(AsyncAsgiTestBase):
             self.assertDictEqual(dict(span.attributes), expected["attributes"])
             self.assertEqual(
                 span.instrumentation_scope.name,
-                "opentelemetry.instrumentation.asgi",
+                SCOPE,
             )
             if "events" in expected:
                 self.assertEqual(len(span.events), len(expected["events"]))
@@ -647,7 +641,7 @@ class TestAsgiApplication(AsyncAsgiTestBase):
         await self.send_default_request()
         outputs = await self.get_all_output()
         self.validate_outputs(outputs)
-        span_list = self.memory_exporter.get_finished_spans()
+        span_list = self.get_finished_spans()
         server_span = span_list[-1]
         assert server_span.kind == SpanKind.SERVER
         span_duration_nanos = server_span.end_time - server_span.start_time
@@ -674,7 +668,7 @@ class TestAsgiApplication(AsyncAsgiTestBase):
             self.seed_app(app)
             await self.send_default_request()
             await self.get_all_output()
-            span_list = self.memory_exporter.get_finished_spans()
+            span_list = self.get_finished_spans()
             self.assertTrue(span_list)
             for span in span_list:
                 for excluded_span in excluded_spans:
@@ -706,7 +700,7 @@ class TestAsgiApplication(AsyncAsgiTestBase):
             return expected
 
         self.validate_outputs(outputs, modifiers=[add_body_and_trailer_span])
-        span_list = self.memory_exporter.get_finished_spans()
+        span_list = self.get_finished_spans()
         server_span = span_list[-1]
         assert server_span.kind == SpanKind.SERVER
         span_duration_nanos = server_span.end_time - server_span.start_time
@@ -767,7 +761,7 @@ class TestAsgiApplication(AsyncAsgiTestBase):
         self.assertEqual(response_body["body"], b"*")
         self.assertEqual(response_start["status"], 200)
 
-        span_list = self.memory_exporter.get_finished_spans()
+        span_list = self.get_finished_spans()
         self.assertEqual(len(span_list), 0)
 
     async def test_behavior_with_scope_server_as_none(self):
@@ -1091,7 +1085,7 @@ class TestAsgiApplication(AsyncAsgiTestBase):
         await self.send_default_request()
         response_start, response_body, *_ = await self.get_all_output()
 
-        span = self.memory_exporter.get_finished_spans()[-1]
+        span = self.get_finished_spans()[-1]
         self.assertEqual(trace_api.SpanKind.SERVER, span.kind)
 
         self.assertEqual(response_body["body"], b"*")
@@ -1131,7 +1125,7 @@ class TestAsgiApplication(AsyncAsgiTestBase):
         await self.send_input({"type": "websocket.receive", "text": "ping"})
         await self.send_input({"type": "websocket.disconnect"})
         await self.get_all_output()
-        span_list = self.memory_exporter.get_finished_spans()
+        span_list = self.get_finished_spans()
         self.assertEqual(len(span_list), 6)
         expected = [
             {
@@ -1205,7 +1199,7 @@ class TestAsgiApplication(AsyncAsgiTestBase):
         await self.send_input({"type": "websocket.receive", "text": "ping"})
         await self.send_input({"type": "websocket.disconnect"})
         await self.get_all_output()
-        span_list = self.memory_exporter.get_finished_spans()
+        span_list = self.get_finished_spans()
         self.assertEqual(len(span_list), 6)
         expected = [
             {
@@ -1278,7 +1272,7 @@ class TestAsgiApplication(AsyncAsgiTestBase):
         await self.send_input({"type": "websocket.receive", "text": "ping"})
         await self.send_input({"type": "websocket.disconnect"})
         await self.get_all_output()
-        span_list = self.memory_exporter.get_finished_spans()
+        span_list = self.get_finished_spans()
         self.assertEqual(len(span_list), 6)
         expected = [
             {
@@ -1368,7 +1362,7 @@ class TestAsgiApplication(AsyncAsgiTestBase):
         await self.send_input({"type": "websocket.disconnect"})
         _, socket_send, *_ = await self.get_all_output()
 
-        span = self.memory_exporter.get_finished_spans()[-1]
+        span = self.get_finished_spans()[-1]
         self.assertEqual(trace_api.SpanKind.SERVER, span.kind)
 
         trace_id = format_trace_id(span.get_span_context().trace_id)
@@ -1390,7 +1384,7 @@ class TestAsgiApplication(AsyncAsgiTestBase):
         app = otel_asgi.OpenTelemetryMiddleware(simple_asgi)
         self.seed_app(app)
         await self.send_default_request()
-        span_list = self.memory_exporter.get_finished_spans()
+        span_list = self.get_finished_spans()
         self.assertEqual(len(span_list), 0)
 
     async def test_hooks(self):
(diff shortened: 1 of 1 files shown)


instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -185,6 +187,33 @@ def _to_dict(value: object):
...
+    if isinstance(tool, dict):
+        return tool
+    if hasattr(tool, "to_json_dict"):
+        try:
+            return tool.to_json_dict()
+        except TypeError:
+            pass
+    if hasattr(tool, "model_dump"):
+        try:
+            return tool.model_dump()
```
RKest: Consider `tool.model_dump(exclude_none=True)`

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -185,6 +187,33 @@ def _to_dict(value: object):
...
+        return {
+            "function": {
+                "name": getattr(tool, "__name__", str(type(tool))),
+                "description": doc.strip(),
+            }
+        }
+    try:
+        return {"value": str(tool)}
+    except Exception:
+        return {"error": "failed to serialize tool definition"}
```
RKest: Consider `return {"error": "failed to serialize tool definition {type(tool)=}"}`

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
wikaaaaa: Done

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
wikaaaaa: Done

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -503,6 +555,7 @@ def _maybe_log_completion_details(
             input_messages,
             output_messages,
             system_instructions,
+            tool_definitions,
```
DylanRussell, line 647: I'm wondering if we should gate this attribute behind the `ContentCapture` flag or just always include it. It's not really sensitive user content that's part of the conversation. @aabmass WDYT ?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -185,6 +187,35 @@ def _to_dict(value: object):
     return json.loads(json.dumps(value))
 
 
+def _to_tool_definition(tool: ToolUnionDict) -> MessagePart:
+    if isinstance(tool, dict):
```
DylanRussell: `ToolUnionDict` can take on 4 different types: https://github.com/googleapis/python-genai/blob/965395b2f640a7d92a6df3d03020f4d15fe6b2fa/google/genai/types.py#L4273 -- can we explicitly do if/else statements on those 4 types instead of what you have ? -- and can we also have test cases for those 4 types (I haven't looked at the tests yet so sorry if you already did it)...

wikaaaaa: I addressed this in the new commit, the if/else statements correspond to the tool types. And I added test cases for each. 

One thing im not sure of, what to do with mcp client sessions, we can either use asyncio like i did in the commit or just return a message saying we cant serialize it if were in the sync execution. wdyt?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -185,6 +189,49 @@ def _to_dict(value: object):
...
+    try:
+        return {"raw_definition": json.loads(json.dumps(tool))}
+    except Exception:
+        return {
+            "error": f"failed to serialize tool definition, tool type={type(tool).__name__}"
+        }
+
+
+def _to_tool_definition(tool: ToolUnionDict) -> MessagePart:
+    if isinstance(tool, McpClientSession):
```
DylanRussell: Looks like the gen AI SDK only supports passing `McpClientSession` when the async `generate_content` is called: https://github.com/googleapis/python-genai/blob/main/google/genai/models.py#L5188 also they have a util that calls ListTools and does some conversion: https://github.com/googleapis/python-genai/blob/main/google/genai/models.py#L7006 but it unfortunately is in a private package we probably shouldn't depend on.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
DylanRussell: Yeah see my other content, the SDK throws an error if they try to use it in the sync version, so probably we should just skip it if we see it in that case

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
DylanRussell: I wonder if we should call ListTools or not, it seems a bit heavyweight maybe. Not sure what else we can do though ? Anything interesting on the `Session` object itself that we could put instead ?

wikaaaaa: Do you mean overall or just in the sync case? 

There is get_server_capabilities() which returns the server capabilities received during initialization, but it only provides information if there are any tools to call, nothing more specific. (https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/session.py#L198, https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/types/_types.py#L388)

DylanRussell: I mean overall not sure if calling `ListTools` is a good idea from within instrumentation code.. Lets get someone else's opinion. Doesn't look like there's anything on the session that's interesting to put into the log/span, so I don't see a good alternative other than ignoring it

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -185,6 +188,50 @@ def _to_dict(value: object):
     return json.loads(json.dumps(value))
 
 
+def _to_tool_definition_common(tool: ToolUnionDict) -> MessagePart:
+    if isinstance(tool, dict):
+        return tool
+
```
DylanRussell, line 240: I'm still a bit confused by these if clauses 

We've handled the `dict` and `mcpsession` types by this point. So that leaves the `Callable` and `mcptypes.Tool` types.. Can we not have 2 more if statements for those 2 types ? Why do we check for the __name__ attr ? Why do we try/except on model_dump and then fall back to json serialization ?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -185,6 +188,50 @@ def _to_dict(value: object):
...
+        return {"raw_definition": json.loads(json.dumps(tool))}
+    except Exception:
+        return {
+            "error": f"failed to serialize tool definition, tool type={type(tool).__name__}"
+        }
+
+
+def _to_tool_definition(tool: ToolUnionDict) -> MessagePart:
+    if isinstance(tool, McpClientSession):
+        return {
```
DylanRussell: I don't think we need this `error` because the gen AI SDK itself is going to fail and return an error (see my link in the other comment).. We will propagate that error to the telemetry. I would just ignore / skip this case

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
wikaaaaa: I tried to make it more clear in the newest commit.

So now we have dict and tool case for ToolDict, im not sure exactly how and why but the ToolDict i create in tests ends up being a tool once it reached to_tool_definitions, the transition likely happens somewhere inside google.genai library. We cant do `isinstance(tool, ToolDict)` becuase  TypedDict does not support instance and class checks . 

Then we have a if for callable, mcp tool and mcp session. 

And i left a default path anyway and the exception handling in case a tool type is added and we miss it.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
wikaaaaa: Done, removed the error.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
DylanRussell: I'm ok with putting it behind this flag. resolving

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/nonstreaming_base.py
```diff
@@ -423,6 +488,68 @@ def test_new_semconv_record_completion_as_log(self):
...
+                                        {
+                                            "name": "mock_tool",
+                                            "description": "Description of mock tool.",
+                                        },
+                                    )
+                                },
+                            ),
+                            "TOOL_DEFINITIONS_ASYNC": (
+                                {
+                                    "function": {
```
DylanRussell: Why nest some tools under a key called "function" ? I wonder if we should get rid of that. And maybe we should add a "type" field at the top level with the function type (McpTool, Callable etc..) WDYT ?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
DylanRussell: OK great it's pretty clear to me now.

This LGTM now

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
DylanRussell: I'm ok leaving it in for now.. Resolving

DylanRussell (approved):

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/generate_content/nonstreaming_base.py
wikaaaaa: Deleted the unecessary function key. 

I think we should skip the top level function types for now, I dont see a clear impact of adding them, and the tool definitions can be large in size as is, so I would leave it.

instrumentation-genai/opentelemetry-instrumentation-google-genai/pyproject.toml
```diff
@@ -41,6 +41,7 @@ dependencies = [
   "opentelemetry-instrumentation >=0.58b0, <2",
   "opentelemetry-semantic-conventions >=0.58b0, <2",
   "opentelemetry-util-genai >= 0.2b0, <0.3b0",
+  "mcp; python_version >= '3.10'",
```
aabmass: Why is it needed here and why only for 3.10+? Since it's an optional, import I think we can remove this.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -82,8 +86,29 @@
...
 
+_is_mcp_imported = False
+if typing.TYPE_CHECKING:
+    from mcp import ClientSession as McpClientSession
+    from mcp import Tool as McpTool
+
+    is_mcp_imported = True
+else:
+    McpClientSession: typing.Type = Any
+    McpTool: typing.Type = Any
```
aabmass: `typing.Type` is deprecated https://docs.python.org/3/library/typing.html#typing.Type but I also don't think it's right here or Any. Does it work if you just remove these two lines?

aabmass (approved):
Thanks for the PR! FYI this attribute will be more "typed" soon: https://github.com/open-telemetry/semantic-conventions/pull/3378

Not blocking.

instrumentation-genai/opentelemetry-instrumentation-google-genai/pyproject.toml
wikaaaaa: Only for 3.10+ because all versions of mcp depend on pyhton >= 3.10. Thats the reason the import is optional, to support older python versions. So I added the depenedency here since I didnt want mcp to be optional with python 3.10+. 

Should it be completely optional, even for newer python versions?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
wikaaaaa: It works without those lines, thank you for catching that. I removed them in the newest commit.

instrumentation-genai/opentelemetry-instrumentation-google-genai/pyproject.toml
aabmass: Is it an optional dependency in google-genai? If yes the instrumentation should also be optional.

wikaaaaa: Done. Removed the mcp import from pypoject

PR #4155, opened 2026-02-01
feat: Add sync streaming support for Anthropic instrumentation

### LINKED ISSUE #3949: Add OpenTelemetry instrumentation for the Anthropic Claude Python SDK

Add OpenTelemetry instrumentation for the Anthropic Claude Python SDK (anthropic package) to support tracing of Messages API calls, following the same pattern as the existing OpenAI instrumentation.

Create a new instrumentation package opentelemetry-instrumentation-anthropic (or similar naming) that:

1. Instruments the Anthropic Messages API (anthropic.messages.create())
2. Captures request/response attributes following GenAI semantic conventions
3. Emits log events for messages (with opt-in content capture)
4. Records metrics for operation duration and token usage
5. Supports both sync and async operations
6. Handles streaming responses

_No response_

### Additional Context

1. Follow the same structure as opentelemetry-instrumentation-openai-v2
2. Use gen_ai.system = "anthropic"
3. Map Anthropic's stop_reason (single value) to gen_ai.response.finish_reasons (array format)
4. Handle Anthropic's content block structure (array of content blocks vs OpenAI's string/array)
5. Support opt-in content capture via OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT

Yes

### References

1. https://docs.claude.com/en/api/messages

vasantteja:
# Description

This PR adds sync streaming support for the Anthropic instrumentation. It enables telemetry capture for:

1. **`Messages.create(stream=True)`** - Streaming responses via the create method with stream parameter
2. **`Messages.stream()`** - The dedicated streaming method that returns a `MessageStreamManager`

Key changes:
- Added `StreamWrapper` class to wrap `Stream[RawMessageStreamEvent]` and extract telemetry from streaming chunks
- Added `MessageStreamManagerWrapper` to wrap `MessageStreamManager` context manager
- Added `MessageWrapper` for non-streaming response telemetry extraction
- Renamed `MessageCreateParams` to `MessageRequestParams` to reflect broader API coverage
- Modified `messages_create` to use manual lifecycle management (`start_llm`/`stop_llm`) instead of context manager to support both streaming and non-streaming

Fixes #3949 partially.

## Type of change

# How Has This Been Tested?

Added comprehensive tests for sync streaming functionality:

- [x] `test_sync_messages_create_streaming` - Tests streaming with context manager
- [x] `test_sync_messages_create_streaming_iteration` - Tests direct iteration without context manager
- [x] `test_sync_messages_create_streaming_connection_error` - Tests error handling for streaming
- [x] `test_sync_messages_stream_basic` - Tests `Messages.stream()` method
- [x] `test_sync_messages_stream_with_params` - Tests stream with additional parameters (temperature, top_p, top_k)
- [x] `test_sync_messages_stream_token_usage` - Tests token usage capture in streaming
- [x] `test_sync_messages_stream_connection_error` - Tests error handling for stream method

All tests use VCR cassettes for reproducible HTTP interaction replay.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/.github/workflows/core_contrib_test_0.yml
+++ b/.github/workflows/core_contrib_test_0.yml
@@ -20,6 +20,10 @@ env:
   CORE_REPO_SHA: ${{ inputs.CORE_REPO_SHA }}
   CONTRIB_REPO_SHA: ${{ inputs.CONTRIB_REPO_SHA }}
   PIP_EXISTS_ACTION: w
+  CORE_REPO_API: ${{ github.workspace }}/opentelemetry-python/opentelemetry-api
+  CORE_REPO_SDK: ${{ github.workspace }}/opentelemetry-python/opentelemetry-sdk
+  CORE_REPO_SEMCONV: ${{ github.workspace }}/opentelemetry-python/opentelemetry-semantic-conventions
+  CORE_REPO_TEST_UTILS: ${{ github.workspace }}/opentelemetry-python/tests/opentelemetry-test-utils
 
 jobs:
 
--- a/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/core_contrib_test.yml.j2
+++ b/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/core_contrib_test.yml.j2
@@ -20,6 +20,10 @@ env:
   CORE_REPO_SHA: ${% raw %}{{ inputs.CORE_REPO_SHA }}{% endraw %}
   CONTRIB_REPO_SHA: ${% raw %}{{ inputs.CONTRIB_REPO_SHA }}{% endraw %}
   PIP_EXISTS_ACTION: w
+  CORE_REPO_API: ${% raw %}{{ github.workspace }}{% endraw %}/opentelemetry-python/opentelemetry-api
+  CORE_REPO_SDK: ${% raw %}{{ github.workspace }}{% endraw %}/opentelemetry-python/opentelemetry-sdk
+  CORE_REPO_SEMCONV: ${% raw %}{{ github.workspace }}{% endraw %}/opentelemetry-python/opentelemetry-semantic-conventions
+  CORE_REPO_TEST_UTILS: ${% raw %}{{ github.workspace }}{% endraw %}/opentelemetry-python/tests/opentelemetry-test-utils
 
 jobs:
   {%- for job_data in job_datas %}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- Fix intermittent `Core Contrib Test` CI failures caused by GitHub git CDN SHA propagation lag by installing core packages from the already-checked-out local copy instead of a second git clone
+  ([#4305](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4305))
+
 ## Version 1.40.0/0.61b0 (2026-03-04)
 
 ### Added
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/CHANGELOG.md
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Align AgentSpanData test stubs and span processor with real OpenAI Agents SDK;
+  remove non-existent `operation`, `description`, `agent_id`, and `model` fields.
+  ([#4229](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4229))
 - Document official package metadata and README for the OpenAI Agents instrumentation.
   ([#3859](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3859))
 
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
@@ -1519,17 +1519,8 @@ class GenAISemanticProcessor(TracingProcessor):
                     return GenAIOperationName.CHAT
             return GenAIOperationName.TEXT_COMPLETION
         if _is_instance_of(span_data, AgentSpanData):
-            # Could be create_agent or invoke_agent based on context
-            operation = getattr(span_data, "operation", None)
-            normalized = (
-                operation.strip().lower()
-                if isinstance(operation, str)
-                else None
-            )
-            if normalized in {"create", "create_agent"}:
-                return GenAIOperationName.CREATE_AGENT
-            if normalized in {"invoke", "invoke_agent"}:
-                return GenAIOperationName.INVOKE_AGENT
+            # The OpenAI Agents SDK AgentSpanData has no "operation" field;
+            # agent spans always represent invoke_agent.
             return GenAIOperationName.INVOKE_AGENT
         if _is_instance_of(span_data, FunctionSpanData):
             return GenAIOperationName.EXECUTE_TOOL
@@ -1831,24 +1822,20 @@ class GenAISemanticProcessor(TracingProcessor):
         if name:
             yield GEN_AI_AGENT_NAME, name
 
-        agent_id = (
-            self.agent_id
-            or getattr(span_data, "agent_id", None)
-            or self._agent_id_default
-        )
+        # agent_id and description are not available on the OpenAI Agents SDK
+        # AgentSpanData; only use user-configured overrides.
+        agent_id = self.agent_id or self._agent_id_default
         if agent_id:
             yield GEN_AI_AGENT_ID, agent_id
 
-        description = (
-            self.agent_description
-            or getattr(span_data, "description", None)
-            or self._agent_description_default
-        )
+        description = self.agent_description or self._agent_description_default
         if description:
             yield GEN_AI_AGENT_DESCRIPTION, description
 
-        model = getattr(span_data, "model", None)
-        if not model and agent_content:
+        # The OpenAI Agents SDK AgentSpanData has no "model" field; fall back to
+        # the model aggregated from child generation/response spans.
+        model = None
+        if agent_content:
             model = agent_content.get("request_model")
         if model:
             yield GEN_AI_REQUEST_MODEL, model
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/tests/stubs/agents/tracing/__init__.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/tests/stubs/agents/tracing/__init__.py
@@ -35,12 +35,9 @@ __all__ = [
 @dataclass
 class AgentSpanData:
     name: str | None = None
+    handoffs: list[str] | None = None
     tools: list[str] | None = None
     output_type: str | None = None
-    description: str | None = None
-    agent_id: str | None = None
-    model: str | None = None
-    operation: str | None = None
 
     @property
     def type(self) -> str:
@@ -200,8 +197,16 @@ def generation_span(**kwargs: Any):
 
 
 @contextmanager
-def agent_span(**kwargs: Any):
-    data = AgentSpanData(**kwargs)
+def agent_span(
+    name: str,
+    handoffs: list[str] | None = None,
+    tools: list[str] | None = None,
+    output_type: str | None = None,
+    **kwargs: Any,
+):
+    data = AgentSpanData(
+        name=name, handoffs=handoffs, tools=tools, output_type=output_type
+    )
     span = _PROVIDER.create_span(data, parent=_CURRENT_TRACE)
     span.start()
     try:
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/tests/test_tracer.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/tests/test_tracer.py
@@ -171,40 +171,31 @@ def test_function_span_records_tool_attributes():
         exporter.clear()
 
 
-def test_agent_create_span_records_attributes():
+def test_agent_invoke_span_records_attributes():
     instrumentor, exporter = _instrument_with_provider()
 
     try:
         with trace("workflow"):
             with agent_span(
-                operation="create",
                 name="support_bot",
-                description="Answers support questions",
-                agent_id="agt_123",
-                model="gpt-4o-mini",
+                handoffs=["escalation_bot"],
+                tools=["search"],
+                output_type="str",
             ):
                 pass
 
         spans = exporter.get_finished_spans()
-        create_span = next(
+        invoke_span = next(
             span
             for span in spans
             if span.attributes[GenAI.GEN_AI_OPERATION_NAME]
-            == GenAI.GenAiOperationNameValues.CREATE_AGENT.value
+            == GenAI.GenAiOperationNameValues.INVOKE_AGENT.value
         )
 
-        assert create_span.kind is SpanKind.CLIENT
-        assert create_span.name == "create_agent support_bot"
-        assert create_span.attributes[GEN_AI_PROVIDER_NAME] == "openai"
-        assert create_span.attributes[GenAI.GEN_AI_AGENT_NAME] == "support_bot"
-        assert (
-            create_span.attributes[GenAI.GEN_AI_AGENT_DESCRIPTION]
-            == "Answers support questions"
-        )
-        assert create_span.attributes[GenAI.GEN_AI_AGENT_ID] == "agt_123"
-        assert (
-            create_span.attributes[GenAI.GEN_AI_REQUEST_MODEL] == "gpt-4o-mini"
-        )
+        assert invoke_span.kind is SpanKind.CLIENT
+        assert invoke_span.name == "invoke_agent support_bot"
+        assert invoke_span.attributes[GEN_AI_PROVIDER_NAME] == "openai"
+        assert invoke_span.attributes[GenAI.GEN_AI_AGENT_NAME] == "support_bot"
     finally:
         instrumentor.uninstrument()
         exporter.clear()
@@ -425,7 +416,7 @@ def test_agent_name_override_applied_to_agent_spans():
 
     try:
         with trace("workflow"):
-            with agent_span(operation="invoke", name="support_bot"):
+            with agent_span(name="support_bot"):
                 pass
 
         spans = exporter.get_finished_spans()
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/tests/test_z_span_processor_unit.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/tests/test_z_span_processor_unit.py
@@ -156,19 +156,14 @@ def test_operation_and_span_naming(processor_setup):
         == sp.GenAIOperationName.EMBEDDINGS
     )
 
-    agent_create = AgentSpanData(operation=" CREATE ")
+    # AgentSpanData always maps to invoke_agent (no operation field in OpenAI Agents SDK)
+    agent_data = AgentSpanData(name="bot")
     assert (
-        processor._get_operation_name(agent_create)
-        == sp.GenAIOperationName.CREATE_AGENT
-    )
-
-    agent_invoke = AgentSpanData(operation="invoke_agent")
-    assert (
-        processor._get_operation_name(agent_invoke)
+        processor._get_operation_name(agent_data)
         == sp.GenAIOperationName.INVOKE_AGENT
     )
 
-    agent_default = AgentSpanData(operation=None)
+    agent_default = AgentSpanData()
     assert (
         processor._get_operation_name(agent_default)
         == sp.GenAIOperationName.INVOKE_AGENT
@@ -315,26 +310,19 @@ def test_attribute_builders(processor_setup):
     agent_span = AgentSpanData(
         name="helper",
         output_type="json",
-        description="desc",
-        agent_id="agent-123",
-        model="model-x",
-        operation="invoke_agent",
     )
     agent_attrs = _collect(
         processor._get_attributes_from_agent_span_data(agent_span, None)
     )
     assert agent_attrs[sp.GEN_AI_AGENT_NAME] == "helper"
-    assert agent_attrs[sp.GEN_AI_AGENT_ID] == "agent-123"
-    assert agent_attrs[sp.GEN_AI_REQUEST_MODEL] == "model-x"
+    assert sp.GEN_AI_AGENT_ID not in agent_attrs
+    assert sp.GEN_AI_REQUEST_MODEL not in agent_attrs
     assert agent_attrs[sp.GEN_AI_OUTPUT_TYPE] == sp.GenAIOutputType.TEXT
 
     # Fallback to aggregated model when span data lacks it
     agent_span_no_model = AgentSpanData(
         name="helper-2",
         output_type="json",
-        description="desc",
-        agent_id="agent-456",
-        operation="invoke_agent",
     )
     agent_content = {
         "input_messages": [],
@@ -435,9 +423,7 @@ def test_span_lifecycle_and_shutdown(processor_setup):
     parent_span = FakeSpan(
         trace_id="trace-1",
         span_id="span-1",
-        span_data=AgentSpanData(
-            operation="invoke", name="agent", model="gpt-4o"
-        ),
+        span_data=AgentSpanData(name="agent"),
         started_at="2024-01-01T00:00:00Z",
         ended_at="2024-01-01T00:00:02Z",
     )
@@ -476,7 +462,7 @@ def test_span_lifecycle_and_shutdown(processor_setup):
     linger_span = FakeSpan(
         trace_id="trace-2",
         span_id="span-3",
-        span_data=AgentSpanData(operation=None),
+        span_data=AgentSpanData(),
         started_at="2024-01-01T00:00:06Z",
     )
     processor.on_span_start(linger_span)
@@ -518,7 +504,6 @@ def test_chat_span_renamed_with_model(processor_setup):
         trace_id=trace.trace_id,
         span_id="agent-span",
         span_data=AgentSpanData(
-            operation="invoke_agent",
             name="Agent",
         ),
         started_at="2025-01-01T00:00:00Z",
--- a/tests/opentelemetry-docker-tests/tests/test-requirements.txt
+++ b/tests/opentelemetry-docker-tests/tests/test-requirements.txt
@@ -4,18 +4,17 @@ asgiref==3.8.1
 async-timeout==4.0.3
 asyncpg==0.29.0
 attrs==23.2.0
-bcrypt==4.1.2
+bcrypt==5.0.0
 billiard==4.2.0
 celery==5.3.6
 certifi==2024.7.4
-cffi==1.16.0
+cffi==2.0.0
 chardet==3.0.4
 click==8.1.7
 click-didyoumean==0.3.0
 click-plugins==1.1.1
 click-repl==0.3.0
-cryptography==44.0.1
-Deprecated==1.2.14
+cryptography==45.0.7
 distro==1.9.0
 Django==4.2.17
 dnspython==2.6.1
@@ -34,23 +33,21 @@ jsonschema==3.2.0
 kombu==5.3.5
 mysql-connector-python==8.3.0
 mysqlclient==2.1.1
-opencensus-proto==0.1.0
 packaging==24.0
-paramiko==3.4.0
+paramiko==4.0.0
 pluggy==1.4.0
 prometheus_client==0.20.0
 prompt-toolkit==3.0.43
-protobuf==3.20.3
 # prerequisite: install libpq-dev (debian) or postgresql-devel (rhel), postgresql (mac)
 # see https://www.psycopg.org/docs/install.html#build-prerequisites
 # you might have to install additional packages depending on your OS
 psycopg==3.1.18
 psycopg2==2.9.9
 psycopg2-binary==2.9.9
-pycparser==2.21
+pycparser==3.0
 pymongo==4.6.3
 PyMySQL==0.10.1
-PyNaCl==1.5.0
+PyNaCl==1.6.2
 # prerequisite: install unixodbc
 pyodbc==5.0.1
 pyrsistent==0.20.0
@@ -61,7 +58,6 @@ python-dotenv==0.21.1
 pytz==2024.1
 PyYAML==5.3.1
 redis==5.0.1
-remoulade==3.2.0
 requests==2.25.0
 six==1.16.0
 SQLAlchemy==1.4.52
--- a/tox.ini
+++ b/tox.ini
@@ -464,10 +464,10 @@ envlist =
 
 [testenv]
 test_deps =
-  opentelemetry-api@{env:CORE_REPO}\#egg=opentelemetry-api&subdirectory=opentelemetry-api
-  opentelemetry-semantic-conventions@{env:CORE_REPO}\#egg=opentelemetry-semantic-conventions&subdirectory=opentelemetry-semantic-conventions
-  opentelemetry-sdk@{env:CORE_REPO}\#egg=opentelemetry-sdk&subdirectory=opentelemetry-sdk
-  opentelemetry-test-utils@{env:CORE_REPO}\#egg=opentelemetry-test-utils&subdirectory=tests/opentelemetry-test-utils
+  opentelemetry-api@{env:CORE_REPO_API}
+  opentelemetry-semantic-conventions@{env:CORE_REPO_SEMCONV}
+  opentelemetry-sdk@{env:CORE_REPO_SDK}
+  opentelemetry-test-utils@{env:CORE_REPO_TEST_UTILS}
 deps =
   lint: -r dev-requirements.txt
   coverage: pytest
@@ -790,6 +790,10 @@ setenv =
   ; i.e: CORE_REPO_SHA=dde62cebffe519c35875af6d06fae053b3be65ec tox -e <env to test>
   CORE_REPO_SHA={env:CORE_REPO_SHA:main}
   CORE_REPO=git+https://github.com/open-telemetry/opentelemetry-python.git@{env:CORE_REPO_SHA}
+  CORE_REPO_API={env:CORE_REPO_API:{env:CORE_REPO}\#egg=opentelemetry-api&subdirectory=opentelemetry-api}
+  CORE_REPO_SDK={env:CORE_REPO_SDK:{env:CORE_REPO}\#egg=opentelemetry-sdk&subdirectory=opentelemetry-sdk}
+  CORE_REPO_SEMCONV={env:CORE_REPO_SEMCONV:{env:CORE_REPO}\#egg=opentelemetry-semantic-conventions&subdirectory=opentelemetry-semantic-conventions}
+  CORE_REPO_TEST_UTILS={env:CORE_REPO_TEST_UTILS:{env:CORE_REPO}\#egg=opentelemetry-test-utils&subdirectory=tests/opentelemetry-test-utils}
   UV_CONFIG_FILE={toxinidir}/tox-uv.toml
   PIP_CONSTRAINTS={toxinidir}/test-constraints.txt
   UV_BUILD_CONSTRAINT={toxinidir}/test-constraints.txt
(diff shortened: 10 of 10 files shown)


instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
```diff
@@ -57,22 +64,60 @@ def traced_method(
...
+        instance: "Messages",
+        args: tuple[Any, ...],
+        kwargs: dict[str, Any],
+    ) -> MessageStreamManagerWrapper:
+        params = extract_params(*args, **kwargs)
+        attributes = get_llm_request_attributes(params, instance)
+        request_model = str(
+            attributes.get(GenAIAttributes.GEN_AI_REQUEST_MODEL)
+            or params.model
+            or "unknown"
```
lmolkova: I don't think it's necessary - if it's unknown, let's not populate it

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
```diff
@@ -57,22 +64,60 @@ def traced_method(
...
+        )
 
-            return result
+        invocation = LLMInvocation(
+            request_model=request_model,
+            provider="anthropic",
```
lmolkova: we should have a constant for it  

https://github.com/open-telemetry/opentelemetry-python/blob/b4c13279187941b561de5d62e6483a91759ba123/opentelemetry-semantic-conventions/src/opentelemetry/semconv/_incubating/attributes/gen_ai_attributes.py#L442

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
```diff
@@ -143,3 +152,219 @@ def get_llm_request_attributes(
...
+            invocation.response_model_name = self._message.model
+
+        if self._message.id:
+            invocation.response_id = self._message.id
+
+        if self._message.stop_reason:
+            invocation.finish_reasons = [self._message.stop_reason]
+
+        if self._message.usage:
+            invocation.input_tokens = self._message.usage.input_tokens
```
lmolkova: is it aligned with https://github.com/open-telemetry/semantic-conventions/blob/573ec5f14bc2713a75f352f3fce4d7e50c942937/docs/gen-ai/anthropic.md?plain=1#L130 ?

lmolkova:
tagging @anirudha who was interested to review the PR :)

anirudha:
Thanks. Taking a look today

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
vasantteja: @lmolkova Extremely sorry, these span attributes were not aligned as per the doc you shared. I was creating span attributes roughly based on this [doc](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/) which is believe is not appropriate for anthropic instrumentation. Thanks for catching this Liudmila.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
```diff
@@ -143,3 +220,274 @@ def get_llm_request_attributes(
...
+
+class StreamWrapper(Iterator["RawMessageStreamEvent"]):
+    """Wrapper for Anthropic Stream that handles telemetry.
+
+    This wrapper wraps Stream[RawMessageStreamEvent] returned by
+    Messages.create(stream=True). It processes streaming chunks,
+    extracts telemetry data, and ensures the span is properly ended
+    when the stream is consumed.
+    """
+
```
anirudha: If someone does `with stream: for chunk in stream`, the iterator hitting `StopIteration` calls `_finalize_invocation()` at L270, then `__exit__` → `close()` calls it again here. That's two `span.end()` + two `otel_context.detach()` calls.

A simple `self._finalized` bool guard would fix it:
```python
def _finalize_invocation(self) -> None:
    if self._finalized:
        return
    self._finalized = True
    ...
```

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
```diff
@@ -143,3 +220,274 @@ def get_llm_request_attributes(
...
+            self._input_tokens = input_tokens
+        if output_tokens is not None:
+            self._output_tokens = output_tokens
+        if cache_creation_input_tokens is not None:
+            self._cache_creation_input_tokens = cache_creation_input_tokens
+        if cache_read_input_tokens is not None:
+            self._cache_read_input_tokens = cache_read_input_tokens
+
+    def _process_chunk(self, chunk: "RawMessageStreamEvent") -> None:
+        """Extract telemetry data from a streaming chunk."""
```
anirudha: Same double-call risk here — no guard against `__exit__` firing more than once. Less likely to hit in practice but same pattern, same fix.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
```diff
@@ -143,3 +220,274 @@ def get_llm_request_attributes(
...
+            self._invocation.output_tokens = self._output_tokens
+        if self._cache_creation_input_tokens is not None:
+            self._invocation.attributes[
+                _GEN_AI_USAGE_CACHE_CREATION_INPUT_TOKENS
+            ] = self._cache_creation_input_tokens
+        if self._cache_read_input_tokens is not None:
+            self._invocation.attributes[
+                _GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS
+            ] = self._cache_read_input_tokens
+
```
anirudha: If `get_final_message()` fails because the stream wasn't fully consumed, this silently swallows it and `stop_llm` still runs with empty response attrs. Nobody will know why their spans have no response data. A `logger.debug` or `logger.warning` here would save someone a debugging headache.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
```diff
@@ -14,65 +14,119 @@
...
             request_model=request_model,
-            provider="anthropic",
+            provider=ANTHROPIC,
             attributes=attributes,
         )
 
-        with handler.llm(invocation) as invocation:
+        is_streaming = kwargs.get("stream", False)
+
+        # Use manual lifecycle management for both streaming and non-streaming
```
anirudha, line 104: Minor style thing — `MessageWrapper.__init__` calls `stop_llm` internally, so the span ends inside a constructor. The OpenAI instrumentation does this inline in the patch function instead. Not blocking, just a readability note — might be clearer to extract+finalize here at the call site rather than hiding it in the constructor.

anirudha (commented):
Tests all pass locally. Nice work overall — the wrapper separation is clean. One bug to fix (double finalize), rest are suggestions.

Note: `conftest.py` isn't in this diff so I can't leave a line comment, but `scrub_response_headers` is a no-op and all new cassettes leak `anthropic-organization-id: 455ea6be-bd92-4199-83ec-0c6b39c5c169`. Worth scrubbing that or adding it to `filter_headers`.

Also, the PR description says `Fixes #3949` but async streaming isn't covered. Totally fine to scope this to sync only, but `Fixes` will auto-close the issue on merge. Maybe `Partially addresses #3949` instead?

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
vasantteja: Makes sense. Let me refactor it. I will ping you once it's done.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
vasantteja: Makes sense. Let me refactor it. I will ping you once it's done.

anirudha:
looks good, thanks @vasantteja !

instrumentation-genai/opentelemetry-instrumentation-anthropic/tests/conftest.py
```diff
@@ -128,14 +136,19 @@ def instrument_no_content(tracer_provider, logger_provider, meter_provider):
...
 
 @pytest.fixture(scope="function")
 def instrument_with_content(tracer_provider, logger_provider, meter_provider):
     """Instrument Anthropic with content capture enabled."""
+    _OpenTelemetrySemanticConventionStability._initialized = False
     os.environ.update(
-        {OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT: "True"}
+        {
+            OTEL_SEMCONV_STABILITY_OPT_IN: "gen_ai_latest_experimental",
+            OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT: "SPAN_ONLY",
```
Cirilla-zmh, line 156: Maybe you should add some tests for `EVENT_ONLY`? I believe we don’t need exhaustive tests for `EVENT_ONLY` since it's well-covered in `SPAN_ONLY`. But let's add at least one test for basic verification and regression purposes.

vasantteja: @Cirilla-zmh added the test, please take a look when you get time.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
```diff
@@ -14,65 +14,138 @@
...
-            attributes.get(GenAIAttributes.GEN_AI_REQUEST_MODEL)
-            or params.model
-            or "unknown"
+        request_model_attribute = attributes.get(
+            GenAIAttributes.GEN_AI_REQUEST_MODEL
+        )
+        request_model = (
+            request_model_attribute
+            if isinstance(request_model_attribute, str)
+            else params.model or ""
```
lmolkova: no need to have an empty string, we can keep it unset if model is not provided

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
```diff
@@ -26,26 +29,328 @@
...
+    return (
+        total_input_tokens,
+        output_tokens,
+        cache_creation_input_tokens,
+        cache_read_input_tokens,
+    )
+
+
+def should_capture_content() -> bool:
```
lmolkova: nit: would it make sense to add it to utils? I think it's ok to do as a follow up

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
```diff
@@ -143,3 +450,315 @@ def get_llm_request_attributes(
...
+            invocation.finish_reasons = [finish_reason]
+
+        if self._message.usage:
+            (
+                input_tokens,
+                output_tokens,
+                cache_creation_input_tokens,
+                cache_read_input_tokens,
+            ) = _extract_usage_tokens(self._message.usage)
+            invocation.input_tokens = input_tokens
```
lmolkova: it seems in case of anthropic, input tokens exclude in response cached tokens and we need to sum them to get total input tokens https://github.com/trask/semantic-conventions/blob/9ab7f8401c9b1fe0bf6c59be5efb40c31189b3d5/docs/gen-ai/anthropic.md?plain=1#L130

If it's not the case, could you please send a PR to update semconv?

instrumentation-genai/opentelemetry-instrumentation-anthropic/tests/test_sync_messages.py
```diff
@@ -300,3 +387,610 @@ def test_multiple_instrument_uninstrument_cycles(
...
+    assert_span_attributes(
+        spans[0],
+        request_model=model,
+        response_id=response_id,
+        response_model=response_model,
+        input_tokens=input_tokens,
+        output_tokens=output_tokens,
+        finish_reasons=[normalize_stop_reason(stop_reason)]
+        if stop_reason
+        else None,
```
lmolkova, line 449: If we don't have a finish reason, it should be "error" - this way we know that response was not finished and it indicates some form of error

lmolkova (commented):
Looks great, left a few minor comments

instrumentation-genai/opentelemetry-instrumentation-anthropic/tests/test_sync_messages.py
lmolkova: let's ignore it for now - turns out we don't have it documented in semconv and there is an active discussion https://github.com/open-telemetry/semantic-conventions/issues/2919

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
vasantteja: @lmolkova Yeah you are right anthropic's input tokens don't include cache_creation_input_tokens and cache_read_input_tokens hence we have to sum them up for getting the total input tokens.

Reference: https://platform.claude.com/docs/en/build-with-claude/prompt-caching#tracking-cache-performance

I believe semconv is right with input tokens field calculation. Can I resolve this if you're okay?

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
vasantteja: Yeah sure let me add it.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
lmolkova: I now see you're calculating the sum. Thanks for checking!

instrumentation-genai/opentelemetry-instrumentation-anthropic/tests/test_sync_messages.py
```diff
@@ -86,6 +122,27 @@ def assert_span_attributes(
...
+def _skip_if_cassette_missing_and_no_real_key(request):
+    cassette_path = (
+        Path(__file__).parent / "cassettes" / f"{request.node.name}.yaml"
+    )
+    api_key = os.getenv("ANTHROPIC_API_KEY")
+    if not cassette_path.exists() and api_key == "test_anthropic_api_key":
+        pytest.skip(
+            f"Cassette {cassette_path.name} is missing. "
+            "Set a real ANTHROPIC_API_KEY to record it."
+        )
```
aabmass, line 144: Why would the cassette be missing?

instrumentation-genai/opentelemetry-instrumentation-anthropic/tests/cassettes/test_sync_messages_create_stop_reason.yaml
aabmass, line 1: Why are there several new requests in here? It doesn't look like `test_sync_messages_create_stop_reason` was changed to make more request

In case you haven't seen it, you can pass [VCR record mode](https://vcrpy.readthedocs.io/en/latest/usage.html#record-modes) with `--vcr-record=<mode>`

instrumentation-genai/opentelemetry-instrumentation-anthropic/tests/test_sync_messages.py
vasantteja: The _skip_if_cassette_missing_and_no_real_key guard is called only by the tool_use and thinking tests — whose cassettes may not exist if the SDK version is too old to record them. It prevents a confusing failure (dummy key hitting the real API) by cleanly skipping the test with a message to set a real key. Lmk if you want to remove this.

instrumentation-genai/opentelemetry-instrumentation-anthropic/tests/cassettes/test_sync_messages_create_stop_reason.yaml
vasantteja: Honeslty thanks for this. I was using this `pytest instrumentation-genai/opentelemetry-instrumentation-anthropic/tests --vcr-record=all` and it was appending my requests. I was rerunning this everytime I make a change just to ensure everything is good. This bulked up the cassettes. I deleted the cassettes and recreated them.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/wrappers.py
```diff
@@ -0,0 +1,332 @@
...
+        self.close()
+        return False
+
+    def close(self) -> None:
+        if hasattr(self._stream, "close"):
+            self._stream.close()
+        self._finalize_invocation()
+
+
+class MessageStreamManagerWrapper:
```
MikeGoldsmith: I don't see where this is used - `Messages.stream()` is not wired up like `StreamWrapper` is.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
```diff
@@ -14,65 +14,88 @@
...
-from .utils import (
+from .messages_extractors import (
     extract_params,
+    get_input_messages,
     get_llm_request_attributes,
+    get_system_instruction,
+)
+from .wrappers import (
+    MessageWrapper,
+    StreamWrapper,
```
MikeGoldsmith: I think we're missing `MessageStreamManagerWrapper` here.

instrumentation-genai/opentelemetry-instrumentation-anthropic/tests/conftest.py
```diff
@@ -115,8 +120,12 @@ def vcr_config():
 @pytest.fixture(scope="function")
 def instrument_no_content(tracer_provider, logger_provider, meter_provider):
     """Instrument Anthropic without content capture."""
+    _OpenTelemetrySemanticConventionStability._initialized = False
```
MikeGoldsmith, line 123: This should be in a finally block otherwise it can leak into other tests.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/wrappers.py
```diff
@@ -0,0 +1,332 @@
...
+    from anthropic.lib.streaming import (
+        MessageStream,
+        MessageStreamManager,
+    )
+    from anthropic.types import Message, RawMessageStreamEvent
+
+_logger = logging.getLogger(__name__)
+
+
+class MessageWrapper:
```
MikeGoldsmith, line 56: Could this be a function instead of a class?

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/wrappers.py
```diff
@@ -0,0 +1,332 @@
...
+        self._handler.stop_llm(self._invocation)
+
+    def _finalize_error(self, exc_type: Any, exc_val: Any) -> None:
+        if self._finalized:
+            return
+        self._finalized = True
+        self._handler.fail_llm(
+            self._invocation,
+            Error(
+                message=str(exc_val) if exc_val else str(exc_type),
```
MikeGoldsmith: I think this should be `exc_type.__name__` instead.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
```diff
@@ -14,65 +14,88 @@
...
-
-            return result
-
-    return traced_method
+        # Use manual lifecycle management for both streaming and non-streaming
+        handler.start_llm(invocation)
+        try:
+            result = wrapped(*args, **kwargs)
+            if is_streaming:
+                return StreamWrapper(result, handler, invocation)  # type: ignore[arg-type]
```
MikeGoldsmith: Can we do anything to remove the `# type: ignore`s in this func?

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
```diff
@@ -14,65 +14,88 @@
...
+        request_model_attribute = attributes.get(
+            GenAIAttributes.GEN_AI_REQUEST_MODEL
+        )
+        request_model = (
+            request_model_attribute
+            if isinstance(request_model_attribute, str)
+            else params.model
         )
 
+        capture_content = should_capture_content()
```
MikeGoldsmith: I think this means we're calling `should_capture_content()` on every request, right? Can we move this `instrument()`?

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/wrappers.py
```diff
@@ -0,0 +1,332 @@
...
+                self._message
+            )
+
+    @property
+    def message(self) -> "Message":
+        """Return the wrapped Message object."""
+        return self._message
+
+
+class StreamWrapper(Iterator["RawMessageStreamEvent"]):
```
MikeGoldsmith: I think we need to add a __getattr__ delegate to this class as the Anthropic Stream contains other things a user might try to use.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/wrappers.py
vasantteja: It could be a function but I redesigned it as a class as I wanted to maintain code parity with all other wrappers.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/wrappers.py
vasantteja: Yep good catch. I forgot to remove it when I was yanking other instrumentation. Right now it's dead code. I am removing it.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
vasantteja: Removed it.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/wrappers.py
vasantteja: Removed it.

MikeGoldsmith (approved):
🚀

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/wrappers.py
```diff
@@ -0,0 +1,235 @@
...
+        except Exception as exc:
+            self._handler.fail_llm(
+                self._invocation, Error(message=str(exc), type=type(exc))
+            )
+            raise
+
+    def __enter__(self) -> "StreamWrapper":
+        return self
+
+    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> bool:
```
nagkumar91: Potential bug: `__next__` can call `fail_llm(...)` on stream errors, but `_finalized` is not set on that path, and `__exit__` always calls `close()` which then calls `_finalize_invocation()` -> `stop_llm(...)`. That can finalize the same invocation twice (fail then stop) and potentially produce incorrect success telemetry after an error. Also, because `__exit__` ignores `exc_type`, exceptions raised by user code inside `with stream:` are currently treated as successful completion.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/wrappers.py
```diff
@@ -0,0 +1,235 @@
...
+        """Return the wrapped Message object."""
+        return self._message
+
+
+class StreamWrapper(Iterator["RawMessageStreamEvent"]):
+    """Wrapper for Anthropic Stream that handles telemetry."""
+
+    def __init__(
+        self,
+        stream: "Stream[RawMessageStreamEvent]",
```
aabmass: You might be able to remove a lot of these types wrapped in strings if you add `from __future__ import annotations` to the file

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/wrappers.py
```diff
@@ -0,0 +1,235 @@
...
+        self._stop_reason: Optional[str] = None
+        self._input_tokens: Optional[int] = None
+        self._output_tokens: Optional[int] = None
+        self._cache_creation_input_tokens: Optional[int] = None
+        self._cache_read_input_tokens: Optional[int] = None
+        self._capture_content = capture_content
+        self._content_blocks: dict[int, dict[str, Any]] = {}
+        self._finalized = False
+
+    def _update_usage(self, usage: Any | None) -> None:
```
aabmass: Does this need to be `Any`?

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/wrappers.py
```diff
@@ -0,0 +1,235 @@
...
+            self._output_tokens = output_tokens
+        if cache_creation_input_tokens is not None:
+            self._cache_creation_input_tokens = cache_creation_input_tokens
+        if cache_read_input_tokens is not None:
+            self._cache_read_input_tokens = cache_read_input_tokens
+
+    def _process_chunk(self, chunk: "RawMessageStreamEvent") -> None:
+        """Extract telemetry data from a streaming chunk."""
+        if chunk.type == "message_start":
+            message = getattr(chunk, "message", None)
```
aabmass: Is there a reason you need getattr through here? Or can you do `chunk.message`

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
```diff
@@ -14,65 +14,95 @@
...
+            else [],
             attributes=attributes,
         )
 
-        with handler.llm(invocation) as invocation:
-            result = wrapped(*args, **kwargs)
-
-            if result.model:
-                invocation.response_model_name = result.model
+        is_streaming = kwargs.get("stream", False)
```
aabmass: Is it possible to make this type safe by moving it into the `extract_params()` function's return value?

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
```diff
@@ -14,65 +14,95 @@
...
-
-            return result
-
-    return traced_method
+        # Use manual lifecycle management for both streaming and non-streaming
+        handler.start_llm(invocation)
+        try:
+            result = wrapped(*args, **kwargs)
+            if is_streaming:
+                stream_result = cast("Stream[RawMessageStreamEvent]", result)
```
aabmass: Generally prefer to do type narrowing rather than cast. Can you do an isinstance check here?

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
```diff
@@ -14,65 +14,95 @@
...
+                return StreamWrapper(
+                    stream_result, handler, invocation, capture_content
+                )
+            message_result = cast("Message", result)
```
aabmass: If you do the type narrowing above, the typechecker will be able to prove that it must be `Message`.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/messages_extractors.py
```diff
@@ -0,0 +1,202 @@
...
+
+    return (
+        total_input_tokens,
+        output_tokens,
+        cache_creation_input_tokens,
+        cache_read_input_tokens,
+    )
+
+
+def get_input_messages(messages: Any) -> list[InputMessage]:
```
aabmass: is it possible to avoid all the `Any` through this file?

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/wrappers.py
vasantteja: Agreed!! This abstraction is little error prone. I rewrote it to mimic openai responses stream wrapper incorporating your other suggestion on that PR.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/wrappers.py
vasantteja: It doesn't need to be `Any`. Removed Any type for this and other places I used `Any` in this file.

instrumentation-genai/opentelemetry-instrumentation-anthropic/tests/test_sync_messages.py
vasantteja: I removed this as we discussed.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
```diff
@@ -111,6 +111,18 @@ def should_emit_event() -> bool:
...
+def should_capture_content() -> bool:
+    """Return True when content conversion should be performed."""
+    if not is_experimental_mode():
+        return False
+    mode = get_content_capturing_mode()
+    if mode == ContentCapturingMode.NO_CONTENT:
+        return False
+    if mode == ContentCapturingMode.EVENT_ONLY and not should_emit_event():
+        return False
+    return True
```
aabmass, line 123: @DylanRussell do we have this util already?

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/messages_extractors.py
```diff
@@ -0,0 +1,228 @@
...
+GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS = "gen_ai.usage.cache_read.input_tokens"
+
+
+def extract_usage_tokens(
+    usage: Usage | MessageDeltaUsage | None,
+) -> tuple[int | None, int | None, int | None, int | None]:
+    if usage is None:
+        return None, None, None, None
+
+    input_tokens = as_int(getattr(usage, "input_tokens", None))
```
aabmass: It looks like `Usage` is typed here. Can we avoid using getattr() since it is not type safe?

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/messages_extractors.py
```diff
@@ -0,0 +1,228 @@
...
+
+GEN_AI_USAGE_CACHE_CREATION_INPUT_TOKENS = (
+    "gen_ai.usage.cache_creation.input_tokens"
+)
+GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS = "gen_ai.usage.cache_read.input_tokens"
+
+
+def extract_usage_tokens(
+    usage: Usage | MessageDeltaUsage | None,
+) -> tuple[int | None, int | None, int | None, int | None]:
```
aabmass: Consider using a quick dataclass for the return type, or please document the order of the tuple

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
aabmass, line 1: I'm finding `getattr()`, `hasattr()`, `cast()`, lots of use of `Any`, and utils like in this file to be a bit of an anti pattern for typed python code. It looks like anthropic SDK is typed though. Can we try to avoid this and "hand hold" the type checker a bit here? I know in some cases it's annoying, but it makes it much easier for maintainers to read and refactor as needed.

In some cases when the code being instrumented is untyped, stuff like this pops up. Please push back if that's the case and it can't be done.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
vasantteja: @aabmass We don't have this in util. This is specifically used for understanding whether we should we do the work of converting messages/content blocks into structured parts. I couldn't see any function which is having same functionality as this one in util.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/messages_extractors.py
vasantteja: Extracted them into a dataclass.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
```diff
@@ -14,65 +14,116 @@
...
-                invocation.output_tokens = result.usage.output_tokens
-
-            return result
-
-    return traced_method
+        # Use manual lifecycle management for both streaming and non-streaming
+        handler.start_llm(invocation)
+        try:
+            result = wrapped(*args, **kwargs)
+            if is_streaming:
```
aabmass: Could we remove this check altogether and just operate based on the instance type?

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
```diff
@@ -14,65 +14,116 @@
...
+                if not isinstance(result, AnthropicStream):
+                    raise TypeError(
```
aabmass: Best not to raise exceptions from instrumentations, and rather to log about the invariant violation and fallback to the un-instrumented path if needed.

Please file a follow up issue to address this, I don't want to block this PR anymore

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
```diff
@@ -12,134 +12,230 @@
...
+        mime_type=media_type if isinstance(media_type, str) else None,
+        modality=modality,
+        content=decoded,
     )
 
 
-def set_server_address_and_port(
-    client_instance: "Messages", attributes: dict[str, Any]
+def _convert_dict_block_to_part(
+    block: Mapping[str, object],
```
aabmass: Would be more appropriate to use `Any` instead of `object` here. It looks a bit unavoidable

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
DylanRussell: So this function is returning True when content captured mode is experimental and includes SPAN, and False otherwise.. Is that the intent ? If so it seems we should rename it to `should_capture_content_on_spans_in_experimental_mode` ?

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/utils.py
vasantteja: can I keep it as `object` instead of `Any`? I read `object` is more defensive and it acts as a guardrail against future unsafe access.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
vasantteja: I removed it.

instrumentation-genai/opentelemetry-instrumentation-anthropic/src/opentelemetry/instrumentation/anthropic/patch.py
vasantteja: We are throwing exceptions only in two places. I removed them as it's a minor refactor.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py
vasantteja: I renamed the function as suggested.

PR #4158, opened 2026-02-01
feat: add ability to configure flush behavior for AWS lambda instrumentation

### LINKED ISSUE #4157: Provide alternate strategies for exporting metrics with AWS Lambda

The current AWS Lambda instrumentation library flushes all metrics after each invocation, which results in metric data-points being generated after each invocation, regardless of the configured export interval. This issue becomes increasingly problematic for Lambda functions with short execution durations (e.g. ~100ms) since even datapoints with the same attributes, will not have the ability to be aggregated if they span more than a single invocation. In addition to being operationally inefficient, because almost all observability vendors charge based on the number of datapoints ingested, this pattern results in dramatic increases in costs.  

There are several changes which can be employed to help reduce the number of metric datapoints being generated. The two initial changes are as follows:

1. Provide a configuration option/environment variable to disable force flushing metrics after each invocation. 
2. Provide a configuration option/environment variable to register a signal handler for the `SIGTERM` shutdown event to flush all metrics before the Lambda instance is cleaned up. (see https://aws.amazon.com/blogs/compute/python-3-12-runtime-now-available-in-aws-lambda/)
3. Either: add an enhancement to the existing [PeriodicExportingMetricReader](https://opentelemetry-python.readthedocs.io/en/latest/sdk/metrics.export.html#opentelemetry.sdk.metrics.export.PeriodicExportingMetricReader) to add the ability to suspend/resume metric collection to ensure that metric collection stops before the Lambda environment is potentially frozen OR create a new `SuspendableExportingMetricReader` which has the same functionality. 

Metric aggregation in the collector, however there seems to be limited support for this pattern. 

### Additional Context

_No response_

Yes

### Tip

herin049:
# Description

Adds ability to configure flush behavior in AWS Lambda instrumentation. 

Fixes #4157 

## Type of change

# How Has This Been Tested?

```
tox run -e $(tox --listenvs | grep aws-lambda | tr '\n' ',')
```

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
@@ -40,17 +40,35 @@ Usage
 
     AwsLambdaInstrumentor().instrument()
 
+Configuration
+-------------
+
+Force Flush
+***********
+By default, the instrumentation force flushes both traces and metrics after each Lambda invocation.
+This behavior can be controlled with the following environment variables:
+
+* ``OTEL_INSTRUMENTATION_AWS_LAMBDA_FORCE_FLUSH``: Enables or disables force flushing for all signal types. Defaults to ``true``.
+* ``OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_TRACES``: Enables or disables force flushing for traces. Defaults to ``true``.
+* ``OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_METRICS``: Enables or disables force flushing for metrics. Defaults to ``true``.
+* ``OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_TIMEOUT``: The timeout in milliseconds for force flushing. Defaults to ``30000``.
+
+A specific signal type will be force flushed if either the general ``OTEL_INSTRUMENTATION_AWS_LAMBDA_FORCE_FLUSH``
+or the specific environment variable for that signal type is set to true.
+
 API
 ---
 
 The `instrument` method accepts the following keyword args:
 
-tracer_provider (TracerProvider) - an optional tracer provider
-meter_provider (MeterProvider) - an optional meter provider
-event_context_extractor (Callable) - a function that returns an OTel Trace
-Context given the Lambda Event the AWS Lambda was invoked with
-this function signature is: def event_context_extractor(lambda_event: Any) -> Context
-for example:
+* ``tracer_provider`` (TracerProvider) - an optional tracer provider
+* ``meter_provider`` (MeterProvider) - an optional meter provider
+* ``event_context_extractor`` (Callable) - a function that returns an OTel Trace Context given the Lambda Event the AWS Lambda was invoked with this function signature is: def event_context_extractor(lambda_event: Any) -> Context
+* ``force_flush`` (bool) - enables or disables force flushing for all signal types. Defaults to True.
+* ``flush_traces`` (bool) - enables or disables force flushing for traces. Defaults to True.
+* ``flush_metrics`` (bool) - enables or disables force flushing for metrics. Defaults to True.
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -122,6 +139,27 @@ def custom_event_context_extractor(lambda_event):
...
+)
+OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_TRACES = (
+    "OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_TRACES"
+)
+OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_METRICS = (
+    "OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_METRICS"
+)
+
+
+def _get_env_bool(env_var: str, default: bool) -> bool:
```
tammy-baylis-swi, line 153: I'm surprised there isn't something like this already in contrib nor the SDK repo -- I couldn't find one either.

tammy-baylis-swi (approved):
Lgtm

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -40,17 +40,34 @@ def lambda_handler(event, context):
...
 
+Configuration
+-------------
+
+Force Flush
+***********
+By default, the instrumentation force flushes both traces and metrics after each Lambda invocation.
+This behavior can be controlled with the following environment variables:
+
+* ``OTEL_INSTRUMENTATION_AWS_LAMBDA_FORCE_FLUSH``: Enables or disables force flushing for all signal types. Defaults to ``true``.
```
xrmx, line 51: If these are Python specific they should be `OTEL_PYTHON_` prefixed, e.g. `OTEL_PYTHON_AWS_LAMBDA_FORCE_FLUSH`

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -122,6 +139,27 @@ def custom_event_context_extractor(lambda_event):
...
+    "OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_METRICS"
+)
+
+
+def _get_env_bool(env_var: str, default: bool) -> bool:
+    value = os.environ.get(env_var)
+    if value is None:
+        return default
+    value = value.strip().lower()
+    if value in ("y", "yes", "t", "true", "on", "1"):
```
xrmx, line 158: We should just care about `true` per:
https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#boolean

herin049: The issue I see with that is that we cannot allow explicitly setting this option to `false`. Should I make this a number (i.e. `0/1`) instead?

tammy-baylis-swi: @xrmx thoughts? (on recently stale PR 🙂)

emdneto:
Thanks for the PR!

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
xrmx: Sorry, missed this. I think I'm missing the issue with false

PR #4175, opened 2026-02-05
aiohttp-server: fix HTTP error inconsistencies

### LINKED ISSUE #4153: HTTP error inconsistencies in `opentelemetry-instrumentation-aiohttp-server`
### Describe your environment

Package version: Version 0.60b1 of `opentelemetry-instrumentation-aiohttp-server`

### What happened?

All exceptions used for web responses by AIOHTTP server are being transmitted as errors. However, exceptions such as `web.HTTPFound` are not indicitive of real errors.

For example, the mechanism to implement redirects is to raise a `web.HTTPFound` exception. More information in the docs: https://docs.aiohttp.org/en/stable/web_quickstart.html#redirects.

I suspect the impact is mostly isolated to this `web.HTTPFound` exception. However, technically AIOHTTP _does_ expose exceptions for 2xx responses as well.

### Steps to Reproduce

Implement a route that uses the `web.HTTPFound` exception to implement a redirect.

### Expected Result

The instrumentation should not be transmitting mere HTTP redirects as errors.

The guidance outlined by the HTTP semantic conventions seem to confirm this intuition: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status.

### Actual Result

The AIOHTTP server instrumentor transmits all instances of `web.HTTPException` with error semantics. Specifically, `status` is set on the span.

This can be seen in the source here: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/v0.60b1/instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py#L440)

### Additional context

The behavior of the instrumentation directly contradicts the guidance [here](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status):

> [Span Status](https://opentelemetry.io/docs/specs/otel/trace/api/#set-status) MUST be left unset if HTTP status code was in the 1xx, 2xx or 3xx ranges, unless there was another error (e.g., network error receiving the response body; or 3xx codes with max redirects exceeded), in which case status MUST be set to Error.

Instead, it transmits ALL responses, even 2xx and 3xx, with `status` in the span.

Anecdotally, this is producing thousands of errors in the APM dashboard of our observability tool. However, of course, an HTTP 302 is not actually indicitive of an actual error.

None

### Tip

emdneto:
# Description

Not all exceptions are errors.

Fixes #4153

## Type of change

# Checklist:


instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py
```diff
@@ -469,6 +463,8 @@ async def _middleware(request, handler):
             context=extract(request, getter=getter),
             kind=trace.SpanKind.SERVER,
             attributes=request_attrs,
+            set_status_on_exception=False,
+            record_exception=False,
```
agroebe, line 467: Thanks for the quick fix!

Out of curiosity, am I correct in saying that this was the main fix? It seems that `set_status_on_exception` as `True` would guarantee every exception would get reported 😉

emdneto: yup. that's correct @agroebe

PR #4181, opened 2026-02-09
Align gen_ai.tool_definitions  with JSON Schema and add it to completion hook 

wikaaaaa:
# Description

This PR aligns `gen_ai.tool_definitions` with JSON Schema proposed in https://github.com/open-telemetry/semantic-conventions/pull/3378. 

This PR also adds `gen_ai.tool_definitions` to completion hook and configures exporting the attribute to GCS. Its configured the same way as system instructions, it's being hashed, since it might be large and often doesn't differ. 

Its a continuation of #4142, which added `gen_ai.tool_definitions` to `gen_ai.client.inference.operation.details`.

## Type of change

# How Has This Been Tested?

- [x]  Manual tests: run adk web and inspected the traces and data send to the storage bucket
- [x] Unit tests 

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
```diff
@@ -247,27 +294,39 @@ def on_completion(
...
         def to_dict(
-            dataclass_list: list[types.InputMessage]
+            data_list: list[types.InputMessage]
             | list[types.OutputMessage]
             | list[types.MessagePart],
         ) -> JsonEncodeable:
-            return [asdict(dc) for dc in dataclass_list]
+            response: JsonEncodeable = []
+            for data in data_list:
+                if isinstance(data, dict):
```
DylanRussell: In what case will it be a dict ? Don't you always serialize the  tool definitions to a sting ? And otherwise it's a dataclass ?

wikaaaaa: The tool definitions are serialized into dictionaries, and in a fallback to a string.

RKest (approved):

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
aabmass: @DylanRussell ping on this

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
```diff
@@ -86,6 +94,34 @@ def is_system_instructions_hashable(
     )
 
 
+def is_tool_definitions_hashable(
+    tool_definitions: list[types.MessagePart] | None,
```
aabmass: Why is tool definitions of `types.MessagePart` and can you give an example of what you would pass here?

I think we should have a dedicated type for this that callers have to convert to before calling this API. The other types in this API are part of the semantic conventions https://github.com/open-telemetry/opentelemetry-python-contrib/blob/93bea2dde78494d454f63248dc7759115830a93b/util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py#L46-L50

Could we align with https://github.com/open-telemetry/semantic-conventions/pull/3378?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
DylanRussell: Yeah it makes sense for it to be this way, I confused myself thinking the hashed tool definitions (which I think is always a string) would be passed to this.. 

But this function takes the tool definitions as they are on the attribute, and in that case it'll be a dict or fallback to a string: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py#L237

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
wikaaaaa: Done. I created a separate types.ToolDefintions and aligned with https://github.com/open-telemetry/semantic-conventions/pull/3378

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
wikaaaaa: the comment is now outdated, since I added the types.ToolDefinitions

instrumentation-genai/opentelemetry-instrumentation-google-genai/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 - Fix bug in how tokens are counted when using the streaming `generateContent` method.  ([#4152](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4152)).
 - Add `gen_ai.tool.definitions` attribute to `gen_ai.client.inference.operation.details` log event ([#4142](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4142)).
+- Add `gen_ai.tool_definitions` to completion hook ([#4181](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4181))
```
DylanRussell, line 11: I think we have to bump the util-genai dependency in pyproject.toml: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation-genai/opentelemetry-instrumentation-google-genai/pyproject.toml#L43 -- but I guess we have to do a release of the util gen AI library first ? I'm a bit confused about what the process is supposed to be here @aabmass

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -208,68 +202,168 @@ def _to_dict(value: object):
...
+) -> dict[str, AttributeValue]:
+    if tool_def == []:
+        return {}
+
+    result = []
+    for td in tool_def:
+        if isinstance(td, FunctionToolDefinition):
+            result.append(
+                FunctionToolDefinition(
+                    name=td.name,
```
DylanRussell: nit: this could just be 1 long dict/list comprehension

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -563,44 +659,95 @@ def _maybe_update_error_type(self, response: GenerateContentResponse):
...
+        if self._content_recording_enabled in [
+            ContentCapturingMode.EVENT_ONLY,
+            ContentCapturingMode.SPAN_AND_EVENT,
+        ]:
+            event.attributes = {
+                **(event.attributes or {}),
+                **completion_details_attributes,
+            }
+        else:
+            tool_def_without_params_attr = _tool_def_without_parameters_attr(
```
DylanRussell: nit: just inline this below instead of storing in a variable for 1 line (here and below)

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -137,6 +137,26 @@ class Uri:
     type: Literal["uri"] = "uri"
 
 
+@dataclass()
+class FunctionToolDefinition:
+    """Represents a function tool definition sent to the model"""
+
+    name: str
+    description: str | None
+    parameters: Any | None  # or dict[str, Any]
```
DylanRussell: I'm pretty sure Any includes None ?

util/opentelemetry-util-genai/CHANGELOG.md
```diff
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Add `gen_ai.tool_definitions` to completion hook ([#4181](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4181))
```
DylanRussell, line 10: This is duplicated here and below

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
```diff
@@ -78,14 +87,43 @@ class CompletionRefs:
...
-        isinstance(x, types.Text) for x in system_instruction
+    return bool(message_parts) and all(
+        isinstance(x, types.Text) for x in message_parts
     )
 
 
+def is_tool_definitions_hashable(
+    tool_definitions: list[types.ToolDefinition] | None,
+) -> bool:
+    return bool(tool_definitions)
```
DylanRussell: nit: just inline bool(tool_definitions), I don't think this needs to be a function. Also do we care if the tool definitions are hashable if this list is empty ? I'm not sure we need to do this check at all

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
```diff
@@ -166,19 +204,26 @@ def done(future: Future[None]) -> None:
...
+        if is_message_part_list_hashable(system_instruction):
             # Get a hash of the text.
             system_instruction_hash = hashlib.sha256(
                 "\n".join(x.content for x in system_instruction).encode(  # pyright: ignore[reportUnknownMemberType, reportAttributeAccessIssue, reportUnknownArgumentType, reportCallIssue, reportArgumentType]
                     "utf-8"
                 ),
                 usedforsecurity=False,
             ).hexdigest()
+
+        tool_definitions_hash = None
```
DylanRussell: nit: update `hash_tool_definitions` to accept an empty list and return None in that case in order to make this code a little cleaner..

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
```diff
@@ -191,6 +236,10 @@ def _calculate_ref_path(
                 self._base_path,
                 f"{system_instruction_hash or uuid_str}_system_instruction.{self._format}",
             ),
+            tool_definitions_ref=posixpath.join(
```
DylanRussell, line 236: do we still want to stamp the ref  / do an upload if tool definitions is None ?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -563,44 +659,95 @@ def _maybe_update_error_type(self, response: GenerateContentResponse):
...
-    def _maybe_get_tool_definitions(self, config):
+    def _maybe_get_tool_definitions(self, config) -> list[ToolDefinition]:
         if (
             self.sem_conv_opt_in_mode
             != _StabilityMode.GEN_AI_LATEST_EXPERIMENTAL
         ):
-            return None
+            return []
 
         tool_definitions = []
```
DylanRussell: nit I think it could be a list comprehension ? ` return [de for de in _to_tool_definition(tool) for tool in _config_to_tools(config) if de` (maybe also below, but not sure with the `await` call..)

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -208,68 +202,168 @@ def _to_dict(value: object):
...
     if _is_mcp_imported and isinstance(tool, McpClientSession):
         result = await tool.list_tools()
-        return [t.model_dump(exclude_none=True) for t in result.tools]
+        return [_model_dump_to_tool_definition(t) for t in result.tools]
 
     return _to_tool_definition_common(tool)
 
 
+def _tool_def_without_parameters_attr(
+    tool_def: list[ToolDefinition],
```
DylanRussell, line 342: Do we just want to use `FunctionToolDefinition`directly  Instead of using `ToolDefinition` everywhere ? I don't think we want to use the `GenericToolDefinition` at all ?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
```diff
@@ -208,68 +202,168 @@ def _to_dict(value: object):
...
-        return _mcp_tool_to_tool_definition(tool)
+        return [_mcp_tool_to_tool_definition(tool)]
 
-    try:
-        return {"raw_definition": json.loads(json.dumps(tool))}
-    except Exception:  # pylint: disable=broad-exception-caught
-        return {
-            "error": f"failed to serialize tool definition, tool type={type(tool).__name__}"
-        }
+    return [
```
DylanRussell, line 316: Ah I see we have a Generic tool definition here.. So do we really want to add something to the telemetry in case of a tool or params that we fail to serialize ? I feel like it makes more sense to leave it out and log an error/warning instead. Maybe worth bringing up that question in the sem conv PR. Maybe @aabmass has thoughts on this

wikaaaaa: We have Generic tools also in _tool_to_tool_definition. 

I wasnt sure which approach would be better, including an error/ leaving the tool out entirely. I was thinking more from the evals perspective, that maybe the information that another tool was there, but just wasnt serialialized successfully might be useful. 
But both approaches make sense to me. 

@aabmass wdyt?

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
wikaaaaa: We do use GenericToolDefinition in serializing genai Tools.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
wikaaaaa: Done

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
wikaaaaa: done

util/opentelemetry-util-genai/CHANGELOG.md
wikaaaaa: fixed

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
wikaaaaa: Maybe we could pass `true` instead of `bool(tool_defintions)`, but I still would rather leave `bool(tool_defintions)`, since its clearer in my opinion

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
wikaaaaa: done

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_upload/completion_hook.py
wikaaaaa: If its empty it wont be uploaded anyway due to the logic in on_completion. The same logic is done for the other attributes, this function just returns the paths, and later on there is the logic that filters out empty attributes

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
wikaaaaa: done

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
wikaaaaa: > Maybe worth bringing up that question in the sem conv PR

Sure, but I wouldn't want to block this PR on a discussion under the sem conv PR. Can we go with one of the options now, and if the sem conv PR decides on a different approach, we will align with it in a future PR

instrumentation-genai/opentelemetry-instrumentation-google-genai/CHANGELOG.md
aabmass: Yes we need to bump the version and release util-genai, which we can totally do. So I think this PR needs to be split

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/requirements.oldest.txt
```diff
@@ -25,10 +25,11 @@ opentelemetry-api==1.37.0
...
 opentelemetry-instrumentation==0.58b0
-opentelemetry-util-genai[upload]==0.2b0
+# opentelemetry-util-genai[upload]==0.2b0
 
 fsspec==2025.9.0
 
 # Install locally from the folder. This path is relative to the
 # root directory, given invocation from "tox" at root level.
 -e instrumentation-genai/opentelemetry-instrumentation-google-genai
+-e util/opentelemetry-util-genai[upload]
```
aabmass, line 37: Can you add a TODO to revert this? This makes the CI green but it doesn't capture the reality this no longer works with the "oldest" util-genai dep listed in `pyproject.toml`.

instrumentation-genai/opentelemetry-instrumentation-google-genai/CHANGELOG.md
aabmass: Left a comment above we can submit this as one unit and then change back `instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/requirements.oldest.txt`. but in the future it would be better to split such PRs up.

aabmass: @wikaaaaa @DylanRussell I can release util-genai and then send a follow up PR to update this stuff. If you can just file an issue and TODO comments to make sure we don't miss anything that would be great.

instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py
aabmass: I still think it would be better to capture this after google-genai itself converts to JSON schema, which would avoid this problem altogether. But that's a problem for another day, so I'm fine with whatever.

instrumentation-genai/opentelemetry-instrumentation-google-genai/CHANGELOG.md
wikaaaaa: Issue filed - https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4221 and comments added in newest commit.

instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/requirements.oldest.txt
wikaaaaa: done

PR #4183, opened 2026-02-09
feat(lambda): read cloud.account.id from symlink in Lambda detector

RaphaelManke:
## Summary

- Reads `cloud.account.id` from the symlink at `/tmp/.otel-aws-account-id` created by the OTel Lambda Extension
- Uses `os.readlink()` to read the symlink target (the AWS account ID)
- Logs at debug level if the symlink doesn't exist (`except OSError` guard)

## Depends on

- open-telemetry/opentelemetry-lambda#2127 (extension creates the symlink)

## Changes

- `_lambda.py` — Added readlink logic with `_ACCOUNT_ID_SYMLINK_PATH` constant
- Tests: happy path, leading zeros preserved, missing symlink handled gracefully

## Test plan

- [x] Unit test: symlink exists → `cloud.account.id` set correctly
- [x] Unit test: leading zeros preserved as string
- [x] Unit test: missing symlink → attribute absent, no error

### DIFF
--- a/sdk-extension/opentelemetry-sdk-extension-aws/tests/resource/test__lambda.py
+++ b/sdk-extension/opentelemetry-sdk-extension-aws/tests/resource/test__lambda.py
@@ -1,12 +1,15 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
+import os
+import tempfile
 import unittest
 from collections import OrderedDict
 from unittest.mock import patch
 
 from opentelemetry.sdk.extension.aws.resource._lambda import (  # pylint: disable=no-name-in-module
     AwsLambdaResourceDetector,
+    _ACCOUNT_ID_SYMLINK_PATH,
 )
 from opentelemetry.semconv.resource import (
     CloudPlatformValues,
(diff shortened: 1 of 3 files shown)


sdk-extension/opentelemetry-sdk-extension-aws/tests/resource/test__lambda.py
```diff
@@ -61,3 +64,115 @@ def test_simple_create(self):
...
+            "AWS_LAMBDA_FUNCTION_MEMORY_SIZE": f"{MockLambdaResourceAttributes[ResourceAttributes.FAAS_MAX_MEMORY]}",
+        },
+        clear=True,
+    )
+    def test_account_id_from_symlink(self):
+        """When the account ID symlink exists, cloud.account.id is set."""
+        symlink_path = None
+        try:
+            tmpdir = tempfile.mkdtemp()
+            symlink_path = os.path.join(tmpdir, ".otel-account-id")
```
herin049: Can we just make this a fixture, preferably one that uses the existing `tmp_path` fixture from Pytest?

RaphaelManke: Done — created an `account_id_symlink` pytest fixture that uses `tmp_path` and returns a factory. Converted the symlink tests to plain pytest functions to receive the fixture natively.

tammy-baylis-swi:
Thanks for this! Please do these for lint/formatting and commit the fixes:

* `tox -e lint-sdk-extension-aws`
* `tox -e precommit`

emdneto:
Thanks for the PR!

PR #4195, opened 2026-02-13
Try out basedpyright with baseline file

aabmass:
Enables typechecking on python packages, examples, tests in the repo using basedpyright's baseline feature https://docs.basedpyright.com/latest/benefits-over-pyright/baseline/

Downsides to this approach:
- VSCode's default python settings doesn't recognize this file and will complain about tons of errors. You can recommend basedpyright LSP instead https://docs.basedpyright.com/latest/installation/ides/#vscode-vscodium
- The `.basedpyright/baseline.json` is pretty enormous
- Typechecking becomes much slower (more files). Takes 28 seconds on my machine.

Benefit:
- All new code even in existing files will be forced to type check
- Don't litter codebase with `# type: ignore`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1100,7 +1100,7 @@ commands =
 [testenv:typecheck]
 deps =
   -c {toxinidir}/dev-requirements.txt
-  pyright
+  basedpyright
   {[testenv]test_deps}
   {toxinidir}/opentelemetry-instrumentation
   {toxinidir}/util/opentelemetry-util-http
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -1115,4 +1115,4 @@ deps =
   {toxinidir}/instrumentation/opentelemetry-instrumentation-aiohttp-client[instruments]
```
aabmass, line 1115: I imagine many errors could be removed by adding all new instrumentations here too.

PR #4208, opened 2026-02-16
openai-agents: Handle various types of span_data.response.output parts

### LINKED ISSUE #4185: [OpenAI Agents] Tool calls show in output as text with ResponseFunctionToolCall()
### Describe your environment

OS: macOS 26.2 
Python version: Python 3.13.7
Package version: opentelemetry-instrumentation-openai-agents-v2 0.1.0

### What happened?

Following the sample code in README.rst, `gen_ai.output.messages` contains 
```
"[{\"role\":\"assistant\",\"parts\":[{\"type\":\"text\",\"content\":\"ResponseFunctionToolCall(arguments='{\\\"city\\\":\\\"Barcelona\\\"}', call_id='call_Zo4kTaBuq5Xj8TVnWF6Ydswv', name='get_weather', type='function_call', id='fc_0e75eae8ddb954c201698acde1e7948195bcd77e035521c1bc', status='completed')\"}]}]"
```

### Steps to Reproduce

Follow sample code from https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/README.rst#usage, use ConsoleSpanExporter or otherwise to log the value of `gen_ai.output.messages`

### Expected Result

[Gen AI semantic conventions docs](https://opentelemetry.io/docs/specs/semconv/gen-ai/non-normative/examples-llm-calls/#genai-client-spans-when-content-capturing-is-enabled-on-span-attributes) suggest that gen_ai.output.messages parts instead of type=text, it should be type=tool_call with id, name and arguments parameters.

### Actual Result

Output shows as type=text, with ResponseFunctionToolCall() instead

### Additional context

[opentelemetry-instrumentation-openai-agents](https://github.com/traceloop/openllmetry/tree/main/packages/opentelemetry-instrumentation-openai-agents) from openllmetry identifies this scenario correctly, however it uses older gen_ai.completions semantics.

None

### Tip

adammw:
# Description

Handle various defined types of span_data.response.output parts, and preference parsing the parts over using the consolidated output_text attribute. Some `finish_reason` logic has also been touched, however I cannot find documentation that they have ever been returned from the Responses API.

| OpenAI Message Part Type         | GenAI/OTel Part Type         | Name                |
|----------------------------------|------------------------------|-------------------------------|
| message                         | text†                         | —                             |
| message (refusal)               | refusal*                     | —                             |
| reasoning                        | reasoning                    | —                             |
| compaction                       | compaction*                  | —                             |
| file_search_call                 | tool_call                    | file_search                   |
| function_call                    | tool_call                    | _function_name_                 |
| web_search_call                  | tool_call                    | web_search                    |
| computer_call                    | tool_call                    | computer                      |
| image_generation_call            | tool_call_response           | image_generation              |
| code_interpreter_call            | tool_call + tool_call_response| code_interpreter              |
| local_shell_call                 | tool_call                    | local_shell                   |
| shell_call                       | tool_call                    | shell                         |
| shell_call_output                | tool_call_response           | shell                         |
| apply_patch_call                 | tool_call                    | apply_patch                   |
| apply_patch_call_output          | tool_call_response           | apply_patch                   |
| mcp_call                         | tool_call + tool_call_response| mcp_call                      |
| mcp_list_tools                   | tool_call + tool_call_response| mcp_list_tools                |
| mcp_approval_request             | tool_call                    | mcp_approval_request          |
| custom_tool_call                 | tool_call                    | _tool_name_                     |
| (fallback: content string attribute) | text                    | —                             |

†text schema does not define annotations which are added as a custom key to preserve as much metadata as possible
*custom/non-standard - no equivalent exists in current [GenAI Semantic Conventions schema](https://github.com/open-telemetry/semantic-conventions/blob/v1.39.0/docs/gen-ai/gen-ai-output-messages.json#L154)

Fixes #4185

## Type of change

# How Has This Been Tested?

- [x] Updated unit tests to exercise new codepaths
- [x] Manually executed with OpenAI Web Search tool and confirmed tool_call parts in the OTel trace output.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -873,6 +884,373 @@ def _normalize_messages_to_role_parts(
...
+                })
+            return parts
+        if part_type == "file_search_call":  # ResponseFileSearchToolCall
+            return [{
+                "type": "tool_call",
+                "name": "file_search",
+                "id": getattr(item, "id", None),
+                "arguments": ({
+                    "queries": ["readacted"]
+                    if not self.include_sensitive_data
```
kylehounslow: Seems that when `include_sensitive_data=False`, the arguments here become `{"queries": ["readacted"]}` (a `dict` wrapping a `list`), but when `True`, `getattr(item, "queries", None)` returns a bare `list`. 
Question:
* Would consumers parsing this output see different shapes depending on the sensitivity setting? Should the `True`path also wrap in `{"queries": ...}` to keep the structure consistent?

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -873,6 +884,373 @@ def _normalize_messages_to_role_parts(
...
+            parts = []
+            content = getattr(item, "content", None)
+            if isinstance(content, Sequence):
+                for c in content:
+                    content_type = getattr(c, "type", None)
+                    if content_type == "output_text":
+                        parts.append({
+                            "type": "text",
+                            "annotations": (  # out of spec but useful
+                                ["readacted"]
```
kylehounslow: nit: `"readacted"` → `"redacted"`? Looks like this may be inherited from the existing codebase, so maybe a separate cleanup PR if the project wants to address it

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/tests/test_tracer.py
```diff
@@ -513,9 +792,219 @@ def __init__(self) -> None:
         )
         assert response.attributes[GenAI.GEN_AI_USAGE_INPUT_TOKENS] == 42
         assert response.attributes[GenAI.GEN_AI_USAGE_OUTPUT_TOKENS] == 9
-        assert response.attributes[GenAI.GEN_AI_RESPONSE_FINISH_REASONS] == (
-            "stop",
-        )
+
+        # Check output messages are properly normalized
```
kylehounslow, line 846: Nice coverage of all the output types! Related to my comment above: 
Since the instrumentor defaults to `capture_message_content=SPAN_AND_EVENT` (i.e. `include_sensitive_data=True`) when no `kwarg` is passed, seems all these assertions exercise the non-redacted path. Would it be worth adding a companion test with `capture_message_content=False` to verify the redaction paths produce the expected shape for a few of the more complex types (e.g., `mcp_call`, `code_interpreter_call`)?

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
adammw: think this was a typo or confusion of the scope of the if statement, added brackets to make it more obvious. the shape should remain the same.

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -873,6 +884,373 @@ def _normalize_messages_to_role_parts(
...
+            }]
+        elif part_type == "function_call":  # ResponseFunctionToolCall
+            return [{
+                "type": "tool_call",
+                "name": getattr(item, "name", None),
+                "id": getattr(item, "id", None),
+                "arguments": (
+                    "readacted"
+                    if not self.include_sensitive_data
+                    else safe_json_loads(getattr(item, "arguments", None))
```
adammw: Speaking of output shape, this will vary here depending on sensitivity, but I couldn't think of a better way to do it.

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
kylehounslow: > think this was a typo or confusion of the scope of the if statement, added brackets to make it more obvious. the shape should remain the same.

Thanks for the quick fix, it's clear to me now!

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
kylehounslow: Ack, probably unavoidable here without adding a lot of complexity. I think consumers would need to handle both cases regardless. Not a blocker.

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
adammw: Turns out none of this is required - when `include_sensitive_data=False` the method never gets called so all this logic can go.

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/tests/test_tracer.py
adammw: Added the `test_response_span_records_redacted_response_attributes` test but the redaction is basically absolute - the output message attribute is not included at all - so none of that code is necessary.

nagkumar91 (commented):
Overall great PR — comprehensive coverage of all the output types. Two minor observations below.

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -371,10 +372,20 @@ def safe_json_dumps(obj: Any) -> str:
...
+    except (TypeError, ValueError) as e:
+        logger.warning("Failed to serialize object to JSON: %s", e)
         return str(obj)
 
 
+def safe_json_loads(s: str) -> Any:
+    """Safely parse JSON string (fallback to original string)."""
+    try:
+        return json.loads(s)
+    except json.JSONDecodeError as e:
```
nagkumar91: `safe_json_loads` catches `json.JSONDecodeError` but `json.loads(None)` raises `TypeError`, not `JSONDecodeError`.

While the OpenAI SDK type definitions for `ResponseFunctionToolCall`, `McpCall`, and `McpApprovalRequest` all declare `arguments: str` (required, non-optional), there are edge cases where `None` could arrive:

1. **Streaming incomplete objects** — there are [known SDK issues](https://github.com/openai/openai-python/issues/2723) where fields like `arguments` can be missing/None during streaming when the `function_call_arguments.done` event is not yet received.
2. **Defensive `getattr` usage** — the callers all use `getattr(item, "arguments", None)` with an explicit `None` default, which means if the attribute is missing for any reason (future SDK changes, non-standard objects), `None` gets passed through.

Since this function is named `safe_json_loads`, it would be good to make it actually safe against `None`:

```python
except (json.JSONDecodeError, TypeError) as e:
```

Or add an early return:
```python
if s is None:
    return None
```

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -873,10 +884,353 @@ def _normalize_messages_to_role_parts(
...
+                    "arguments": {
+                        "server": getattr(item, "server_label", None),
+                        "tool_name": getattr(item, "name", None),
+                        "tool_args": (
+                            safe_json_loads(getattr(item, "arguments", None))
+                        ),
+                    },
+                }
+            ]
+            if getattr(item, "output", None) or getattr(item, "error", None):
```
nagkumar91: nit: `""` (empty string) is falsy in Python, so if an MCP tool returns an empty string as a valid response, this condition evaluates to `False` and no `tool_call_response` part is emitted — silently dropping the response.

```python
# Current:
if getattr(item, "output", None) or getattr(item, "error", None):

# Safer:
if getattr(item, "output", None) is not None or getattr(item, "error", None) is not None:
```

Same applies to the `mcp_list_tools` handler a few lines below. Probably unlikely in practice but worth considering for correctness.

nagkumar91 (changes requested):

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
adammw: added both, since pydantic type hinting didn't like passing none to json.loads() and relying on the TypeError. but catching it in case it does happen at runtime (other non none types).

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
adammw: added

nagkumar91 (approved):
Both comments addressed — `safe_json_loads` now has the None guard + TypeError catch, and the MCP output/tools checks use `is not None` instead of truthiness. Also nice bonus adding `hasattr(output, "to_dict")` guards on the `.to_dict()` calls. LGTM.

lzchen (approved):

hectorhdzg (approved):
LGTM

lzchen:
@adammw 

Please fix conflicts and we can get this merged.

lmolkova:
thanks for the contribution @adammw !

We're moving GenAI instrumentations to the new repository - https://github.com/open-telemetry/opentelemetry-python-genai.

I'm closing this PR here, but you're welcome to reopen it in the new repo.

PR #4214, opened 2026-02-18
Add log and metrics provider to langchain

wrisa:
# Description

Add log and metrics provider in langchain's call to genai utils telemetry handler.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,4 +1,4 @@
 [codespell]
 # skipping auto generated folders
-skip = ./.tox,./.mypy_cache,./docs/_build,./target,*/LICENSE,./venv
+skip = ./.tox,./.mypy_cache,./docs/_build,./target,*/LICENSE,./venv,*/cassettes
 ignore-words-list = ot


instrumentation-genai/opentelemetry-instrumentation-langchain/CHANGELOG.md
```diff
@@ -7,8 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-  ([https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3889](#3889))
-
 - Added span support for genAI langchain llm invocation.
-  ([#3665](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3665))
\ No newline at end of file
+  ([#3665](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3665))
+- Added support to call genai utils handler for langchain LLM invocations.
+  ([#3889](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3889))
+- Added log and metrics provider to langchain genai utils handler
+  ([#4214](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4214))
```
keith-decker: File should have a newline at the end

pmcollins (commented):
Production code changes are quite small and they LGTM. Kudos for thorough testing, but IMO, if they don't already, these tests should live in genai utils rather than here, as they are testing that code. For this PR, I think a spot check or even just a test that verifies the provider wiring would suffice.

pmcollins (approved):

instrumentation-genai/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/__init__.py
```diff
@@ -70,12 +70,16 @@ def _instrument(self, **kwargs: Any):
         Enable Langchain instrumentation.
         """
         tracer_provider = kwargs.get("tracer_provider")
+        meter_provider = kwargs.get("meter_provider")
+        logger_provider = kwargs.get("logger_provider")
 
         telemetry_handler = get_telemetry_handler(
```
nagkumar91, line 76: Potential issue: `get_telemetry_handler(...)` is a singleton cache, so after the first instrumentation call, later calls with different `tracer_provider`/`meter_provider`/`logger_provider` will still reuse the original handler and ignore new providers. Tests work around this by deleting `_default_handler`, but production `_uninstrument()` only unwraps and does not clear/reset the singleton. Could we clear/recreate the handler on `uninstrument()` (or when providers differ) to avoid stale telemetry wiring after re-instrumentation?

JWinermaSplunk: +1 for clear in uninstrument

instrumentation-genai/opentelemetry-instrumentation-langchain/tests/test_llm_call.py
```diff
@@ -5,15 +5,24 @@
...
 
 
-# span_exporter, start_instrumentation, chat_openai_gpt_3_5_turbo_model are coming from fixtures defined in conftest.py
+# span_exporter, metric_reader, log_exporter, start_instrumentation, chat_openai_gpt_3_5_turbo_model are coming from fixtures defined in conftest.py
 @pytest.mark.vcr()
-@pytest.mark.parametrize("capture_content", ["SPAN_ONLY", "NO_CONTENT"])
+@pytest.mark.parametrize(
+    "capture_content",
+    ["SPAN_ONLY", "NO_CONTENT", "SPAN_AND_EVENT", "EVENT_ONLY"],
+)
```
aabmass, line 21: @wrisa it seems like you could re-use the same cassette across all these test runs, is that correct?

IMO that helps keep PRs a little less cluttered. Does this work?

```py
import pytest
import requests

# don't use the mark
@pytest.mark.parametrize("endpoint", ["/users", "/posts", "/comments"])
def test_api_endpoints(vcr, endpoint):
    # All parametrized cases will read from/write to "shared_api_cassette.yaml"
    with vcr.use_cassette("shared_api_cassette.yaml"):
        response = requests.get(f"https://jsonplaceholder.typicode.com{endpoint}")
        assert response.status_code == 200
```

aabmass:
Can you update the spellcheck config? https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/.codespellrc

PR #4216, opened 2026-02-19
opentelemetry-instrumentation-aws-lambda: fix improper handling of header casing

### LINKED ISSUE #4106: W3CBaggagePropagator fails to extract baggage when header name is title-cased (e.g., "Baggage" vs "baggage")
### Describe your environment

- **OS:** Amazon Linux 2 (AWS Lambda)
- **Python version:** 3.11
- **SDK version:** opentelemetry-sdk==1.33.0
- **Relevant packages:** opentelemetry-instrumentation-aws-lambda

### What happened?

The `W3CBaggagePropagator` fails to extract baggage context when the HTTP header name is title-cased (`Baggage`) instead of lowercase (`baggage`). This occurs when requests pass through AWS API Gateway, which normalizes header names to title case.

Per [RFC 7230](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2) and the [W3C Baggage specification](https://www.w3.org/TR/baggage/#header-name), HTTP header names are case-insensitive. The propagator should perform case-insensitive header lookup.

### Steps to Reproduce

1. Deploy a Python Lambda function with OpenTelemetry instrumentation and W3CBaggagePropagator
2. Send a request with `baggage` header via two different methods:
   - **Method A:** Via AWS API Gateway (e.g., using k6 load test tool)
   - **Method B:** Via AWS Lambda Test UI (direct invocation)
3. Observe that baggage extraction works in one case but not the other

### Expected Result

Baggage should be extracted regardless of header name casing (`baggage`, `Baggage`, `BAGGAGE`, etc.)

### Actual Result

**Via AWS Lambda Test UI (works) - header stays lowercase:**

```
Request headers: {'baggage': 'city_id=123456,env=MY_ENV,root_endpoint=my-service-handler'}
Extracted baggage: {'city_id': '123456', 'env': 'MY_ENV', 'root_endpoint': 'my-service-handler'}
```

**Via API Gateway with k6 (fails) - header is title-cased:**

```
Request headers: {'Baggage': 'city_id=123456,env=MY_ENV,root_endpoint=my-service-handler'}
Extracted baggage: {}  # Empty - baggage not extracted
```

AWS API Gateway normalizes HTTP header names to title case (`baggage` → `Baggage`), causing the extraction to fail.

### Additional context

_No response_

Yes

### Tip

herin049:
# Description
Fixes issue where request headers are extracted in a case-sensitive manner, resulting in issues with context propagation for API Gateway requests.

Fixes #4106

## Type of change

# How Has This Been Tested?

```
tox run -e $(tox --listenvs | grep opentelemetry-aws-lambda | tr '\n' ',')
```

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -216,20 +218,21 @@ def _set_api_gateway_v1_proxy_attributes(
     span.set_attribute(HTTP_METHOD, lambda_event.get("httpMethod"))
 
     if lambda_event.get("headers"):
-        if "User-Agent" in lambda_event["headers"]:
+        headers = {k.casefold(): v for k, v in lambda_event["headers"].items()}
```
emdneto: don't we have any helper to normalize headers in util-http?

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -176,7 +176,9 @@ def _default_event_context_extractor(lambda_event: Any) -> Context:
         )
     if not isinstance(headers, dict):
         headers = {}
-    return get_global_textmap().extract(headers)
+    return get_global_textmap().extract(
```
xrmx, line 169: When I did this in another code base I've only done the header normalization for v1 using this to decide if that was the case:

```
def should_normalize(event):
    request_context = event.get("requestContext", {})
    return ("elb" in request_context or "requestId" in request_context) and "http" not in request_context
```

Maybe here it would be impractical since we are picking more stuff from the headers.

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
xrmx: I think we have only calls to lower(), we have an helper to normalize them as attributes, e.g.

 ```
    key = header.lower().replace("-", "_")
    return f"http.request.header.{key}"
```

Comment (herin049) [reply instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py]: Ideally, we can add a `CIDict` to `util-http`, but I figured that this is a simple enough fix. Looking through some of the other instrumentation libraries (e.g. Kafka if I recall right), these header casing bugs might be scattered elsewhere. Let me know what you think, I can either add another method to `util-http` or add a `CIDict` implementation.

File: instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py (hunk 2 above)
Comment (herin049) [reply instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py:169]: We'd want to normalize regardless, no? For example, even if you're using API GW v2, it's possible that a caller still might not provide headers as lowercase.

**Comment by herin049** [comment]
I've opted to add a `CIDict` to `opentelemetry-instrumentation`, since this will likely have uses elsewhere and is not limited to just HTTP.

File: opentelemetry-instrumentation/src/opentelemetry/instrumentation/cidict.py [hunk 3]
```diff
@@ -0,0 +1,93 @@
...
+    Tuple,
+    TypeVar,
+    Union,
+)
+
+KT = TypeVar("KT")
+VT = TypeVar("VT")
+
+
+class CIDict(MutableMapping[KT, VT], Generic[KT, VT]):
```
MikeGoldsmith: I don't think we need the extra Generic definition, MutableMap should make this generic.

```suggestion
class CIDict(MutableMapping[KT, VT]):
```


Also, would this be useful in the core repo so other instrumentation libraries can use it? eg Flask, Django, FastAPI, etc. Maybe we make this a private class here and open a PR to add to core, then replace when available?

MikeGoldsmith (approved):
Looks good to me. I've left a couple of non-blocking suggestions for the CIDict class.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/cidict.py
herin049: Removed the redundant superclass. On your second comment, I figure that leaving this in the instrumentation library should be sufficient. Most Http libraries have the headers object implemented as a case-insensitive dictionary anyways, so we do not need to explicitly handle converting the headers to a case insensitive header dictionary. Furthermore, since all instrumentation libraries depend on `opentelemetry-instrumentation` anyways, I personally don't see to much value adding this to core, but I'm certainly still open to it. Let me know what you think.

MikeGoldsmith: It was just a thought, no strong opinion. Happy to keep in here for now 😄

MikeGoldsmith (approved):

aabmass:
Good to close the open conversation threads?

herin049:
> Good to close the open conversation threads?

@aabmass I think so. @xrmx do you want to take another look quick?

aabmass (approved):

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
aabmass: @xrmx just waiting on this thread to merge

emdneto:
Thanks for the PR!

xrmx (approved):
Need to move to changelog fragment

PR #4218, opened 2026-02-19
GenAI Utils | Update ToolCall Type

keith-decker:
# Description

Update the tool type to match semantic conventions for [GenAI Spans](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-spans.md#execute-tool-span). 

TelemetryHandler functions for start_tool and stop_tool coming in next PR. 

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [X] basic setters and getter functions

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/instrumentation-genai/opentelemetry-instrumentation-google-genai/pyproject.toml
+++ b/instrumentation-genai/opentelemetry-instrumentation-google-genai/pyproject.toml
@@ -39,9 +39,9 @@ classifiers = [
   "Programming Language :: Python :: 3.14",
 ]
 dependencies = [
-  "opentelemetry-api ~=1.37",
-  "opentelemetry-instrumentation >=0.58b0, <2",
-  "opentelemetry-semantic-conventions >=0.58b0, <2",
+  "opentelemetry-api ~=1.39",
+  "opentelemetry-instrumentation >=0.60b0, <2",
+  "opentelemetry-semantic-conventions >=0.60b0, <2",
   "opentelemetry-util-genai >= 0.4b0.dev, <0.5b0",
 ]
 
--- a/instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/requirements.oldest.txt
+++ b/instrumentation-genai/opentelemetry-instrumentation-google-genai/tests/requirements.oldest.txt
@@ -21,10 +21,10 @@ pytest-vcr==1.0.2
 
 google-auth==2.15.0
 google-genai==1.32.0
-opentelemetry-api==1.37.0
-opentelemetry-sdk==1.37.0
-opentelemetry-semantic-conventions==0.58b0
-opentelemetry-instrumentation==0.58b0
+opentelemetry-api==1.39.0
+opentelemetry-sdk==1.39.0
+opentelemetry-semantic-conventions==0.60b0
+opentelemetry-instrumentation==0.60b0
 -e util/opentelemetry-util-genai[upload]
 
 fsspec==2025.9.0
--- a/instrumentation-genai/opentelemetry-instrumentation-vertexai/pyproject.toml
+++ b/instrumentation-genai/opentelemetry-instrumentation-vertexai/pyproject.toml
@@ -26,9 +26,9 @@ classifiers = [
   "Programming Language :: Python :: 3.14",
 ]
 dependencies = [
-  "opentelemetry-api ~= 1.37",
-  "opentelemetry-instrumentation ~= 0.58b0",
-  "opentelemetry-semantic-conventions ~= 0.58b0",
+  "opentelemetry-api ~= 1.39",
+  "opentelemetry-instrumentation ~= 0.60b0",
+  "opentelemetry-semantic-conventions ~= 0.60b0",
   "opentelemetry-util-genai >= 0.4b0.dev, <0.5b0",
 ]
 
--- a/instrumentation-genai/opentelemetry-instrumentation-vertexai/tests/requirements.oldest.txt
+++ b/instrumentation-genai/opentelemetry-instrumentation-vertexai/tests/requirements.oldest.txt
@@ -65,10 +65,10 @@ grpcio>=1.75.1 ; python_version >= "3.14"
 shapely==2.0.6 ; python_version < "3.10"
 shapely==2.1.2 ; python_version >= "3.10"
 # when updating, also update in pyproject.toml
-opentelemetry-api==1.37
-opentelemetry-sdk==1.37
-opentelemetry-semantic-conventions==0.58b0
-opentelemetry-instrumentation==0.58b0
+opentelemetry-api==1.39
+opentelemetry-sdk==1.39
+opentelemetry-semantic-conventions==0.60b0
+opentelemetry-instrumentation==0.60b0
 # opentelemetry-util-genai[upload]==0.2b0
 -e util/opentelemetry-util-genai[upload]
 fsspec==2025.9.0


keith-decker:
I think the docker tests failure is unrelated to this PR.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -55,6 +55,40 @@ class ToolCall:
     type: Literal["tool_call"] = "tool_call"
```
DylanRussell, line 69: if we are to have 2 different types here, should we also have 2 different `type` names ? Otherwise why not just have  1 type that encompasses both ?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -158,7 +192,15 @@ class GenericToolDefinition:
...
-    Text, ToolCall, ToolCallResponse, Blob, File, Uri, Reasoning, Any
+    Text,
+    ToolCallRequest,
+    ToolCall,
+    ToolCallResponse,
+    Blob,
+    File,
+    Uri,
+    Reasoning,
+    Any,
```
DylanRussell: Just a drive by comment not related to this PR, we should remove `Any` from this union, otherwise what's the point of this ? I think Any was included to capture `GenericPart`, butI think we should just have that as a separate type (like maybe `GenericPart` = Any as an alias)

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
keith-decker: I initially started with one type, but ended up blowing up the vertex tests with extra data. It looks like vertexai uses toolcall as part of the body of an LLM request. 

I separated these into two types to distinguish between toolcallrequests that are part of an LLM invocation, and a tool execution that may be stand alone.

What are your thoughts on that approach?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
keith-decker: removing Any might improve readability, since it does look weird to have this many defined types only to have the final catch all for Any. 

I'd be ok with adding an alias for GenericPart and then defining that more as the spec evolves.

aabmass: Yes we really need to fix it. `X | Y | Any` is just `Any`. But I think we need to tweak the semconv like you mentioned.

And ya, a short term fix that might work for the python code is to use [`NewType`](https://docs.python.org/3/library/typing.html#distinct) for a transparent wrapper. [Playground PoC](https://pyright-play.net/?pythonVersion=3.12&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCCKcANFAHICmA7gCqKUBQjA4pSpSEgMZQC8FGvQSUAFACI2HLt3FkwAIwBWlbjACUzeCKgBZOAFUUSMCn5QAzjBBQAPphT57Uzj2YATSsFgBDANaUFgD6AK7GpqIAHgBceobhKOrRjFCpUAB0mVr%2BgaEJEgAWlAA2xWDimjA5wWEmKKIAjABMAMyajADEUJSRImrdIOAgjFUBNfkA2gC67V3UuH4WUNRIMAXdvcU8q1BgCPioI9V5daIuMqLT6u2MQA)

aabmass:
I'm a little confused.`ToolCall` is just a wrapper as part of GenAI util for instrumentation purposes? This is separate from the ToolCallRequest and ToolCallResponse json schemas in [here](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/non-normative/models.ipynb) which defines an actual data format.

Is that right?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
DylanRussell: Sounds reasonable but what's in the sem convs ?  I think we want these types to match the sem coonvs

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -55,6 +55,40 @@ class ToolCall:
     type: Literal["tool_call"] = "tool_call"
 
 
+@dataclass()
+class ToolCall(ToolCallRequest):
```
wrisa: should it inherit `class ToolCall(GenAIInvocation)` instead ? `ToolCallRequest` can be separate and ok to have its attributes duplicated in ToolCall?

keith-decker: Refactored to extend GenAIInvocation instead. Separated ToolCallRequest to align with message parts

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
keith-decker: This type here refers to the "type" in the message part for ToolCallRequest (https://github.com/open-telemetry/semantic-conventions/blob/95c6e0a3c6803c6546fc66847855761e9c90f9c6/docs/gen-ai/non-normative/models.ipynb?plain=1) 

The type in the new ToolCall refers to the type defined here (https://github.com/open-telemetry/semantic-conventions/blob/95c6e0a3c6803c6546fc66847855761e9c90f9c6/docs/gen-ai/gen-ai-spans.md?plain=1#L618) which is a span level attribute instead of part of the message.

keith-decker:
> I'm a little confused.`ToolCall` is just a wrapper as part of GenAI util for instrumentation purposes? This is separate from the ToolCallRequest and ToolCallResponse json schemas in [here](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/non-normative/models.ipynb) which defines an actual data format.
> 
> Is that right?

@aabmass As discussed in the SIG call on thursday:
The types here called it ToolCall instead of ToolCallRequest like in the "GenAI messages Python models" link [here](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/non-normative/models.ipynb). Changing the names aligns them with that part of the semconvs, and opens ToolCall type for use with the execute_tool span conventions [here](https://github.com/open-telemetry/semantic-conventions/blob/95c6e0a3c6803c6546fc66847855761e9c90f9c6/docs/gen-ai/gen-ai-spans.md?plain=1#L618)

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
DylanRussell: The type is weirdly named tool_call instead of tool_call_request in the sem conv.. oh well.

That second type https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/#execute-tool-span -- seems like it's just supposed to be a string though unless I'm missing something ??

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/utils.py
```diff
@@ -308,21 +310,15 @@ def request_to_events(
         yield user_event(role=content.role, content=request_content)
 
 
-@dataclass
```
DylanRussell, line 311: Not sure where  this was moved but we can put BlobPart into the Gen AI util types and also add it to Message part --- that's how it exists in the sem convs (https://github.com/open-telemetry/semantic-conventions/blob/95c6e0a3c6803c6546fc66847855761e9c90f9c6/docs/gen-ai/non-normative/models.ipynb?plain=1)

That way we don't need to wrap it in a GenericPart (it should be it's own part)

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -30,6 +30,17 @@
 ContextToken: TypeAlias = Token[Context]
```
DylanRussell, line 30: Not sure if you saw but https://github.com/open-telemetry/semantic-conventions/pull/3038/changes was merged.. This adds 4 new types:

GenericServerToolCall/GenerticServerToolCallResponse (not added to MessagePart.. maybe because GenericPart is already there)

also ServerToolCallPart/ServerToolCallResponsePart (added to MessagePart)


Maybe we can use this PR to just get completely in sync with https://github.com/open-telemetry/semantic-conventions/blob/95c6e0a3c6803c6546fc66847855761e9c90f9c6/docs/gen-ai/non-normative/models.ipynb?plain=1

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/utils.py
keith-decker: Further down the file, I replace the vertexai defined `BlobPart` and `FileDataPart` with the `Blob` and `Uri` dataclasses from GenAI Utils. 

It looks like the vertexai classes were created before these message part classes were created in GenAI Utils. 

I used Uri instead of File because it looks like this was expecting a url from some cloud storage. 

Can you clarify what changes you are looking for? I'm probably missing some context here.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
keith-decker: Added the ServerToolCall and and ServerToolCallResponse to the types as well.

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/utils.py
DylanRussell: Can you remove FileDataPart and BlobPart from instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/message.py -- and import / use the Uri and Blob parts from the gen AI util library ?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
DylanRussell: Ok nice. looks like we're caught up to what's in that file although I didn't check too closely

instrumentation-genai/opentelemetry-instrumentation-vertexai/src/opentelemetry/instrumentation/vertexai/utils.py
DylanRussell: Thanks

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -158,7 +207,17 @@ class GenericToolDefinition:
 ToolDefinition = Union[FunctionToolDefinition, GenericToolDefinition]
 
 MessagePart = Union[
-    Text, ToolCall, ToolCallResponse, Blob, File, Uri, Reasoning, Any
+    Text,
+    ToolCallRequest,
+    "ToolCall",
```
DylanRussell: Why put ToolCall inside MessagePart if it's not to be used for a MessagePart ?

keith-decker: It was leftover from Anthropic using ToolCall. I've updated the instrumentation to use ToolCallRequest.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -256,6 +314,53 @@ class LLMInvocation(GenAIInvocation):
...
+
+    # Execution tracking fields (used for execute_tool spans):
+    # gen_ai.tool.type - Tool type: "function", "extension", or "datastore"
+    tool_type: str | None = None
+    # gen_ai.tool.description - Description of what the tool does
+    tool_description: str | None = None
+    # gen_ai.tool.call.result - Result returned by the tool (Opt-In, may contain sensitive data)
+    tool_result: Any = None
+    # error.type - Error type if the tool call failed
+    error_type: str | None = None
```
lmolkova: do we usually expose it? it's recorded in `fail_llm` and such and if we want to make it settable, it should be common across all operations

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
```diff
@@ -42,8 +53,11 @@ class ContentCapturingMode(Enum):
 
 
 @dataclass()
-class ToolCall:
-    """Represents a tool call requested by the model
+class ToolCallRequest:
```
lmolkova, line 56: this is a breaking change, which is common and not a problem on its own, but it means that we should consider releasing all the ai instrumentations now, otherwise they'll end up being incompatible and impossible to use together

aabmass: related https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4232

lmolkova: I left a comment on https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4232#issuecomment-4057494460, I think we should not block this PR - this is just one of many  changes that would be problematic from version compatibility, but from now on we should make sure all GenAI libs depend on the same version of utils. One way would be to release them together and I'm advocating for releasing them along with this repo, but that's just to minimize effort.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py
keith-decker: Moved to the base class, since all genai models have it as conditionally required

util/opentelemetry-util-genai/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Enrich ToolCall type ([#4218](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4218))
```
Cirilla-zmh: I think this is a breaking change, because `ToolCall` is a class exposed to external users. The change here could cause errors in other tools that depend on `opentelemetry-util-genai`.

keith-decker: ToolCall still exposes the same properties as before, it is now extending GenAI Invocation.

The previous ToolCall in utils did not match semconvs. It should have been ToolCallRequest.

Cirilla-zmh: > The previous ToolCall in utils did not match semconvs. It should have been ToolCallRequest.

Exactly. What I mean is that users may have instantiated `ToolCall`. When they upgrade `util-genai`, the interpreter may throw an error because it cannot find the class. At the least, this should be clearly noted in the changelog as a breaking change.

keith-decker: Noted in changelog!

PR #4229, opened 2026-02-20
fix(openai-agents): align AgentSpanData stubs and span processor with real SDK

nagkumar91:
## Description

Fixes test stubs and span processor code that referenced `AgentSpanData` fields (`operation`, `description`, `agent_id`, `model`) which **never existed** in the real [OpenAI Agents SDK](https://github.com/openai/openai-agents-python/blob/main/src/agents/tracing/span_data.py#L33-L56). The real SDK `AgentSpanData` only has: `name`, `handoffs`, `tools`, `output_type`.

Reported by @adammw in https://github.com/open-telemetry/opentelemetry-python-contrib/commit/fa5f44ebdc7af385fe9cea9f723e418e22c5cade#r177563121

### Test stubs (`tests/stubs/agents/tracing/__init__.py`)
- Removed `operation`, `description`, `agent_id`, `model` from stub `AgentSpanData`
- Added `handoffs` field to match real SDK
- Updated `agent_span()` helper to match real SDK signature (positional `name` arg)

### Span processor (`span_processor.py`)
- Simplified `_get_operation_name()`: agent spans always return `invoke_agent` (real SDK has no `operation` field to distinguish create vs invoke)
- Removed `getattr(span_data, "agent_id")`, `getattr(span_data, "description")`, `getattr(span_data, "model")` from `_get_attributes_from_agent_span_data()`
- **Kept**: user-configured overrides (`self.agent_id`, `self.agent_description`) and model fallback from child span aggregation

### Tests
- Replaced `test_agent_create_span_records_attributes` with `test_agent_invoke_span_records_attributes`
- Removed all uses of non-existent `operation=`, `description=`, `agent_id=`, `model=` kwargs from `AgentSpanData()` and `agent_span()` calls

All 103 tests pass.

### DIFF
--- a/.github/workflows/core_contrib_test_0.yml
+++ b/.github/workflows/core_contrib_test_0.yml
@@ -20,6 +20,10 @@ env:
   CORE_REPO_SHA: ${{ inputs.CORE_REPO_SHA }}
   CONTRIB_REPO_SHA: ${{ inputs.CONTRIB_REPO_SHA }}
   PIP_EXISTS_ACTION: w
+  CORE_REPO_API: ${{ github.workspace }}/opentelemetry-python/opentelemetry-api
+  CORE_REPO_SDK: ${{ github.workspace }}/opentelemetry-python/opentelemetry-sdk
+  CORE_REPO_SEMCONV: ${{ github.workspace }}/opentelemetry-python/opentelemetry-semantic-conventions
+  CORE_REPO_TEST_UTILS: ${{ github.workspace }}/opentelemetry-python/tests/opentelemetry-test-utils
 
 jobs:
 
--- a/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/core_contrib_test.yml.j2
+++ b/.github/workflows/generate_workflows_lib/src/generate_workflows_lib/core_contrib_test.yml.j2
@@ -20,6 +20,10 @@ env:
   CORE_REPO_SHA: ${% raw %}{{ inputs.CORE_REPO_SHA }}{% endraw %}
   CONTRIB_REPO_SHA: ${% raw %}{{ inputs.CONTRIB_REPO_SHA }}{% endraw %}
   PIP_EXISTS_ACTION: w
+  CORE_REPO_API: ${% raw %}{{ github.workspace }}{% endraw %}/opentelemetry-python/opentelemetry-api
+  CORE_REPO_SDK: ${% raw %}{{ github.workspace }}{% endraw %}/opentelemetry-python/opentelemetry-sdk
+  CORE_REPO_SEMCONV: ${% raw %}{{ github.workspace }}{% endraw %}/opentelemetry-python/opentelemetry-semantic-conventions
+  CORE_REPO_TEST_UTILS: ${% raw %}{{ github.workspace }}{% endraw %}/opentelemetry-python/tests/opentelemetry-test-utils
 
 jobs:
   {%- for job_data in job_datas %}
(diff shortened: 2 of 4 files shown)


instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -1519,17 +1519,8 @@ def _get_operation_name(self, span_data: Any) -> str:
                     return GenAIOperationName.CHAT
             return GenAIOperationName.TEXT_COMPLETION
         if _is_instance_of(span_data, AgentSpanData):
-            # Could be create_agent or invoke_agent based on context
-            operation = getattr(span_data, "operation", None)
```
hectorhdzg, line 1523: I'm a little confused why this was added in the first place, is this something that was there previously and removed in openai-agents sdk?

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -1831,24 +1822,20 @@ def _get_attributes_from_agent_span_data(
         if name:
             yield GEN_AI_AGENT_NAME, name
 
-        agent_id = (
-            self.agent_id
-            or getattr(span_data, "agent_id", None)
-            or self._agent_id_default
-        )
+        # agent_id and description are not available on the real SDK
```
hectorhdzg: It took me a while to figure out the term "real SDK" used in multiple places here, maybe using something like OpenAI Agents SDK would be more convenient

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -1831,24 +1822,20 @@ def _get_attributes_from_agent_span_data(
...
+        agent_id = self.agent_id or self._agent_id_default
         if agent_id:
             yield GEN_AI_AGENT_ID, agent_id
 
-        description = (
-            self.agent_description
-            or getattr(span_data, "description", None)
-            or self._agent_description_default
-        )
+        description = self.agent_description or self._agent_description_default
```
hectorhdzg, line 1831: This will use OpenAIAgentsInstrumentor.instrument method arguments, I would assume this is not ideal, is there a plan for OpenAI Agents SDK to provide richer data in AgentSpanData?

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
nagkumar91: It was added by mistake

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
nagkumar91: I’ll make this change

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
nagkumar91: Not that I know of. There’s not been much activity on the python repo for their agents

lzchen (approved):

PR #4246, opened 2026-02-21
fix(threading): attribute error when run is called w/o start

### LINKED ISSUE #4245: instrumentation-threading breaks when `Thread.run()` is called without `Thread.start()`
### Describe your environment

OS: MacOS 15
Python version: 3.13.9
Package version: `0.60b1` / `main`

### What happened?

With `ThreadingInstrumentor()` enabled, if a thread is created and immediately run, the program will throw an exception.

### Steps to Reproduce

```python
from threading import Thread
from opentelemetry.instrumentation.threading import ThreadingInstrumentor

def main():
    ThreadingInstrumentor().instrument()
    t = Thread(target=lambda: print("Hello world!"))
    t.run()

if __name__ == "__main__":
    main()
```

### Expected Result

Program prints out "Hello world!" and exits

### Actual Result

```
  File "/Users/din/threading-repro/main.py", line 10, in <module>
    main()
    ~~~~^^
  File "/Users/din/threading-repro/main.py", line 7, in main
    t.run()
    ~~~~~^^
  File "/Users/din/threading-repro/.venv/lib/python3.13/site-packages/opentelemetry/instrumentation/threading/__init__.py", line 150, in __wrap_threading_run
    token = context.attach(instance._otel_context)
                           ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Thread' object has no attribute '_otel_context'
```

### Additional context

I know this isn't a completely valid way of using python `threading.Thread`s, and that `run()` is a base method meant to behave somewhat like protected methods, i.e. can be overriden, but shouldn't be called directly. For example, attempting to `join` a thread like this will result in `RuntimeError: cannot join thread before it is started`.

However:
- The repro is still a completely valid Python program and doesn't fail
- The current instrumentation somehow breaks with frameworks like celery, maybe they run threads directly without starting them, and their framework works.

Yes

### Tip

dinmukhamedm:
# Description

Fixes an issue where a thread run without being started would result in an attribute error.

Options considered:
Option | Pros | Cons | Chosen
--- | --- | --- | ---
Call the wrapped function directly on error | Simplest option; least intrusive | Defeats the purpose of this instrumentation; Slight risk of missing other errors | No
Only instrument `run`, not `start` |  Simpler code; works for most default cases | `Thread.run` is designed to be overriden by custom child classes of `Thread`, in which case instrumentation __may__ not work | No
Capture the context at `run` if not captured at `start` | Most precise fix for the specific breaking use case | `Thread.run` is designed to be overriden by custom runners, in which case instrumentation __may__ not work | Yes

I am still open to any of the above options, and would appreciate review / further discussion.

Fixes #4245 

## Type of change

# How Has This Been Tested?

- [ ] Test that the original AttributeError is removed
- [ ] Test that properly started threads propagate context

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-threading/src/opentelemetry/instrumentation/threading/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-threading/src/opentelemetry/instrumentation/threading/__init__.py
@@ -147,12 +147,8 @@ class ThreadingInstrumentor(BaseInstrumentor):
     ) -> R:
         token = None
         try:
-            # `or` instead of default for type checking
-            ctx = (
-                getattr(instance, "_otel_context", None)
-                or context.get_current()
-            )
-            token = context.attach(ctx)
+            if hasattr(instance, "_otel_context"):
+                token = context.attach(instance._otel_context)
             return call_wrapped(*args, **kwargs)
         finally:
             if token is not None:


herin049:
It might be better to just move the context initialization into `__init__`

dinmukhamedm:
> It might be better to just move the context initialization into `__init__`

@herin049 I don't think that's the intended behaviour. It's common to initialize threads in one place in code and start them elsewhere. The intent is for the spawned thread to inherit the context of the place where it is `start`-ed, not where it is initialized.

Consider this pseudocode example

```python
worker_threads = []
for i in range(10):
    t = threading.Thread(target=lambda: notify_downstream())
    worker_threads.append(t)

# further initialization
# ...

for t in worker_threads:
    # we want to capture this context, not the one
    # above at initialization
    with tracer.start_as_current_span(name=f"process_{i}"):
        t.start() # or t.run()
```

Happy to discuss further, of course.

herin049:
> > It might be better to just move the context initialization into `__init__`
> 
> @herin049 I don't think that's the intended behaviour. It's common to initialize threads in one place in code and start them elsewhere. The intent is for the spawned thread to inherit the context of the place where it is `start`-ed, not where it is initialized.
> 
> Consider this pseudocode example
> 
> ```python
> worker_threads = []
> for i in range(10):
>     t = threading.Thread(target=lambda: notify_downstream())
>     worker_threads.append(t)
> 
> # further initialization
> # ...
> 
> for t in worker_threads:
>     # we want to capture this context, not the one
>     # above at initialization
>     with tracer.start_as_current_span(name=f"process_{i}"):
>         t.start() # or t.run()
> ```
> 
> Happy to discuss further, of course.

Yep, you're right here, not sure what I was thinking with my original suggestion.

That being said, one potential issue that we might encounter with the current fix is if `thread.run()` is called directly more than once. For example:

```python
with tracer.start_as_current_span(name="run_1"):
        t.run()
with tracer.start_as_current_span(name="run_2"):
        t.run()
```

With the current approach, the second call to `t.run()` would incorrectly have "run_1" as the parent span instead of "run_2". I doubt this will ever be an issue in practice, but the following fix does handle this scenario:


```python

token = None
try:
    if hasattr(instance, "_otel_context"):
        token = context.attach(instance._otel_context)
    return call_wrapped(*args, **kwargs)
finally:
    if token is not None:
        context.detach(token)
```

Furthermore, I also don't think it really makes sense to initialize `instance._otel_context` to `context.get_current()` if `thread.start()` was never called. It is probably more straightforward to simply leave it uninitialized and not attach another context. Let me know what you think.

dinmukhamedm:
> Furthermore, I also don't think it really makes sense to initialize instance._otel_context to context.get_current() if thread.start() was never called. It is probably more straightforward to simply leave it uninitialized and not attach another context. Let me know what you think.

I think it still make sense to attach in `run`, but not store it on the `instance`, sort of like what you did in your example. I'll push an update later today

instrumentation/opentelemetry-instrumentation-threading/src/opentelemetry/instrumentation/threading/__init__.py
```diff
@@ -147,7 +147,10 @@ def __wrap_threading_run(
     ) -> R:
         token = None
         try:
-            token = context.attach(instance._otel_context)
+            if getattr(instance, "_otel_context", None) is not None:
+                token = context.attach(instance._otel_context)
+            else:
+                token = context.attach(context.get_current())
```
xrmx: If the case is that you don't have the attribute you don't need the if and can just get the current context as default:

```suggestion
            ctx = getattr(instance, "_otel_context", context.get_current())
            token = context.attach(ctx)
```

instrumentation/opentelemetry-instrumentation-threading/tests/test_threading.py
```diff
@@ -14,9 +14,10 @@
 
 import threading
 from concurrent.futures import (  # pylint: disable=no-name-in-module; TODO #4199
+    Future,
     ThreadPoolExecutor,
 )
-from typing import List
+from typing import List, Union
```
xrmx: Instead of using unions please add `from __future__ import annotations` and then just use the `|`, e.g. `int | float`

instrumentation/opentelemetry-instrumentation-threading/src/opentelemetry/instrumentation/threading/__init__.py
herin049: Do we even want to attach a context if we're just attaching `context.get_current()` (which is essentially a no-op)

dinmukhamedm: yeah, good point. I will remove this

herin049:
LGTM!

xrmx:
@dinmukhamedm Please fix the conflicts, thanks

dinmukhamedm:
> Please fix the conflicts, thanks

done

xrmx:
@dinmukhamedm please pull main. Opening PRs from company repositories where we cannot push on our own is painful since for merging we require  the branch to include latest main.

dinmukhamedm:
@xrmx updated and invited you as an outside collaborator to the fork, just in case

Issue #4202, opened 2026-02-13
Document OTEL_SEMCONV_STABILITY_OPT_IN


xrmx:
All the http instrumentation now supports the `OTEL_SEMCONV_STABILITY_OPT_IN` env variable but we're not documenting it. We should document in each http instrumentation documentation that if they are setting the env var to:
- http - emit the new, stable HTTP and networking conventions, and stop emitting the old experimental HTTP and networking conventions that the instrumentation emitted previously.
- http/dup - emit both the old and the stable HTTP and networking conventions, allowing for a seamless transition.

And that by default the old experimental HTTP and networking conventions are emitted.

Also we should probably document this in opentelemetry.io somewhere.

rite7sh:
Hi! I’d like to work on this issue.
Quick clarification , should the `OTEL_SEMCONV_STABILITY_OPT_IN` section be added to **all HTTP-related instrumentation docs** (Flask, FastAPI, Requests, HTTPX, Django, ASGI/WSGI, etc.), or is there a specific subset you’d like covered?
Also, should this be documented only in the contrib instrumentation `.rst` files for now, or do you want a follow-up change in opentelemetry.io as well?
Happy to start once the scope is confirmed. Thanks

xrmx:
@rite7sh I would work to get one of your 3 already open PR merged before opening new ones, thanks.

rite7sh:
@xrmx affirmative, thanks for the heads up .

dipak0000812:
Hi @xrmx,
I checked the HTTP instrumentation docs (Flask, FastAPI, Requests, HTTPX) and I don’t see OTEL_SEMCONV_STABILITY_OPT_IN documented there.
I’d like to add a consistent section explaining the behavior for http and http/dup, and clarify the default behavior.
I’ll follow the existing env var documentation pattern.
Please let me know if I should limit this to contrib docs first or also prepare a follow-up for opentelemetry.io.

xrmx:
@dipak0000812 opentelemetry.io is another repo and we can update that only after the next release.

PR #4254, opened 2026-02-24
Document OTEL_SEMCONV_STABILITY_OPT_IN environment variable

thakoreh:
## Description

Fixes issue #4202 - Document the OTEL_SEMCONV_STABILITY_OPT_IN environment variable.

## Changes

Added a new subsection under Semantic Convention status explaining:
- How to use OTEL_SEMCONV_STABILITY_OPT_IN
- Available options (http, http/dup)
- Default behavior (old experimental conventions)
- Example usage with multiple opt-ins

## Documentation Added

- Explains http vs http/dup modes
- Shows example environment variable usage
- Documents default behavior


README.md
```diff
@@ -103,6 +103,31 @@ To release a package as `1.0` stable, the package:
...
 
+### Opting into stable semantic conventions
+
+You can control which semantic conventions are used by setting the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable. This is useful for transitioning from experimental to stable semantic conventions.
```
xrmx: I think this is the wrong place, I think we should have these to every http instrumentation rst documentation and maybe on opentelemetry.io

xrmx (changes requested):

README.md
tammy-baylis-swi: I agree with xrmx. It will be a bigger update, but the individual docs are a better place for details especially for users doing manual instrumentation. For example: one update in the [Flask readthedocs](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/flask/flask.html#configuration) already linked from [Flask readme](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-flask/README.rst)

thakoreh:
Thanks for the feedback @xrmx @tammy-baylis-swi! You're right - the documentation belongs in the individual instrumentation READMEs rather than the root README. I'll update this PR to add the env var documentation to the Flask instrumentation README and link it from there.

Let me refactor this to target the specific instrumentation docs.

emdneto:
Thanks for the PR!

pmcollins:
Thanks for taking this on. Looks like Django and Starlette need similar treatment -- can you take a look at those instrumentations?

Also, the instrumentation-specific docs are helpful, but should the overall behavior of `OTEL_SEMCONV_STABILITY_OPT_IN` also be documented? Could mention all of the values it accepts and the fact that it accepts comma separated values. Could be a separate PR.

thakoreh:
Thanks for the review. I updated the PR to add the same OTEL_SEMCONV_STABILITY_OPT_IN documentation to Django and Starlette, expanded the existing instrumentation docs to mention comma-separated opt-in values, and added a Towncrier changelog fragment.

I kept the broader/general documentation in this PR limited to the instrumentation docs, since the opentelemetry.io/global behavior docs may be better handled separately.

tammy-baylis-swi (approved):
Lgtm! Maintainers will also have to review

.changelog/4254.added
```diff
@@ -0,0 +1 @@
+`opentelemetry-instrumentation`: document OTEL_SEMCONV_STABILITY_OPT_IN
```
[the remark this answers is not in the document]
tammy-baylis-swi: This PR covers multiple instrumentations so maybe for conciseness:

```suggestion
`opentelemetry-instrumentation-*`: document `OTEL_SEMCONV_STABILITY_OPT_IN` for HTTP instrumentations
```

.changelog/4254.added
aabmass, line 1: Sorry for the churn but I think we can remove this altogether since it's a docs only change. I've updated the PR labels to skip the changelog check and you can delete this file

instrumentation/opentelemetry-instrumentation-aiohttp-client/src/opentelemetry/instrumentation/aiohttp_client/__init__.py
```diff
@@ -178,6 +178,27 @@ def response_hook(span: Span, params: typing.Union[
 Note:
     The environment variable names used to capture HTTP headers are still experimental, and thus are subject to change.
 
+Stable Semantic Conventions
```
aabmass, line 181: This snippet looks the same across all files. Would it work to move it to a separate RST file and use the [include directive](https://docutils.sourceforge.io/docs/ref/rst/directives.html#include)?

aabmass (approved):

Issue #1180, opened 2022-07-05
Thread-ConsumeBidirectionalStream caught unexpected exception 'generator' object has no attribute 'add_done_callback' and will exit.


mharoonkwentra:
**Notes**
We are trying to use Opentelemetry to instrument grpc to get traces of functions. We observed this error on local once, and we couldn't reproduce it again, but it's always on production server. Furthermore, we are using GCP monitoring suite (stack drive) to check traces.

**Environment:**
Django==3.2.6
Python==3.7

grpcio-tools==1.32.0
grpcio==1.33.2

opentelemetry-api==1.4.1
opentelemetry-sdk==1.4.1
opentelemetry-instrumentation-grpc==0.23b2

**Instrumentation code:**
        set_global_textmap(CloudTraceFormatPropagator())
        # sampling sends a portion of generated spans based on a 1/X population size
        sample_rate = int(os.getenv('SAMPLE_RATE', '1'))
        sampler = TraceIdRatioBased(1 / sample_rate)
        tracer_provider = TracerProvider(sampler=sampler)
        cloud_trace_exporter = CloudTraceSpanExporter()
        tracer_provider.add_span_processor(SimpleSpanProcessor(cloud_trace_exporter))
        trace.set_tracer_provider(tracer_provider)
        tracer = trace.get_tracer(__name__)
        GrpcInstrumentorClient().instrument() # when commented error isn't reproduced
except (ValueError, NotImplementedError, DefaultCredentialsError) as exc:
    print(exc)  # Handle errors here

**The error**
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/google/api_core/bidi.py", line 637, in _thread_main
    self._bidi_rpc.open()
  File "/usr/local/lib/python3.7/site-packages/google/api_core/bidi.py", line 287, in open
    call._wrapped.add_done_callback(self._on_call_done)
AttributeError: 'generator' object has no attribute 'add_done_callback'

AkselAllas:
I have the same issue. Not using django. Maybe a grpcio issue?

drice:
I've found this occurs when using StreamingPullManager from google-cloud-pubsub, as in this sample:

```python
import google.cloud.pubsub_v1 as pubsub_v1
from opentelemetry.instrumentation.grpc import GrpcInstrumentorClient


def sample_streaming_pull():
    # Create a client
    client = pubsub_v1.SubscriberClient()
    future = client.subscribe(
[...]

PR #4259, opened 2026-02-25
Fix gRPC client interceptor breaking bidirectional streaming (#1180)

juandelacruz-calvo:
## Summary

- Fix bidirectional streaming RPCs raising `AttributeError: 'generator' object has no attribute 'add_done_callback'` by routing bidi streams through `_intercept` instead of the generator-based `_intercept_server_stream` (issue #1180)
- Add regression test `test_stream_stream_preserves_call_interface` verifying the bidi stream response preserves the `grpc.Call` interface (`add_done_callback`, `cancel`, `is_active`)
- Add CHANGELOG entry under Unreleased/Fixed

## Root Cause

In `_client.py`, `intercept_stream()` routed all RPCs with `is_server_stream=True` into `_intercept_server_stream()`, including bidirectional streams (where both `is_client_stream` and `is_server_stream` are `True`). `_intercept_server_stream` is a generator function (uses `yield from`), so calling it returns a bare Python generator that lacks the `grpc.Call` interface (`add_done_callback`, `cancel`, `is_active`, etc.).

Downstream code — notably `google.api_core.bidi.BidiRpc.open()` used by Google Cloud Pub/Sub — expects a real gRPC call object and crashes:

```
AttributeError: 'generator' object has no attribute 'add_done_callback'
```

## The Fix

One condition change in `intercept_stream()`:

```python
# Before:
if client_info.is_server_stream:

# After:
if client_info.is_server_stream and not client_info.is_client_stream:
```

This ensures only unary-stream RPCs use the generator-based `_intercept_server_stream`. Bidi streams fall through to `_intercept`, which preserves the `grpc.Call`/`grpc.Future` interface.

| RPC Type | Before | After |
|---|---|---|
| Unary-Stream | `_intercept_server_stream` | `_intercept_server_stream` (unchanged) |
| Stream-Unary | `_intercept` | `_intercept` (unchanged) |
| Bidi (Stream-Stream) | `_intercept_server_stream` (broken) | `_intercept` (fixed) |

## Files Changed

| File | Change |
|---|---|
| `instrumentation/.../grpc/_client.py` | One-line condition fix |
| `instrumentation/.../tests/test_client_interceptor.py` | Add `test_stream_stream_preserves_call_interface` |
| `CHANGELOG.md` | Add bugfix entry under Unreleased/Fixed |

## Test plan

- [ ] Existing `test_stream_stream` continues to pass (span correctness)
- [ ] New `test_stream_stream_preserves_call_interface` passes — verifies `grpc.Call` attributes on bidi response
- [ ] Existing unary-stream tests pass (no regression from condition change)
- [ ] Existing stream-unary tests pass

Made with [Cursor](https://cursor.com)

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ For more information about the maintainer role, see the [community repository](h
 - [Jeremy Voss](https://github.com/jeremydvoss), Microsoft
 - [Keith Decker](https://github.com/keith-decker), Cisco/Splunk
 - [Liudmila Molkova](https://github.com/lmolkova), Grafana Labs
-- [Lukas Hering](https://github.com/herin049), Capital One
+- [Lukas Hering](https://github.com/herin049), Oracle
 - [Owais Lone](https://github.com/owais), Splunk
 - [Pablo Collins](https://github.com/pmcollins), Splunk
 - [Sanket Mehta](https://github.com/sanketmehta28), Cisco
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/CHANGELOG.md
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3715](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3715))
 - Add wrappers for OpenAI Responses API streams and response stream managers
   ([#4280](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4280))
+- Add async wrappers for OpenAI Responses API streams and response stream managers
+  ([#4325](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4325))
 
 ## Version 2.3b0 (2025-12-24)
 
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/response_wrappers.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/response_wrappers.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import logging
-from contextlib import ExitStack, contextmanager
+from contextlib import AsyncExitStack, ExitStack, contextmanager
 from types import TracebackType
 from typing import TYPE_CHECKING, Callable, Generator, Generic, TypeVar
 
@@ -53,6 +53,8 @@ if TYPE_CHECKING:
         ResponseStreamEvent,
     )
     from openai.lib.streaming.responses._responses import (
+        AsyncResponseStream,
+        AsyncResponseStreamManager,
         ResponseStream,
         ResponseStreamManager,
     )  # pylint: disable=no-name-in-module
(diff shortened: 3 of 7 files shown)


juandelacruz-calvo:
@xrmx

I have created an even simpler PR to fix the issues with PubSub that hopefully should be simpler to get approval for 🙏

juandelacruz-calvo:
For anyone needing a fix, this monkey patch should do:

```
    from opentelemetry.instrumentation.grpc._client import OpenTelemetryClientInterceptor

    _original_intercept_stream = OpenTelemetryClientInterceptor.intercept_stream

    def _patched_intercept_stream(self, request_or_iterator, metadata, client_info, invoker):
        if client_info.is_server_stream and client_info.is_client_stream:
            return self._intercept(request_or_iterator, metadata, client_info, invoker)
        return _original_intercept_stream(self, request_or_iterator, metadata, client_info, invoker)

    OpenTelemetryClientInterceptor.intercept_stream = _patched_intercept_stream
```

tammy-baylis-swi:
Thanks for contributing! Please could you make sure the existing tests all still pass or are adjusted. For example when `tox -e py313-test-instrumentation-grpc-1` there are 2 failures at status code asserts.

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py
```diff
@@ -60,6 +60,13 @@ def callback(response_future):
             code = response_future.code()
             if code != grpc.StatusCode.OK:
                 rpc_info.error = code
+                span.set_attribute(RPC_GRPC_STATUS_CODE, code.value[0])
+                span.set_status(
+                    Status(
+                        status_code=StatusCode.ERROR,
+                        description=str(code),
```
lmolkova: any need to do it? The code is already present in the attribute and in the new semconv version we'll even switch to string representation https://github.com/open-telemetry/semantic-conventions/blob/v1.40.0/docs/rpc/rpc-spans.md

lmolkova (approved):
LGTM @juandelacruz-calvo !

I'm working on stabilizing OTel RPC semantic conventions (which involves a bunch of changes to this instrumentation in opt-in more).

Would you be interested in taking a look at the prototype https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4279 ? Would really appreciate your thoughts.

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py
juandelacruz-calvo: I'll get back to you on this next week, I'm pretty tight at the moment, thanks @lmolkova

juandelacruz-calvo: I have updated the PR, hopefully it is ok now

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py
```diff
@@ -60,6 +60,12 @@ def callback(response_future):
             code = response_future.code()
             if code != grpc.StatusCode.OK:
                 rpc_info.error = code
+                span.set_attribute(RPC_GRPC_STATUS_CODE, code.value[0])
```
xrmx, line 63: Shouldn't tests be updated after this?

juandelacruz-calvo: Adding a negative test, thanks!

PR #4268, opened 2026-02-27
pika: Fix `messaging.destination` attribute

baodrate:
# Description

These two attributes are mistakenly swapped around. ~~destination (exchange/routing-key) was used as span name and~~ `messaging.destination` was set to consumer-tag 

## Type of change

# How Has This Been Tested?

```sh
tox run -e $(tox --listenvs | grep pika | tr '\n' ',')
```

Also, instrumented app with updated library and observed that `messaging.destination` is no longer `ctag...`

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/utils.py
+++ b/instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/utils.py
@@ -157,7 +157,7 @@ def _get_span(
         kind=span_kind,
     )
     if span.is_recording():
-        _enrich_span(span, channel, properties, task_name, operation)
+        _enrich_span(span, channel, properties, destination, operation)
     return span
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `_key_serializer`/`_value_serializer` directly instead of the internal `_serialize` method
   whose signature changed in 0.13 from `(topic, key, value)` to `(key, value, headers)`
   ([#4379](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4379))
+- `opentelemetry-instrumentation-pika`: Fix `messaging.destination` attribute
+  ([#4268](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4268))
 
 ### Breaking changes
 


instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/utils.py
```diff
@@ -148,11 +148,11 @@ def _get_span(
         return None
     task_name = properties.type if properties.type else task_name
     span = tracer.start_span(
-        name=_generate_span_name(destination, operation),
+        name=_generate_span_name(task_name, operation),
```
MikeGoldsmith: I don't think we should change this. The span name should be `{destination} {operation}` but this would change it.


```suggestion
        name=_generate_span_name(destination, operation),
```

MikeGoldsmith (changes requested):
Thanks @baodrate - sorry this got lost and took a while to get back to it.

The second change is all that's needed, we don't need to change the name. You'll also need to add a changelog entry.

instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/utils.py
baodrate: You're right, I think I was mistaken. But the existing behavior seems to contradict both the library code and the OpenTelemetry standard[^1]? (unless you mean we shouldn't change it to avoid unexpected changes for users)

[^1]: Disclaimer: I'm an opentelemetry noob.

I based this off the [signature of `_generate_span_name`](https://github.com/baodrate/opentelemetry-python-contrib/blob/8fec6b672eb7326c405fe65bbe574b3a6bf75157/instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/utils.py#L159-L161):

```python
def _generate_span_name(task_name: str, operation: Optional[MessagingOperationValues]) -> str:
```

But according to the docs for [_Semantic conventions for messaging spans_](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-name):

> The span name SHOULD be `{messaging.operation.name} {destination}` (see below for the exact definition of the [`{destination}`](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#destination-placeholder) placeholder).

And [_Semantic conventions for RabbitMQ_](https://opentelemetry.io/docs/specs/semconv/messaging/rabbitmq/):

> * **On the producer side**: `{exchange}:{routing key}` when both values are present and non-empty. When only one is available, only that value SHOULD be used. E.g., `{exchange}` or `{routing key}`. Otherwise: `amq.default` when the default exchange is used and no routing key is provided
> **On the consumer side**: `{exchange}:{routing key}:{queue}` when all values are present and non-empty. If any has an empty value (e.g., the default exchange is used) it SHOULD be omitted. For cases when `{routing key}` and `{queue}` are equal, only one of them SHOULD be used, e.g., `{exchange}:{routing key}`.

So if I'm reading correctly it seems the change should be something like:

```diff
diff --git a/instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/utils.py b/instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/utils.py
index 184cc5607..123792b2d 100644
--- a/instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/utils.py
+++ b/instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/utils.py
@@ -69,9 +69,8 @@ def _decorate_callback(
             tracer,
             channel,
             properties,
-            destination=(
-                method.exchange if method.exchange else method.routing_key
-            ),
+            # TODO: capture queue name in PikaInstrumentor._decorate_basic_consume()
+            destination=_get_dest(method.exchange, method.routing_key),
             span_kind=SpanKind.CONSUMER,
             task_name=task_name,
             operation=MessagingOperationValues.RECEIVE,
@@ -112,7 +111,7 @@ def _decorate_basic_publish(
             tracer,
             channel,
             properties,
-            destination=exchange if exchange else routing_key,
+            destination=_get_dest(exchange, routing_key),
             span_kind=SpanKind.PRODUCER,
             task_name="(temporary)",
             operation=None,
@@ -134,7 +133,6 @@ def _decorate_basic_publish(
 
     return decorated_function
 
-
 def _get_span(
     tracer: Tracer,
     channel: Optional[Channel],
@@ -148,7 +146,7 @@ def _get_span(
         return None
     task_name = properties.type if properties.type else task_name
     span = tracer.start_span(
-        name=_generate_span_name(task_name, operation),
+        name=_generate_span_name(destination, operation),
         kind=span_kind,
     )
     if span.is_recording():
@@ -157,11 +155,25 @@ def _get_span(
 
 
 def _generate_span_name(
-    task_name: str, operation: Optional[MessagingOperationValues]
+    destination: str, operation: Optional[MessagingOperationValues]
 ) -> str:
     if not operation:
-        return f"{task_name} send"
-    return f"{task_name} {operation.value}"
+        return f"send {destination}"
+    return f"{operation.value} {destination}"
+
+
+def _get_dest(exchange: Optional[str], routing_key: Optional[str], queue: Optional[str]):
+    parts: list[str] = []
+    if exchange:
+        parts.append(exchange)
+    if routing_key:
+        parts.append(routing_key)
+    if queue and queue != routing_key:
+        parts.append(queue)
+    if not parts:
+        parts.append("amq.default")
+    destination = ":".join(parts)
+    return destination
 
 
 def _enrich_span(
@@ -239,11 +251,7 @@ class ReadyMessagesDequeProxy(ObjectProxy):
                     self._self_tracer,
                     None,
                     properties,
-                    destination=(
-                        method.exchange
-                        if method.exchange
-                        else method.routing_key
-                    ),
+                    destination=_get_dest(method.exchange, method.routing_key),
                     span_kind=SpanKind.CONSUMER,
                     task_name=self._self_queue_consumer_generator.consumer_tag,
                     operation=MessagingOperationValues.RECEIVE,
```

Is there something I'm missing?

baodrate: In either case the call to `_generate_span_name` is wrong though. Fixed and pushed

baodrate:
fixed the call to `_generate_span_name()`, added the changelog line, and rebased on main. 

questions about semconv left [in comment thread](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4268#discussion_r3052558350)

instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/utils.py
emdneto: @baodrate This instrumentation was originally implemented based on semantic convention version 1.11.0. In newer versions, the guidance changed and now recommends using the format you mentioned earlier: {messaging.operation.name} {destination}.

We should eventually align this instrumentation with the latest semantic conventions once they become stable, or earlier if we decide to support the OTEL_SEMCONV_STABILITY_OPT_IN approach for messaging.

emdneto (commented):
I’m afraid I couldn’t clearly see what issue this change is fixing. Could you elaborate a bit more, perhaps by updating the existing tests or adding a new one to demonstrate the expected behavior?

instrumentation/opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/utils.py
baodrate: understood, thank you

emdneto:
Thanks for the PR!

PR #4281, opened 2026-03-03
GenAI Utils | Allow passing external Span to the manual LLMInvocation

Fiery-Fenix:
# Description

This PR allows to pass already active Span to `TelemetryHandler` via `LLMInvocation` instead of always creating a new Span during `start_llm` call.
This will allow to control Span lifecycle outside of GenAI utils which might be useful for some cases.

Primary goal for this change - to allow usage of GenAI Utills in `botocore` instrumentation for `bedrock` instrumentation. Main feature of `botocore` instrumentation that it consist of 2 layers:
- main instrumentation with wrappers implementation, which manage Span lifecycle (start/stop)
- "extensions" that are populating Span with useful data

To use `instrumentation-genai-util` in `botocore` instrumentation we'll either need to do a big refactoring of the `botocore` instrumentation to move Span lifecycle control to "extensions" or to allow passing external Span to `TelemetryHandler`, keeping control on the caller side.

This is also can be useful for instrumenting Streaming GenAI calls when Span lifecycle should be managed on the same level as stream lifecycle control.

## Type of change

# How Has This Been Tested?

New test was added.

```
$ tox -e py39-test-util-genai,py310-test-util-genai,py311-test-util-genai,py312-test-util-genai,py313-test-util-genai,py314-test-util-genai
  py39-test-util-genai: OK (12.91=setup[6.41]+cmd[6.51] seconds)
  py310-test-util-genai: OK (16.36=setup[7.09]+cmd[9.27] seconds)
  py311-test-util-genai: OK (18.19=setup[8.96]+cmd[9.23] seconds)
  py312-test-util-genai: OK (15.13=setup[8.66]+cmd[6.47] seconds)
  py313-test-util-genai: OK (18.22=setup[7.69]+cmd[10.52] seconds)
  py314-test-util-genai: OK (18.83=setup[9.22]+cmd[9.61] seconds)
```

- [x] No.

# Checklist:


util/opentelemetry-util-genai/tests/test_utils.py
```diff
@@ -861,6 +861,59 @@ def test_emits_llm_event_by_default_for_span_and_event(self):
...
+            name="external operation", kind=trace.SpanKind.INTERNAL
+        )
+
+        invocation = LLMInvocation(
+            span=external_span,
+            request_model="manual-model",
+            input_messages=[message],
+            provider="test-provider",
+            attributes={"external": True},
+        )
```
aabmass, line 755: Can you explain a little more on the use case? Since the span is returned already, why not just use use the returned span?

Fiery-Fenix: Sure, main purpose was to make adoption of `instrumentation-genai-utils` into `instrumentation-botocore` without big changes on botocore instrumentation side.

In botocore instrumentation Span is started [here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py#L319). Please take a note that this Span is started for any supported "extension" here, not only for Bedrock instrumentation, but also for S3/Lambda/DocumentDB/etc.
At the same time actual LLM Invocation instrumentation is located deeper - in [bedrock extension](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/bedrock.py) which is actually called few lines [below](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py#L327).
Please take a note that bedrock extension can handle both synchronous and stream invokes that bring another complexity for `instrumentation-genai-utils`.

My idea is to keep top-level instrumentation as is, keeping Span lifecycle management there, at the same time run TelemetryHandler.start_llm/TelemetryHandler.stop_llm exactly in bedrock extension using the provided active Span. This will keep code change more local and will not break code responsibility by introducing dependency on GenAI TelemetryHandler on the botocore instrumentation top level.

aabmass: Hmm I see. I don't love it for complicating the API, I think this will pretty much only be used for Bedrock.

It's a bit of an anti pattern to update the span name after starting the span, since the final name isn't available to samplers. But it is what it is. I'll defer to @xrmx who wrote the bedrock code though if he thinks it's worth refactoring.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -137,16 +137,20 @@ def start_llm(
     ) -> LLMInvocation:
         """Start an LLM invocation and create a pending span entry."""
         # Create a span and attach it as current; keep the token to detach later
-        span = self._tracer.start_span(
-            name=f"{invocation.operation_name} {invocation.request_model}",
-            kind=SpanKind.CLIENT,
-        )
+        # If the span is already set on the invocation - use it instead of creating a new one
+        # This allows users to control the span outside of the handler if needed
```
aabmass: Can we move it into the docstring instead?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -181,7 +187,9 @@ def fail_llm(  # pylint: disable=no-self-use
         _maybe_emit_llm_event(self._logger, span, invocation, error)
         # Detach context and end span
         otel_context.detach(invocation.context_token)
-        span.end()
+        # End the span only if it was created by the handler
+        if invocation.end_span_on_exit:
+            span.end()
```
aabmass: Does `end_span_on_exit` need to be public, or is it only used internally here?

Another option that might be cleaner is to update LLMInvocation to store an ExitStack for cleanup, since you can dynamically push callbacks onto it.

util/opentelemetry-util-genai/tests/test_utils.py
Fiery-Fenix: Actually, current implementation of the `TelemetryHandler` [forcedly overwrites](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py#L200) Span name after it's [creation](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py#L140), I haven't touched that part.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
Fiery-Fenix: As this field is used by `TelemetryHandler` to determine if it's required to close the span - I believe yes, it should be public field of `LLMInvocation`. Usage of this field is very similar to existing `LLMInvocation` field `context_token` - it's also seems to be private, but never used by `LLMInvocation` class itself, only by `TelemetryHandler`.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -138,18 +138,24 @@ def start_llm(
...
+                name=f"{invocation.operation_name} {invocation.request_model}",
+                kind=SpanKind.CLIENT,
+            )
+            invocation.span = span
+            invocation.end_span_on_exit = True
         # Record a monotonic start timestamp (seconds) for duration
         # calculation using timeit.default_timer.
         invocation.monotonic_start_s = timeit.default_timer()
-        invocation.span = span
         invocation.context_token = otel_context.attach(
```
wrisa: Do we want to skip attaching context in case of external span started as [start_as_current_span](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py#L319) will do it already ? This would be botocore specific ? What if the external span was started as `start_span`, can we detect that here and attach/not-attach depending on the way span was started?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -165,7 +171,9 @@ def stop_llm(self, invocation: LLMInvocation) -> LLMInvocation:  # pylint: disab
         _maybe_emit_llm_event(self._logger, span, invocation)
         # Detach context and end span
         otel_context.detach(invocation.context_token)
```
wrisa: In case of external span, is there a chance of span getting detached here sooner than intended ?

wrisa (commented):
Create a prototype and see if botocore/bedrock instrumentation works correctly with genAI utils.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
Fiery-Fenix: That's really a good point, I have added a few more checks in code (for Non-Recording Span and for Active Span).
Also added few more tests to cover this corner cases.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
Fiery-Fenix: If external Span is already attached to context (is active) - than no, I have added a special check for it.
As for not active Span - potentially it's possible, but the caller should do it **intentionally**.
The most interesting part that the same detachment could be done even in current implementation (on main branch):
```
invocation = LLMInvocation(...)
handler.start_llm(invocation)
#Here the harmful context detach part 
otel_context.detach(invocation.context_token)
handler.stop_llm(invocation) # the result is RuntimeError "Failed to detach context"
```

Unfortunately, at the moment I couldn't find a way how to prevent this, would be grateful for your help

util/opentelemetry-util-genai/tests/test_utils.py
lmolkova: We need to migrate all instrumentations to use genai utils properly sooner than later. There does not seem to be a good long-term need for external span on util's invocations and we should not introduce one just to minimize changes in botocore. Let's rip the bandaid and update to common utils asap.

PR #4286, opened 2026-03-03
Add Structlog Processor as Instrumentation

JWinermaSplunk:
# Description

Fixes https://github.com/open-telemetry/opentelemetry-python/issues/2993

## Type of change

# How Has This Been Tested?

- [x] test_structlog.py

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
```diff
@@ -0,0 +1,382 @@
...
+    51: SeverityNumber.FATAL2,
+    52: SeverityNumber.FATAL3,
+    53: SeverityNumber.FATAL4,
+}
+
+
+def std_to_otel(levelno: int) -> SeverityNumber:
+    """
+    Map python log levelno to OTel log severity number.
+    """
```
tammy-baylis-swi: This duplicates some of `logging` instrumentation. What about moving reusable helpers to `opentelemetry-instrumentation` at some point?

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
```diff
@@ -0,0 +1,382 @@
...
+    if levelno < 10:
+        return SeverityNumber.UNSPECIFIED
+    if levelno > 53:
+        return SeverityNumber.FATAL4
+    return _STD_TO_OTEL[levelno]
+
+
+class StructlogHandler:
```
tammy-baylis-swi: Could this be simplified by subclassing an existing type in `structlog` (which I don't know that well 🙂 )?

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
```diff
@@ -0,0 +1,382 @@
...
+    """
+    A structlog processor that translates structlog events into OpenTelemetry LogRecords.
```
tammy-baylis-swi: Nit: the docstrings could be more consistent in referring to this as a `handler` instead of a `processor`, the latter in my mind being something like SimpleLogRecordProcessor.

tammy-baylis-swi:
Thank you for undertaking this!

opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap_gen.py
```diff
@@ -188,6 +188,10 @@
         "library": "starlette >= 0.13",
         "instrumentation": "opentelemetry-instrumentation-starlette==0.62b0.dev",
     },
+    {
+        "library": "structlog ~= 21.1",
```
pmcollins: `>= 21.1` to match the version constraint in package.py?

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
```diff
@@ -0,0 +1,382 @@
...
+        # Insert before the last processor, assumed to be the renderer.
+        if current_processors:
+            insert_position = len(current_processors) - 1
+        else:
+            insert_position = 0
+
+        current_processors.insert(insert_position, processor)
+
+        # Reconfigure structlog with the new processor chain
+        structlog.configure(processors=current_processors)
```
pmcollins: During auto-instrumentation, instrumentors will run first, then the user's code will run, and if that code calls `structlog.configure`, looks like it will overwrite this config. The solution may be to monkeypatch `structlog.configure`.

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
JWinermaSplunk: Understood, would this be a good thing to expose in this PR or should we open a separate one?

tammy-baylis-swi: This PR might be a good time to introduce fewer lines overall.

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
JWinermaSplunk: Believe I resolved this by intercepting the original call, injecting the handler into the processors list, then calling the original once. But, I will leave this comment unresolved for context just in case I did not.

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
```diff
@@ -0,0 +1,353 @@
...
+
+from opentelemetry._logs import (
+    LogRecord,
+    NoOpLogger,
+    get_logger,
+    get_logger_provider,
+)
+from opentelemetry.context import get_current
+from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
+from opentelemetry.instrumentation.log_utils import std_to_otel
```
tammy-baylis-swi, line 39: I think this has not been checked in

tammy-baylis-swi:
I think a rebase is needed. This PR is now reverting some unrelated docker-test fixes ([this other PR](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4339))

JWinermaSplunk:
> I think a rebase is needed. This PR is now reverting some unrelated docker-test fixes ([this other PR](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4339))

Believe I went ahead and rebased, but feel free to let me know if it still looks off!

CHANGELOG.md
```diff
@@ -251,7 +251,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#3912](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3912))
 
 ### Added
-
+- `opentelemetry-instrumentation-structlog`: Add new package providing `StructlogHandler` and `StructlogInstrumentor` to bridge structlog into the OpenTelemetry Logs SDK pipeline with trace context correlation and exception capture.
```
tammy-baylis-swi: Something like this:

```suggestion
- `opentelemetry-instrumentation-structlog`: Add new package providing `StructlogHandler` and `StructlogInstrumentor` to bridge structlog into the OpenTelemetry Logs SDK pipeline with trace context correlation and exception capture. `opentelemetry-instrumentation`: adds shared `log_utils`
```

tammy-baylis-swi (approved):

dodofarm:
Hey team,
any timeline on this PR? This is a great addition that we've been anticipating for quite some time!

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
```diff
@@ -0,0 +1,384 @@
...
+            dt = datetime.fromisoformat(value)
+            if dt.tzinfo is None:
+                dt = dt.replace(tzinfo=timezone.utc)
+            return int(dt.timestamp() * 1e9)
+        except ValueError:
+            return None
+    return None
+
+
+class StructlogHandler:
```
pmcollins: Would it be preferable to call this a `StructlogProcessor` since that's what structlog calls them? Also, below, we have `processor = StructlogHandler(logger_provider=logger_provider)`.

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
```diff
@@ -0,0 +1,384 @@
...
+    """
+    if value is None:
+        return None
+    if isinstance(value, (int, float)):
+        return int(value * 1e9)
+    if isinstance(value, str):
+        try:
+            dt = datetime.fromisoformat(value)
+            if dt.tzinfo is None:
+                dt = dt.replace(tzinfo=timezone.utc)
```
pmcollins: Not sure we want this because we could be classifying a local timestamp as UTC here. For example, one of the default structlog processors is `TimeStamper(fmt="%Y-%m-%d %H:%M:%S", utc=False)` which doesn't include a `%z`, so the emitted string doesn't include a timezone component, tzinfo is None, and we'd be then shifting the time to UTC.

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
```diff
@@ -0,0 +1,384 @@
...
+
+        # Reconfigure structlog with the new processor chain
+        structlog.configure(processors=current_processors)
+
+        # Store reference for uninstrumentation
+        StructlogInstrumentor._processor = processor
+
+        # Wrap structlog.configure so that if user code calls it after
+        # instrumentation, the handler is re-inserted into the new chain.
+        StructlogInstrumentor._original_configure = structlog.configure
```
pmcollins: Can we convert these class variables to instance variables?

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
```diff
@@ -0,0 +1,384 @@
...
+
+        def _patched_configure(**kwargs):
```
pmcollins: nit: since this is a closure I don't think you need the leading underscore.

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
```diff
@@ -0,0 +1,384 @@
...
+            return None
+
+        structlog.configure = _patched_configure
+
+    def _uninstrument(self, **kwargs):
+        """
+        Remove the StructlogHandler from structlog's processor chain.
+        """
+        if StructlogInstrumentor._processor is None:
+            return
```
pmcollins: I'm not sure this test is necessary -- looks like the only way `_processor` can be None is if the user calls uninstrument without calling instrument, but seems like that's guarded against by the parent class. Which means the class variable can go away and patched_configure can just refer to `_processor` as a local (or just `processor`) since it's a closure.

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
```diff
@@ -0,0 +1,384 @@
...
+                if not any(
+                    isinstance(p, StructlogHandler) for p in processors
+                ):
+                    insert_position = max(len(processors) - 1, 0)
+                    processors.insert(
+                        insert_position, StructlogInstrumentor._processor
+                    )
+                    kwargs["processors"] = processors
+            original = StructlogInstrumentor._original_configure
+            if original is not None:
```
pmcollins: Can `original` be None here? Might be an unnecessary test.

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
```diff
@@ -0,0 +1,384 @@
...
+        # observed_timestamp is when the SDK received the event (always now).
+        # timestamp is when the event occurred; use the structlog "timestamp"
+        # field if present and parseable (UNIX float or ISO 8601 string),
+        # otherwise leave as None and let the SDK fill it in.
+        observed_timestamp = time_ns()
+        timestamp = _parse_structlog_timestamp(event_dict.get("timestamp"))
+
+        # Get the log level and map to OTel severity
+        level_str = event_dict.get("level", "info")
+        levelno = _STRUCTLOG_LEVEL_TO_LEVELNO.get(level_str.lower(), 20)
```
pmcollins: Looks like 20 maps to INFO -- might be more appropriate to use Otel's UNSPECIFIED level that maps to `0`. You may be able to use `SeverityNumber.UNSPECIFIED` in the API package.

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
```diff
@@ -0,0 +1,384 @@
...
+        attributes = {
+            k: v
+            for k, v in event_dict.items()
+            if k not in _STRUCTLOG_RESERVED_KEYS
+        }
+
+        # Handle exception information
+        exc_info = event_dict.get("exc_info")
+
+        if exc_info is True:
```
pmcollins, line 175: Might want to add a comment here that `is True` is required to prevent matching on a non-empty tuple or exception which are also truthy. (Pycharm linter is flagging this)

pmcollins (commented):
Hey Josh, this instrumentation seems to work fine, but I added some comments.

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
JWinermaSplunk: I might be wrong, but isn't this the same behavior as existing instrumentations that use `BaseInstrumentor` and therefore would be protected by that singleton guard?

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
JWinermaSplunk: Fair point, I can adjust the line below, but I went with handler instead of processor to follow the logging handler from logging instrumentation, I believe. Thoughts?

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
pmcollins: Yeah, my preference is for variables to have the narrowest scope possible, but since it doesn't appear to affect functionality, will defer to your preference.

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
pmcollins: The LoggingHandler name makes sense because it's a subclass of logging.Handler. In this case the class doesn't extend anything -- it's just a callable that structlog considers a processor. I don't see a need to match terminology with that library, but won't block on this if you disagree.

JWinermaSplunk:
commenting

emdneto:
Thanks for the PR!

aabmass:
@JWinermaSplunk there are some unresolved comment threads. Can you either resolve them with an explanation or reply for more feedback?

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
JWinermaSplunk: Hi @tammy-baylis-swi,
Apologies it took me awhile to look into this, but I believe Structlog processors are usually callables and not normally built by subclassing. So, the current approach should be consistent with the structlog-native pattern.

JWinermaSplunk:
Hi @aabmass,
Apologies, I believe they have all been resolved now but please let me know if not!

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+            # If the user is supplying a processors list, ensure our processor
+            # is included before passing it to the original configure.
+            StructlogInstrumentor._is_configured_by_app = True
+            if args and "processors" not in kwargs:
+                processors = args[0]
+                if processors is not None:
+                    args = (ensure_processor(processors), *args[1:])
+            elif kwargs.get("processors") is not None:
+                kwargs["processors"] = ensure_processor(kwargs["processors"])
+            return StructlogInstrumentor._original_configure(*args, **kwargs)
```
pmcollins, line 389: Looks like this code block is the same as the one on line 391. Can they be factored out?

pmcollins (commented):
What do you think about adding test coverage for both the oldest and latest supported `structlog` versions?

The package indicates `structlog >= 21.1`, but the test requirements pin `structlog~=21.1`, so CI only exercises the lower 21.x line. Looks like other instrumentations with broad dependency ranges split this into version-factor tox envs, for example `starlette-{oldest,latest}` with `test-requirements.oldest.txt` and `test-requirements.latest.txt`.

For this package, I think the same pattern would fit:

- `py3{10,11,12,13,14}-test-instrumentation-structlog-{oldest,latest}`
- `pypy3-test-instrumentation-structlog-{oldest,latest}`

instrumentation/opentelemetry-instrumentation-structlog/src/opentelemetry/instrumentation/structlog/__init__.py
JWinermaSplunk: resolved to `ensure_processor_in_args`

pmcollins (approved):
LGTM -- thanks @JWinermaSplunk.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/log_utils.py
```diff
@@ -0,0 +1,65 @@
...
+    48: SeverityNumber.ERROR4,
+    49: SeverityNumber.ERROR4,
+    50: SeverityNumber.FATAL,
+    51: SeverityNumber.FATAL2,
+    52: SeverityNumber.FATAL3,
+    53: SeverityNumber.FATAL4,
+}
+
+
+def std_to_otel(levelno: int) -> SeverityNumber:
```
xrmx, line 54: We should probably add a test for this

instrumentation/opentelemetry-instrumentation-structlog/tests/test_structlog.py
```diff
@@ -0,0 +1,783 @@
...
+        self.logger_provider = LoggerProvider()
+        self.logger_provider.add_log_record_processor(
+            SimpleLogRecordProcessor(self.exporter)
+        )
+
+        # Configure structlog with OTel processor
+        self.processor = StructlogProcessor(
+            logger_provider=self.logger_provider
+        )
+        structlog.configure(
```
xrmx, line 43: GPT suggests to also tests log emission without calling structlog.configure first:

```
+    def test_instrument_without_prior_structlog_configuration_emits_logs(self):
+        structlog.reset_defaults()
+
+        StructlogInstrumentor().instrument(
+            logger_provider=self.logger_provider
+        )
+
+        structlog.get_logger().info("zero config message")
+
+        logs = self.exporter.get_finished_logs()
+        self.assertEqual(len(logs), 1)
+        self.assertEqual(logs[0].log_record.body, "zero config message")
```

Not sure we have already a testcase we can put this in.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/log_utils.py
JWinermaSplunk: added `test_log_utils.py`!

xrmx (approved):

JWinermaSplunk:
Hi @xrmx,
Precommit was failing on merge from master, so I went ahead and reran and pushed. Should be good to go now.
Thank you!

xrmx:
> Hi @xrmx, Precommit was failing on merge from master, so I went ahead and reran and pushed. Should be good to go now. Thank you!

Puzzled by this since CI did run fine on main, anyway checked that `tox -e precommit` gives the very same result as yours. Thanks!

PR #4315, opened 2026-03-07
OpenAI v2: completion hook support and minor fixes

lmolkova:
From spec: https://github.com/open-telemetry/semantic-conventions/blob/v1.40.0/docs/gen-ai/gen-ai-spans.md#uploading-content-to-external-storage

- Call completion hook in telemetry handler (if passed in constructor)
- Add should_capture_content method to handler (so that instrumentations don't need to implement their own)
- Test against latest OpenAI and fix issues (new Omit type for values that were not provided)


util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -178,7 +229,17 @@ def fail_llm(  # pylint: disable=no-self-use
         _apply_error_attributes(invocation.span, error)
         error_type = getattr(error.type, "__qualname__", None)
         self._record_llm_metrics(invocation, span, error_type=error_type)
-        _maybe_emit_llm_event(self._logger, span, invocation, error)
+        log_record = _maybe_build_llm_event_record(span, invocation, error)
+        if self._completion_hook is not None:
```
JWinermaSplunk: Are 233-240 worth abstracting?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -159,7 +200,17 @@ def stop_llm(self, invocation: LLMInvocation) -> LLMInvocation:  # pylint: disab
         span = invocation.span
         _apply_llm_finish_attributes(span, invocation)
         self._record_llm_metrics(invocation, span)
-        _maybe_emit_llm_event(self._logger, span, invocation)
+        log_record = _maybe_build_llm_event_record(span, invocation)
+        if self._completion_hook is not None:
```
JWinermaSplunk: Are 204-211 worth abstracting?

instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/test_completion_hook.py
```diff
@@ -0,0 +1,130 @@
...
+    finally:
+        instrumentor.uninstrument()
+
+    default_hook.on_completion.assert_called_once()
+    kwargs = default_hook.on_completion.call_args.kwargs
+    assert kwargs["inputs"]
+    assert kwargs["outputs"]
+    assert kwargs["span"] is not None
+
+    # Content goes to the hook only — not to span attributes or log records
```
JWinermaSplunk: Does this comment look right?

JWinermaSplunk (approved):

instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/test_completion_hook.py
lmolkova: good catch!

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -243,9 +300,20 @@ def _fail(self, invocation: _T, error: Error) -> _T:
                 self._record_llm_metrics(
                     invocation, span, error_type=error_type
                 )
-                _maybe_emit_llm_event(
-                    self._logger, span, invocation, error_type
+                log_record = _maybe_build_llm_event_record(
+                    span, invocation, error_type
                 )
+                if self._completion_hook is not None:
+                    self._completion_hook.on_completion(
```
MikeGoldsmith: Should we be calling `self._call_completion_hook(invocation, span, log_record)` here instead? We do in `_stop`.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/span_utils.py
```diff
@@ -401,11 +399,12 @@ def _get_embedding_response_attributes(
 __all__ = [
     "_apply_llm_finish_attributes",
     "_apply_error_attributes",
+    "_maybe_build_llm_event_record",
     "_get_llm_common_attributes",
     "_get_llm_request_attributes",
     "_get_llm_response_attributes",
     "_get_llm_span_name",
-    "_maybe_emit_llm_event",
+    "_maybe_build_llm_event_record",
```
MikeGoldsmith: Did we mean to replace `_maybe_emit_llm_event`? There are two `_maybe_build_llm_event_record` entries now.

MikeGoldsmith (commented):
Sorry for delay getting this reviewed. It got lost and wasn't added to the PR review board.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
lmolkova: great catch!

MikeGoldsmith (approved):

lmolkova:
I'm going to put it on hold and rebase on top of https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4315 given the amount of conflicts it's going to introduce.

Nik-Reddy (commented):
Reviewed thoroughly - this is a clean and well-structured refactor. A few observations:

**What I like:**
- Moving `should_capture_content()` into `TelemetryHandler` is the right call. Individual instrumentations (like the Cohere one I'm adding in #4418) shouldn't need to manage content capture logic themselves - this centralizes it properly.
- The `_maybe_build_llm_event_record` rename (from `_maybe_emit_llm_event`) is a good separation of concerns - building the record vs emitting it lets the completion hook inspect it first.
- Handling `openai.Omit` in `value_is_set()` is a nice forward-compatibility fix for newer OpenAI SDK versions.
- `_openai_response_format_to_output_type` correctly maps `json_object`/`json_schema` to `json`.

**One question:**
In the `TelemetryHandler.__init__` content capture logic (lines ~164-177), the fallback for non-experimental mode checks `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT == "true"` only. Should this also handle the newer `span_only`/`event_only`/`span_and_event` values for consistency, or is that intentionally left as experimental-only?

**Note:** This PR supersedes my #4417 (content capture fix) - I'll close that one and reference this instead. Happy to help test once merged.

LGTM overall.

Nik-Reddy (approved):

lmolkova:
> In the TelemetryHandler.__init__ content capture logic (lines ~164-177), the fallback for non-experimental mode checks OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT == "true" only. Should this also handle the newer span_only/event_only/span_and_event values for consistency, or is that intentionally left as experimental-only?

@Nik-Reddy it makes sense, update the PR to support new valued on legacy path

MikeGoldsmith (approved):
Looks good - I think we need to address a couple of the copilot points:
- `_call_completion_hook()` is called inside `_apply_finish()`, we should add a try/except
- use `GenAiOutputTypeValues.JSON.value` instead of `"json"` in utils.py

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py
```diff
@@ -61,6 +63,7 @@ def __init__(
         self._tracer = tracer
         self._metrics_recorder = metrics_recorder
         self._logger = logger
+        self._completion_hook = completion_hook
         self._operation_name: str = operation_name
```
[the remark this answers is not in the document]
lmolkova: good concern, but we never released it, so it's ok

lmolkova:
@MikeGoldsmith thanks! I think completion hook should take care of exceptions - if it comes straight from user, we should be wrapping it into exception handling and then using here. We're going to call it a lot, it's easier to wrap once inside than make every caller handle errors. I can look into this in a separate PR.

UPDATE: Follow up: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4506

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/__init__.py
```diff
@@ -107,22 +104,19 @@ def _instrument(self, **kwargs):
...
-        content_mode = (
-            get_content_capturing_mode()
-            if latest_experimental_enabled
-            else ContentCapturingMode.NO_CONTENT
-        )
         handler = TelemetryHandler(
             tracer_provider=tracer_provider,
             meter_provider=meter_provider,
             logger_provider=logger_provider,
+            completion_hook=kwargs.get("completion_hook")
```
lzchen, line 146: Nit: Could you add a simple example in `examples` of a completion_hook being passed during instrumenation?

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/__init__.py
```diff
@@ -52,14 +52,11 @@
 from opentelemetry.metrics import get_meter
 from opentelemetry.semconv.schemas import Schemas
 from opentelemetry.trace import get_tracer
+from opentelemetry.util.genai.completion_hook import load_completion_hook
```
lzchen, line 90: I know it's not part of this PR but a few comments on `load_completion_hook`.

1. Can we log a warning when a name is passed into the envvar but no match is found in the entry point?
2. When a loaded object does not implement `CompletionHook`, let's use `logger.warning` instead of `debug` since this seems like a real configuration bug.
3. When an exception is caught, the log says the noop is being used by the iteration continues for the rest of the entry points (technically would be ok unless duplicate names).

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -105,6 +117,31 @@ def __init__(
             logger_provider,
             schema_url=schema_url,
         )
+        self._completion_hook = completion_hook or _NoOpCompletionHook()
```
lzchen, line 120: Nit: Should we have the `handler` more lean and put this logic into `utils`? Not a strong opinion.

```
self._capture_content = utils.content_enabled() or not isinstance(
            self._completion_hook, _NoOpCompletionHook
        )

def utils.content_enabled():
     return get_content_capturing_mode() != ContentCapturingMode.NO_CONTENT
```

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
```diff
@@ -105,6 +117,31 @@ def __init__(
...
+        else:
+            content_enabled = os.environ.get(
+                OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT, ""
+            ).lower() in (
+                "true",
+                "span_only",
+                "event_only",
+                "span_and_event",
+            )
+        self._capture_content = content_enabled or not isinstance(
```
lzchen, line 134: Is there any way we can make the configuration story more obvious to users?

To be able to get fully configured content, a user has to:

1. Set sem conv env var to experimental
2. Set capture message content env var to one of the capture modes
3. Set a completion hook entry point that points to a VALID completion hook.

I know we have the env vars listed in README of genai utils package but a user of an instrumentation package like this wouldn't really know to look there. wdyt?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py
```diff
@@ -219,4 +222,16 @@ def _apply_finish(self, error: Error | None = None) -> None:
         attributes.update(self._get_content_attributes_for_span())
         attributes.update(self.attributes)
         self.span.set_attributes(attributes)
+        self._call_completion_hook()
         self._metrics_recorder.record(self)
+
+    def _call_completion_hook(
```
lzchen: Nit: Can this be abstracted into `GenAiInvocation`? For tool and embedding can be noops

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py
```diff
@@ -219,4 +222,16 @@ def _apply_finish(self, error: Error | None = None) -> None:
...
+        self,
+    ) -> None:
+        self._completion_hook.on_completion(
```
lzchen: What if the hook raises an exception? I think `_finish` in `GenAiInvocation` only catches exceptions related to detaching context.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py
```diff
@@ -219,4 +222,16 @@ def _apply_finish(self, error: Error | None = None) -> None:
...
+            inputs=self.input_messages,
+            outputs=self.output_messages,
+            system_instruction=self.system_instruction,
```
lzchen, line 228: This might be on purpose or a dumb question but how come `tool_definitions` isn't being passed in here for `Agent` and `inference`?

lzchen (changes requested):
Overall LGTM but just some questions :)

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
lmolkova: Handler is the per-instrumentation config container - hook and content flag travel with it, no global state needed. Pulling content_enabled into utils means instrumentations either carry the flag themselves or re-read the env each time. What does the split buy us?

lmolkova: or did you mean to extract 
```
        self._completion_hook = completion_hook or _NoOpCompletionHook()
        if is_experimental_mode():
            content_enabled = (
                get_content_capturing_mode() != ContentCapturingMode.NO_CONTENT
            )
        else:
            content_enabled = os.environ.get(
                OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT, ""
            ).lower() in (
                "true",
                "span_only",
                "event_only",
                "span_and_event",
            )
        self._capture_content = content_enabled or not isinstance(
            self._completion_hook, _NoOpCompletionHook
        )
```

to a separate method?

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/__init__.py
lmolkova: Good catches! I'll open a follow-up PR for load_completion_hook

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py
lmolkova: I think we should not handle hook exceptions on every call-site. Proposing this https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4506 (wrapt user-provided callback and handle stuff there)

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py
lmolkova: ~~tool definitions are not really sensitive and we didn't support them in the hook - we can start to, but probably not in this PR?~~

Ah, i didn't even know it takes tool definitions

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
lzchen: Yeah the above can be put into something like `utils.is_content_enabled()` and then we can just call it via 
```
self._capture_content = utils.content_enabled() or not isinstance(
            self._completion_hook, _NoOpCompletionHook
        )
```
but it is a nit so feel free to ignore if we feel like it's fine!

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py
lzchen: Yeah separate pr problem thanks.

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/__init__.py
lmolkova: added custom_hook example

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
lmolkova: updated docstring and readme files with more details

util/opentelemetry-util-genai/src/opentelemetry/util/genai/handler.py
lmolkova: I think the right fix would be to make get_content_capturing_mode not raise if experimental mode was not enabled, I'd prefer to do it separately

lzchen (approved):

PR #4358, opened 2026-03-25
fix #4278: proxy all methods in ProxiedProducer/Consumer

JMGalvao:
## Description

Fixes #4278

`ProxiedProducer` and `ProxiedConsumer` inherited from the confluent-kafka C-extension classes (`Producer`/`Consumer`) but never called `super().__init__()`. Any method not explicitly defined on the proxy class fell through to the base class with an uninitialized C handle, causing `RuntimeError: Handle has been closed` / `Consumer closed`.

The fix removes inheritance and uses pure composition with `__getattr__` to delegate all undefined method calls to the wrapped instance.

## Type of Change

## Testing

Added two regression tests:
- `test_proxied_producer_delegates_undefined_methods` — calls `list_topics()` on `ProxiedProducer` and verifies it delegates to the underlying producer
- `test_proxied_consumer_delegates_undefined_methods` — calls `assignment()` on `ProxiedConsumer` and verifies it delegates to the underlying consumer

## Does This Require a Core Repo Change?

- No

## Checklist

- [x] Changelogs updated for any user-facing changes
- [x] Unit tests added covering the changes


tammy-baylis-swi:
Thanks for this! Please pull as I did a changelog resolve before triggering workflows.

Please then could you address the lint issues with:
* `tox -e lint-instrumentation-confluent-kafka`
* `tox -e precommit`

tammy-baylis-swi (approved):
Thanks again! Maintainers will also have a look

MikeGoldsmith (approved):

instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/__init__.py
```diff
@@ -151,7 +151,7 @@ def close(self):  # pylint: disable=useless-super-delegation
         return super().close()
 
 
-class ProxiedProducer(Producer):
```
xrmx, line 154: Isn't this going to break any isinstance() that may be used by downstream users? I think the right solution would be to use wrapt object proxies instead.

MikeGoldsmith: Yeah, I missed this before. This is right, it would be a breaking change and any `isinstance` calls would now return `False`.

I agree using wrapt would be a better solution.

MikeGoldsmith (changes requested):
As per @xrmx's comment, this would be a breaking change. We should consider using wrapt instead to keep the type hierarchy the same.

JMGalvao:
Addressed the feedback — replaced the `__getattr__` composition approach with `wrapt.ObjectProxy` as suggested.

**What changed:**
- `ProxiedProducer` and `ProxiedConsumer` now inherit from `wrapt.ObjectProxy`, so `isinstance(proxied, Producer)` / `isinstance(proxied, Consumer)` return `True` again
- Span-tracking state (`_current_consume_span`, `_current_context_token`) stored with `_self_` prefix (wrapt's convention for proxy-local attrs) and exposed via properties — no changes to `utils.py` needed
- All explicit delegation methods removed; `ObjectProxy.__getattr__` handles them automatically
- Tests updated to use `assertIsInstance` instead of `__class__` equality, with new assertions verifying `isinstance` compatibility

PR #4364, opened 2026-03-26
feat(opentelemetry-instrumentation-botocore): Instrument kinesis records

mosheshaham-dash0:
# Description

This PR adds tracecontext to kinesis PutRecord requests by injecting them into the message payload.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

ran manual testing and added unit testing

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/kinesis.py
```diff
@@ -0,0 +1,166 @@
...
+        stream_name = cls._extract_stream_name(call_context)
+        call_context.span_name = f"{stream_name} send"
+        attributes[SpanAttributes.MESSAGING_DESTINATION_NAME] = stream_name
+
+    @classmethod
+    def before_service_call(cls, call_context: _AwsSdkCallContext, span: Span):
+        cls._inject_span_into_entry(call_context.params)
+
+    @classmethod
+    def _inject_span_into_entry(cls, entry: MutableMapping[str, Any]):
```
herin049, line 94: This functionality should at the very least be opt-in. There is no guarantee that the user will be serializing data a JSON. Furthermore, from a performance/efficiency point of view, it is very wasteful to serialize JSON just to immediately deserialize it.

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/kinesis.py
```diff
@@ -0,0 +1,166 @@
...
+from opentelemetry.trace.span import Span
+
+_logger = logging.getLogger(__name__)
+
+################################################################################
+# Kinesis operations
+################################################################################
+
+
+class _KinesisOperation(abc.ABC):
```
herin049, line 39: I don't really see the value added here for introducing this `ABC` and the inheritance hierarchy, especially given that we are only instrumenting two operations here. In my opinion, this makes the implementation details much harder to reason about. Can we refactor this to just use a simple set of helper functions?

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/kinesis.py
```diff
@@ -0,0 +1,166 @@
...
+        for record in call_context.params.get("Records", []):
+            cls._inject_span_into_entry(record)
+
+
+################################################################################
+# Kinesis extension
+################################################################################
+
+_OPERATION_MAPPING: Dict[str, _KinesisOperation] = {
+    op.operation_name(): op
```
herin049: This seems overkill, I would just explicitly list out the operations here.

herin049 (changes requested):

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/kinesis.py
mosheshaham-dash0: you are right, i fixed it. (btw, this was copied from the sns.py insturmentation)

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/kinesis.py
mosheshaham-dash0: agree. introduced the `ENABLE_KINESIS_INSTRUMENTATION` env var. by default disabled.

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/kinesis.py
mosheshaham-dash0: actually, this pattern was copied from other instrumentations in the package, like sns, lambda, dynamodb. I don't think it's too complicated and i think it's better than instrumentations that don't use this pattern like bedrock.

Issue #3032, opened 2024-11-20
Create database client instrumentation docker-test


tammy-baylis-swi:
### What problem do you want to solve?

Current unit and integration tests for instrumentors are using `unittest` mock objects in place of the real database driver frameworks, e.g. [mysqlclient](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/4e992dd60a384184ed6f0ad91ae59b896efc38d6/instrumentation/opentelemetry-instrumentation-mysqlclient/tests/test_mysqlclient_integration.py#L31-L39), [psycopg2](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/4e992dd60a384184ed6f0ad91ae59b896efc38d6/instrumentation/opentelemetry-instrumentation-psycopg2/tests/test_psycopg2_integration.py#L69-L77). These let us check how the drivers are being called, but not how the drivers are behaving while being instrumented. The latter is especially important for testing existing and future sqlcommenting functionality.

### Describe the solution you'd like

Create a docker-test suite that will check existing db client instrumentor functionality such as:
1. sqlcommenting of query statements made by driver
2. span context matches sqlcomment content
3. `db.statement` attribute matches sqlcomment
4. configuration of above

sqlcommenting is currently implemented for:

1. [Django ORM](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/django/django.html#sqlcommenter)
2. [Flask 'ORM'](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/flask/flask.html#sqlcommenter)
3. [SQLAlchemy](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/sqlalchemy/sqlalchemy.html#sqlcommenter)
4. [psycopg2](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/psycopg/psycopg.html#sqlcommenter)
5. [psycopg](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/psycopg2/psycopg2.html#sqlcommenter)

### Describe alternatives you've considered

Neither MySQL nor PostgreSQL drivers can query sqlite3, else we could write unit/integration tests that read from file.

There is a 3rd party package I saw called [testing.mysqld](https://pypi.org/project/testing.mysqld/) but we're not currently using that dep and it's been a while since its last release so I'd hesitate in adding it.

### Additional Context

[...]

PR #4367, opened 2026-03-27
Remove duplicate query logging that breaks Django's assertNumQueries

federicobond:
The SqlCommenter _QueryWrapper manually appends the SQL string to queries_log when a CursorDebugWrapper is active. However, CursorDebugWrapper.debug_sql already logs the query as a dict with "sql" and "time" keys. This causes two problems:

1. Queries are double-counted in queries_log
2. The raw string entries cause a TypeError in Django's assertNumQueries, which expects dict entries with a "sql" key

Remove the manual append and let Django handle query logging.

## Type of change

# How Has This Been Tested?

- [x] Tested against a Django project that was failing a test case:
```
         with self.assertNumQueries(14):
              ^^^^^^^^^^^^^^^^^^^^^^^^^
     E   TypeError: string indices must be integers, not 'str'
```

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/sqlcommenter_middleware.py
```diff
@@ -122,8 +121,4 @@ def __call__(self, execute: Type[T], sql, params, many, context) -> T:
         #  * https://github.com/basecamp/marginalia/issues/61
         #  * https://github.com/basecamp/marginalia/pull/80
```
tammy-baylis-swi: While we're here, I think this comment block can be removed. They were associated with a line that already doesn't exist from https://github.com/open-telemetry/opentelemetry-python-contrib/pull/896/changes#diff-06ab7e9a7ae280afa68f406d92297deccf9c50c8adde710249825a8fb57709d9R105

tammy-baylis-swi:
Thanks for catching this! Could unit test coverage be added here on OTel Contrib side? Maybe it could borrow from your other-project test case that does the `self.assertNumQueries(14)`.

instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/sqlcommenter_middleware.py
federicobond: Done!

federicobond:
@tammy-baylis-swi I tried but found it hard to come up with a meaningful test that it does NOT mess with the query log. Could be something like this, but still feels a bit artificial:

```python
connection = MagicMock()
connection.queries_log = deque()
connection.force_debug_cursor = False

with CaptureQueriesContext(connection) as ctx:
    qw_instance = _QueryWrapper(requests_mock)
    execute_mock = MagicMock()
    context = {"cursor": MagicMock(), "connection": connection}
    qw_instance(
        execute_mock, "SELECT 1", MagicMock(), MagicMock(), context
    )

self.assertEqual(len(ctx), 0)
```

tammy-baylis-swi:
> @tammy-baylis-swi I tried but found it hard to come up with a meaningful test that it does NOT mess with the query log. Could be something like this, but still feels a bit artificial:

Agree that it's quite mock-y @federicobond , which reminds me that we've been wanting to add proper db client testing for a while but haven't gotten to it with all else going on: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3032

Maybe a quick unit test is better than nothing for now? Would be good to get other Approvers' thoughts.

federicobond:
I'm not sure it's worth it to add a test here to be honest. There was never a test for the original query log behavior to begin with.

tammy-baylis-swi (approved):

federicobond:
This issue is still relevant. Pull request is just waiting for a second approval

tammy-baylis-swi:
Maintainers, please have a look.

emdneto:
Thanks for the PR!

federicobond:
Done, let me know if I've correctly added the changelog entry.

lzchen (approved):

PR #4368, opened 2026-03-27
fix #4251: wrap BackgroundTasks in their own span

### LINKED ISSUE #4251: [fastapi] BackgroundTasks produce child spans that outlive their closed parent span
### Describe your environment

- OS: Windows 11
- Python version: 3.12.4
- `opentelemetry-instrumentation-fastapi`: 0.61b0.dev0 (main)
- `opentelemetry-instrumentation-httpx`: 0.61b0.dev0 (main)
- `opentelemetry-sdk`: 1.40.0.dev0 (main)
- FastAPI: 0.115.x

### What happened?

When using FastAPI's `BackgroundTasks`, any instrumented work that runs inside a background task (outbound HTTP calls, database queries, etc.) produces child spans whose `parent_id` points to the originating request span, which has already been closed by the time the background task executes.

This means child spans end later than their parent, producing a broken trace timeline in backends like Jaeger or Grafana Tempo. It also means there is no dedicated span wrapping the background task itself, so there is no way to 
distinguish work done during the request/response cycle from work done asynchronously after the response was sent.

### Steps to Reproduce

1. Install dependencies:
```
pip install opentelemetry-sdk fastapi uvicorn httpx
pip install opentelemetry-instrumentation-fastapi opentelemetry-instrumentation-httpx
```

2. Create your fast API app; e.g `app.py` with the following:
```python
import httpx
from fastapi import BackgroundTasks, FastAPI
from opentelemetry.sdk.trace.export import SimpleSpanProcessor, ConsoleSpanExporter

provider = TracerProvider()

app = FastAPI()


async def background_notify(user_id: int):
    async with httpx.AsyncClient() as client:
        await client.get("https://httpbin.org/get", timeout=5)


@app.post("/checkout")
async def checkout(background_tasks: BackgroundTasks):
    background_tasks.add_task(background_notify, user_id=42)
    return {"status": "processing"}
```

3. Run the server: `uvicorn app:app --port 8000`
4. Send a request: `curl -X POST http://localhost:8000/checkout`
5. Observe the span output in the server terminal.

### Expected Result

A dedicated span wrapping the background task execution should be created as a child of the request span. Any spans produced inside the task (httpx calls, database queries, etc.) should be children of that wrapper span, not direct children of the already-closed server span.

This would make it possible to identify background work as a distinct unit in a trace and would prevent child spans from extending past their parent's `end_time`.

### Actual Result

The `POST /checkout` server span closes as soon as the response is sent:
```json
{
    "name": "POST /checkout",
    "context": {
        "trace_id": "0xf48fb97cb6a9fe003ba52b729243ca43",
        "span_id": "0xe66e1014d69a4e10"
    },
    "kind": "SpanKind.SERVER",
    "start_time": "2026-02-23T15:02:25.666097Z",
    "end_time": "2026-02-23T15:02:25.819538Z"
}
```

The background task then produces an httpx span roughly 2 seconds later. That 
span's `parent_id` points to the already-closed server span:
```json
{
    "name": "GET",
    "context": {
        "trace_id": "0xf48fb97cb6a9fe003ba52b729243ca43",
        "span_id": "0x4762a09bd4a43e3e"
    },
    "kind": "SpanKind.CLIENT",
    "parent_id": "0xe66e1014d69a4e10",
    "start_time": "2026-02-23T15:02:27.077417Z",
    "end_time": "2026-02-23T15:02:28.871654Z"
}
```

The child span ends at `15:02:28`, roughly 2 seconds after its parent closed at `15:02:25`. There is also no span representing the background task itself.

### Additional context

The trace context is being carried into background tasks correctly via `contextvars` (the same `trace_id` is present in both the request handler and the background task). The problem is not context loss. The problem is that no wrapper span is created for the background task, so all work done inside it attaches directly to a parent that is already finished.

(If I remember correctly) A similar pattern is handled in `opentelemetry-instrumentation-celery`, where task execution is wrapped in its own span. Patching `starlette.background.BackgroundTask.__call__` inside the `FastAPIInstrumentor` instrument/uninstrument lifecycle would be one approach to solving this.

No

### Tip

davidgss04:
# Description

Fixes #4251 

FastAPI background tasks were not properly instrumented. While request spans were created correctly, background tasks executed after the response without their own span. As a result, any spans created inside a background task were incorrectly attached directly to the request span, which had already finished, leading to broken trace hierarchies and inaccurate timing information.

The fix patches `BackgroundTask.__call__` to wrap its execution in a dedicated span. This ensures that each background task runs within its own span, correctly parented to the request span, and that any spans created inside the task are properly nested.

The patch is applied only once using a guard (`hasattr`) to avoid double instrumentation, and the original method is restored during `uninstrument_app` to prevent side effects.

## Type of change

# Testing
Added three regression tests:

- `test_background_task_span_parents_inner_spans` — verifies that background tasks create a wrapper span and that spans created inside the task are correctly parented

- `test_uninstrument_app_restores_background_task_call` — ensures the original `BackgroundTask.__call__` is restored after uninstrumentation

- `test_background_task_span_not_duplicated_on_double_instrument_app` — verifies that repeated instrumentation does not create duplicate spans or apply the patch multiple times

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


tammy-baylis-swi:
Thanks for this! The fix to `Background` tracing makes sense to me and there is no semconv requirement to add HTTP attributes to that particular span. Metrics generated will be the same.

Please could you also commit fixes from:
* `tox -e precommit`
* `tox -e lint-instrumentation-fastapi`

tammy-baylis-swi:
Thanks @davidgss04 ! Forgot to also mention: please add an entry to `CHANGELOG.md` linking this PR.

davidgss04:
Good evening @tammy-baylis-swi,

I hope you're doing well. Although, I did not mention it before, I’m a final-year CS student and I’m working on this bug as part of my final project.

I noticed that `main` was merged into my branch, which introduced a merge commit authored by you. Unfortunately, for the purposes of my project evaluation, all commits in the branch need to be authored only by me, otherwise it affects my grading.

To address this, I’m planning to reset my branch to my last commit and force-push it to remove the merge commit, and then proceed with the update of the `CHANGELOG.md`.

Please let me know if you have any concerns with this approach.

Thank you again for your help and review!

tammy-baylis-swi (approved):
Thanks again! This lgtm. 👍  The maintainers will also have to review this.

tammy-baylis-swi (approved):

davidgss04:
Good evening @tammy-baylis-swi,

Regarding the failing tests, I noticed that they are all related to Python 3.9 environments. To investigate, I synced my local repository with `upstream/main` and ran the same failing tox environments locally. The failures occur during dependency resolution, as the current `opentelemetry-api` (from `main`) now requires Python >=3.10, making the Python 3.9 environments unsatisfiable.

Based on this, it seems the issue is not related to the changes in this PR, but rather to the current test configuration or dependency versions for Python 3.9.

I also observed that some recent passing PRs do not appear to run Python 3.9 tests.

Do you have any guidance on how this should be handled?

aabmass (commented):
I'm OK with this PR but I think I remember seeing something related to this in FastAPI upstream. If the background tasks are explicitly run by FastAPI with an isolated contextvars.Context instance, they will end up detached from the request by default. IIRC FastAPI made this the default behavior.

PR #4373, opened 2026-03-28
feat(asyncpg): add capture_connection_cleanup option to silence pool cleanup spans

RiyaChaturvedi37:
Closes #4264

# Description
Adds a new `exclude_queries` option to `AsyncPGInstrumentor` to allow 
silencing specific queries from being traced.

When using a connection pool, asyncpg executes cleanup statements after
each invocation:
- SELECT pg_advisory_unlock_all()
- CLOSE ALL
- UNLISTEN *
- RESET ALL

These can pollute traces. Users can now exclude them:

```python
AsyncPGInstrumentor(exclude_queries=AsyncPGInstrumentor._CLEANUP_QUERIES).instrument()
```

The option accepts any set of queries, making it flexible for other use cases too.

## Type of change


instrumentation/opentelemetry-instrumentation-asyncpg/tests/test_asyncpg_wrapper.py
```diff
@@ -144,3 +144,32 @@ async def exec_mock(*args, **kwargs):
 
         spans = self.memory_exporter.get_finished_spans()
         self.assertEqual(len(spans), 0)
+
+    def test_capture_connection_cleanup_false(self):
```
tammy-baylis-swi, line 151: Could we add more coverage for the `default` setting, and for when explicitly set to `True`?

CHANGELOG.md
tammy-baylis-swi, line 1: This diff is too big! 😅 Please could you revert to original state, then add one line to link this PR.

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
tammy-baylis-swi, line 1: Please could you update the documentation at the top of this file about the new config, which will render into [this doc](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/asyncpg/asyncpg.html).

tammy-baylis-swi:
Thank you for this @RiyaChaturvedi37 ! I have added some comments

CHANGELOG.md
RiyaChaturvedi37: > This diff is too big! 😅 Please could you revert to original state, then add one line to link this PR.

Fixed! Reverted to original state and added just one line.

instrumentation/opentelemetry-instrumentation-asyncpg/tests/test_asyncpg_wrapper.py
RiyaChaturvedi37: Added two more tests:
- `test_capture_connection_cleanup_true_by_default` - verifies default behavior
- `test_capture_connection_cleanup_explicitly_true` - verifies explicit True setting

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
RiyaChaturvedi37: Added usage example for `capture_connection_cleanup=False` in the module docstring.

RiyaChaturvedi37:
Note: The large diff (+2440/-2389) is due to Windows line ending (CRLF) differences in the existing file on my local machine. The actual code changes are minimal — only 3 files with the new feature, tests, and changelog entry. I apologize for the noise!

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
```diff
@@ -35,6 +35,7 @@
 
     # You can optionally pass a custom TracerProvider to AsyncPGInstrumentor.instrument()
     AsyncPGInstrumentor().instrument()
+    AsyncPGInstrumentor(capture_connection_cleanup=False).instrument()
```
tammy-baylis-swi: Nit: Just so it's extra clear that users shouldn't call `instrument()` twice:


```suggestion
    # OR opt into not tracing connection pool cleanup
    AsyncPGInstrumentor(capture_connection_cleanup=False).instrument()
```

tammy-baylis-swi (approved):
Thank you for fixes, lgtm! Just suggested a small docs update. The Maintainers will also have to review this.

RiyaChaturvedi37:
> Thank you for fixes, lgtm! Just suggested a small docs update. The Maintainers will also have to review this.

Done! Added the clarifying comment as suggested.

instrumentation/opentelemetry-instrumentation-asyncpg/src/opentelemetry/instrumentation/asyncpg/__init__.py
```diff
@@ -36,6 +36,9 @@
     # You can optionally pass a custom TracerProvider to AsyncPGInstrumentor.instrument()
     AsyncPGInstrumentor().instrument()
 
+    # OR opt into not tracing connection pool cleanup
+    AsyncPGInstrumentor(capture_connection_cleanup=False).instrument()
```
xrmx: Two thoughts on this:
- if we are going to make this instrumentation specific I think a better api could be to take a sequence of queries to match so that this can be used for other purposes as well, e.g. `AsyncPGInstrumentor(exclude_queries=AsyncPGInstrumentor._CLEANUP_QUERIES).instrument()`
- Have you thought on doing the filtering in the sampler? We added a RuleBasedSampler recently, if we have enough information in the span attributes that can be used as well to drop spans. A bit more overhead but can can be centrally configured and works for every instrumentation :)

RiyaChaturvedi37: Thank you for the feedback @xrmx! 

I like the `exclude_queries` approach as it's more flexible and can be used for other purposes too. I'll implement it as:
`AsyncPGInstrumentor(exclude_queries=AsyncPGInstrumentor._CLEANUP_QUERIES).instrument()`

Regarding the RuleBasedSampler approach - would you prefer I go that route instead, or should I proceed with the `exclude_queries` parameter?

xrmx: Discussed this at last Python SIG and it's fine to solve this in the instrumentation. RuleBasedSampler approach may be a bit too hardcore and drop also some downstream spans.

RiyaChaturvedi37:
Still actively working on this! Will resolve the CHANGELOG conflict shortly.

RiyaChaturvedi37:
Still actively working on this! Just resolved the CHANGELOG conflict and all checks are passing. 

Also updated the implementation based on @xrmx's feedback — replaced `capture_connection_cleanup` with the more flexible `exclude_queries` parameter:

```python
AsyncPGInstrumentor(exclude_queries=AsyncPGInstrumentor._CLEANUP_QUERIES).instrument()
```

This allows users to exclude any custom set of queries, not just the cleanup ones.

emdneto:
Thanks for the PR!

PR #4380, opened 2026-03-31
ci: add GHA to add PRs to project board when marked ready for review

MikeGoldsmith:
## Description

The [Python PR digest board](https://github.com/orgs/open-telemetry/projects/88) uses a built-in GitHub project workflow to auto-add PRs. However, that workflow only fires on PR \`opened\` and \`reopened\` events — it does not fire when a draft PR is converted to ready for review.

This means any PR opened as a draft and later marked ready for review is silently skipped and never added to the board.

This GHA covers the gap by triggering on \`opened\`, \`reopened\`, and \`ready_for_review\`, while skipping drafts. Since it uses \`actions/add-to-project\` (which is idempotent), PRs opened as non-drafts are safe — they'll just be a no-op on the second trigger.

Mirrors: open-telemetry/opentelemetry-python#5026

## Security: \`pull_request_target\` trigger

This workflow uses \`pull_request_target\` rather than \`pull_request\`. This is intentional: \`pull_request\` does not have access to repository secrets for fork PRs (\`Secret source: None\`), so the otelbot token would never resolve. \`pull_request_target\` runs in the base repo context and can access secrets.

The workflow contains **no \`actions/checkout\` step** — it never executes any code from the fork, only uses the PR node ID to make a single GitHub API call. See open-telemetry/opentelemetry-python#4955 for the full security discussion.

## Prerequisites for org admins

Before this workflow will function after merge, an org admin must grant otelbot the Projects permission:

- [x] Org admin grants otelbot **"Projects: read/write"** permission at https://github.com/organizations/open-telemetry/settings/apps/otelbot


MikeGoldsmith:
Please add the skip-changelog label.

tammy-baylis-swi (approved):
Sgtm, I fit into this use case. 🙂 Also fine with me that contrib and core PRs are on the same board.

MikeGoldsmith:
FYI we're waiting on the following issue to make sure the credentials the action runs as has the correct access:
- https://github.com/open-telemetry/community/issues/3351

MikeGoldsmith:
Updated with new credentials from community issue - this is ready to merge now 🎉

lzchen (approved):

PR #4399, opened 2026-04-06
Enabled flake8-print plugin rules for ruff linter

JayeshHire:
# Description

Enabled flake8-print plugin rules for ruff linter. These set of rules will throw errors when `print` statements or `pprint` statements are found in the code. If someone has to add debugs while development they should use logger to print the debugs. As sensitive data can be leaked via `print` statements to the stdout. `print` and `pprint` cannot be configured by the users like logs can be configured. This rule adds a check to warn contributors about their use of `print` and `pprint` statements inside the code and mandates them to remove these statements from the code.

1. scripts/*
2. \*\*/tests/*
3. \*\*/examples/*
4. opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py

contributes to [opentelemetry-python#4227](https://github.com/open-telemetry/opentelemetry-python/issues/4227) 

# How Has This Been Tested?

- [x] Test A
> This has been tested by running the below command
> uv run ruff check

# Checklist:


CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   
 ### Added
 
+- - Enabled the flake8-print plugin rules for ruff linter. These rules throw warnings over the use of `print` and `pprint` statements.
+([4399](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4399))
```
```suggestion
([#4399](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4399))
```

instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py
```diff
@@ -188,7 +188,7 @@ def _safe_end_processing_span(receipt_handle: str) -> None:
     @staticmethod
     def _extract_queue_name_from_url(queue_url: str) -> str:
         # A Queue name cannot have the `/` char, therefore we can return the part after the last /
-        return queue_url.split("/")[-1]
+        return queue_url.rsplit("/", maxsplit=1)[-1]
```
emdneto: Was that needed for this PR?

JayeshHire: That was the automatic change caused because of the `ruff check --fix`. This line was throwing `PLC0207` error code which checks for unnecessary splits of a string. I thought it's a normal change. But after your comment when saw it, this change is defeating the purpose of the function. Hence, in the recent commit I have undone my change and I have appended `# noqa: PLC0207` comment for this line.

emdneto:
Thanks for the PR!

PR #4414, opened 2026-04-12
fix(sdk-extension-aws): replace deprecated aws-auth ConfigMap check with JWT iss claim detection in AwsEksResourceDetector

alimx07:
# Description

Fixes `AwsEksResourceDetector` on EKS clusters using the Access Entries API, where the `aws-auth` ConfigMap no longer exists.

Previously `_is_eks()` made an HTTP request to the Kubernetes API to check for the `aws-auth` ConfigMap in the `kube-system` namespace. This request returns 404 on modern clusters that use the Access Entries API, causing detection to fail even when running on EKS.

The fix decodes the pod service-account JWT locally and checks whether the `iss` claim matches the EKS OIDC issuer pattern (`oidc.eks.*`). No network call is needed, the token is already present at the standard Kubernetes service-account path.

## Type of change

# How Has This Been Tested?

- Unit tests added covering:
  - EKS OIDC issuer in JWT `iss` claim → platform detected as `aws_eks`
  - Non-EKS issuer in JWT `iss` claim → returns empty resource
- All existing `AwsEksResourceDetectorTest` tests continue to pass (5/5)
- Manually verified on a live EKS 1.35 cluster with both IRSA and non-IRSA pods

- [x] No.

# Checklist:


sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/eks.py
```diff
@@ -32,6 +33,7 @@
 
 _TOKEN_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/token"
 _CERT_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+_EKS_OIDC_ISSUER = "oidc.eks."
```
MikeGoldsmith: This could capture other claims in the substring check, eg `evil.oidc.eks.com`. Maybe a tighter regex might work better?

Maybe something like:
`r'https://oidc\.eks\.[^.]+\.amazonaws\.com/'` (note `[^.]+` is intentional to match the region segment)

sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/eks.py
```diff
@@ -60,11 +62,17 @@ def _get_k8s_cred_value():
...
+    parts = cred_value.removeprefix("Bearer ").split(".")
+    if len(parts) != 3:
+        return False
+    try:
+        seg = parts[1]
+        payload = json.loads(
+            base64.urlsafe_b64decode(seg + "=" * (-len(seg) % 4))
+        )
+    except Exception:
+        return False
```
MikeGoldsmith, line 79: We should log that an error happened instead of silently ignoring.

sdk-extension/opentelemetry-sdk-extension-aws/tests/resource/test_eks.py
```diff
@@ -138,3 +151,52 @@ def test_if_no_eks_paths_should_not_raise(
...
+
+    @patch(
+        "opentelemetry.sdk.extension.aws.resource.eks._get_k8s_cred_value",
+        return_value=_bearer_jwt({"iss": "https://accounts.google.com"}),
+    )
+    @patch(
+        "opentelemetry.sdk.extension.aws.resource.eks._is_k8s",
+        return_value=True,
+    )
+    def test_non_eks_jwt_returns_empty(self, mock_is_k8s, mock_get_k8s_cred_value):
```
MikeGoldsmith: Can we add a test for a malformed JWT to test the exception handler?

MikeGoldsmith (changes requested):
Overall looks good - thanks @alimx07. I've left some suggestions we should address before accepting.

sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/eks.py
xrmx: Also I guess linting will fail here because the exception catched is too broad so you should limit to json decode error and base64 decode error.

alimx07: > Also I guess linting will fail here because the exception catched is too broad so you should limit to json decode error and base64 decode error.

Thanks @xrmx,  I tried a few scenarios and saw that `JSONDecodeError`, `UnicodeDecodeError`, and `binascii.Error` can all be raised here.
So I’m thinking of just catching ValueError since all of them inherit from it. That should still cover the same decoding/parsing failures and keeps things simpler without extra imports.

Do you think that’s fine (linter is fine with ValueError), or should I explicitly list the three?

alimx07:
@MikeGoldsmith @xrmx

Thanks for the review. I’ve added the suggested changes. let me know if there’s anything else you’d like me to adjust.

sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/eks.py
```diff
@@ -32,8 +34,10 @@
 
 _TOKEN_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/token"
 _CERT_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
-
-
+_EKS_OIDC_RE = re.compile(
+    r"^https://oidc\.eks\.[^.]+\.amazonaws\.com(?:\.cn)?/id/[A-F0-9]{32}$",
+    re.ASCII,
+)
```
MikeGoldsmith, line 40:
```suggestion
)

```

MikeGoldsmith (commented):
Looks good. I'd still like to see a test for invalid JWTs, eg incorrect number of parts (!= 3) or an invalid base64 encoding.

alimx07:
> Looks good. I'd still like to see a test for invalid JWTs, eg incorrect number of parts (!= 3) or an invalid base64 encoding.

@MikeGoldsmith  Done, I have added two tests simulating wrong parts count and invalid json payload.

MikeGoldsmith (approved):
Thanks @alimx07

alimx07:
Hi @MikeGoldsmith, just checking, I noticed the PR is still marked as `Reviewed PRs that need fixes` in the PR digest, not `approved`. Is there anything still pending or needs to be addressed on my side?

MikeGoldsmith:
> Hi @MikeGoldsmith, just checking, I noticed the PR is still marked as `Reviewed PRs that need fixes` in the PR digest, not `approved`. Is there anything still pending or needs to be addressed on my side?

Ah, no - just the automation didn't move it. I'll do it now 👍🏻

alimx07:
Just waiting for maintainers.

emdneto:
Thanks for the PR!

alimx07:
Thanks @emdneto, I have added the new changes.

sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/eks.py
```diff
@@ -49,11 +55,18 @@ def _get_k8s_cred_value():
 
 
 def _is_eks(cred_value):
-    return _aws_http_request(
```
emdneto, line 52: Is this method 100% removed or just deprecated? By changing this, we can ensure backwards compatibility for old clusters that only have the authentication mode as a CONFIGMAP or API_CONFIGMAP?

alimx07: This method is just deprecated as stated [here](https://docs.aws.amazon.com/eks/latest/userguide/auth-configmap.html), So `API_CONFIGMAP` or `CONFIG_MAP` still have `aws_auth` configmap while new `API` doesn't. 
Is this backward compatibile , I think yes as this leans on SA token which already found in pods (our logic in `is_k8s`) and according to my info, the `iss` will be always in the same format on managed eks.

alimx07: Hey @emdneto , I double check and I think we are okay.
Only one edge case: a pod with `automountServiceAccountToken: false` won't have the SA token mounted, but that's not an `is_eks` concern at this moment as the detector fails at the `is_k8s` step first.

Also the `collecter-contrib` repo adopted similar changes recently, see [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45885). they add some other env Vars signals first also, but `iss` already cover this, they are just short path.

emdneto: I'm fine then. Thank you

emdneto (approved):

PR #4417, opened 2026-04-13
fix(openai): Handle content capture with span_and_event mode

### LINKED ISSUE #4038: GenAI instrumentation not capturing content message
### Describe your environment

OS: MacOS Tahoe 26.1
Python version: Python 3.12
Package version: 0.60b1

### What happened?

We recently added an LLM mock service in the [OTel Demo](https://github.com/open-telemetry/opentelemetry-demo).
The [product-reviews service](https://github.com/open-telemetry/opentelemetry-demo/tree/main/src/product-reviews), it generates spans following the sem conv.
Initially I was seeing some spans with old sem conv, but when I set `OTEL_SEMCONV_STABILITY_OPT_IN=gen_ai_latest_experimental` it worked fine.

The issue is that message content is not part of the spans, even though we have set `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT= span_and_event` set.

Checking the [specs](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#gen-ai-input-messages), looks like we are following the [right pattern](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/product-reviews/product_reviews_server.py#L181-L184).
Are we missing anything?

### Steps to Reproduce

```
git clone git@github.com:open-telemetry/opentelemetry-demo.git
cd opentelemetry-demo
```

Modify the [`docker-compose.yml`](https://github.com/open-telemetry/opentelemetry-demo/blob/main/docker-compose.yml#L538) file to comment old format and add the new format:
```
      #- OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true
      - OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=span_and_event
      - OTEL_SEMCONV_STABILITY_OPT_IN=gen_ai_latest_experimental
```

start the demo

```
make start
```

Navigate to <http://localhost:8080/jaeger/ui/> and check the spans for service `product-reviews`. (it may take a minute)

### Expected Result

I expect seeing spans events with message content

### Actual Result

No span contains any content message events.

### Additional context

If you want to keep the demo running and just test changes on code or env vars of product-reviews, you can change the code and run:

```
make redeploy service=product-reviews
```

No

### Tip

Nik-Reddy:
# Description

When `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` is set to `span_and_event` (or other valid `ContentCapturingMode` values like `span_only`, `event_only`), the openai-v2 instrumentation silently drops all message content. This is because `is_content_enabled()` only checks for the string `"true"` — any other value is treated as disabled.

This affects the OTel Demo's product-reviews service and any user following the current semantic conventions documentation.

Fixes #4038

## Root Cause

`is_content_enabled()` in `utils.py`:
`python
return os.environ.get(OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT, "").lower() == "true"
`

The newer `ContentCapturingMode` enum values (`span_and_event`, `span_only`, `event_only`) were never recognized — only the legacy `"true"` string worked.

## Changes

- **`utils.py`**: Updated `is_content_enabled()` to recognize all valid `ContentCapturingMode` enum values as content-enabled, while preserving backward compatibility with the `"true"` string
- **`conftest.py`**: Added `span_and_event` test parameter to verify the legacy semconv path works with newer env var values

## Type of change

# How Has This Been Tested?

All 102 content-related tests pass (57 sync + 45 async), including the new `span_and_event` on legacy path test cases.

```
pytest tests/ -v -k content
```

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
@@ -37,6 +37,7 @@ from opentelemetry.semconv.attributes import (
 )
 from opentelemetry.trace.status import Status, StatusCode
 from opentelemetry.util.genai.types import (
+    ContentCapturingMode,
     InputMessage,
     LLMInvocation,
     OutputMessage,
@@ -55,7 +56,16 @@ def is_content_enabled() -> bool:
         OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT, "false"
     )
 
-    return capture_content.lower() == "true"
+    lower = capture_content.lower()
+    if lower == "true":
+        return True
+
+    # Support newer ContentCapturingMode enum values (e.g. span_and_event)
+    try:
+        mode = ContentCapturingMode[lower.upper()]
+        return mode != ContentCapturingMode.NO_CONTENT
+    except KeyError:
+        return False
 
 
 def extract_tool_calls(item, capture_content):
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/test_is_content_enabled.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/test_is_content_enabled.py
@@ -1,0 +1,70 @@
+# Copyright The OpenTelemetry Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests for is_content_enabled() covering all ContentCapturingMode values."""
+
+import pytest
+
+from opentelemetry.instrumentation.openai_v2.utils import (
+    OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT,
+    is_content_enabled,
+)
+
+
+@pytest.mark.parametrize(
+    "env_value",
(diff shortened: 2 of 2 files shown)


instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/conftest.py
```diff
@@ -118,7 +118,7 @@ def vcr_config():
 
 @pytest.fixture(
     scope="function",
-    params=[(True, "span_only"), (False, "True")],
+    params=[(True, "span_only"), (False, "True"), (False, "span_and_event")],
```
xrmx: This does not match the comment on the fixture. Maybe we want to do some unit testing on is_content_enabled or test that chat_completions_create_v_old / async_chat_completions_create_v_old are called with the proper value dependeing on the env var on instrument() instead.

Nik-Reddy: You're right — reverted the conftest change and added dedicated unit tests for \is_content_enabled()\ instead. New \	est_is_content_enabled.py\ covers all valid \ContentCapturingMode\ values (\span_and_event\, \span_only\, \vent_only\), legacy \	rue\/\alse\ strings, empty string, unset env var, and invalid values. 12 test cases total. Force-pushed with the fix.

lmolkova (commented):
I think it's being addressed in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4315, could you please review that PR instead?

Nik-Reddy:
Rebased on latest main. All review feedback addressed, reverted conftest changes and added 12 focused unit tests. Review thread resolved. Ready for review.

Nik-Reddy:
Closing this PR - @lmolkova's #4315 addresses the content capture issue as part of a larger refactor that moves `should_capture_content()` into the shared `TelemetryHandler`. This is a cleaner approach since it centralizes content capture logic for all instrumentations rather than fixing it per-instrumentation.

I've reviewed #4315 and left feedback there. The key change is that `TelemetryHandler` now owns content mode resolution, which also benefits new instrumentations like Cohere (#4418) that won't need to implement their own content capture logic.

Superseded by #4315.

Issue #3932, opened 2025-11-11
Metric: gen_ai.server.time_to_first_token


spurs30:
### What problem do you want to solve?

In Semantic Conventions 1.38.0, there is a standard definition for the TTFT (Time to First Transaction) metric in the metrics section. However, the current agent does not yet implement this metric. Is there any plan to implement this metric in the future? If not, what would be the approach to customizing the implementation of this metric?

### Describe the solution you'd like

1. Follow-up plan for achieving the TTFT metric.
2. Developer's approach to customizing the TTFT metric.

### Would you like to implement a fix?

None

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

xrmx:
First thing is that implemented stuff is per-instrumentation and not globally, second thing is that this is a *server* metric and AFAICS we only have client instrumentations.

spurs30:
Yes, currently, the agent only has client-side metrics. How can we define this metric for a specific instrumentation? For example, vLLM, SGLang.

Nik-Reddy:
Hi @spurs30, thanks for raising this. I've opened a PR that implements the `gen_ai.server.time_to_first_token` metric for the OpenAI v2 instrumentation: #4415.

To address @xrmx's point about this being a *server* metric: my implementation measures TTFT from the **client's perspective** , specifically, the elapsed time from request initiation to the first content chunk received during streaming. While this includes network latency and isn't a pure server-side measurement, it's the best approximation available from a client instrumentation and is the approach taken by other observability SDKs for this metric.

**What's implemented:**
- `gen_ai.server.time_to_first_token` histogram (seconds) with semconv v1.38.0 bucket boundaries
- Recorded only for successful streaming responses
- First-token detection in the stream wrapper, wired into both sync and async paths
- 8 test cases covering streaming, non-streaming exclusion, and tool-call streaming

For true server-side TTFT (e.g., vLLM, SGLang instrumentations), a separate server instrumentation would be needed, but this covers the common client-side use case.

Would appreciate a review when you get a chance. @xrmx, any feedback on the approach would be welcome as well.

Nik-Reddy:
[...]

PR #4419, opened 2026-04-13
feat: Add vLLM instrumentation with server-side TTFT/TPOT metrics

Nik-Reddy:
# Description

Implements OpenTelemetry instrumentation for [vLLM](https://github.com/vllm-project/vllm), the most widely used open-source LLM inference engine. This is a **server-side** instrumentation — the first in this repo — addressing the gap identified in #3932 where the community requested TTFT support for inference servers like vLLM and SGLang.

Unlike existing GenAI instrumentations in this repo (which are all client-side), this instruments the server/inference side, recording true server-side metrics that reflect actual model inference performance without network latency.

Ref #3932

## Changes

**New package:** `instrumentation-genai/opentelemetry-instrumentation-vllm/`

**Instrumented methods:**
| Method | Description |
|---|---|
| `vllm.LLM.generate()` | Offline/batch text generation |
| `vllm.LLM.chat()` | Chat completions |

**Server-side metrics (the key differentiator):**
| Metric | Type | Description |
|---|---|---|
| `gen_ai.server.time_to_first_token` | Histogram (s) | Time from request to first output token |
| `gen_ai.server.time_per_output_token` | Histogram (s) | Time per output token after the first |
| `gen_ai.server.request.duration` | Histogram (s) | Total server-side request duration |
| `gen_ai.client.operation.duration` | Histogram (s) | Operation duration |
| `gen_ai.client.token.usage` | Counter | Token usage (input/output) |

**Spans** use `SpanKind.SERVER` with full GenAI semantic convention attributes:
- `gen_ai.system` = `vllm`
- `gen_ai.operation.name` = `chat` / `generate`
- `gen_ai.request.model`, `gen_ai.request.max_tokens`, `gen_ai.request.temperature`
- `gen_ai.response.finish_reasons`, `gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens`

**Design decisions:**
- vLLM is an **optional** dependency — the package installs without GPU hardware and fails gracefully at `instrument()` time if vLLM is not available
- All tests are **mock-based** (no GPU required for CI)
- Bucket boundaries follow semconv v1.38.0 specification

## Type of change

# How Has This Been Tested?

35 mock-based tests covering spans, metrics (TTFT, TPOT, token usage, duration), error handling, uninstrument, and utility functions.

```
cd instrumentation-genai/opentelemetry-instrumentation-vllm
pytest tests/ -v
```

- [x] No.

# Checklist:

- [x] Documentation has been updated (README.rst included)

### DIFF
--- a/instrumentation-genai/opentelemetry-instrumentation-vllm/src/opentelemetry/instrumentation/vllm/__init__.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-vllm/src/opentelemetry/instrumentation/vllm/__init__.py
@@ -1,0 +1,105 @@
+# Copyright The OpenTelemetry Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+vLLM server-side instrumentation supporting `vllm`, it can be enabled by
+using ``VLLMInstrumentor``.
+
+.. _vllm: https://pypi.org/project/vllm/
+
+Unlike other GenAI instrumentations in this repository which are client-side,
+this is a **server-side** instrumentation for the vLLM inference engine.
+It records server-side metrics including time-to-first-token (TTFT) and
+time-per-output-token (TPOT).
+
+Metrics and span attributes follow the OpenTelemetry GenAI semantic conventions:
+https://opentelemetry.io/docs/specs/semconv/gen-ai/
+
+Usage
+-----
+
+.. code:: python
+
+    from vllm import LLM
+    from opentelemetry.instrumentation.vllm import VLLMInstrumentor
+
+    VLLMInstrumentor().instrument()
+
+    llm = LLM(model="meta-llama/Llama-2-7b-hf")
+    outputs = llm.generate(["Hello, world!"])
+
+API
+---
+"""
+
+import logging
+from typing import Collection
+
+from wrapt import wrap_function_wrapper
+
+from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
+from opentelemetry.instrumentation.vllm.package import _instruments
+from opentelemetry.metrics import get_meter
+from opentelemetry.trace import get_tracer
+
+from .instruments import Instruments
+from .patch import chat_wrapper, generate_wrapper
+
+logger = logging.getLogger(__name__)
+
+
+class VLLMInstrumentor(BaseInstrumentor):
+    """OpenTelemetry instrumentor for vLLM inference engine.
+
+    Wraps ``vllm.LLM.generate()`` and ``vllm.LLM.chat()`` to emit
+    OpenTelemetry spans and metrics following GenAI semantic conventions.
+
+    Key server-side metrics:
+      - ``gen_ai.server.time_to_first_token`` — TTFT histogram
+      - ``gen_ai.server.time_per_output_token`` — TPOT histogram
+      - ``gen_ai.client.operation.duration`` — operation duration
+      - ``gen_ai.client.token.usage`` — token usage histogram
+    """
+
+    def instrumentation_dependencies(self) -> Collection[str]:
+        return _instruments
+
+    def _instrument(self, **kwargs):
+        """Enable vLLM instrumentation."""
+        tracer_provider = kwargs.get("tracer_provider")
+        tracer = get_tracer(__name__, "", tracer_provider)
+
+        meter_provider = kwargs.get("meter_provider")
+        meter = get_meter(__name__, "", meter_provider)
+        instruments = Instruments(meter)
+
+        wrap_function_wrapper(
+            module="vllm",
+            name="LLM.generate",
+            wrapper=generate_wrapper(tracer, instruments),
+        )
+
+        wrap_function_wrapper(
+            module="vllm",
+            name="LLM.chat",
+            wrapper=chat_wrapper(tracer, instruments),
+        )
+
+    def _uninstrument(self, **kwargs):
+        import vllm  # pylint: disable=import-outside-toplevel
+
+        from opentelemetry.instrumentation.utils import unwrap
+
+        unwrap(vllm.LLM, "generate")
+        unwrap(vllm.LLM, "chat")
--- a/instrumentation-genai/opentelemetry-instrumentation-vllm/src/opentelemetry/instrumentation/vllm/patch.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-vllm/src/opentelemetry/instrumentation/vllm/patch.py
@@ -1,0 +1,234 @@
+# Copyright The OpenTelemetry Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Patch functions for vLLM instrumentation."""
+
+from __future__ import annotations
+
+from timeit import default_timer
+from typing import Any, List, Optional
+
+from opentelemetry.semconv._incubating.attributes import (
+    gen_ai_attributes as GenAIAttributes,
+)
+from opentelemetry.trace import SpanKind, Tracer
+
+from .instruments import Instruments
+from .utils import (
+    VLLM_SYSTEM,
+    extract_finish_reasons,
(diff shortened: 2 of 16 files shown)


instrumentation-genai/opentelemetry-instrumentation-vllm/src/opentelemetry/instrumentation/vllm/patch.py
```diff
@@ -0,0 +1,233 @@
...
+        span_attributes = get_request_attributes(
+            operation_name=operation_name,
+            model=model,
+            **sp_dict,
+        )
+
+        span_name = get_span_name(operation_name, model)
+        with tracer.start_as_current_span(
+            name=span_name,
+            kind=SpanKind.SERVER,
```
MikeGoldsmith: I think you're hosting the vLLM in-process, so I think SpanKind.INTERNAL is the right kind for that. We're not making a request to an external process / service / server.

instrumentation-genai/opentelemetry-instrumentation-vllm/src/opentelemetry/instrumentation/vllm/utils.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    metrics = getattr(request_output, "metrics", None)
+    if metrics is None:
+        return None, None
+
+    ttft = None
+    tpot = None
+
+    # vLLM stores these as attributes on the metrics object
+    if hasattr(metrics, "time_to_first_token"):
```
MikeGoldsmith: Are these attribute names correct? Looking at vLLM's metrics, I think these should be `first_token_latency` and `mean_time_per_output_token`.

instrumentation-genai/opentelemetry-instrumentation-vllm/src/opentelemetry/instrumentation/vllm/patch.py
```diff
@@ -0,0 +1,233 @@
...
+
+    # Client-side operation duration
+    instruments.client_operation_duration_histogram.record(
+        duration, attributes=metric_attributes
+    )
+
+    # Server-side request duration (same as client for local inference)
+    instruments.server_request_duration_histogram.record(
+        duration, attributes=metric_attributes
+    )
```
MikeGoldsmith: These metrics record the same duration - is that intentional? Is this something we expect to be different if we support other vLLM configurations?

instrumentation-genai/opentelemetry-instrumentation-vllm/src/opentelemetry/instrumentation/vllm/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        """Enable vLLM instrumentation."""
+        tracer_provider = kwargs.get("tracer_provider")
+        tracer = get_tracer(__name__, "", tracer_provider)
+
+        meter_provider = kwargs.get("meter_provider")
+        meter = get_meter(__name__, "", meter_provider)
+        instruments = Instruments(meter)
+
+        wrap_function_wrapper(
+            module="vllm",
```
MikeGoldsmith, line 88: `vllm` isn't a valid semconv value for `gen_ai.system` yet. Please can you open an issue / PR to have it added?

MikeGoldsmith (changes requested):
Looks like a good start - thanks @Nik-Reddy.

I've left some queries and suggestions. We also need to verify semconv includes vllm, plus add tox files, CI tests, changelog.

instrumentation-genai/opentelemetry-instrumentation-vllm/src/opentelemetry/instrumentation/vllm/patch.py
Nik-Reddy: You're right — changed to `SpanKind.INTERNAL` in both `generate_wrapper` and `chat_wrapper`. Updated tests accordingly.

instrumentation-genai/opentelemetry-instrumentation-vllm/src/opentelemetry/instrumentation/vllm/utils.py
Nik-Reddy: Good catch — renamed to `first_token_latency` and `mean_time_per_output_token` to match vLLM's actual metric names.

instrumentation-genai/opentelemetry-instrumentation-vllm/src/opentelemetry/instrumentation/vllm/patch.py
Nik-Reddy: Not intentional — TTFT (time from request to first token) and TPOT (mean inter-token latency) measure fundamentally different things. Clarified the docstring and added inline comments distinguishing the two metrics.

instrumentation-genai/opentelemetry-instrumentation-vllm/src/opentelemetry/instrumentation/vllm/__init__.py
Nik-Reddy: Added a TODO comment noting this needs a semantic-conventions issue/PR. Will open one to register `vllm` as a valid `gen_ai.system` value.

Nik-Reddy:
Hi @MikeGoldsmith, thank you for the thorough review! I've addressed all four comments:

1. Changed SpanKind to `INTERNAL` (not making external requests)
2. Renamed attributes to `first_token_latency` and \mean_time_per_output_token\ to match vLLM metrics
3. Clarified TTFT vs TPOT distinction with docstrings and inline comments
4. Added TODO noting `llm` needs to be registered in semantic conventions

Will open a semconv issue/PR for the `gen_ai.system` value separately. Ready for re-review!

Nik-Reddy:
Rebased on latest main. All 4 review threads from @MikeGoldsmith have been addressed:
- Fixed SpanKind to INTERNAL for server-side instrumentation
- Corrected attribute names to use semantic conventions
- Added documentation for TTFT/TPOT metric semantics
- Added semconv TODO for pending attribute stabilization

All threads resolved. Ready for re-review.

MikeGoldsmith:
Thanks for the updates @Nik-Reddy.

There are a few things we still to address:

For the client/server duration comment — I think there was a misunderstanding. My question was about `client_operation_duration_histogram` and `server_request_duration_histogram` both recording the same `duration` value (not about TTFT/TPOT). The comment in the code even says "same as client for local inference". Is recording both intentional, and do we expect them to diverge for other vLLM configurations?

Also:
- Can you share a link to the semconv issue/PR for registering `vllm` as a `gen_ai.system` value?
- The package is still missing `tox.ini` entries to wire it into CI. Please can you add those? You can use the one of the other libraries as a reference, eg `openai`.

lmolkova (changes requested):
@Nik-Reddy can you please discuss in https://cloud-native.slack.com/archives/C06KR7ARS3X before adding new instrumentations?

Adding new instrumentation requires finding component owners that are ready to maintain it in the long run.

I'm not an expert on vLLM side, but I don't think this instrumentation belongs in this repo. 

If you ask your AI to review this and ask it to question its usefullness, it'd give you something like this

vLLM already ships both natively:                                                                                                                                                                                                                                                                                                                                     
  - Prometheus /metrics endpoint with vllm:time_to_first_token_seconds, TPOT, iteration tokens, etc. (vllm/v1/metrics/loggers.py)                                                                                                                                                                                                                                       
  - OpenTelemetry tracing in vllm/tracing/utils.py with GEN_AI_LATENCY_TIME_TO_FIRST_TOKEN — emitted from vllm/v1/engine/output_processor.py                                                                                                                                                                                                                            
  - Docs: https://github.com/vllm-project/vllm/blob/main/docs/usage/metrics.md                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                        
  What the PR actually instruments:                                                                                                                                                                                                                                                                                                                                     
  - vllm.LLM.generate() and vllm.LLM.chat() — these are the offline/batch Python API, not the serving path (vllm serve → OpenAI-compatible HTTP server). So the "server-side instrumentation — the first in this repo" framing is misleading; it's still an in-process library wrap, just of a different entry point.                                                   
                                                                                                                                                                                                                                                                                                                                                                        
  Resulting issues:
  1. Duplicates metrics vLLM already emits, with different names/semantics (risk of drift from upstream).                                                                                                                                                                                                                                                               
  2. Misses the serving code path entirely — where TTFT/TPOT actually matter in production.                                                                                                                                                                                                                                                                             
  3. Reviewer already flagged gen_ai.system=vllm isn't registered in semconv, missing tox/CI wiring, and duplicate client/server duration recording.                                                                                                                                                                                                                    
  4. Upstream is the natural home for this: improving vLLM's built-in OTel emitter (which already exists) would benefit everyone without a monkey-patch layer.                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                        
  The right path is probably: contribute missing semconv attributes upstream to vLLM, not add a wrapper here. Worth raising on the PR/issue #3932 before more work goes in.

Nik-Reddy:
@MikeGoldsmith All three points addressed.

Removed server_request_duration. It was recording the same value as client_operation_duration for local vLLM inference, so it was purely redundant.

Added semconv links in the module docstring and next to the gen_ai.system TODO, so the rationale for each attribute is traceable.

Wired tox.ini following the same pattern used by the other GenAI instrumentations, with oldest and latest requirement files. Tests pass in both configurations.

lmolkova (changes requested):
@Nik-Reddy you did not address points from my last comments. Please don't re-request review unless you tried to.

lmolkova:
thanks for the contribution @Nik-Reddy !

We're moving GenAI instrumentations to the new repository - https://github.com/open-telemetry/opentelemetry-python-genai. Before reopening it in the repo, please  explore vLLM native instrumentation.

PR #4423, opened 2026-04-13
fix(confluent-kafka): populate bootstrap.servers span attributes

### LINKED ISSUE #4104: [confluent-kafka] Dead code: extract_bootstrap_servers is never called, missing broker address in spans
## Describe the bug

The `KafkaPropertiesExtractor.extract_bootstrap_servers()` method exists in `utils.py` but is never called anywhere in the instrumentation. As a result, Kafka spans are missing broker address attributes (`messaging.url`, `server.address`, `server.port`).

## Steps to reproduce

1. Instrument a Python application using `opentelemetry-instrumentation-confluent-kafka`
2. Produce or consume messages to/from Kafka
3. Inspect the generated spans

## Expected behavior

Kafka spans should include broker address information:
- `messaging.url` with the full bootstrap servers string
- `server.address` with the primary broker host
- `server.port` with the broker port

## Actual behavior

Kafka spans only contain:
- `messaging.system`
- `messaging.destination`
- `messaging.kafka.partition`
- `messaging.destination_kind`
- `messaging.operation`

No broker address information is available.

## Evidence

In `utils.py`, the `KafkaPropertiesExtractor` class has:

```python
class KafkaPropertiesExtractor:
    @staticmethod
    def extract_bootstrap_servers(instance):
        return instance.config.get("bootstrap_servers")  # Never called!
```

But this method is never invoked in `__init__.py` where spans are created.

### Trace Data Example

Here's an actual Kafka span from a production environment:

```json
{
  "messaging.system": "kafka",
  "messaging.destination": "backend_log",
  "messaging.kafka.partition": 0,
  "messaging.destination_kind": "queue",
  "messaging.operation": "receive"
}
```

The broker address (e.g., `b-1.example-kafka.amazonaws.com:9092`) is completely missing.

## Impact

- **Service dependency graphs** (e.g., Grafana Service Map with Tempo) cannot show Kafka broker endpoints
- **Multi-cluster Kafka environments** cannot be distinguished in observability tools
- **Chaos engineering tools** cannot determine which Kafka cluster is being targeted
- **Dependency analysis** cannot determine external service addresses for Kafka

## Proposed fix

1. Pass `bootstrap_servers` to `_enrich_span()`
2. Set `messaging.url`, `server.address`, and `server.port` attributes

Example fix in `_enrich_span()`:

```python
def _enrich_span(
    span,
    topic,
    partition: Optional[int] = None,
    offset: Optional[int] = None,
    operation: Optional[MessagingOperationValues] = None,
    bootstrap_servers: Optional[str] = None,  # Add parameter
):
    # ... existing code ...
    
    # Add: Set bootstrap servers attributes
    if bootstrap_servers:
        span.set_attribute(SpanAttributes.MESSAGING_URL, bootstrap_servers)
        first_broker = bootstrap_servers.split(",")[0].strip()
        if ":" in first_broker:
            host, port = first_broker.rsplit(":", 1)
            span.set_attribute(SpanAttributes.SERVER_ADDRESS, host)
            span.set_attribute(SpanAttributes.SERVER_PORT, int(port))
        else:
            span.set_attribute(SpanAttributes.SERVER_ADDRESS, first_broker)
```

And call `extract_bootstrap_servers()` in `wrap_produce()`, `wrap_poll()`, `wrap_consume()`.

## Environment

- `opentelemetry-instrumentation-confluent-kafka` version: 0.50b0 (also verified in v0.60b1)
- Python version: 3.11
- confluent-kafka version: 2.x

## Would you like a PR?

I'm happy to submit a PR to fix this issue if the approach sounds reasonable.

alliasgher:
## Description

`KafkaPropertiesExtractor.extract_bootstrap_servers` was defined in `utils.py` but was never actually invoked anywhere in the instrumentation, so confluent-kafka spans were missing `messaging.url`, `server.address`, and `server.port` even though the extraction helper existed for it. This PR wires the helper up end-to-end.

Changes:

- `AutoInstrumentedProducer` / `AutoInstrumentedConsumer` now capture the config dict passed to their constructor, and `ProxiedProducer` / `ProxiedConsumer` surface the wrapped instance's config if one is present.
- `extract_bootstrap_servers` accepts both `"bootstrap.servers"` (the dotted key confluent-kafka uses) and `"bootstrap_servers"`, tolerates list-valued configs, and returns `None` when no config is attached instead of raising `AttributeError`.
- `wrap_produce` / `wrap_poll` / `wrap_consume` now pass the bootstrap servers through to `_enrich_span`, which sets `messaging.url`, `server.address`, and `server.port` (parsed from the first broker in the list).

Producer spans and consumer `<topic> process` spans both get the new attributes.

Fixes #4104

## Type of change

## How Has This Been Tested?

- Added `test_producer_sets_bootstrap_servers_attributes` and `test_consumer_sets_bootstrap_servers_attributes` that exercise the proxy path end-to-end and assert the new attributes.
- Ran the full `test_instrumentation.py` suite (16 tests pass).

## Does This PR Require a Contrib Repo Change?

- [x] No.

## Checklist:


instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/utils.py
```diff
@@ -115,12 +129,37 @@ def _get_links_from_records(records):
     return links
 
 
+def _set_bootstrap_servers_attributes(span, bootstrap_servers):
+    """Populate messaging.url, server.address, server.port from a
+    bootstrap.servers string (e.g. ``host1:9092,host2:9092``)."""
+    if not bootstrap_servers:
+        return
+
+    span.set_attribute(SpanAttributes.MESSAGING_URL, bootstrap_servers)
```
tammy-baylis-swi: Thanks for submitting this!

I know the issue asks for `messaging.url` but it's [deprecated](https://github.com/open-telemetry/opentelemetry-python/blob/9c48362c1d29acc74c333354c2c314bd3b71ef8e/opentelemetry-semantic-conventions/src/opentelemetry/semconv/trace/__init__.py#L1309-L1312) so I don't think we should add it now. Just keep adding `server.address` and `server_port`.

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- `opentelemetry-instrumentation-confluent-kafka`: Populate `messaging.url`, `server.address` and `server.port` span attributes from the producer/consumer `bootstrap.servers` config; previously `KafkaPropertiesExtractor.extract_bootstrap_servers` was defined but never called
+  ([#4104](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4104))
```
tammy-baylis-swi: Repo convention is to link the PR


```suggestion
  ([#4423](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4423))
```

instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/utils.py
alliasgher: Done, removed messaging.url.

CHANGELOG.md
alliasgher: Done.

instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/__init__.py
```diff
@@ -367,10 +394,14 @@ def wrap_produce(func, instance, tracer, args, kwargs):
...
             )
+            bootstrap_servers = (
+                KafkaPropertiesExtractor.extract_bootstrap_servers(instance)
+            )
             _enrich_span(
                 span,
                 topic,
                 operation=MessagingOperationTypeValues.RECEIVE,
+                bootstrap_servers=bootstrap_servers,
             )  # Replace
```
MikeGoldsmith: Should this be a publish instead of receive?

```suggestion
            _enrich_span(
                span,
                topic,
                operation=MessagingOperationTypeValues.PUBLISH,
                bootstrap_servers=bootstrap_servers,
            )  # Publish
```

MikeGoldsmith (approved):
Looks good, just noticed one thing that wasn't changed here but very close and would be good to fix. We can do in a separate PR if you prefer.

instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/__init__.py
alliasgher: Applied in this PR since it is the same file — the second `_enrich_span` inside `wrap_produce` also runs in a producer span and now uses `PUBLISH` to match the first path and the semantic conventions.

tammy-baylis-swi (approved):
Thanks for revisions!

xrmx:
This clashes with #4358

alliasgher:
Thanks @xrmx — you're right, #4358 restructures `ProxiedProducer`/`ProxiedConsumer` away from inheritance to a `__getattr__` proxy. My diff only adds a `self.config = getattr(..., "config", None)` line to each of their `__init__`s, so the two should coexist cleanly, but happy to wait for #4358 to merge first and then rebase this on top so there's no merge mess.

Let me know if you'd prefer the opposite order.

PR #4433, opened 2026-04-14
fix(flask): wrap wsgi_app call in try/except to prevent active_requests gauge leak

alliasgher:
## Description

Wrap the `wsgi_app` call in `try/finally` so that `http.server.active_requests` is decremented when the wrapped WSGI app raises. Without this, every exception leaves the gauge permanently above the real number of in-flight requests, causing steady drift upward over the process lifetime.

Fixes #4403

## Type of change

## How Has This Been Tested?

Added `test_active_requests_decremented_on_error` in `test_programmatic.py` which drives a handler that raises and asserts the gauge ends at 0.

## Checklist


liyaka:
Thanks for the quick fix! We hit this exact bug in production — the leaked gauge was preventing our Kubernetes HPA from scaling down.

One suggestion: consider using `try/finally` instead of `try/except` to keep the decrement in a single code path. With the current approach, `active_requests_counter.add(-1, ...)` exists in two places (the `except` block and line 458), which is easy to get out of sync during future refactors.

The WSGI instrumentation already uses this pattern:

```python
# wsgi/__init__.py
try:
    ...
    return _end_span_after_iterating(iterable, span, token)
except Exception as ex:
    raise
finally:
    self.active_requests_counter.add(-1, active_requests_count_attrs)
```

For Flask it would look like:

```python
try:
    result = wsgi_app(wrapped_app_environ, _start_response)
    # ...duration recording unchanged...
    return result
finally:
    active_requests_counter.add(-1, active_requests_count_attrs)
```

This also removes the original decrement at line 458, so there's exactly one decrement path.

Also worth adding a test to prevent regression — I have one in #4437 (`test_active_requests_counter_decremented_on_error`) that could be adapted here.

alliasgher:
Good catch — updated to use `try/finally` so the decrement is in a single code path regardless of whether wsgi_app succeeds or raises @liyaka

alliasgher:
Added the regression test test_active_requests_counter_decremented_on_error. It hits /hello/500 (which raises ValueError internally), collects http.server.active_requests, and asserts the value is back to 0 after both requests complete.

MikeGoldsmith (approved):
Thanks for this fix @alliasgher - looks good to me.

liyaka:
Any estimation for this fix to be released?
thank you!

alliasgher:
Pushed bb17ad0 to fix broken links in `CHANGELOG.md` that were failing `check-links` CI.

CHANGELOG.md
```diff
@@ -1162,15 +1167,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Support `aio_pika` 9.x (([#1670](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1670])
+- Support `aio_pika` 9.x ([#1670](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1670))
```
MikeGoldsmith, line 1322: While it's good to clean up the changelog, let's not mix that in with your changes, we should do that in a different PR 👍🏻

MikeGoldsmith (approved):

CHANGELOG.md
alliasgher: Removed in 6df6ce6 — reset CHANGELOG.md to current main and kept only this PR's entry.

liyaka:
@alliasgher looks like PR is approved - can you please solve the conflicts and merge it?
We are waiting for this fix

alliasgher:
@liyaka no conflicts. only repo maintainers have merge access; I can't merge it

liyaka:
@MikeGoldsmith any estimation for this PR merge and release of the fix?
Thank you!

emdneto:
Thanks for the PR!

tammy-baylis-swi (approved):
Lgtm, please address the comment about changelog.

liyaka:
@alliasgher could you please address the comment so PR can be merged?

CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#4360](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4360))
 - `opentelemetry-instrumentation-aiohttp-server`: Use `canonical` attribute of the `Resource` as a span name
   ([#3896](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3896))
+- `opentelemetry-instrumentation-flask`: Wrap `wsgi_app` call in try/finally so the `http.server.active_requests` gauge is decremented when a request raises, preventing a long-term gauge drift
```
xrmx: this needs to be moved to a fragment

daniefdz:
Thanks for the fix @alliasgher. We also hit this bug and are waiting for it.

emdneto (approved):

PR #4434, opened 2026-04-14
fix(celery): coerce non-string values to str in CeleryGetter.get()

alliasgher:
## Description

`CeleryGetter.get()` previously returned header values verbatim. When a downstream library sets a non-string value in the header map (e.g. an int), propagator machinery downstream fails because it expects `str`. Coerce non-string values with `str()` so context propagation is robust.

Fixes #4341

## Type of change

## Checklist


MikeGoldsmith:
This is a duplicate of #4360. Closing this one in favour of that one as it's already approved. Thanks @alliasgher.

PR #4435, opened 2026-04-14
fix(confluent-kafka): use PUBLISH operation for producer spans

alliasgher:
## Description

Producer spans in the Confluent Kafka instrumentation used `CREATE` as the messaging operation name. The semantic conventions define `PUBLISH` for message send operations. Use `PUBLISH` so the span's `messaging.operation.name` attribute follows the spec.

Fixes #4418

## Type of change

## Checklist


MikeGoldsmith:
This will conflict with #4423 where I suggested the same fix as part of other work. I see you did it there, so going to close this PR.

PR #3478, opened 2025-05-08
Add Valkey Instrumentation

sightseeker:
# Description

This PR introduces a new Valkey instrumentation plugin for the OpenTelemetry Python Contrib library.
users can now automatically capture and export Valkey trace data alongside existing OpenTelemetry spans. This fills the gap for applications leveraging Valkey’s Client library, enabling end-to-end distributed tracing without manual span management.

Notes:
- Since the `valkey-search` module is not yet GA, related tests are intentionally omitted in this PR.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)  
- [x] New feature (non-breaking change which adds functionality)  
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)  
- [x] This change requires a documentation update  

# How Has This Been Tested?

- Added unit tests under `instrumentation/opentelemetry-instrumentation-valkey/tests/` covering:
  - automatic span creation on Valkey client calls  
  - context propagation across async boundaries  
  - correct attribute mapping (e.g. `db.statement`, `db.valkey.args_length`, `db.valkey.database_index`, `net.peer.name`)  
- Manual end-to-end test with [example app](https://github.com/sightseeker/opentelemetry-instrumentation-valkey-example-app) instrumented via `opentelemetry-instrumentation-valkey`, verified spans in Jaeger backend  
- Test configuration:
  - `tox -e py38-test-instrumentation-valkey`
  - `tox -e py39-test-instrumentation-valkey`
  - `tox -e py310-test-instrumentation-valkey`
  - `tox -e py311-test-instrumentation-valkey`
  - `tox -e py312-test-instrumentation-valkey`
  - `tox -e py313-test-instrumentation-valkey`
  - `tox -e pypy3-test-instrumentation-valkey`
- Lint configuration:
    - `tox -e lint-instrumentation-valkey`

# Does This PR Require a Core Repo Change?

- [ ] Yes. – Link to PR:  
- [x] No.

# Checklist:

- [x] Followed the style guidelines of this project  
- [x] Changelogs have been updated (`CHANGELOG.md` entry under “New Features”)  
- [x] Unit tests have been added  
- [x] Documentation has been updated (README and Sphinx docs)

PR #4436, opened 2026-04-14
fix(redis): extract ClusterPipeline commands from _execution_strategy in redis-py 6+

alliasgher:
## Description

In redis-py 6+, `ClusterPipeline` moved its pending command list from a direct attribute to an internal `_execution_strategy` object. The instrumentation still read the old attribute path and therefore saw an empty command list, so span attributes for pipelined cluster commands were blank.

Extract the command list from `_execution_strategy` when present so the span attributes reflect the actual pipelined commands.

Fixes #4420

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bump `pylint` to `4.0.5`
   ([#4244](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4244))
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


MikeGoldsmith (approved):
Looks good - thanks @alliasgher.

Worth noting #3478 is also modifying this func, so we need to be aware whichever comes second will have a conflict to resolve.

alliasgher:
> Looks good - thanks @alliasgher.
> 
> Worth noting #3478 is also modifying this func, so we need to be aware whichever comes second will have a conflict to resolve.

@MikeGoldsmith approved but still marked stale

emdneto:
Thanks for the PR!

PR #4448, opened 2026-04-16
fix(pyramid): add missing http.response.status_code in duration metrics

dominod-dev:
# Description

The new convention for metrics use http.response.status_code in metrics for http.server.request.duration metric:

https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration

Only the old convention had the attribute for pyramid, but the new one was missing.

Did not touch the changelog as I was not sure if i needed to create an issue first to reference it or if a PR is enough.

## Type of change

# How Has This Been Tested?

http.responses.status_code was missing from my metrics, this PR fixed it.

A test was added in to check that the attribute is correctly present.

- [x] tox

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


herin049 (approved):

emdneto:
Thanks for the PR!

instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py
```diff
@@ -254,9 +255,8 @@ def trace_tween(request):
             status = getattr(response, "status", status)
             status_code = otel_wsgi._parse_status_code(status)
             if status_code is not None:
-                duration_attrs[HTTP_STATUS_CODE] = (
-                    otel_wsgi._parse_status_code(status)
-                )
+                duration_attrs[HTTP_STATUS_CODE] = status_code
```
xrmx, line 258: What about setting the old field only in the old path and the new one for only the new one? While we introduce a change of shape in the telemetry at least we make them compliant.
And so the changelog would be something like "In duration metrics with the stable http semantic conventions we are now sending the correct `http.response.status_code` instead of `http.status_code`"

herin049: I think this actually is ok because of the logic in 262-277. @dominod-dev can you add some tests to validate that this new attribute is not present for the legacy histogram?

herin049 (changes requested):

instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py
dominod-dev: Yeah from what I've understood, later code clean up the duration dict that the PR modify to retain only the correct metrics from the respective conventions. So the old `http.status_code` was never in the new convention and the new code does not add the `http.response.status_code` in the old one either.

I added some test to confirm it. Tell me if I still need to update the changelog to something more explicit. I could also only modify the duration dict depending on the convention if you think it is cleaner but right now it should match the intended behavior I think.

.changelog/4448.fixed
```diff
@@ -0,0 +1 @@
+`opentelemetry-instrumentation-pyramid`: add missing http.response.status_code in duration metrics
```
xrmx:
```suggestion
`opentelemetry-instrumentation-pyramid`: add missing `http.response.status_code` in duration metrics for stable http semantic conventions
```

instrumentation/opentelemetry-instrumentation-pyramid/src/opentelemetry/instrumentation/pyramid/callbacks.py
xrmx: Thanks!

xrmx (approved):

herin049 (approved):

herin049:
Looks like typecheck is failing.

PR #4389, opened 2026-04-02
Enhance LangChain GenAI semconv tracing

nagkumar91:
# Description

This PR expands the LangChain instrumentor toward semconv-first GenAI coverage and folds in the review/test follow-up fixes needed to make the package consistent with contrib expectations.

The main changes are:

- semconv-aligned model, agent, workflow, tool, and retriever tracing
- richer message/tool/retriever event emission and improved LangGraph/W3C parent propagation
- review follow-up fixes for callback/span handling plus VCR compatibility with current `pytest-recording`
- final changelog, typecheck, and spellcheck cleanup for the package

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

- [x] `python3 -m pytest instrumentation-genai/opentelemetry-instrumentation-langchain/tests`
- [x] `uv tool run ruff check src tests` (from `instrumentation-genai/opentelemetry-instrumentation-langchain`)
- [x] `tox -e typecheck`
- [x] `tox -e spellcheck`

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR:
- [x] No.

# Checklist:

- [x] Followed the style guidelines of this project
- [x] Changelogs have been updated
- [x] Unit tests have been added
- [ ] Documentation has been updated

## Notes

- Package changelog updated at `instrumentation-genai/opentelemetry-instrumentation-langchain/CHANGELOG.md`.
- `test_gemini` is skipped when `tests/cassettes/test_gemini.yaml` is absent and no real Google API key is available for recording.
- Bedrock cassette replay now ignores path drift caused by newer inference-profile/model request shapes while keeping the recorded response assertions intact.

PR #4450, opened 2026-04-16
feat(langchain): add semconv attributes, operation mapping, and content recording modules [1/5]

nagkumar91:
# Description

Add foundation modules for enhanced LangChain GenAI semantic convention tracing. These are internal building-block modules that will be consumed by subsequent PRs in this series.

### `semconv_attributes.py`
Per-operation attribute matrix based on OTel GenAI semantic conventions. Single source of truth for which attributes apply to which operations (chat, text_completion, invoke_agent, execute_tool, invoke_workflow, retrieval). Defines requirement levels (REQUIRED, CONDITIONALLY_REQUIRED, RECOMMENDED, OPT_IN) following the [GenAI semconv spec](https://opentelemetry.io/docs/specs/semconv/gen-ai/).

### `operation_mapping.py`
Callback-to-semconv operation mapping. Maps each LangChain callback to the correct GenAI semantic convention operation name. Includes heuristic classification for `on_chain_start` callbacks (agents vs workflows vs internal plumbing) and LangGraph marker recognition.

### `content_recording.py`
Thin integration layer over the shared genai content capture utilities. Provides clear APIs for the callback handler to decide what content should be recorded on spans and events.

## Tests
- `test_operation_mapping.py` — comprehensive tests for chain classification, agent name resolution, LangGraph markers
- `test_content_recording.py` — tests for content policy modes, redaction behavior

## PR Series
This is **PR 1 of 5** breaking down [#4389](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4389) into smaller reviewable units:
1. **Foundation modules** (this PR)
2. Provider inference, message formatting, W3C propagation utilities
3. Span manager enhancements + Event emitter
4. Callback handler — model, chain, agent callbacks + wiring
5. Callback handler — tool, retriever callbacks + E2E tests

## Type of change

## Checklist

### DIFF
--- a/instrumentation-genai/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/content_recording.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/content_recording.py
@@ -1,0 +1,109 @@
+# Copyright The OpenTelemetry Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Thin integration layer over the shared genai content capture utilities.
+
+Provides clear APIs for the LangChain callback handler to decide what
+content should be recorded on spans and events.
+"""
+
+from typing import Optional
+
+from opentelemetry.util.genai.types import ContentCapturingMode
+from opentelemetry.util.genai.utils import (
+    get_content_capturing_mode,
+    is_experimental_mode,
+    should_emit_event,
+)
+
+
+class ContentPolicy:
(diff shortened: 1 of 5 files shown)


instrumentation-genai/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/semconv_attributes.py
```diff
@@ -0,0 +1,313 @@
+# Copyright The OpenTelemetry Authors
```
lzchen, line 1: I'm curious as to why we need this module and not just take a dependency on the semantic conventions package?

instrumentation-genai/opentelemetry-instrumentation-langchain/src/opentelemetry/instrumentation/langchain/content_recording.py
```diff
@@ -0,0 +1,109 @@
+# Copyright The OpenTelemetry Authors
```
lzchen, line 1: It's quite difficult to validate the correctness of these new files without seeing the functions within them being actually used end to end and, in addition, how they will be used in the existing code. For someone not super familiar with the inner workings of langchain, it would be good to see a simple scenario (like chat) in which a patched method with span creation, recording metrics is done end to end before adding all the other mechanisms (like invoke agent, etc). Would you mind building the solution piece by piece so that each PR as a standalone adds end to end functionality so it's easier to review/test?

nagkumar91:
Closing per author request.

PR #4456, opened 2026-04-16
openai-agents-v2: use genai-utils for env var constant, histogram factories, and meter_provider

nagkumar91:
## Summary

Three improvements to align `opentelemetry-instrumentation-openai-agents-v2` with `opentelemetry-util-genai`:

### 1. Env var constant — single source of truth
Remove `_CONTENT_CAPTURE_ENV = "OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT"` from `__init__.py` and import the constant from `opentelemetry.util.genai.environment_variables`.

### 2. Fix silently-ignored `meter_provider`
`GenAISemanticProcessor._init_metrics()` previously called `get_meter(...)` without any provider, so a caller-supplied `meter_provider` was silently ignored. This PR:
- Adds `meter_provider: Optional[MeterProvider] = None` to `GenAISemanticProcessor.__init__`
- Threads it through to `_init_metrics(meter_provider)`
- Passes it to `get_meter()` along with the package's actual `__version__` (was hardcoded as `"0.1.0"`)

### 3. Semconv bucket boundaries via genai-utils histogram factories
Replace inline `meter.create_histogram()` calls with `create_duration_histogram(meter)` and `create_token_histogram(meter)` from `opentelemetry.util.genai.instruments`. This adds the semconv-specified explicit bucket boundaries that were previously missing from the agent's metrics.

## Testing

All 103 existing tests pass.


lzchen (approved):

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
```diff
@@ -515,10 +522,10 @@ def __init__(
         # Metrics configuration
         self._metrics_enabled = metrics_enabled
         self._meter = None
-        self._duration_histogram: Optional[Histogram] = None
-        self._token_usage_histogram: Optional[Histogram] = None
+        self._duration_histogram: Optional[object] = None
```
herin049, line 525: Why are changing from `Optional[Histogram]` to `Optional[object]` here? We should keep the type annotation as is if possible.

herin049 (changes requested):

instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/span_processor.py
lzchen, line 423: I think `_resolve_content_mode` in `instrument` can be replaced by https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py#L45

lzchen, line 622: Might be missing something here but what does this function do? Is recording events not implemented yet?

lzchen, line 623: We can also use utils -> https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py#L68 for checking to emit events in conjunction with the current logic.

lzchen (changes requested):

lmolkova:
thanks for the contribution @nagkumar91  !

We're moving GenAI instrumentations to the new repository - https://github.com/open-telemetry/opentelemetry-python-genai.

I'm closing this PR here, but you're welcome to reopen it in the new repo.

Issue #1050, opened 2022-04-12
opentelemetry-instrument command can cause recursive creation of subprocesses


aabmass:
**Describe your environment**

Python3.9, linux.

**Steps to reproduce**
Using `opentelemetry-instrument` with any exporter or instrumentation which invokes a python subprocess **during initialization**. For example, the `opentelemetry-exporter-gcp-trace` exporter may invoke the `gcloud` (written in python) command in a subprocess to get project information and authentication tokens. The subprocess will then try to autoinstrument, creating a recursive loop of subprocesses being created.

**What is the expected behavior?**
Auto-instrumentation should not apply to subprocesses created in the `initialize()` phase of auto-instrumentation. The `PYTHONPATH` environment variable should have the `sitecustomize.py` dirname stripped out at the beginning of `sitecustomize.py`. This would prevent subprocesses from being autoinstrumented during setup, which can cause a loop.

**What is the actual behavior?**
`PYTHONPATH` is correctly stripped later on to avoid this https://github.com/open-telemetry/opentelemetry-python-contrib/blob/e9f83e1292b0fe5f3478c9b23f3b5a5508481e68/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py#L120-L125

However, any subprocesses created in [these lines](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/e9f83e1292b0fe5f3478c9b23f3b5a5508481e68/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py#L114-L117) will cause a loop.

**Additional context**
I can write a repro if necessary.

PR #4469, opened 2026-04-21
feat: update auto-instrumentation to re-inject instrumentation path after init

herin049:
# Description

Updates auto-instrumentation behavior to re-inject instrumentation path after init. This fixes auto-instrumentation with subprocesses (e.g. Uvicorn workers)

Fixes https://github.com/open-telemetry/opentelemetry-python/issues/4927

## Type of change

# How Has This Been Tested?

```bash
uv run tox -e py312-test-opentelemetry-instrumentation-wrapt2
```

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


CHANGELOG.md
```diff
@@ -56,6 +56,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#4049](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4049))
 - `opentelemetry-instrumentation-sqlalchemy`: implement new semantic convention opt-in migration
   ([#4110](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110))
+- `opentelemetry-instrumentation`: add ability to auto-instrument subprocesses with `OTEL_PYTHON_AUTO_INSTRUMENTATION_INSTRUMENT_SUBPROCESSES` environment variable
```
MikeGoldsmith: Needs moving up to unreleased.

MikeGoldsmith (approved):
Looks good - thanks @herin049. Need to move changelog.

DylanRussell (approved):

emdneto (approved):
Nice. If merged, I think it's also valid to document this option in opentelemetry.io docs as well + explaining the behavior to users.

opentelemetry-instrumentation/README.rst
```diff
@@ -114,6 +114,14 @@ If set by the user to `patch_all` , opentelemetry instrument will call the geven
 This is considered experimental but can be useful to instrument gevent applications.
 e.g OTEL_PYTHON_AUTO_INSTRUMENTATION_EXPERIMENTAL_GEVENT_PATCH=patch_all
 
+* ``OTEL_PYTHON_AUTO_INSTRUMENTATION_INSTRUMENT_SUBPROCESSES``
+
+By default, auto-instrumentation removes the auto-instrumentation ``sitecustomize.py`` directory
```
aabmass: Would be good to state why this is the default. For context https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1050

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
```diff
@@ -129,8 +130,17 @@ def initialize(*, swallow_exceptions: bool = True) -> None:
...
+            OTEL_PYTHON_AUTO_INSTRUMENTATION_INSTRUMENT_SUBPROCESSES, ""
+        )
+        .strip()
+        .lower()
+        == "true"
+    )
+
     # prevents auto-instrumentation of subprocesses if code execs another python process
-    if "PYTHONPATH" in environ:
+    if "PYTHONPATH" in environ and not instrument_subprocesses:
```
aabmass: For context https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1050

As I was thinking about it, it would probably be safe to re-patch the `PYTHONPATH` after the OTel initialization is finished, when sitecustomize finishes running. Then we could avoid exposing this option at all and make it the default again.

I might be missing some edge cases where the shallow behavior is still preferred and safer. WDYT?

herin049: Agreed, I like this approach. Will update the PR.

aabmass: Thanks! I think we just need some test coverage

herin049: Refactored to use the approach discussed.

opentelemetry-instrumentation/tests/auto_instrumentation/test_initialize.py
aabmass, line 1: Is it possible to add a test that repros the original issue of subprocess loop, just to be sure the fix is good and no regressions?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
```diff
@@ -157,3 +146,37 @@ def initialize(*, swallow_exceptions: bool = True) -> None:
...
+    # where subprocesses spawned in the initialization phase execute the
+    # initialization phase again, spawning more subprocesses.
+    if python_path is not None:
+        environ["PYTHONPATH"] = _python_path_without_directory(
+            python_path, filedir, pathsep
+        )
+
+    try:
+        _initialize(swallow_exceptions=swallow_exceptions)
+    finally:
```
aabmass, line 176: Would it work to just save the original value of PYTHONPATH envvar and restore it here?

aabmass (approved):

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
herin049: I wanted to handle the edge case where `PYTHONPATH` is mutated during the initialization phase. Let me know if you think this is overkill, and I can remove this logic.

opentelemetry-instrumentation/tests/auto_instrumentation/test_initialize.py
herin049: Good idea, will update the tests.

emdneto:
Thanks for the PR!

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/__init__.py
aabmass: SG and makes sense.

opentelemetry-instrumentation/tests/auto_instrumentation/test_initialize.py
aabmass: SG, please move to "ready to merge" once it's done

tammy-baylis-swi (approved):
Awesome

xrmx (approved):
pylint is complaining

PR #4482, opened 2026-04-23
docs(google-genai): document config recording environment variables

RiyaChaturvedi37:
## Summary
Documents the configuration for enabling config recording in the 
Google GenAI SDK instrumentation.

## Changes
Added a new "Configuration recording" section to the README.rst 
explaining:
- `OTEL_GOOGLE_GENAI_GENERATE_CONTENT_CONFIG_INCLUDES` — controls 
  which config fields are included as span attributes
- `OTEL_GOOGLE_GENAI_GENERATE_CONTENT_CONFIG_EXCLUDES` — controls 
  which config fields are excluded from span attributes
- Default behavior (no fields recorded by default)
- Example usage for both variables
- How they work together as allowlist/denylist pair

Fixes #3438

PR #4490, opened 2026-04-26
fix(sqlcommenter): prepend comment to beginning of query

### LINKED ISSUE #3583: sqlcommenter: put the comment at the beginning of the query
Hi!

[sqlcommenter][1] appends the comment to the end of the query. I would propose prepending the comment at the beginning of the query instead.

The comment at the end is problematic when the query is really long, and the comment is truncated from database query logs. The very long queries are often produced by Django, and the query logs (e.g. top slow queries) have limited length in AWS RDS.

[1]:https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/sqlcommenter_utils.py#L27

What do you think?

RiyaChaturvedi37:
## Problem
The SQL comment was appended to the end of the query, which caused 
truncation issues in database query logs for long queries (e.g. 
queries produced by Django in AWS RDS logs).

## Fix
Changed the comment position from end to beginning of the query.

Before:
```sql
SELECT * FROM users /*key='value'*/;
```

After:
```sql
/*key='value'*/ SELECT * FROM users;
```

## Testing
The existing tests have been updated to reflect the new comment 
position at the beginning of queries.

Fixes #3583


RiyaChaturvedi37:
After implementing this change, I noticed it affects many existing tests and could be a breaking change for users who parse SQL logs expecting comment at the end. Should I proceed with updating all tests, or would maintainers prefer a different approach such as a configuration option to control comment position?

MikeGoldsmith (changes requested):
A few concerns:

1. **Breaking change** — moving the comment from end to start changes behaviour for any downstream system that expects `sqlcommenter` comments at the end of queries. The [sqlcommenter spec](https://google.github.io/sqlcommenter/) defines comments at the end. This needs discussion on the issue (#3583) before we change the default behaviour — maybe an option to control position rather than switching it outright.

2. **Semicolon test looks wrong** — the expected output for the semicolon case is `" /*...*/ Select 1"` (no semicolon), but the input was `"Select 1;"`. The semicolon is being dropped.

3. **Leading space** — the prepended output starts with a space before the comment: `" /*...*/ SELECT 1"`.

4. **Line endings** — same as #4482, the diff shows the entire file rewritten due to CRLF line endings. Please configure your editor/git to use Unix line endings (`git config core.autocrlf input`).

tammy-baylis-swi:
Hi @RiyaChaturvedi37 , thanks for working on this. I realized in my above comments that supporting the original requested feature is going to involve changes to several [instrumentation libraries](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/cbd7f657f8c20cca165c3a6ff16b556cda548e51/instrumentation) and not just this util:

* DB-API
* Django
* Flask
* SQLAlchemy
* psycopg2
* psycopg
* mysql
* mysqlclient
* PyMySQL

Are you interested and available for submitting several PRs for this?

RiyaChaturvedi37:
Hi @tammy-baylis-swi, yes I'm interested and available! I'd be happy to submit PRs for each of these instrumentation libraries. Should I start with one library first for review, then proceed to the others once the approach is confirmed? Or would you prefer I submit them all at once?

tammy-baylis-swi:
> Hi @tammy-baylis-swi, yes I'm interested and available! I'd be happy to submit PRs for each of these instrumentation libraries. Should I start with one library first for review, then proceed to the others once the approach is confirmed? Or would you prefer I submit them all at once?

Awesome! I'd say start with this PR + another PR for one library first, to confirm the approach like you said and to make it easier for the Maintainers. Please could you also note your plan on the issue https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3583

instrumentation-genai/opentelemetry-instrumentation-google-genai/CHANGELOG.md
```diff
@@ -33,7 +33,7 @@ Ensure same sem conv attributes are on the log and span. Fix an issue where the
 - Implement the new semantic convention changes made in https://github.com/open-telemetry/semantic-conventions/pull/2179.
 A single event (`gen_ai.client.inference.operation.details`) is used to capture Chat History. This is opt-in,
 an environment variable OTEL_SEMCONV_STABILITY_OPT_IN needs to be set to `gen_ai_latest_experimental` to see them ([#3386](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3386))
-- Support CompletionHook for upload to cloud storage. 
+- Support CompletionHook for upload to cloud storage.
```
lmolkova, line 36: this is not the right changelog location, I believe this change should go to  CHANGELOG.md in the root of the repo. Pleae also make sure to follow the format of previous entries that include link to PR

RiyaChaturvedi37:
Fixed @lmolkova! Moved the changelog entry to the root CHANGELOG.md and included the PR link as requested.

emdneto:
Thanks for the PR!

tammy-baylis-swi:
Thanks @RiyaChaturvedi37 for the progress. Please could you address the changelog as well as the remaining feedback on this PR?:

1. https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4490/changes#r3190827101
2. https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4490/changes#r3190844770

RiyaChaturvedi37:
Closing in favor of #4628 which has a clean git history without CHANGELOG.md modifications. The changelog is handled via .changelog/4490.added fragment.

PR #4494, opened 2026-04-27
openai-v2: default empty string for GEN_AI_REQUEST_MODEL on missing model

lzchen:
When model is not provided in kwargs, default to empty string instead of None for the GEN_AI_REQUEST_MODEL attribute to avoid invalid attribute values. This can occur for some custom implementations of OpenAi client like [AzureOpenAi](https://github.com/openai/openai-python/blob/e507a4ebeea4c3f93cd48986014a3e2ca79230c2/src/openai/lib/azure.py#L90).

### DIFF
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/CHANGELOG.md
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/CHANGELOG.md
@@ -29,7 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   extraction improvements
   ([#4337](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4337))
 - Default empty string for `GEN_AI_REQUEST_MODEL` attribute on missing model.
-  ([#4337](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4337))
+  ([#4494](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4494))
 
 ## Version 2.3b0 (2025-12-24)
 
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-v2/CHANGELOG.md
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-v2/CHANGELOG.md
@@ -28,7 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add strongly typed Responses API extractors with validation and content
   extraction improvements
   ([#4337](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4337))
-- Default empty string for `GEN_AI_REQUEST_MODEL` attribute on missing model.
+- Default empty string for `gen_ai.request.model` attribute on missing model.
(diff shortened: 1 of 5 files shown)


rads-1996 (approved):

xrmx (commented):
LGTM but maybe we should add a test to avoid regressions during refactorings

xrmx:
> LGTM but maybe we should add a test to avoid regressions during refactorings

Or alternatively drop the attribute since it's "conditionally required". That would also affect the span name

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
```diff
@@ -214,7 +214,7 @@ def get_llm_request_attributes(
 
     attributes = {
         GenAIAttributes.GEN_AI_OPERATION_NAME: operation_name,
-        GenAIAttributes.GEN_AI_REQUEST_MODEL: kwargs.get("model"),
+        GenAIAttributes.GEN_AI_REQUEST_MODEL: kwargs.get("model", ""),
```
lmolkova: why ""  and not None? Would it result in span name being `"chat "` ?

instrumentation-genai/opentelemetry-instrumentation-openai-v2/CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add strongly typed Responses API extractors with validation and content
   extraction improvements
   ([#4337](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4337))
+- Default empty string for `GEN_AI_REQUEST_MODEL` attribute on missing model.
```
lmolkova:
```suggestion
- Default empty string for `gen_ai.request.model` attribute on missing model.
```

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
lzchen: I'm okay with dropping the attribute too, just depends on if we like "chat" better than "chat " for span name?

hectorhdzg (approved):
LGTM

lzchen:
@xrmx 

Added some unit tests.

herin049 (approved):

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/utils.py
xrmx: It's not specified in the genai semconvs but I think it should be "chat", so better drop the attribute if it's not set and handle the missing attribute when contructing the span name.

lzchen: @xrmx 

Done.

lmolkova (approved):

PR #4500, opened 2026-04-29
genai-util: refactor streams to a generic ABC wrapper.

eternalcuriouslearner:
# Description

This change introduces shared stream wrapper base classes in util-genai and updates the OpenAI v2 chat streaming wrappers to use that common lifecycle shape.

The new SyncStreamWrapper and AsyncStreamWrapper centralize stream iteration, close/context-manager handling, passthrough behavior, double-finalization protection, and failure handling. OpenAI chat streaming now uses dedicated ChatStreamWrapper and AsyncChatStreamWrapper classes that mirror the Messages and Responses wrapper structure while continuing to propagate SDK/API exceptions and user exceptions unchanged.

This also moves chat choice buffering into a separate chat_buffers.py module to avoid import cycles, switches the new chat path to LLMInvocation/InferenceInvocation lifecycle methods, and adds tests for stream wrapper behavior and sync/async chat exception propagation.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- Added unit coverage for sync/async stream wrapper lifecycle behavior.
- Added VCR-backed sync and async chat tests for API exception propagation.
- Added sync and async streaming tests for user exception propagation.
- Ran OpenAI v2 chat test slices against oldest and latest dependency envs.
- Ran OpenAI v2 lint successfully.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -0,0 +1,193 @@
...
+import logging
+from abc import ABC, abstractmethod
+from types import TracebackType
+from typing import Any, Generic, Literal, TypeVar
+
+ChunkT = TypeVar("ChunkT")
+_logger = logging.getLogger(__name__)
+
+
+class SyncStreamWrapper(ABC, Generic[ChunkT]):
```
lmolkova: could you please add a docstring explaining how one can use this class? also add a few short sentences in agents.md telling agents to use these helpers?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def _finalize_failure(self, error: BaseException) -> None:
+        if self._finalized:
+            return
+        self._fail_stream(error)
+        self._finalized = True
+
+    @abstractmethod
+    def _process_chunk(self, chunk: ChunkT) -> None:
+        """Process one stream chunk for telemetry."""
```
lmolkova, line 157: should these methods be public (no underscore) ?

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
```diff
@@ -0,0 +1,221 @@
...
+        if error:
+            self.invocation.fail(error)
+        else:
+            self.invocation.stop()
+        self._started = False
+
+
+class AsyncChatStreamWrapper(
+    AsyncStreamWrapper[Any],
+    ChatStreamWrapper,
```
lmolkova: it seems there is a diamond inheritance problem here.

Here's some details and repros from my AI

`AsyncChatStreamWrapper` inherits from both `AsyncStreamWrapper` and `ChatStreamWrapper`, where `ChatStreamWrapper` already inherits from `SyncStreamWrapper`. The two base classes — `SyncStreamWrapper` and `AsyncStreamWrapper` — are near-duplicates: same `__init__` body, same `__getattr__`, same `_finalize_*` helpers, and overlapping public surface        
  (`close`, iteration, context manager) where one is sync and the other async.
                                                                                                                                                                                                                                                                                                                                                                        
  This works today only because:
  1. The two `__init__` bodies are identical, so it doesn't matter which one runs.
  2. The parent order in the class declaration happens to put the async base first, so `close` resolves to the async version.                                                                                                                                                                                                                                           
  3. The constructor chain is hand-wired (`ChatStreamWrapper.__init__` calls `SyncStreamWrapper.__init__` directly, never `super()`).                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                        
  Each of those is a latent failure waiting on the first refactor.                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                        
  ## Repro 1: parent of one wrapper silently never runs its constructor                                                                                                                                                                                                                                                                                                 
                  
  ```python                                                                                                                                                                                                                                                                                                                                                             
  class SyncBase: 
      def __init__(self, stream):
          self.stream = stream
          self._finalized = False

  class AsyncBase:
      def __init__(self, stream):                                                                                                                                                                                                                                                                                                                                       
          self.stream = stream
          self._finalized = False                                                                                                                                                                                                                                                                                                                                       
          # Imagine someone adds async-specific state later:
          self._loop = "async-only"                                                                                                                                                                                                                                                                                                                                     
   
  class ChatLogic(SyncBase):                                                                                                                                                                                                                                                                                                                                            
      def __init__(self, stream, invocation):
          SyncBase.__init__(self, stream)         # exactly what the PR does
          self.invocation = invocation                                                                                                                                                                                                                                                                                                                                  
   
  class AsyncChat(AsyncBase, ChatLogic):                                                                                                                                                                                                                                                                                                                                
      def __init__(self, stream, invocation):
          ChatLogic.__init__(self, stream, invocation)                                                                                                                                                                                                                                                                                                                  
   
  a = AsyncChat("s", "inv")                                                                                                                                                                                                                                                                                                                                             
  print(hasattr(a, "_loop"))   # False — AsyncBase.__init__ never ran
  ```                                                                                                                                                                                                                                                                                                                                                                   
   
  Today the two `__init__`s are byte-identical so nobody notices. The day someone adds a field to `AsyncStreamWrapper.__init__`, the async wrapper silently ships without it — no error, just a missing attribute that surfaces later as `AttributeError` or wrong telemetry.                                                                                           
                  
  ## Repro 2: parent-order in the class declaration is load-bearing                                                                                                                                                                                                                                                                                                     
                  
  ```python                                                                                                                                                                                                                                                                                                                                                             
  import asyncio  

  class SyncBase:
      def close(self):
          return "sync"

  class AsyncBase:
      async def close(self):
          return "async"
                                                                                                                                                                                                                                                                                                                                                                        
  class ChatLogic(SyncBase):
      pass                                                                                                                                                                                                                                                                                                                                                              
                  
  # What the PR has — async base listed first:
  class AsyncChat(AsyncBase, ChatLogic):
      pass

  # A future refactor swaps the order, e.g. to put the "logic" mixin first:
  class AsyncChatRefactored(ChatLogic, AsyncBase):                                                                                                                                                                                                                                                                                                                      
      pass
                                                                                                                                                                                                                                                                                                                                                                        
  async def main():
      print(await AsyncChat().close())          # "async" — works
      print(await AsyncChatRefactored().close())  # TypeError: object str can't be used in 'await'                                                                                                                                                                                                                                                                      
   
  asyncio.run(main())                                                                                                                                                                                                                                                                                                                                                   
  ```                                                                                                                                                                                                                                                                                                                                                                   
   
  Method resolution walks the parents left-to-right. The async wrapper happens to be listed first, so `AsyncBase.close` (async) wins. Anyone reordering the parents — for stylistic reasons, to add another mixin, to "put the logic class first" — silently breaks every `await wrapper.close()` call.                                                                 
                  
  ## Repro 3: sync iteration protocol leaks onto the async object                                                                                                                                                                                                                                                                                                       
                  
  ```python                                                                                                                                                                                                                                                                                                                                                             
  class SyncBase: 
      def __iter__(self): return self
      def __next__(self): return next(self.stream)

  class AsyncBase:
      def __aiter__(self): return self
      async def __anext__(self): return await self.stream.__anext__()                                                                                                                                                                                                                                                                                                   
   
  class ChatLogic(SyncBase): pass                                                                                                                                                                                                                                                                                                                                       
  class AsyncChat(AsyncBase, ChatLogic): pass
                                                                                                                                                                                                                                                                                                                                                                        
  w = AsyncChat()
  # Both protocols exist on the same object:                                                                                                                                                                                                                                                                                                                            
  print(hasattr(w, "__next__"), hasattr(w, "__anext__"))   # True True
  ```                                                                                                                                                                                                                                                                                                                                                                   
   
  The async wrapper claims to support `for chunk in w:`. It will fail eventually (calling `next()` on an async iterator throws), but the failure is deep in the iteration protocol — not a clear "this object is async-only." Static analyzers and IDE autocomplete see both protocols and offer both.

lmolkova: here's the fix it suggests

Extract the chat-specific hooks into a mixin that has **no `__init__`, no iteration methods, and no `close`/context-manager methods**. Combine it once with each iteration base.                                                                                                                                                                                      
   
  ```python                                                                                                                                                                                                                                                                                                                                                             
  # chat_wrappers.py

  class _ChatStreamLogic:
      """Chat-specific hooks shared by sync and async wrappers.
                                                                                                                                                                                                                                                                                                                                                                        
      Intentionally has no __init__, no iteration protocol, no close/__enter__/
      __aenter__ — those belong to the iteration base class it's mixed with.                                                                                                                                                                                                                                                                                            
      """                                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                        
      invocation: InferenceInvocation                                                                                                                                                                                                                                                                                                                                   
      capture_content: bool
      choice_buffers: list
      response_id: Optional[str] = None                                                                                                                                                                                                                                                                                                                                 
      response_model: Optional[str] = None                                                                                                                                                                                                                                                                                                                              
      service_tier: Optional[str] = None                                                                                                                                                                                                                                                                                                                                
      prompt_tokens: Optional[int] = None                                                                                                                                                                                                                                                                                                                               
      completion_tokens: Optional[int] = None                                                                                                                                                                                                                                                                                                                           
   
      # --- hooks called by the iteration base ---                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                        
      def _process_chunk(self, chunk):
          self._set_response_id(chunk)                                                                                                                                                                                                                                                                                                                                  
          self._set_response_model(chunk)                                                                                                                                                                                                                                                                                                                               
          self._set_response_service_tier(chunk)
          self._build_streaming_response(chunk)                                                                                                                                                                                                                                                                                                                         
          self._set_usage(chunk)                                                                                                                                                                                                                                                                                                                                        
   
      def _stop_stream(self) -> None:                                                                                                                                                                                                                                                                                                                                   
          self._cleanup()
                                                                                                                                                                                                                                                                                                                                                                        
      def _fail_stream(self, error: BaseException) -> None:
          self._cleanup(error)                                                                                                                                                                                                                                                                                                                                          
                  
      # --- chat-specific helpers (unchanged from the PR) ---
                                                                                                                                                                                                                                                                                                                                                                        
      def _set_response_id(self, chunk): ...
      def _set_response_model(self, chunk): ...                                                                                                                                                                                                                                                                                                                         
      def _set_response_service_tier(self, chunk): ...
      def _build_streaming_response(self, chunk): ...                                                                                                                                                                                                                                                                                                                   
      def _set_usage(self, chunk): ...
      def _set_output_messages(self): ...                                                                                                                                                                                                                                                                                                                               
   
      def _cleanup(self, error: Optional[BaseException] = None):                                                                                                                                                                                                                                                                                                        
          self.invocation.response_model_name = self.response_model                                                                                                                                                                                                                                                                                                     
          self.invocation.response_id = self.response_id                                                                                                                                                                                                                                                                                                                
          self.invocation.input_tokens = self.prompt_tokens                                                                                                                                                                                                                                                                                                             
          self.invocation.output_tokens = self.completion_tokens                                                                                                                                                                                                                                                                                                        
          if self.service_tier:                                                                                                                                                                                                                                                                                                                                         
              self.invocation.attributes.update(
                  {OpenAIAttributes.OPENAI_RESPONSE_SERVICE_TIER: self.service_tier},                                                                                                                                                                                                                                                                                   
              )                                                                                                                                                                                                                                                                                                                                                         
          self._set_output_messages()                                                                                                                                                                                                                                                                                                                                   
          if error:                                                                                                                                                                                                                                                                                                                                                     
              self.invocation.fail(error)
          else:
              self.invocation.stop()                                                                                                                                                                                                                                                                                                                                    
   
      def parse(self):                                                                                                                                                                                                                                                                                                                                                  
          """Called when using with_raw_response with stream=True."""
          return self                                                                                                                                                                                                                                                                                                                                                   
   
                                                                                                                                                                                                                                                                                                                                                                        
  class ChatStreamWrapper(_ChatStreamLogic, SyncStreamWrapper[Any]):
      def __init__(
          self,                                                                                                                                                                                                                                                                                                                                                         
          stream: Stream,
          invocation: InferenceInvocation,                                                                                                                                                                                                                                                                                                                              
          capture_content: bool,
      ):                                                                                                                                                                                                                                                                                                                                                                
          super().__init__(stream)              # → SyncStreamWrapper.__init__
          self.invocation = invocation                                                                                                                                                                                                                                                                                                                                  
          self.capture_content = capture_content
          self.choice_buffers = []                                                                                                                                                                                                                                                                                                                                      
   
                                                                                                                                                                                                                                                                                                                                                                        
  class AsyncChatStreamWrapper(_ChatStreamLogic, AsyncStreamWrapper[Any]):
      def __init__(                                                                                                                                                                                                                                                                                                                                                     
          self,
          stream: AsyncStream,                  # was incorrectly typed as Stream                                                                                                                                                                                                                                                                                       
          invocation: InferenceInvocation,                                                                                                                                                                                                                                                                                                                              
          capture_content: bool,
      ):                                                                                                                                                                                                                                                                                                                                                                
          super().__init__(stream)              # → AsyncStreamWrapper.__init__
          self.invocation = invocation                                                                                                                                                                                                                                                                                                                                  
          self.capture_content = capture_content
          self.choice_buffers = []                                                                                                                                                                                                                                                                                                                                      
                  

  __all__ = ["AsyncChatStreamWrapper", "ChatStreamWrapper"]
  ```                                                                                                                                                                                                                                                                                                                                                                   
   
  ## Why this works                                                                                                                                                                                                                                                                                                                                                     
                  
  - **No diamond.** Each concrete wrapper has exactly one iteration base in its MRO. `SyncStreamWrapper` and `AsyncStreamWrapper` never coexist in the same class.
  - **`super().__init__(stream)` resolves to exactly one parent.** Adding async-only state to `AsyncStreamWrapper.__init__` later automatically reaches `AsyncChatStreamWrapper`. Repro 1 is gone.                                                                                                                                                                      
  - **Parent order is no longer load-bearing.** The mixin doesn't define `close`, `__iter__`, `__aiter__`, etc. — there's nothing to collide with the iteration base. Repro 2 is gone.                                                                                                                                                                                  
  - **Each wrapper exposes exactly one iteration protocol.** `AsyncChatStreamWrapper` has only `__aiter__`/`__anext__`/`__aenter__`/`__aexit__` (from `AsyncStreamWrapper`). Trying `for chunk in async_wrapper:` fails immediately with a clear protocol error. Repro 3 is gone.                                                                                       
  - **The forwarding overrides** (`_process_chunk`, `_stop_stream`, `_fail_stream` in the original `AsyncChatStreamWrapper`) are deleted — MRO finds them on `_ChatStreamLogic` directly.                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                        
  ## Things to double-check while making the change                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                        
  - `SyncStreamWrapper` / `AsyncStreamWrapper` are `ABC`s with `_process_chunk` / `_stop_stream` / `_fail_stream` marked `@abstractmethod`. Concrete impls living on a non-`ABC` mixin still satisfy the abstract contract — Python checks abstractness at instantiation by walking the MRO. Verified: instantiating `ChatStreamWrapper(...)` / 
  `AsyncChatStreamWrapper(...)` will not raise `TypeError: Can't instantiate abstract class`.                                                                                                                                                                                                                                                                           
  - The old `ChatStreamWrapper` set `self._started = True` in `__init__` and used it as a re-entry guard inside `cleanup`. That guard is now redundant — `SyncStreamWrapper._finalize_success/_finalize_failure` already short-circuit on `self._finalized`. Drop `_started` entirely; don't carry it into the mixin.
  - Fix the type hint on `AsyncChatStreamWrapper.__init__`: today the PR types `stream: Stream` (sync) on the async wrapper. Use `openai.AsyncStream` (or `Any` if the import is awkward).

eternalcuriouslearner: This is a good catch. I am kind of thinking about the problem but I am hitting a wall with a good long term fix. I am going to look more closely and refactor this a little to remove this. Thanks for the detailed explanation. This is pretty helpful.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
eternalcuriouslearner: Not exactly. _process_chunk is called internally by `SyncStreamWrapper.__next__` and `AsyncStreamWrapper.__anext__`; users interact with the wrapper through iteration, context manager methods, and close(), not by processing chunks directly.

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
eternalcuriouslearner: @lmolkova I shifted bulk of code to a mixin as you suggested. Can you please take another look?

lmolkova: thank you, it looks good!

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/patch.py
```diff
@@ -859,83 +817,3 @@ def cleanup(self, error: Optional[BaseException] = None):
...
-
-    def cleanup(self, error: Optional[BaseException] = None):
-        if not self._started:
-            return
-
-        self.invocation.response_model_name = self.response_model
-        self.invocation.response_id = self.response_id
-        self.invocation.input_tokens = self.prompt_tokens
-        self.invocation.output_tokens = self.completion_tokens
-        self.invocation.finish_reasons = self.finish_reasons
```
lzchen, line 916: Nit: We recorded `finish_reasons` even if the list was empty before and now we only record if there are entries. Is this intentional?

eternalcuriouslearner: Yes @lzchen we are not populating `finish_reasons` if the list is empty.

lzchen (approved):

CLAUDE.md
```diff
@@ -1 +1,8 @@
 @AGENTS.md
+
+For GenAI streaming wrappers, use `SyncStreamWrapper` and
+`AsyncStreamWrapper` from `opentelemetry.util.genai.stream` instead of
+reimplementing iteration, close/context-manager, and finalization behavior in
+provider packages. Keep provider-specific chunk parsing and telemetry
+finalization in private hooks or a narrow mixin, and do not make async stream
+wrappers inherit from sync stream wrappers.
```
lmolkova: nit, I think it's not necessary, it's already in agents.md

```suggestion
```

AGENTS.md
```diff
@@ -76,6 +76,11 @@ Apply to packages under `instrumentation/` and `instrumentation-genai/`.
 - When catching exceptions from the underlying library to record telemetry, always re-raise the
   original exception unmodified.
 - Do not raise new exceptions in instrumentation/telemetry code.
+- For GenAI streaming wrappers, prefer the shared `SyncStreamWrapper` and `AsyncStreamWrapper`
```
lmolkova: this belongs in ./instrumentation-genai/AGENTS.md, not in the root

lmolkova (approved):

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_buffers.py
```diff
@@ -0,0 +1,52 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+class ToolCallBuffer:
+    def __init__(self, index, tool_call_id, function_name):
```
aabmass: If this is new code, can we add types?

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_buffers.py
```diff
@@ -0,0 +1,52 @@
...
+        self.index = index
+        self.function_name = function_name
+        self.tool_call_id = tool_call_id
+        self.arguments = []
+
+    def append_arguments(self, arguments):
+        if arguments is not None:
```
aabmass, line 22: it doesn't look like arguments is ever None

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
```diff
@@ -0,0 +1,219 @@
...
+    service_tier: Optional[str] = None
+    prompt_tokens: Optional[int] = None
+    completion_tokens: Optional[int] = None
+
+    def _set_response_model(self, chunk):
+        if self.response_model:
+            return
+
+        if getattr(chunk, "model", None):
+            self.response_model = chunk.model
```
aabmass: Why do we need getattr()?

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
```diff
@@ -0,0 +1,219 @@
...
+            if choice.tool_calls_buffers:
+                tool_calls = []
+                for tool_call in choice.tool_calls_buffers:
+                    arguments = None
+                    arguments_str = "".join(tool_call.arguments)
+                    if arguments_str:
+                        try:
+                            arguments = json.loads(arguments_str)
+                        except json.JSONDecodeError:
+                            arguments = arguments_str
```
aabmass, line 126: Ik this code already existed, but shouldn't this join by `","`?

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -0,0 +1,272 @@
...
+    ``super().__init__(stream)`` and implement the three telemetry hooks:
+    ``_process_chunk`` for per-chunk state, ``_stop_stream`` for successful
+    finalization, and ``_fail_stream`` for failure finalization.
+
+    Users should consume subclasses as normal streams, for example with
+    ``for chunk in wrapper`` or ``with wrapper``. The hook methods are called
+    internally by the wrapper lifecycle and are not part of the public API.
+    """
+
+    def __init__(self, stream: Any):
```
aabmass: Can you type stream? it looks like `Iterable[ChunkT]` might work (or some other built in protocol). If not, can you define a new Protocol that describes all the operations it must support

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -0,0 +1,272 @@
...
+                exc_info=True,
+            )
+
+    @abstractmethod
+    def _process_chunk(self, chunk: ChunkT) -> None:
+        """Process one stream chunk for telemetry."""
+
+    @abstractmethod
+    def _stop_stream(self) -> None:
+        """Finalize the stream successfully."""
```
aabmass, line 161: Is the subclass supposed to do the finalization of the wrapped stream? If these are hooks I'd recommend naming them `_on_<event>` instead.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -0,0 +1,272 @@
...
+    ``_process_chunk`` for per-chunk state, ``_stop_stream`` for successful
+    finalization, and ``_fail_stream`` for failure finalization.
+
+    Users should consume subclasses as normal streams, for example with
+    ``for chunk in wrapper`` or ``with wrapper``. The hook methods are called
+    internally by the wrapper lifecycle and are not part of the public API.
+    """
+
+    def __init__(self, stream: Any):
+        self.stream = stream
```
aabmass: Can this be prefixed with underscore? It's possible this can conflict with the public API of the wrapped object

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -0,0 +1,272 @@
...
+    ``_process_chunk`` for per-chunk state, ``_stop_stream`` for successful
+    finalization, and ``_fail_stream`` for failure finalization.
+
+    Users should consume subclasses as normal async streams, for example with
+    ``async for chunk in wrapper`` or ``async with wrapper``. The hook methods
+    remain synchronous telemetry hooks; async stream reads and close handling
+    are owned by this base class.
+    """
+
+    def __init__(self, stream: Any):
```
aabmass: same comment about protocol

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
aabmass, line 1: If this is new code, please add type annotations. It looks like some of it is copied/refactored which I'm OK to ignore to not bloat the scope of this PR.

util/opentelemetry-util-genai/src/opentelemetry/util/genai/stream.py
```diff
@@ -0,0 +1,272 @@
...
+            self._safe_finalize_failure(error)
+            raise
+        try:
+            self._process_chunk(chunk)
+        except Exception as error:  # pylint: disable=broad-exception-caught
+            self._handle_process_chunk_error(error)
+        return chunk
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self.stream, name)
```
aabmass: Also FYI this doesn't work for magic method function calls, so e.g. `len()` or `str()` will break. That's why we usually use wrapt, but I'm ok with this approach for now.

eternalcuriouslearner: I ported the class to use `ObjectProxy`. I realized the pr for bumping wrapt was merged and I was planning to use it in responses wrappers. But I felt it's better to use it here and remove the unwanted `__getattr__` method. Thanks for reminding me.

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_buffers.py
eternalcuriouslearner: @aabmass the guard is the fix for issue #4344. OpenAI-compatible providers (vLLM, TGI) do send arguments=None. I am going leave this as is, hope it's okay with you.

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
eternalcuriouslearner: I believe in streaming responses, tool-call arguments arrive as fragments of one JSON string, so inserting commas between chunks can corrupt valid JSON. Hence I left them as they were written previously.

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/chat_wrappers.py
eternalcuriouslearner: I am using `getattr` only for SDK-version-dependent optional streaming fields: service_tier and usage. Currently, OpenAI older supported versions, especially 1.26.0, do not consistently expose these attributes on every streaming chunk. Direct access can raise AttributeError inside _process_chunk(), which causes
instrumentation chunk processing to be skipped and leaves usage/output telemetry missing. However, I am not using `getattr` for stable fields like id, model, and choices still use direct typed access.

aabmass (approved):
LG thank you!

JamieDanielson (approved):

PR #4501, opened 2026-04-29
openai-v2: use genai-utils histogram factories for metrics

lzchen:
Replace local bucket boundary definitions and histogram creation in instruments.py with `create_duration_histogram`/`create_token_histogram` from `opentelemetry.util.genai.instruments`.


xrmx (commented):
The genai-util descriptions for these histograms are different though, the current ones matches what we have in the python semconv package

lmolkova (approved):

PR #4439, opened 2026-04-15
feat(celery): Add task and worker lifecycle metrics

diogosilva30:
# Description

Implements a subset of the [Prometheus metrics exposed by Celery Flower](https://flower.readthedocs.io/en/latest/prometheus-integration.html#available-metrics) directly in the Celery instrumentation, so users who only need basic task/worker observability no longer need to run Flower as a separate service.

Fixes #3458

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)

### Task-level metrics (in `CeleryInstrumentor`)

Metric tracking wired into the existing `prerun`/`postrun`/`failure`/`retry` handlers:

| Flower metric | Impl. name | Labels | Type |
|---|---|---|---|
| `flower_events_total` | `flower.events.total` | task, type, worker | Counter |
| `flower_task_runtime_seconds` | `flower.task.runtime.seconds` | task, worker | Histogram |
| `flower_worker_number_of_currently_executing_tasks` | `flower.worker.number.of.currently.executing.tasks` | worker | UpDownCounter (gauge) |

### Worker lifecycle metrics (new `CeleryWorkerInstrumentor`)

A separate instrumentor that hooks into `worker_ready` / `worker_shutdown` / `task_received` / `task_revoked` signals in the main worker process:

| Flower metric | Impl. name | Labels | Type |
|---|---|---|---|
| `flower_events_total` | `flower.events.total` | task, type, worker | Counter |
| `flower_worker_online` | `flower.worker.online` | worker | UpDownCounter (gauge) |

Registered as a new `celery_worker` entry point in `pyproject.toml`.

### Omitted metrics

`flower_task_prefetch_time_seconds` and `flower_worker_prefetched_tasks` are **not** implemented. In `pool=prefork` mode (the default), `task_received` fires in the main process while `task_prerun` fires in a child process — there is no shared state to compute the time delta between them. Flower can do this because it consumes Celery events externally in a single process; replicating that inside the worker is not feasible without producer-side instrumentation or protocol changes.

### Bug fix — memory leak in `task_id_to_start_time`

Previously, `task_id_to_start_time` was a class-level dict that was never cleaned up after task completion, growing unbounded over time. It is now instance-scoped and entries are removed in `_trace_postrun`.

### Housekeeping

- Type annotations added across `__init__.py` and `utils.py`
- Typed `CeleryGetter` properly (`Getter[Request]`)
[...]

PR #4505, opened 2026-04-29
fix(celery): type getter and harden utility helpers

diogosilva30:
## Description

Split out of #4439.

This PR keeps only the Celery housekeeping changes:
- type `CeleryGetter` and normalize its returned values
- add null guards and helper typing in `utils.py`
- refine utility helper protocols and context aliases to match actual call sites
- update getter tests to match the typed behavior
- add a Sphinx nitpick exception for `celery.worker.request.Request` used in the documented getter type annotation

## Testing

- `uv run tox -e py311-test-instrumentation-celery -- -q`
- `uv run tox -e docs -- -W`

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/utils.py
+++ b/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/utils.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 
 import logging
 from collections.abc import Mapping
-from typing import TYPE_CHECKING, Any, Optional, Tuple, cast
+from typing import TYPE_CHECKING, Any, Optional, Protocol, cast
 
 from celery import registry  # pylint: disable=no-name-in-module
 from celery.app.task import Task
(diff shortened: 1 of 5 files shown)


MikeGoldsmith (approved):
Thanks @diogosilva30 🚀

diogosilva30:
@MikeGoldsmith rebased from main, and added towncrier fragment instead of changelog edit

emdneto:
Thanks for the PR!

tammy-baylis-swi (approved):

instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/utils.py
```diff
@@ -144,7 +157,7 @@ def attach_context(
     if task is None:
         return
 
-    ctx_dict = getattr(task, CTX_KEY, None)
+    ctx_dict = cast(Optional[ContextDict], getattr(task, CTX_KEY, None))
```
xrmx: Can you please use the `| None` syntax instead? We are going to add ruff rules for enforcing that and you are introducing quite a bit of them.

PR #4530, opened 2026-05-04
fix: batch pip installs in bootstrap for faster installs

### LINKED ISSUE #4484: opentelemetry-bootstrap -a install is unnecessarily slow — recommend piping to pip install in docs

`opentelemetry-bootstrap --action=install` is very slow because it calls `pip install -U --upgrade-strategy=only-if-needed <package>` in a loop — once per discovered instrumentor ([source](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py#L136-L138)).

For a typical project with ~8 instrumentors, this means:

* 8 separate subprocess spawns (each ~0.5s Python startup overhead)
* 8 separate pip loads (~1s each)
* 8 separate dependency resolution passes with -U flag (checking PyPI for upgrades each time)
* 8 separate pip check validations implicitly
In my testing this takes 60–120 seconds in a Docker build.

### Faster alternative
`opentelemetry-bootstrap --action=requirements | pip install -r /dev/stdin`

This is faster because:

* 1 subprocess instead of 8
* 1 pip startup instead of 8
* 1 resolver pass that handles all packages together, instead of 8 independent passes
* No `-U` flag — in a fresh environment (e.g., Docker build), there's nothing to upgrade, so the upgrade check is pure overhead
In my testing this takes 15–25 seconds — a 4–6x speedup.

### Relation to existing docs

The [troubleshooting docs](https://opentelemetry.io/docs/zero-code/python/troubleshooting/#bootstrap-using-uv) already recommend a similar pattern for `uv`:
`uv run opentelemetry-bootstrap -a requirements | uv add --requirement -`
But there's no equivalent recommendation for pip users. Since pip is still the default package manager for most Python users, it would be helpful to document this pattern as well.

### Suggestions
1. Add a "Bootstrap using pip" section in the troubleshooting docs alongside the existing "Bootstrap using uv" section, recommending:
`opentelemetry-bootstrap --action=requirements | pip install -r /dev/stdin`
2. Consider changing `_run_install` to batch all packages into a single `pip install` call instead of looping, which would make `--action=install` fast by default without requiring users to discover this workaround.

### Environment
Python 3.14
opentelemetry-distro (latest)
Docker build context (fresh venv each time, no cache benefit from `-U`)

Manvi2402:
# Description

Previously, `_run_install` in `bootstrap.py` called `pip install` once per 
package in a loop, causing multiple subprocess spawns, pip loads, and 
dependency resolution passes. For a typical project with ~8 instrumentors, 
this took 60-120 seconds.

This change batches all packages into a single `pip install` call, reducing 
it to 15-25 seconds (4-6x speedup).

Fixes #4484

## Type of change

# How Has This Been Tested?

- Ran pre-commit checks — all passed
- Verified syntax with `python -m py_compile`

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


CHANGELOG.md
```diff
@@ -21,7 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#4457](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4457))
 
 ### Fixed
-
+- `opentelemetry-instrumentation`: Batch all packages into a single `pip install` call in `bootstrap.py` to improve performance
+  ([#4484](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4484))
```
herin049: This needs to be updated to link to the PR

herin049 (changes requested):

CHANGELOG.md
Manvi2402: Updated the changelog to link to the PR instead of the issue. Thanks for the review.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py
```diff
@@ -71,7 +71,7 @@ def _sys_pip_install(package):
                 "-U",
```
idan-rahamim-lendbuzz, line 62: Thanks For your contribution !
Can you also add to the `opentelemetry-bootstrap -a install` command a flag support to disable the `-U ` of pip install? 
In a fresh environment (e.g., Docker build), there's nothing to upgrade, so the upgrade check is pure overhead

Manvi2402: Thanks for the suggestion. Added a --no-upgrade flag to disable the -U option for fresh environments like Docker builds.

idan-rahamim-lendbuzz (approved):

MikeGoldsmith (changes requested):
Nice optimization — batching pip installs is a cool thing to see.

A few things to tidy up:

1. **Error message broken for batched packages** — `f'{msg} for package "{packages}"'` will print the list repr (e.g. `for package "['pkg1', 'pkg2']"`). Should join them, e.g. `f'{msg} for packages "{", ".join(packages)}"'`.

2. **`_syscall` wrapper is fragile** — `if packages:` is falsy for `[]`, which would fall through to `func()` with no args and crash. Currently protected by the `if libs:` guard in `_run_install`, but the wrapper shouldn't rely on callers to not pass empty lists.

3. **`--no-upgrade` flag** — this is a separate feature from the batching change. Consider splitting it into its own PR or at minimum mentioning it in the CHANGELOG entry.

4. **CHANGELOG section** — this is a performance improvement, not a bug fix. Should be under a different section.

Manvi2402:
Hi @MikeGoldsmith , all review comments addressed!

Fixed error message to use ', '.join(packages) instead of list repr
Added empty list guard in _syscall wrapper
Removed --no-upgrade flag
Moved CHANGELOG entry to ### Changed section
Updated tests to reflect batched pip install behavior"

emdneto:
Thanks for the PR!

opentelemetry-instrumentation/src/opentelemetry/instrumentation/bootstrap.py
emdneto: is this fixed?

emdneto (commented):
CI failures are not related to this PR. Investigating.

PR #3779, opened 2025-09-24
celery: allow using links instead of child spans for task execution

jlc-christie:
# Description

As described in #3002, the current default behaviour results in all celery tasks that execute are child spans of the code that pushed it on to the broker, which [conflicts with the semantic conventions](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3002#issuecomment-3211249453). 

This PR **doesn't change the default behaviour** since this may be undesirable for consumers expecting the behaviour to be consistent.

Fixes #3002 by providing an optional parameter to the `.instrument()` method to allow for optionally using span links, e.g. 
```python
CeleryInstrumentor().instrument(use_links=True)
```

## Type of change
- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- Unit test added
- Tested E2E locally with jaeger with `use_links=False` and also `use_links=True` and observed the expected behaviour: when the flag is `False` (default) all task executions of a fan-out job are child spans of the code that enqueued it and when `True` they appeared as linked spans. 

# Does This PR Require a Core Repo Change?
- No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [X] Followed the style guidelines of this project
- [X] Changelogs have been updated
- [X] Unit tests have been added
- [X] Documentation has been updated

PR #4537, opened 2026-05-06
celery: allow using links instead of child spans for task execution

### LINKED ISSUE #3002: Should celery add a span link instead?
### Describe your environment

Currently, the celery task that is executing gets the context from the headers, and attaches it to the new span:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/e3ba54b95c68a0b56daccb80cd57f1987714ddf4/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py#L173-L180

I don't think this is the correct behavior. It should actually be a [span link](https://opentelemetry.io/docs/concepts/signals/traces/#span-links) - according to the docs:
> Links exist so that you can associate one span with one or more spans, implying a causal relationship. For example, let’s say we have a distributed system where some operations are tracked by a trace.

Which is exactly this case.

When a web framework triggers a celery task, it shouldn't be on the same trace. It's already a new one... But they are linked through the span link.

### What happened?

...

### Steps to Reproduce

...

### Expected Result

...

### Actual Result

...

### Additional context

_No response_

Yes

jlc-christie:
# Description

Duplicate of closed (stale, lack of reviews) PR #3779 since I lack the permissions to re-open.

As described in #3002, the current default behaviour results in all celery tasks that execute are child spans of the code that pushed it on to the broker, which [conflicts with the semantic conventions](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3002#issuecomment-3211249453). 

This PR **doesn't change the default behaviour** since this may be undesirable for consumers expecting the behaviour to be consistent.

Fixes #3002 by providing an optional parameter to the `.instrument()` method to allow for optionally using span links, e.g. 
```python
CeleryInstrumentor().instrument(use_links=True)
```

## Type of change
- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- Unit test added
- Tested E2E locally with jaeger with `use_links=False` and also `use_links=True` and observed the expected behaviour: when the flag is `False` (default) all task executions of a fan-out job are child spans of the code that enqueued it and when `True` they appeared as linked spans. 

- No.

# Checklist:


jlc-christie:
@tammy-baylis-swi @MikeGoldsmith would either of you be able to take a look at this if you get some spare time please? I've noticed you've both been active with celery instrumentation reviews recently. It's been previously approved by @Kludex on https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3779 but was closed due to no second review a while ago.

emdneto:
Thanks for the PR!

tammy-baylis-swi:
Thank you for your patience @jlc-christie . I think having this as an opt-in to better meet semantic conventions makes sense.

If it's not too much trouble it would be nice if the [docker tests](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/tests/opentelemetry-docker-tests/tests/celery/test_celery_functional.py) also tested this feature. Else I think this looks good after addressing the `towncrier` generation.

jlc-christie:
Thanks @emdneto and @tammy-baylis-swi, wasn't aware of the change log process change so thanks for that. Also added the docker test as requested.

emdneto (commented):
LGTM. I'm more in favor of making this the default behavior to match the spec and treat it like a fix. We can also put this under the opt-in semconv stability gate, so it works for zero-code instrumentation as well.

jlc-christie:
> LGTM. I'm more in favor of making this the default behavior to match the spec and treat it like a fix. We can also put this under the opt-in semconv stability gate, so it works for zero-code instrumentation as well.

Thanks for the review, happy to make those changes if everyone agrees that's the correct approach, but there was some good points on the original [issue](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3002) arguing against making it the default. @tammy-baylis-swi what are your thoughts?

jlc-christie:
sorry to ping you again @tammy-baylis-swi but would be great to get this one resolved and merged 🙏🏼

jlc-christie:
@emdneto any opinions on the previous comment? happy to implement any changes - just looking for strong direction/approval since there are some conflicting opinions and it would be good to get something merged that unblocks the issue of _having_ to have child spans with celery since it makes things like fan-out jobs a horrible UX in most o11y frontends.

emdneto (approved):

MikeGoldsmith (approved):

xrmx (approved):
I think this is fine as a temporary solution. This should not be an option anymore with `OTEL_SEMCONV_STABILITY_OPT_IN=messaging` and the parameter will be eventually deprecated once we have the opt-in support.

instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py
```diff
@@ -126,13 +136,15 @@ def instrumentation_dependencies(self) -> Collection[str]:
 
     def _instrument(self, **kwargs):
         tracer_provider = kwargs.get("tracer_provider")
```
xrmx, line 139:
```suggestion
        # TODO: deprecate this when support for stable semconv is available
        tracer_provider = kwargs.get("tracer_provider")
```

PR #4538, opened 2026-05-06
refactor(genai-util): pass sampling attributes at span creation time

lzchen:
Remove non-sampling parameters from InferenceInvocation.__init__ and pass provider, model, and server attributes to start_span() so that head-based samplers can make decisions on them.

Wil also do the same for other invocation classes in followup prs.

### DIFF
--- a/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
+++ b/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
@@ -55,7 +55,7 @@ class InferenceInvocation(GenAIInvocation):
     context manager rather than constructing this directly.
     """
 
-    def __init__(  # pylint: disable=too-many-locals
+    def __init__(
         self,
         tracer: Tracer,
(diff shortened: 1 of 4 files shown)


util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
```diff
@@ -55,7 +55,7 @@ class InferenceInvocation(GenAIInvocation):
     context manager rather than constructing this directly.
     """
 
-    def __init__(  # pylint: disable=too-many-locals
```
lmolkova, line 58: would it be possible to expand the scope to other invocation types?

lmolkova (approved):

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_inference_invocation.py
lzchen: Yes! I left in the description that I will be creating followup prs for each invocation type.

PR #4542, opened 2026-05-06
fix(auto-instrumentation): print debug startup messages

### LINKED ISSUE #4540: [instrumentation] write auto-instrumentation loader messages to stderr when OTEL_LOG_LEVEL=debug
# Write auto-instrumentation debug messages to stderr when OTEL_LOG_LEVEL=debug

## What problem do you want to solve?

When Python auto-instrumentation runs, it is hard to tell which instrumentors were loaded. These messages are logged, but they are usually not visible when `opentelemetry-instrument` runs even though `OTEL_LOG_LEVEL=debug` may have been set. This is because auto-instrumentation runs from `sitecustomize`, before the application has had a chance to configure logging.

Writing these messages to stderr would be especially useful in k8s/operator-based deployments, where many instrumentation packages may be present in the auto-instrumentation image and users need to understand what happened at startup.

## Expected behavior

When the user explicitly sets `OTEL_LOG_LEVEL=debug` or `OTEL_LOG_LEVEL=trace`, auto-instrumentation should write minimal startup messages to stderr.

For example:

```text
DEBUG:opentelemetry.instrumentation.auto_instrumentation._load:Distribution 'distro' will be configured
DEBUG:opentelemetry.instrumentation.auto_instrumentation._load:Instrumented 'requests'
```

## Proposed solution

Write auto-instrumentation startup messages to stderr when auto-instrumentation runs and `OTEL_LOG_LEVEL=debug` or `OTEL_LOG_LEVEL=trace` is set. Don't attempt to configure application logging. In the unlikely event that the application has already set up logging and then makes an explicit call to `auto_instrumentation.initialize()`, suppress this new output channel to prevent double logging.

## Would you like to implement a fix?

Yes.

pmcollins:
# Description

Writes auto-instrumentation startup debug messages to stderr when `OTEL_LOG_LEVEL=debug` or `OTEL_LOG_LEVEL=trace` is set. This makes it easier to see which distro and instrumentors were loaded during `opentelemetry-instrument` startup.

If Python logging is already configured, the new stderr output is suppressed to avoid double logging.

Fixes #4540

## Type of change

# How Has This Been Tested?

- [x] Ran `python -m unittest opentelemetry-python-contrib/opentelemetry-instrumentation/tests/auto_instrumentation/test_load.py`
- [x] Ran `oteltest` with local packages and confirmed the expected stderr debug messages

- [x] No.

# Checklist:

### DIFF
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
@@ -2,7 +2,14 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from functools import cached_property
-from logging import DEBUG, NOTSET, Logger, LoggerAdapter, getLogger
+from logging import (
+    DEBUG,
+    NOTSET,
+    Logger,
+    LoggerAdapter,
+    NullHandler,
+    getLogger,
+)
 from os import environ
 from sys import stderr
 
@@ -50,20 +57,12 @@ class _OtelLogLevelLoggerAdapter(LoggerAdapter):
     OTEL_LOG_LEVEL asks for debug output and Python logging would not emit them.
     """
 
-    def __init__(self, logger, extra):
-        super().__init__(logger, extra)
-        # This adapter is built when this module is imported. That covers the
-        # sitecustomize path where auto-instrumentation runs before application
-        # logging setup, and the edge case where Python logging is set up and then
-        # auto_instrumentation.initialize() is invoked explicitly. If this adapter
-        # is used after application code can change logging, it should be modified
-        # to not use the cached value in the _logger_emits_debug field.
-        self._logger_emits_debug = self._logger_would_emit(DEBUG)
-
     def debug(self, msg: str, *args: object, **kwargs: object) -> None:
         super().debug(msg, *args, **kwargs)
 
-        if not _otel_log_level_allows_debug() or self._logger_emits_debug:
+        if not _otel_log_level_allows_debug() or self._logger_would_emit(
+            DEBUG
+        ):
             return
 
         message = msg
@@ -81,6 +80,9 @@ class _OtelLogLevelLoggerAdapter(LoggerAdapter):
         logger: Logger | None = self.logger
         while logger:
             for handler in logger.handlers:
+                if isinstance(handler, NullHandler):
+                    continue
+
                 if handler.level == NOTSET or level >= handler.level:
                     return True
 
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
@@ -2,8 +2,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from functools import cached_property
-from logging import getLogger
+from logging import DEBUG, getLogger
 from os import environ
+from sys import stderr
 
 from opentelemetry.instrumentation.dependencies import (
     DependencyConflictError,
@@ -25,6 +26,37 @@ from opentelemetry.util._importlib_metadata import (
 _logger = getLogger(__name__)
 
 SKIPPED_INSTRUMENTATIONS_WILDCARD = "*"
+OTEL_LOG_LEVEL = "OTEL_LOG_LEVEL"
+_DEBUG_LOG_LEVELS = frozenset(("trace", "debug"))
+
+
+def _diagnostic_debug_enabled() -> bool:
+    log_level = environ.get(OTEL_LOG_LEVEL, "").strip().lower()
+    return log_level in _DEBUG_LOG_LEVELS
+
+
+def _logger_handles_debug() -> bool:
+    return _logger.isEnabledFor(DEBUG) and _logger.hasHandlers()
+
+
+def _format_diagnostic_arg(arg: object) -> object:
+    if isinstance(arg, str):
+        return repr(arg)
+
+    return arg
+
+
+def _debug(message: str, *args: object) -> None:
+    _logger.debug(message, *args)
+
+    if not _diagnostic_debug_enabled() or _logger_handles_debug():
+        return
+
+    if args:
+        message = message % tuple(_format_diagnostic_arg(arg) for arg in args)
+
+    stderr.write(f"DEBUG:{__name__}:{message}\n")
+    stderr.flush()
 
 
 class _EntryPointDistFinder:
@@ -56,12 +88,12 @@ def _load_distro() -> BaseDistro:
             if distro_name is None or distro_name == entry_point.name:
                 distro = entry_point.load()()
                 if not isinstance(distro, BaseDistro):
-                    _logger.debug(
+                    _debug(
                         "%s is not an OpenTelemetry Distro. Skipping",
                         entry_point.name,
                     )
                     continue
-                _logger.debug(
+                _debug(
                     "Distribution %s will be configured", entry_point.name
                 )
                 return distro
@@ -89,7 +121,7 @@ def _load_instrumentors(distro):
             break
 
         if entry_point.name in package_to_exclude:
-            _logger.debug(
+            _debug(
                 "Instrumentation skipped for library %s", entry_point.name
             )
             continue
@@ -98,7 +130,7 @@ def _load_instrumentors(distro):
             entry_point_dist = entry_point_finder.dist_for(entry_point)
             conflict = get_dist_dependency_conflicts(entry_point_dist)
             if conflict:
-                _logger.debug(
+                _debug(
                     "Skipping instrumentation %s: %s",
                     entry_point.name,
                     conflict,
@@ -107,13 +139,13 @@ def _load_instrumentors(distro):
 
             # tell instrumentation to not run dep checks again as we already did it above
             distro.load_instrumentor(entry_point, skip_dep_check=True)
-            _logger.debug("Instrumented %s", entry_point.name)
+            _debug("Instrumented %s", entry_point.name)
         except DependencyConflictError as exc:
             # Dependency conflicts are generally caught from get_dist_dependency_conflicts
             # returning a DependencyConflict. Keeping this error handling in case custom
             # distro and instrumentor behavior raises a DependencyConflictError later.
             # See https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3610
-            _logger.debug(
+            _debug(
                 "Skipping instrumentation %s: %s",
                 entry_point.name,
                 exc.conflict,
@@ -123,7 +155,7 @@ def _load_instrumentors(distro):
             # ModuleNotFoundError is raised when the library is not installed
             # and the instrumentation is not required to be loaded.
             # See https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3421
-            _logger.debug(
+            _debug(
                 "Skipping instrumentation %s: %s", entry_point.name, exc.msg
             )
             continue
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
@@ -1,6 +1,7 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
+import sys
 from functools import cached_property
 from logging import (
     DEBUG,
@@ -11,7 +12,6 @@ from logging import (
     getLogger,
 )
 from os import environ
-from sys import stderr
 
 from opentelemetry.instrumentation.dependencies import (
     DependencyConflictError,
@@ -47,6 +47,16 @@ def _format_log_arg(arg: object) -> object:
     return arg
 
 
+def _format_log_message(msg: str, args: tuple[object, ...]) -> str:
+    if not args:
+        return msg
+
+    try:
+        return msg % tuple(_format_log_arg(arg) for arg in args)
+    except Exception:  # pylint: disable=broad-except
+        return msg
+
+
 class _OtelLogLevelLoggerAdapter(LoggerAdapter):
     """Write startup debug messages to stderr when logging would drop them.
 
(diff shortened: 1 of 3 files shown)


opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -36,6 +37,37 @@
...
+
+
+def _format_diagnostic_arg(arg: object) -> object:
+    if isinstance(arg, str):
+        return repr(arg)
+
+    return arg
+
+
+def _debug(message: str, *args: object) -> None:
```
herin049: Instead of defining a helper `_debug` function, it might be worth just creating a logger adapter. e.g. 
```python
class _DiagnosticLoggerAdapter(logging.LoggerAdapter):
    def log(self, level, msg, *args, **kwargs):
        super().log(level, msg, *args, **kwargs)

        env_level = _otel_log_level()
        if env_level is None or level < env_level:
            return 
        if self.logger.isEnabledFor(level) and _handler_would_emit(self.logger, level):
            return
        ...
        # print to stderr 

_logger = _DiagnosticLoggerAdapter(logging.getLogger(__name__), {})
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -36,6 +37,37 @@
...
+_DEBUG_LOG_LEVELS = frozenset(("trace", "debug"))
+
+
+def _diagnostic_debug_enabled() -> bool:
+    log_level = environ.get(OTEL_LOG_LEVEL, "").strip().lower()
+    return log_level in _DEBUG_LOG_LEVELS
+
+
+def _logger_handles_debug() -> bool:
+    return _logger.isEnabledFor(DEBUG) and _logger.hasHandlers()
```
herin049: Note that this check isn't sufficient. The logger could be enabled for `DEBUG` but have all of the handlers set to a higher level. You'd want to iterate through all of the logger handlers and determine if any of them would emit.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -36,6 +37,37 @@
 _logger = getLogger(__name__)
 
 SKIPPED_INSTRUMENTATIONS_WILDCARD = "*"
+OTEL_LOG_LEVEL = "OTEL_LOG_LEVEL"
+_DEBUG_LOG_LEVELS = frozenset(("trace", "debug"))
```
herin049: It might be easier to just define a mapping between OTel log level and Python log level.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
pmcollins: Excellent idea. Done.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
pmcollins: Thanks -- fixed.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
pmcollins: I tried doing this and ended up with a mapping of otel trace to python debug and otel debug to python debug (python doesn't have trace). Not sure if there's enough of a value add to do this?

herin049: The one use-case I can think of is supporting different values of `OTEL_LOG_LEVEL` in the future. I'm fine keeping it as is though if we don't foresee supporting different levels.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -22,9 +23,78 @@
...
+        stderr.flush()
+
+    def _logger_would_emit(self, level: int) -> bool:
+        # If the logger itself would reject this level, don't bother walking handlers.
+        if not self.logger.isEnabledFor(level):
+            return False
+
+        logger: Logger | None = self.logger
+        while logger:
+            for handler in logger.handlers:
```
herin049, line 116: We should probably also ignore `NullHandler` here.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -22,9 +23,78 @@
...
+
+    def __init__(self, logger, extra):
+        super().__init__(logger, extra)
+        # This adapter is built when this module is imported. That covers the
+        # sitecustomize path where auto-instrumentation runs before application
+        # logging setup, and the edge case where Python logging is set up and then
+        # auto_instrumentation.initialize() is invoked explicitly. If this adapter
+        # is used after application code can change logging, it should be modified
+        # to not use the cached value in the _logger_emits_debug field.
+        self._logger_emits_debug = self._logger_would_emit(DEBUG)
```
herin049: Correct me if I'm wrong, but this still doesn't handle the case where 

```python
from opentelemetry.instrumentation.auto_instrumentation import initialize
# configure logging
...
initialize()
```
Although it adds some overhead, maybe it's better to just call `_logger_would_emit` on each emit. WDYT?

herin049 (approved):

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
pmcollins: Fixed.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
pmcollins: Oh, yeah 😅. Not caching anything anymore.

herin049 (approved):

emdneto:
Thanks for the PR!

lzchen (approved):

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -22,9 +30,73 @@
     entry_points,
 )
 
-_logger = getLogger(__name__)
-
 SKIPPED_INSTRUMENTATIONS_WILDCARD = "*"
+OTEL_LOG_LEVEL = "OTEL_LOG_LEVEL"
```
aabmass, line 38: Is this in available for import in any of the `environment.py` files?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -22,9 +30,73 @@
...
+        super().debug(msg, *args, **kwargs)
+
+        if not _otel_log_level_allows_debug() or self._logger_would_emit(
+            DEBUG
+        ):
+            return
+
+        message = msg
+        if args:
+            message = message % tuple(_format_log_arg(arg) for arg in args)
```
aabmass: does this behave the same as usual logging calls like `logging.info()` calls work?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -22,9 +30,73 @@
...
+
+    return arg
+
+
+class _OtelLogLevelLoggerAdapter(LoggerAdapter):
+    """Write startup debug messages to stderr when logging would drop them.
+
+    Auto-instrumentation usually runs from sitecustomize before the
+    application configures logging, so normal logger.debug calls are often
+    not visible even when OTEL_LOG_LEVEL=debug. This adapter keeps normal
```
aabmass: I think the default in python is warning and above. Does this handle just debug or does it also print info level?

aabmass (approved):

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -2,8 +2,16 @@
...
+from logging import (
+    DEBUG,
+    NOTSET,
+    Logger,
+    LoggerAdapter,
+    NullHandler,
+    getLogger,
+)
 from os import environ
+from sys import stderr
```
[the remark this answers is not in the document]
aabmass: 🤷 feel free to resolve lol

i guess it could be true for something like https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stderr

xrmx (commented):
Maybe a silly question: why are we not setting up our own handler?

herin049:
> Maybe a silly question: why are we not setting up our own handler?

Presumably to prevent double logging if there is also a user-defined handler.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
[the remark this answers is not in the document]
pmcollins: Thanks, fixed.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
pmcollins: Unfortunately, the only one I found is in the SDK.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
pmcollins: Not quite -- it's simplified and tweaked a little just for this narrow, local use case. It puts quotes around string arguments so the output looks like this:

```
DEBUG:opentelemetry.instrumentation.auto_instrumentation._load:Distribution 'distro' will be configured
DEBUG:opentelemetry.instrumentation.auto_instrumentation._load:Instrumented 'requests'
```

I added a catch all exceptions guard though.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
pmcollins: It doesn't print at info level, but maybe it should -- idk. WDYT?

pmcollins:
> Maybe a silly question: why are we not setting up our own handler?

Not a silly question -- I'm happy to set up a handler if folks prefer, but my understanding is that that would be more within the purview of the application, and that we, as a library, should just call the logging API and not attempt to configure logging. For this issue, we're in a middle ground, where we want to log, but we don't want to configure logging, and the application typically hasn't had a chance to do so either, so we just print these startup messages to stderr.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
aabmass: I guess that's ok, but we could use a real handler in theory. Just wondering why you didn't take that approach

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
aabmass: If we print debug logs, I would expect debug and higher to also print

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
pmcollins: Yeah, but were you thinking we'd remove the handler after auto-instrumentation? It seems that setting up logging falls under the purview of the application and that as a library we shouldn't do any logging setup. There's a [blurb](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) about it in the Python docs:

>Note It is strongly advised that you do not add any handlers other than [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

We want to print a few log messages, but we're a library and we don't want to set up logging, and the application hasn't had a chance to set up logging, so if the user requested otel debug logs, we print directly to stderr for this brief, early window.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
pmcollins: I guess what I meant is that with this PR, these startup messages are only printed when `OTEL_LOG_LEVEL=debug` or `OTEL_LOG_LEVEL=trace`. If `OTEL_LOG_LEVEL=info`, they are not printed because the startup messages are debug logs.

The adapter only overrides `debug()` because that's the only adapter method that needs a custom fallback. Warnings and errors still go through normal logging, which already prints them when no handlers are configured, and we're not logging at the info level during startup.

Would you prefer that we also override `info()` to future-proof this?

herin049: Based on my original suggestion, it's probably best to just generalize this to just override `log()` and check the log level.

herin049: i.e.
```python
class _DiagnosticLoggerAdapter(logging.LoggerAdapter):
    def log(self, level, msg, *args, **kwargs):
        super().log(level, msg, *args, **kwargs)

        env_level = _otel_log_level()
        if env_level is None or level < env_level:
            return 
        if self.logger.isEnabledFor(level) and _handler_would_emit(self.logger, level):
            return
        ...
```

pmcollins: Thanks @herin049 and @aabmass -- I've pushed a change to override the `log` method instead of just the `debug` method.

pmcollins:
Thanks all for the reviews so far. Since this PR has already received some approvals, I wanted to call attention to a change in the latest update: the fallback logic changed from "would this debug record be emitted by Python logging?" to "has the application configured logging?".

If the application has configured logging, we now skip the stderr fallback and defer to the application’s logger/handler configuration, even if that configuration would drop this specific debug message -- the idea being that the stderr fallback is only for startup messages before application logging exists, not a way to bypass application logging. I'm open to suggestions on this, but I feel that this is a more honest solution as it respects the application's logging configuration if it exists.

All of this should be uncommon in practice, since it requires the application to first configure logging and then explicitly call the auto-instrumentation function, but it's supported for completeness, and I wanted to mention these changes in intent.

PR #4553, opened 2026-05-07
refactor(genai-util): pass sampling attributes at span creation for rest of invocation types

lzchen:
Continuing of https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4538

Extend the pattern from InferenceInvocation to AgentInvocation, WorkflowInvocation, EmbeddingInvocation, and ToolInvocation so that all invocation types pass base attributes to start_span() for head-based sampling decisions.


util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py
lmolkova, line 56: how come agent name is not one of them? It looks like a bug in semconv https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-agent-spans.md

it's part of span name and should be an optional parameter

we should also set proper span name here instead of https://github.com/open-telemetry/opentelemetry-python-contrib/blob/955e488d8db7b9564a40b3a9e6bed3f154a71819/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py#L224

lmolkova: If you have time, @lzchen could you send a pr to  semconv? Otherwise I can send tomorrow. I'd rather go with agent name right away in this PR and not block on semconv.

lzchen: I changed it so we pass in agent name in instantiation but left the logic on `_apply_finish` since agent name is optional on creation and can still be set afterwards. Is this something we want to keep as behavior or should we enforce agent name upon creation (will be breaking).

lzchen: Also https://github.com/open-telemetry/semantic-conventions-genai/pull/107

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py
```diff
@@ -68,17 +67,17 @@ def __init__(
...
-            attributes=attributes,
-            metric_attributes=metric_attributes,
         )
         self.provider = provider
         self.request_model = request_model
         self.server_address = server_address
         self.server_port = server_port
 
-        self.agent_name: str | None = None
+        self.agent_name: str | None = agent_name
```
lmolkova, line 80: can we remove https://github.com/open-telemetry/opentelemetry-python-contrib/blob/47a3d9dd39aaaa136d8532d4033b2cbb1aa8a193/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py#L204 ?

it should be invalid to change operation name after the fact, it could potentially affect sampling decisions and we should not need to update span name.

lmolkova (approved):

util/opentelemetry-util-genai/src/opentelemetry/util/genai/_agent_invocation.py
lzchen: Yeah I was wondering about this. Removed.

PR #4577, opened 2026-05-13
opentelemetry-instrumentation-tornado: reduce cardinality of span names and metrics attributes

xrmx:
# Description

This PR fixes two cardinality issues for the tornado instrumentation:
- in the span name, where we are now using a low-cardinality version of the request path
- in the metrics attributes for stable http semantic conventions. This changes the attributes of some of the metrics dropping out of spec `url.query`, `url.path` in favor of in-spec `http.route`. For older semantic conventions this is not possible since `http.target` is high cardinality by definition.

Since Tornado does not offer us any helper to get the templated name of a route from the handler we need to introduce a basic reimplementation of its routing algorithm. We then reuse a tornado internal method to subsitute the regexp groups in the rule with some names taken from the handler method function parameters, this internal function has been there since ages so I think it's pretty safe to reuse.

Assisted-by: Cursor

## Type of change

# How Has This Been Tested?

- [x] tox

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
```diff
@@ -718,22 +720,28 @@ def _get_attributes_from_request(request, sem_conv_opt_in_mode):
...
+def _get_default_span_name(handler):
     """
-    Default span name is the HTTP method and URL path, or just the method.
+    Default span name is the HTTP method and route, or just the method.
     https://github.com/open-telemetry/opentelemetry-specification/pull/3165
     https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/#name
 
     Args:
-        request: Tornado request object.
+        request: Tornado handler object.
```
lzchen:
```suggestion
        handler: Tornado handler object.
```

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
```diff
@@ -718,22 +720,28 @@ def _get_attributes_from_request(request, sem_conv_opt_in_mode):
...
-    if method and path:
-        return f"{method} {path}"
+    method = handler.request.method
+    path = handler.request.path
+    rule = find_matched_rule(handler)
+    # if there's no rule it may be a 404 so keep the path
+    if rule:
+        route = route_from_rule(rule, handler)
+    else:
+        route = path
```
lzchen: Is it possible that attackers probe random urls that return 404s and then random span names will be generated? Should we do something similar to other [instrumentations](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py#L165) where we use a fixed string like "HTTP 404" or just use {method}?

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/_utils.py
```diff
@@ -0,0 +1,88 @@
...
+        # Callable / connection delegate / other terminal target.
+        return _NOT_FOUND
+    return None
+
+
+def _is_handler_target(target: Any) -> bool:
+    return isinstance(target, type) and issubclass(target, RequestHandler)
+
+
+def route_from_rule(rule: Rule, handler: RequestHandler) -> str | None:
```
aabmass, line 66: nit but IMO it would be easier to follow this if we flatten the branching and use early returns for
- `if not hasattr(rule.matcher, "_find_groups")`
- `if num_params == 0:`

aabmass (approved):

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/_utils.py
```diff
@@ -0,0 +1,88 @@
...
+                f"{{{param}}}"
+                for param in inspect.signature(method).parameters.keys()
+            )
+            if len(method_args) == num_params:
+                route = format_str % method_args
+            else:
+                route = format_str
+        else:
+            # if we don't have parameters to substitute use the path directly
+            route = handler.request.path
```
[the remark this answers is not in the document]
aabmass: I guess this is similar to Leighton's comment above

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
xrmx: Changed to only return the method

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/_utils.py
xrmx: Indeed

instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/_utils.py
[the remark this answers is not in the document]
xrmx: Not sure that's something you'll see in real life but updated and added test.

lzchen (approved):

PR #4103, opened 2026-01-15
feat: propagate custom genai generate_content attributes to logs, in addition to spans

mwiatrowski:
# Description

#3961 added an ability to pass additional attributes to be added to generate_content {model.name} spans in Google GenAI instrumentation, through Open Telemetry context.

This PR extends that functionality by saving those attributes in generated logs as well.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

- [X] Added unit tests.
- [X] Tested the integration with ADK manually: I changed my local ADK fork to save extra attributes in OTel context and verified that they are included in the generated logs.

# Does This PR Require a Core Repo Change?

- [ ] Yes.
- [X] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [X] Followed the style guidelines of this project
- [x] Changelogs have been updated
- [X] Unit tests have been added
- [ ] Documentation has been updated

PR #4581, opened 2026-05-13
[google-genai] Provide a way to attach extra attributes to the operation-details event but not to the span

### LINKED ISSUE #4580: [google-genai] Provide a way to attach extra attributes to the operation-details event but not to the span

`opentelemetry-instrumentation-google-genai` currently exposes a single channel
for callers to attach extra attributes to `generate_content` telemetry:
[`GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY`][key] (added in #3961, #4103).
Whatever a caller puts there is written to **both**:

- the `generate_content {model}` span (via `span.set_attributes(...)`), and
- the `gen_ai.client.inference.operation.details` log event (via the
  `LogRecord` emitted in `_maybe_log_completion_details`).

There is no way to mark a caller-supplied attribute as event-only.

The existing `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` env var
(modes: `NO_CONTENT` / `SPAN_ONLY` / `EVENT_ONLY` / `SPAN_AND_EVENT`) only gates
the standard semconv message-content fields (`gen_ai.input.messages`,
`gen_ai.output.messages`, etc.) — not arbitrary caller-supplied extras.

This matters when callers want to attach attributes that are acceptable for
audit-style log/event consumers but undesirable on broadly-sampled spans —
for example, an end-user identifier (`user.id`). A concrete instance: the
[Google ADK][adk] passes a `common_attributes` dict containing
`user.id` (the `USER_ID` semconv constant from
`opentelemetry.semconv._incubating.attributes.user_attributes`) through this
context key, which causes `user.id` to land on every `generate_content` span
as soon as `opentelemetry-instrumentation-google-genai` is installed.
ADK currently has no way to keep that value off the span without also losing
it from the operation-details event.

[key]: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/generate_content.py#L100
[adk]: https://github.com/google/adk-python

Add a sibling context key (`GENERATE_CONTENT_EVENT_ONLY_EXTRA_ATTRIBUTES_CONTEXT_KEY`) that is read by all four `generate_content` wrappers
(sync/async × streaming/non-streaming) but is **only** merged into the
`gen_ai.client.inference.operation.details` `LogRecord`, never into
`span.set_attributes(...)`.

Behavior:

- Keys attached under the new context key appear on the
  `gen_ai.client.inference.operation.details` event log record.
- Those keys never appear on the `generate_content {model}` span.
- On key collisions with the existing `GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY`,
  the event-only value wins on the event; the span carries the existing
  extra-attributes value.
- The new constant is re-exported from `opentelemetry.instrumentation.google_genai`
  alongside `GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY`.

1. Have callers strip the attribute before calling the SDK and re-attach it
   to log records via a separate path. This is what ADK is being pushed
   toward today, but it requires duplicating the operation-details event
   construction outside contrib (or post-processing the LogRecord), defeating
   the point of having the contrib instrumentation own the event.
2. A new env var (e.g. `OTEL_INSTRUMENTATION_GENAI_DROP_SPAN_ATTRIBUTES`)
   listing attribute keys to strip from spans. Rejected because it puts the
   policy in deployment config rather than in the code that knows the
   attribute's semantics, and because keys are caller-defined there's no
   useful default list contrib could ship.

### Additional Context

_No response_

Yes

### Tip

Achuth17:
Fixes #4580

# Description

Adds a sibling context key,
`GENERATE_CONTENT_EVENT_ONLY_EXTRA_ATTRIBUTES_CONTEXT_KEY`, to
`opentelemetry-instrumentation-google-genai` for attaching caller-supplied
attributes that should be emitted only on the
`gen_ai.client.inference.operation.details` log event and **never** on the
`generate_content {model}` span.

The existing `GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY` (added in #3961
for spans, extended in #4103 for events) writes attributes to **both** the
span and the event with no opt-out. That makes it unsafe for caller-supplied
attributes that are acceptable on event log records but must not land on
broadly-sampled spans, for example, an end-user identifier (`user.id`).

Concrete motivating case: the [Google ADK][adk] passes a
`common_attributes` dict containing the `USER_ID` semconv constant
(string value `"user.id"`, from
`opentelemetry.semconv._incubating.attributes.user_attributes`) through this
context key, which causes `user.id` to land on every `generate_content` span
as soon as `opentelemetry-instrumentation-google-genai` is installed. ADK
has no way today to keep that value off the span without also losing it from
the operation-details event.

The new key follows the exact same naming and access pattern as the
existing `GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY`, keeping the
public API surface uniform. It is read in all four `generate_content`
wrappers (sync/async × streaming/non-streaming) and merged into the
`LogRecord` only — never into `span.set_attributes(...)`. On key collisions
with the existing extras channel, the event-only value wins on the event;
the span carries the existing extra-attributes value.

No new semconv keys are introduced, keys remain caller-defined, like the
existing extras channel. No external dependencies added or changed.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- **Targeted unit tests** (8 new test cases — 2 per wrapper, run for sync +
  async × streaming + non-streaming via the existing shared base classes
  in `nonstreaming_base.py` and `streaming_base.py`):
  - `test_event_only_extra_attributes_not_set_on_span` — verifies an
    attribute attached only via `GENERATE_CONTENT_EVENT_ONLY_EXTRA_ATTRIBUTES_CONTEXT_KEY`
    is absent from `generate_content {model}` span attributes.
  - `test_event_only_extra_attributes_set_on_event_only` — verifies the
    same attribute appears on the
    `gen_ai.client.inference.operation.details` event LogRecord, plus the
    collision-precedence rule (event-only value wins on the event; the
    existing-extras value is what reaches the span).

- **Test commands run:**

  ```sh
  # Targeted (event-only and existing extra-attributes tests, all 4 wrappers):
  uv run tox -e py312-test-instrumentation-google-genai-latest -- -k "event_only or extra_genai_attributes"
  # -> 24 passed, 24 skipped (intentional base-class skips), 0 failed

  # Full package suite (no regressions in pre-existing tests):
  uv run tox -e py312-test-instrumentation-google-genai-latest
  # -> 204 passed, 114 skipped, 0 failed

  # Static analysis:
  uv run tox -e typecheck
  # -> 0 errors, 0 warnings (pyright)

  # Lint + format:
  uvx ruff check  <touched files>
  uvx ruff format --check  <touched files>
  # -> All checks passed; 4 files already formatted
  ```

- [x] No.

# Checklist:

### DIFF
--- a/instrumentation-genai/opentelemetry-instrumentation-google-genai/CHANGELOG.md
+++ b/instrumentation-genai/opentelemetry-instrumentation-google-genai/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Add `GENERATE_CONTENT_EVENT_ONLY_EXTRA_ATTRIBUTES_CONTEXT_KEY` for attaching
+  caller-supplied attributes that are emitted only on the
+  `gen_ai.client.inference.operation.details` log event and never on the
+  `generate_content {model}` span. Use this for attributes (such as end-user
+  identifiers) that should not land on broadly-sampled spans. On key
+  collisions with `GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY`, the
+  event-only value wins on the event.
+  ([#XXXX](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/XXXX))
 -Add `gen_ai.usage.reasoning.output_tokens` attribute to capture thinking tokens on spans/events when the experimental sem conv flag is set. Add thinking tokens to output tokens. ([#4313](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4313))
 -Add `gen_ai.usage.cache_read.input_tokens` attribute to capture cached tokens on spans/events when the experimental sem conv flag is set. ([#4313](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4313))
 
--- a/instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/__init__.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-google-genai/src/opentelemetry/instrumentation/google_genai/__init__.py
@@ -30,12 +30,16 @@ API
 ---
 """
 
-from .generate_content import GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY
+from .generate_content import (
+    GENERATE_CONTENT_EVENT_ONLY_EXTRA_ATTRIBUTES_CONTEXT_KEY,
+    GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY,
+)
 from .instrumentor import GoogleGenAiSdkInstrumentor
 from .version import __version__
 
 __all__ = [
     "GoogleGenAiSdkInstrumentor",
     "GENERATE_CONTENT_EXTRA_ATTRIBUTES_CONTEXT_KEY",
+    "GENERATE_CONTENT_EVENT_ONLY_EXTRA_ATTRIBUTES_CONTEXT_KEY",
     "__version__",
 ]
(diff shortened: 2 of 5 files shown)


DylanRussell:
This looks good but should be opened in https://github.com/open-telemetry/opentelemetry-python-genai where we recently moved the instrumentations..

Also https://github.com/open-telemetry/opentelemetry-python-genai/pull/10 -- you might wait until this is merged but I can also merge my PR on top of yours so that doesn't matter too much

Achuth17:
Moved the PR to https://github.com/open-telemetry/opentelemetry-python-genai/pull/24

PR #4586, opened 2026-05-15
ci: validate changelog fragment filenames

MikeGoldsmith:
## Summary

Extends the existing fragment-existence check so each newly added changelog fragment is also validated by basename against `<PR_NUMBER>.<type>` (where `<type>` ∈ `added | changed | deprecated | removed | fixed`). Catches two failure modes the existence check doesn't:

- **Wrong PR number** in the filename (e.g. `4321.fixed` on PR #4322) — the rendered changelog entry links to the wrong PR.
- **Unknown type** suffix (e.g. `4322.fix`) — towncrier silently treats it as untyped on build.

Also adds a one-line note to the do-not-edit comment in each towncrier-managed `CHANGELOG.md` (root + the 4 independent packages: `opamp-client`, `resource-detector-azure`, `propagator-aws-xray`, `sdk-extension-aws`) flagging that the existing static `## Unreleased` entries pre-date towncrier and need folding into the first towncrier-generated release manually — `towncrier build` inserts new release blocks above that section rather than merging.

Same fixes also being applied to `opentelemetry-python` (open-telemetry/opentelemetry-python#5212) and originally caught in `opentelemetry-python-genai` (open-telemetry/opentelemetry-python-genai#16).

## Test plan

- [ ] CI runs the new validation step on this PR (which has the `Skip Changelog` label, so the workflow as a whole skips — verification of the validation itself is via a follow-up PR that intentionally adds a bad fragment, or by inspection of the workflow logic)
- [ ] Confirm the existence check still gates on at least one fragment being added


.github/workflows/changelog.yml
```diff
@@ -57,6 +59,24 @@ jobs:
             echo "Or add the \"Skip Changelog\" label if this job should be skipped."
             exit 1
           fi
+          invalid=()
+          while IFS= read -r f; do
+            base=$(basename "$f")
+            if [[ ! "$base" =~ ^([0-9]+)\.(added|changed|deprecated|removed|fixed)$ ]]; then
```
emdneto, line 65: Can you put a comment in the towncrier config saying that in case of changes in the nomenclature of changelog files, we also need to update the CI logic here?

emdneto (approved):

opamp/opentelemetry-opamp-client/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ Do *NOT* add changelog entries here!
 
 This changelog is managed by towncrier and is compiled at release time.
 
+The static "## Unreleased" section below pre-dates towncrier; its entries
```
xrmx: Do we still need these hunks since you have converted them already?

xrmx (approved):

.github/workflows/changelog.yml
xrmx: Done

PR #4605, opened 2026-05-19
fix(aws-lambda): support ALB multiValueHeaders

jbfenton:
# Description

Adds support for extracting propagated headers and HTTP request attributes from AWS Lambda ALB events that use `multiValueHeaders`.

This keeps the existing `headers` behaviour unchanged and falls back to the first value for each header in `multiValueHeaders` when needed.

Implements part of an older (closed) PR: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2585 

## Type of change

# How Has This Been Tested?

- [x] `uv run tox -e py312-test-instrumentation-aws-lambda -- instrumentation/opentelemetry-instrumentation-aws-lambda/tests/test_aws_lambda_instrumentation_manual.py -k "alb or multi_value"`
- [x] `uv run tox -e py312-test-instrumentation-aws-lambda`
- [x] `uv run pre-commit run ruff --files instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py instrumentation/opentelemetry-instrumentation-aws-lambda/tests/test_aws_lambda_instrumentation_manual.py`

- [x] No.

# Checklist:


tammy-baylis-swi (approved):
Makes sense to me since `headers` is default, multi-value is opt-in, and they should never co-exist.

xrmx (approved):
@herin049 any chance you can take a look please?

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -229,8 +249,9 @@ def _get_api_gateway_v1_proxy_attributes(
     attributes = {}
     attributes[HTTP_METHOD] = lambda_event.get("httpMethod")
 
-    if lambda_event.get("headers"):
-        headers = CIDict(lambda_event["headers"])
+    headers = _extract_http_headers(lambda_event)
+    if headers:
```
herin049:
```suggestion
    if headers := _extract_http_headers(lambda_event):
```

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -178,20 +178,40 @@ def _default_event_context_extractor(lambda_event: Any) -> Context:
     Returns:
         A Context with configuration found in the event.
     """
-    headers = None
-    try:
-        headers = lambda_event["headers"]
-    except (TypeError, KeyError):
+    headers = _extract_http_headers(lambda_event)
+    if not headers:
```
herin049: Just because `headers` is empty, does not mean that extraction necessarily failed. Personally, I think we should probably just remove the debug log below since it's difficult to determine whether no headers are present or if extraction failed.

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -178,20 +178,40 @@ def _default_event_context_extractor(lambda_event: Any) -> Context:
...
+        return {}
+
+    if not isinstance(multi_value_headers, dict):
+        return {}
+
+    normalized_headers = {}
+    for key, values in multi_value_headers.items():
+        if isinstance(values, list) and values:
+            normalized_headers[key] = values[0]
+    return normalized_headers
```
herin049: Can be simplified a bit.
```suggestion
    if not isinstance(lambda_event, dict):
        return {}

    headers = lambda_event.get("headers")
    if isinstance(headers, dict):
        return headers

    multi_value_headers = lambda_event.get("multiValueHeaders")
    if not isinstance(multi_value_headers, dict):
        return {}

    return {
        key: values[0]
        for key, values in multi_value_headers.items()
        if isinstance(values, list) and values
    }
```

instrumentation/opentelemetry-instrumentation-aws-lambda/tests/test_aws_lambda_instrumentation_manual.py
```diff
@@ -774,9 +775,48 @@ def test_alb_multi_value_header_event_sets_attributes(self):
...
                 FAAS_TRIGGER: "http",
                 HTTP_METHOD: "GET",
+                HTTP_SCHEME: "https",
+                HTTP_USER_AGENT: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6)",
+                NET_HOST_NAME: "lambda-846800462-us-east-2.elb.amazonaws.com",
             },
         )
 
+    def test_alb_multi_value_header_event_extracts_parent_context(self):
+        test_env_patch = mock.patch.dict(
```
herin049: Can we just use the `@patch.dict` decorator here?

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
xrmx: Right, or we can continue to catch the exception here and update _extract_http_headers to get the headers instead of the lambda event

jbfenton:
Thanks folks. I've addressed the review feedback in one follow-up update.

Changes made:
- removed the debug log in `_default_event_context_extractor` so empty or missing headers are no longer reported as extraction failure
- simplified `_extract_http_headers` and expanded it to handle mixed `headers` + `multiValueHeaders` events
- preserved per-key fallback to `multiValueHeaders` when `headers` is incomplete
- updated precedence for overlapping keys so `multiValueHeaders` wins when both carriers provide the same header, matching API Gateway’s documented merged behavior
- applied the small cleanup in `_get_api_gateway_v1_proxy_attributes`
- switched the new env-patching test to `@mock.patch.dict(...)`

Test coverage added/updated:
- no debug log for missing headers
- ALB events with only `multiValueHeaders`
- ALB events with both `headers` and `multiValueHeaders`
- conflicting values in both carriers, including `traceparent`
- updated the API Gateway proxy expectation to match the effective merged host value from `multiValueHeaders`

instrumentation/opentelemetry-instrumentation-aws-lambda/src/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -178,20 +178,34 @@ def _default_event_context_extractor(lambda_event: Any) -> Context:
...
+        CIDict(_extract_http_headers(lambda_event)),
     )
 
 
+def _extract_http_headers(lambda_event: Any) -> dict[str, Any]:
+    if not isinstance(lambda_event, dict):
+        return {}
+
+    headers = lambda_event.get("headers")
+    multi_value_headers = lambda_event.get("multiValueHeaders")
```
xrmx, line 189: So is it possible to have both set at the same time?

jbfenton: I believe so...

For API Gateway proxy events, AWS documents that both can be present in the same event.

AWS doc:
https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html

Relevant note:
> The headers key can only contain single-value headers.  
> The multiValueHeaders key can contain multi-value headers as well as single-value headers.  
> If you specify values for both headers and multiValueHeaders, API Gateway merges them into a single list.

The docs say API Gateway “merges them into a single list” as the effective request semantics, but the event payload still contains both carriers separately. 

Because our instrumentation reads the raw event object, it has to reproduce that merge same behaviour itself (or risk trace propagation / http attributes being wrong).

PR #4627, opened 2026-05-26
fix(kafka): avoid recording incorrect partition when using random partitioner

### LINKED ISSUE #4625: Incorrect partition in opentelemetry-instrumentation-kafka-python
### Describe your environment

OS: Tahoe 26.3
Python version: 3.13.9
Package version: 0.63b0.dev

### What happened?

Before Kafka producer sends data, the instrumentation creates a producer span. When `DefaultPartitioner` is used, key is not set, and a partition is not explicitly set in the producer, I found that opentelemetry-instrumentation-kafka-python records incorrect Kafka partition in the span.

The instrumentation extracts a partition using `instance._partition` and assigns it to the `messaging.kafka.partition` attribute. But this `_partition` method randomly chooses one partition from partition candidates. This partition is not the actual partition where the data is sent, but only used for span recording.

After that, `_partition` is called again inside the Kafka producer `send` method, and this time the selected partition is the actual partition where the data is sent.

Because of this, the partition recorded in the span can be different from the actual partition used by the producer.

### Steps to Reproduce

Reproducible example:
```python
# kafka_test.py

from kafka import KafkaProducer
from kafka.admin import KafkaAdminClient, NewTopic
from opentelemetry.instrumentation.kafka import KafkaInstrumentor
from opentelemetry.sdk.trace.export import SimpleSpanProcessor, SpanExporter, ConsoleSpanExporter


provider = TracerProvider()

TOPIC = "test-topic"
NUM_PARTITIONS = 12
admin = KafkaAdminClient(bootstrap_servers="localhost:9092")
try:
    admin.create_topics([NewTopic(TOPIC, num_partitions=NUM_PARTITIONS, replication_factor=1)])
except Exception:
    pass

def produce_hook(span, args, kwargs):
        value = kwargs.get("value") if "value" in kwargs else (args[1] if len(args) > 1 else None)
        if value is not None:
            span.set_attribute(
                "messaging.kafka.message.value",
                value.decode() if isinstance(value, (bytes, bytearray)) else str(value),
            )
        key = kwargs.get("key") if "key" in kwargs else (args[2] if len(args) > 2 else None)
        if key is not None:
            span.set_attribute(
                "messaging.kafka.message.key",
                key.decode() if isinstance(key, (bytes, bytearray)) else str(key),
            )


KafkaInstrumentor().instrument(produce_hook=produce_hook)
producer = KafkaProducer(bootstrap_servers="localhost:9092")

N = 10

for i in range(N):
    fut = producer.send(TOPIC, value=f"v{i}".encode())
    actual = fut.get(timeout=10).partition
```

### Expected Result

`messaging.kafka.partition` in the code result and `messaging.kafka.partition` in kafka-console-consumer result should be the same

### Actual Result

kafka_test.py result:
```
python kafka_test.py | jq -r '
  "messaging.kafka.message.value : \(.attributes."messaging.kafka.message.value")",
  "messaging.kafka.partition : \(.attributes."messaging.kafka.partition")",
  "------------------------------------"
'

messaging.kafka.message.value : v0
messaging.kafka.partition : null (should be 0)
------------------------------------
messaging.kafka.message.value : v1
messaging.kafka.partition : 6 
------------------------------------
messaging.kafka.message.value : v2
messaging.kafka.partition : 11
------------------------------------
messaging.kafka.message.value : v3
messaging.kafka.partition : 3 (should be 9)
------------------------------------
messaging.kafka.message.value : v4
messaging.kafka.partition : 7 (should be 10)
------------------------------------
messaging.kafka.message.value : v5
messaging.kafka.partition : 4
------------------------------------
messaging.kafka.message.value : v6
messaging.kafka.partition : 4 (should be 7)
------------------------------------
messaging.kafka.message.value : v7
messaging.kafka.partition : 10 (should be 9)
------------------------------------
messaging.kafka.message.value : v8
messaging.kafka.partition : 7 (should be 9)
------------------------------------
messaging.kafka.message.value : v9
messaging.kafka.partition : 9
------------------------------------
```

kafka-console-consumer result:
```
./kafka-console-consumer.sh --bootstrap-server localhost:9092 \
  --topic test-topic \
  --property print.partition=true

Partition:0	v0
Partition:6	v1
Partition:11	v2
Partition:9	v3
Partition:10	v4
Partition:4	v5
Partition:7	v6
Partition:9	v7
Partition:9	v8
Partition:9	v9
```

### Additional context

_No response_

None

### Tip

gauravSsinha:
## Description

Fixes incorrect `messaging.kafka.partition` attribute in producer spans when using the `DefaultPartitioner` without an explicit partition or key.

## Problem

When no explicit partition and no key are provided, the `DefaultPartitioner` randomly selects a partition. The instrumentation calls `instance._partition()` to record this value in the span, but the actual `send()` method calls `_partition()` again independently, potentially selecting a **different** random partition. This causes the span's `messaging.kafka.partition` attribute to not match the actual partition where the message was delivered.

## Fix

Only compute partition in the instrumentation when the result is deterministic:
- **Explicit partition provided** → use it directly
- **Key provided** → partition is determined by key hash (deterministic), safe to compute
- **Neither provided** → omit the `messaging.kafka.partition` attribute rather than record an incorrect value

This is more correct than recording a potentially wrong partition, which could mislead debugging and monitoring.

## Changes

- `KafkaPropertiesExtractor.extract_send_partition()` — refactored to only call `_partition()` when the result is deterministic
- `_enrich_span()` — handles `None` partition gracefully by omitting the attribute

Fixes #4625

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
+++ b/instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
@@ -57,14 +57,34 @@ class KafkaPropertiesExtractor:
 
     @staticmethod
     def extract_send_partition(instance, args, kwargs):
-        """extract partition `send` method arguments, using the `_partition` method in KafkaProducer class"""
+        """Extract partition from `send` method arguments.
+
+        Only returns a partition value when it can be determined deterministically:
+        - When partition is explicitly provided by the caller
+        - When a key is provided (partition is determined by key hash)
+
+        Returns None when partition would be randomly assigned by the
+        DefaultPartitioner, since calling _partition() here would produce a
+        different result than the actual send() call.
+        """
+        partition = KafkaPropertiesExtractor._extract_argument(
+            "partition", 4, None, args, kwargs
+        )
+        # If partition is explicitly set, return it directly
+        if partition is not None:
+            return partition
+
+        key = KafkaPropertiesExtractor.extract_send_key(args, kwargs)
+        # If no key is provided and no explicit partition, the DefaultPartitioner
+        # will randomly select a partition. We cannot predict which one, so
+        # return None to avoid recording an incorrect value.
+        if key is None:
+            return None
+
+        # Key is provided — partition is deterministic (hash-based), safe to compute
         try:
             topic = KafkaPropertiesExtractor.extract_send_topic(args, kwargs)
-            key = KafkaPropertiesExtractor.extract_send_key(args, kwargs)
             value = KafkaPropertiesExtractor.extract_send_value(args, kwargs)
-            partition = KafkaPropertiesExtractor._extract_argument(
-                "partition", 4, None, args, kwargs
-            )
             key_bytes = instance._serialize(
                 instance.config["key_serializer"], topic, key
             )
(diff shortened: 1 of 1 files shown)


tammy-baylis-swi:
Hi @gauravSsinha thanks for this, we'll take a look! Just for next time please keep the admin checklist included in the template (changelog, tests, etc)

PR #4628, opened 2026-05-26
feat(sqlcommenter): add comment_position option to control SQL comment placement

RiyaChaturvedi37:
Problem
The SQL comment was always appended to the end of the query, which caused truncation issues in database query logs for long queries (e.g. queries produced by Django in AWS RDS logs).
Solution
Added a comment_position parameter to _add_sql_comment() that defaults to "end" (preserving existing behavior per OTel spec). Set comment_position="start" to prepend the comment instead.
Usage
python_add_sql_comment(sql, comment_position="start", **meta)
Notes

Default behavior unchanged
Part of a series of PRs adding comment_position support across instrumentation libraries
Replaces #4490

Part of #3583


opentelemetry-instrumentation/src/opentelemetry/instrumentation/sqlcommenter_utils.py
```diff
@@ -5,17 +5,26 @@
 from opentelemetry.instrumentation.utils import _url_quote
 
 
-def _add_sql_comment(sql, **meta) -> str:
+def _add_sql_comment(sql, comment_position="end", **meta) -> str:
```
tammy-baylis-swi, line 8: Pardon my delayed review @RiyaChaturvedi37 and thank you for resuming this!

I had this comment on the other PR: Instead of a string, what if we had it as a case-insensitive boolean string check for false/true and the config was called `comment_prepend=false`? Since I don't think we'll support a "middle".

opentelemetry-instrumentation/src/opentelemetry/instrumentation/sqlcommenter_utils.py
tammy-baylis-swi, line 1: Please could test coverage be added in https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/tests/test_utils.py or a new file.

PR #4640, opened 2026-05-31
docs: document HTTP semantic convention opt-in

popsiclelmlm:
## Summary
Fixes #4202.

Documents `OTEL_SEMCONV_STABILITY_OPT_IN` for HTTP instrumentations by adding a shared docs include and referencing it from each HTTP-related instrumentation page.

## Reproduction
Before this change, the HTTP instrumentation docs did not mention how to opt in to the stable HTTP semantic conventions with `OTEL_SEMCONV_STABILITY_OPT_IN=http` or how to emit both old and stable conventions with `http/dup`.

## Root cause
The instrumentations support the semantic convention stability opt-in, but the readthedocs pages did not expose the configuration behavior.

## Changes
- Add shared HTTP semantic convention stability opt-in documentation.
- Include it from aiohttp client/server, ASGI, Django, Falcon, FastAPI, Flask, HTTPX, Pyramid, Requests, Starlette, Tornado, urllib, urllib3, and WSGI instrumentation docs.

## Tests
- [x] `git diff --check`
- [x] `/tmp/otel-rstcheck-venv/bin/rstcheck --config .rstcheck.cfg docs/_includes/http-semconv-stability.rst docs/instrumentation/aiohttp_client/aiohttp_client.rst docs/instrumentation/aiohttp_server/aiohttp_server.rst docs/instrumentation/asgi/asgi.rst docs/instrumentation/django/django.rst docs/instrumentation/falcon/falcon.rst docs/instrumentation/fastapi/fastapi.rst docs/instrumentation/flask/flask.rst docs/instrumentation/httpx/httpx.rst docs/instrumentation/pyramid/pyramid.rst docs/instrumentation/requests/requests.rst docs/instrumentation/starlette/starlette.rst docs/instrumentation/tornado/tornado.rst docs/instrumentation/urllib/urllib3.rst docs/instrumentation/urllib3/urllib3.rst docs/instrumentation/wsgi/wsgi.rst`

## Screenshots / Logs
Not applicable; documentation only.

PR #4644, opened 2026-06-01
docs: document OTEL_SEMCONV_STABILITY_OPT_IN across HTTP instrumentations

### LINKED ISSUE #4202: Document OTEL_SEMCONV_STABILITY_OPT_IN

mitanshubhoot:
## Summary

All eight HTTP instrumentations already support the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable but it is not mentioned in any of their user-facing documentation. This PR adds a **Semantic Convention Stability** subsection to the module docstring of each affected instrumentation (which is what gets rendered on readthedocs via `automodule`).

**Instrumentations updated:** aiohttp-client, ASGI, Django, FastAPI, Flask, HTTPX, Requests, WSGI

## What was added

Each instrumentation's `__init__.py` module docstring now includes:

```
Semantic Convention Stability
*****************************

This instrumentation supports the ``OTEL_SEMCONV_STABILITY_OPT_IN`` environment
variable, which controls which version of the HTTP semantic conventions is emitted:

* ``http`` — emit the new, stable HTTP and networking conventions only, and stop
  emitting the old experimental HTTP and networking conventions that the
  instrumentation emitted previously.
* ``http/dup`` — emit both the old and the stable HTTP and networking conventions,
  allowing for a seamless transition.
* Default (unset) — emit only the old experimental HTTP and networking conventions.
```

## Rationale for placing docs in module docstrings

The `docs/instrumentation/*/instrumentation.rst` files use `.. automodule::` to render documentation from the module's docstring. All other configuration environment variables (e.g. `OTEL_PYTHON_FLASK_EXCLUDED_URLS`, `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_*`) are already documented in the module docstrings, so this section follows the same pattern and appears naturally in the readthedocs output.

## Changes

- `instrumentation/opentelemetry-instrumentation-{aiohttp-client,asgi,django,fastapi,flask,httpx,requests,wsgi}/src/.../__ init__.py` — add "Semantic Convention Stability" subsection before the closing `API` section
- `.changelog/4202.added` — towncrier fragment

Fixes #4202

Made with [Cursor](https://cursor.com)

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-aiohttp-client/src/opentelemetry/instrumentation/aiohttp_client/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-aiohttp-client/src/opentelemetry/instrumentation/aiohttp_client/__init__.py
@@ -178,6 +178,23 @@ will replace the value of headers such as ``session-id`` and ``set-cookie`` with
 Note:
     The environment variable names used to capture HTTP headers are still experimental, and thus are subject to change.
 
+Semantic Convention Stability
+*****************************
+
+This instrumentation supports the ``OTEL_SEMCONV_STABILITY_OPT_IN`` environment
+variable, which controls which version of the HTTP semantic conventions is emitted:
+
+* ``http`` — emit the new, stable HTTP and networking conventions only, and stop
+  emitting the old experimental HTTP and networking conventions that the
+  instrumentation emitted previously.
+* ``http/dup`` — emit both the old and the stable HTTP and networking conventions,
+  allowing for a seamless transition.
+* Default (unset) — emit only the old experimental HTTP and networking conventions.
+
+For example, to opt in to the stable conventions::
+
+    export OTEL_SEMCONV_STABILITY_OPT_IN=http
+
 API
 ---
 """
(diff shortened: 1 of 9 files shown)


instrumentation/opentelemetry-instrumentation-aiohttp-client/src/opentelemetry/instrumentation/aiohttp_client/__init__.py
```diff
@@ -178,6 +178,23 @@ def response_hook(span: Span, params: typing.Union[
 Note:
     The environment variable names used to capture HTTP headers are still experimental, and thus are subject to change.
 
+Semantic Convention Stability
```
rads-1996, line 181: This header is a little misleading, clearly not setting this env var means emitting old experimental conventions. Maybe we should preface it by saying this how to default works and to get the latest/stable conventions this is how you can set the variable.

lzchen:
@mitanshubhoot 

I think this is a duplicate of https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4640 so will be closing this.

PR #4648, opened 2026-06-02
fix(instrumentation-confluent-kafka): add __getattr__ delegation to ProxiedProducer and ProxiedConsumer

stark256-spec:
## Description

Fixes #4278

`ProxiedProducer` and `ProxiedConsumer` only forward an explicit set of methods to the underlying confluent-kafka objects. Calling any method outside that list — such as `list_topics()`, `assignment()`, `memberid()`, or `set_sasl_credentials()` — raises `AttributeError` on pure-Python mocks or causes a **segmentation fault** on real confluent-kafka C-extension objects, because the C-extension base classes don't implement standard Python attribute lookup.

Add `__getattr__` to both proxy classes so that any attribute or method lookup not satisfied by an explicitly defined proxy method is forwarded to the underlying `_producer` / `_consumer` via `getattr()`. Methods already explicitly proxied are unaffected.

## Type of change

## How Has This Been Tested?

Two new unit tests added to `tests/test_instrumentation.py`:

- `test_proxied_producer_delegates_unproxied_methods` — verifies `ProxiedProducer` forwards a method not in its explicit proxy list via `__getattr__`
- `test_proxied_consumer_delegates_unproxied_methods` — same for `ProxiedConsumer`

- [x] No.


xrmx (changes requested):
I think the proper fix would be to start using wrapt like the other instrumentations are doing, see https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4270

stark256-spec:
Makes sense — I'll switch this to wrapt.ObjectProxy instead of __getattr__.

stark256-spec:
Switched to wrapt.ObjectProxy — thanks for the pointer, it's cleaner than the __getattr__ approach.

Since ObjectProxy forwards anything not explicitly overridden, I was able to drop the whole hand-maintained pass-through list (flush, poll, purge, committed, commit, get_watermark_offsets, offsets_for_times, subscribe) — the proxies now only define the instrumented methods. Net -75/+72.

Two things worth flagging though ( assuming while youre doing diff might be useful):

utils.py reads and writes _current_consume_span / _current_context_token on the instance, so those are properties backed by ObjectProxy's _self_ storage. Otherwise they'd get forwarded onto the wrapped consumer, which has no __dict__.
I had to change assertEqual(x.__class__, ProxiedProducer) to assertIsInstance in the existing tests, since ObjectProxy intentionally reports the wrapped object's __class__. The uninstrument assertions are unchanged and still pass.
Also rebased onto main and added a .changelog/4648.fixed fragment. 
Let me know if you see a concern,, will try to figure out a another approach...
Tests pass, ruff and pylint clean.

instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/__init__.py
```diff
@@ -168,84 +168,78 @@ def close(self):  # pylint: disable=useless-super-delegation
         return super().close()
 
 
-class ProxiedProducer(Producer):
+class ProxiedProducer(wrapt.ObjectProxy):
```
xrmx, line 171: Please use BaseObjectProxy if available becase ObjectProxy always implementes `__iter__` even if the proxied object doesn't, we do this in other instrumentations:

```
try:
    # wrapt 2.0.0+
    from wrapt import BaseObjectProxy
except ImportError:
    from wrapt import ObjectProxy as BaseObjectProxy
```

instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/__init__.py
```diff
@@ -168,84 +168,78 @@ def close(self):  # pylint: disable=useless-super-delegation
...
+    def _current_consume_span(self, value) -> None:
+        self._self_current_consume_span = value
+
+    @property
+    def _current_context_token(self):
+        return self._self_current_context_token
+
+    @_current_context_token.setter
+    def _current_context_token(self, value) -> None:
+        self._self_current_context_token = value
```
xrmx, line 220: I don't see much value on these wrappers, I think we can live with the caller referencing the `._self_` version directly

instrumentation/opentelemetry-instrumentation-confluent-kafka/tests/test_instrumentation.py
```diff
@@ -116,7 +120,9 @@ def test_consumer_commit_method_exists(self) -> None:
         )
 
         consumer = instrumentation.instrument_consumer(consumer)
-        self.assertEqual(consumer.__class__, ProxiedConsumer)
+        # See test_instrument_api: ObjectProxy reports the wrapped __class__.
+        self.assertIsInstance(consumer, ProxiedConsumer)
+        # commit() is no longer declared on the proxy; ObjectProxy forwards it.
```
xrmx, line 125:
```suggestion
```

Looks more like a changelog

xrmx (commented):
Thanks, looks better!

stark256-spec:
Merged latest `main` — the conflicts are resolved and the PR is mergeable again.

The only conflict was in the confluent-kafka proxy classes: `main` recently added two pass-through methods to `ProxiedConsumer` (`get_watermark_offsets`, `offsets_for_times`). Since this PR switches the proxies to `wrapt.ObjectProxy` (per the earlier review — dropping `__getattr__` and the manual pass-throughs), those methods are now forwarded transparently, so I kept the `ObjectProxy` implementation and let it delegate them rather than re-adding the manual overrides.

Verified locally: all 18 package tests pass — including `test_proxied_consumer_delegates_unproxied_methods` / `test_proxied_producer_delegates_unproxied_methods`, which cover exactly that delegation — and `ruff check` is clean.

@xrmx this should be ready for another look whenever you have a chance — thanks!

xrmx:
> @xrmx this should be ready for another look whenever you have a chance — thanks!

There are still 3 open comments from before your last update.

PR #4651, opened 2026-06-02
fix(instrumentation-logging): prevent _uninstrument from corrupting log-record-factory chain

### LINKED ISSUE #3808: opentelemetry-instrumentation-logging: LoggingInstrumentor._uninstrument CORRUPTS the log factory linked list
### Describe your environment

OS: Ubuntu, OS X, all OSs
Python version: 3.11.13
Package version: 0.58b0

### What happened?
Concerning `opentelemetry.instrumentation.logging.LoggingInstrumentor` ...

`LoggingInstrumentor._uninstrument()` corrupts the log factory linked list. The python`logging.logging.setLogRecordFactory()` and `logging.getLogRecordFactory()` are intended to create a linked list, chaining multiple log record factories to augment information in the log records.

The record factory mechanism is also commonly used by applications.

The python logging record factory mechanism does not support removing (i.e. uninstrumenting) of any link in the chain. Once added, it MUST NOT BE REMOVED, or bad things happen.

`LoggingInstrumentor._uninstrument()` is implemented as if it's the only custom log record factory in the entire application.  When it executes, it cuts off any of the custom log factories that were added after it.
As you can see from the current implementation below, it sets  `_old_factory` node as the current log record factory, effectively cutting off any custom application or other libraries' factories from the linked list.
```
```

Example - Before
```
OTHERS--->OTEL--->ORIGINAL
```
Example - After - corrupted linked list where the OTHERS links got cut off by `LoggingInstrumentor._uninstrument()`
```
OTHERS    OTEL--->ORIGINAL
```

### Steps to Reproduce

See description in "What Happened?" section.

### Expected Result

`LoggingInstrumentor._uninstrument()` MUST NOT cut off other custom log factories that were added after `LoggingInstrumentor._instrument()`

### Actual Result

`LoggingInstrumentor._uninstrument()` cuts off other custom log factories that were added after `LoggingInstrumentor._instrument()`.

Additionally, `LoggingInstrumentor._uninstrument()` is called by `LoggingInstrumentor.instrument()`, where it also cuts off other links in the linked list of custom log factories.

### Additional context

If uninstrumenting is needed for some reason, then a safe solution would be to implement this logging augmentation as a python logging filter instead of a log record factory and allow applications to add/remove the logging filter using python logging API. For example:

In opentelementry.instrumentation.logging:
```
class LoggingInstrumentorFilter(logging.Filter):
        def filter(self, record):
            record.otelSpanID = "0"
            record.otelTraceID = "0"
            record.otelTraceSampled = False

            ...

            return True
```

An application can then add and remove the filter to a specific logger or handler. Typical applications would likely add it to the root logger, but that may not be the case for all applications. An application needs to be in control of this. Below example adds and removes the filter to/from the root logger, thereby causing the filter to be executed for ALL loggers. The application has control to which logger or handler to instrument with the filter.
```
        otel_logging_instrumentor_filter = LoggingInstrumentorFilter()
        logging.root.addFilter(otel_logging_instrumentor_filter)

        ...

       # Later, if the application needs to remove this filter for some reason:
       logging.root.removeFilter(otel_logging_instrumentor_filter)
``` 

For example:
```
import logging
from typing import Optional

class LoggingInstrumentorFilter(logging.Filter):
    def __init__(self, *args, **kwargs) -> None:

        # We track the last updated record to avoid redundant updates.
        #
        # NOTE: A filter may be invoked multiple times for the same record in some
        # logging configurations, such as when an application adds the filter
        # to multiple log handlers.
        self._last_updated_record: Optional[logging.LogRecord] = None

    def filter(self, record: logging.LogRecord) -> bool:
        if record is self._last_updated_record:
            # Already updated this record, skip
            return True

        self._last_updated_record = record

        record.otelSpanID = "0"
        record.otelTraceID = "0"
        record.otelTraceSampled = False

        span = get_current_span()

        if span != INVALID_SPAN:
            ctx = span.get_span_context()
            if ctx != INVALID_SPAN_CONTEXT:

        return True
```

In addition, it would be very helpful to add the otelParentSpanID into the log record, using logic like this:
```
from opentelemetry.sdk import trace as sdk_trace
from opentelemetry.trace import INVALID_SPAN, get_current_span

record.otelParentSpanID = "0"

span: sdk_trace.Span = get_current_span()
if span != INVALID_SPAN:
    ctx = span.get_span_context()
    if ctx != INVALID_SPAN_CONTEXT:
         ...

         # NOTE: We can get a NonRecordingSpan that has no parent attribute when
         #  sampling is disabled.
         if (parent_ctx := getattr(span, 'parent', None) ) is not None:
             record.otelParentSpanID = format(parent_ctx.span_id, "016x") 
```

None

### Tip

stark256-spec:
## Description

Fixes #3808

`LoggingInstrumentor._uninstrument()` unconditionally restored `_old_factory` (the factory active before instrumentation), silently dropping any log-record factories installed by other libraries **after** OTel's factory was added.

Example:
```
Before uninstrument:  APP_FACTORY → OTEL_FACTORY → ORIGINAL
After old behaviour:  (APP_FACTORY orphaned)   ORIGINAL
```

**Fix:** store a reference to our own factory closure (`_our_factory`). In `_uninstrument`, compare the current factory against `_our_factory` by identity. Only restore `_old_factory` when we are still at the head of the chain. If another factory was installed on top of ours, skip the restore — our factory remains in the chain, but no data is lost.

## Type of change

## How Has This Been Tested?

Logic reviewed against the described failure scenario. Existing test suite passes.

- [x] No.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/__init__.py
@@ -135,6 +135,7 @@ class LoggingInstrumentor(BaseInstrumentor):  # pylint: disable=empty-docstring
     """
 
     _old_factory = None
+    _our_factory = None
     _log_hook = None
     _logging_handler = None
 
@@ -214,6 +215,7 @@ class LoggingInstrumentor(BaseInstrumentor):  # pylint: disable=empty-docstring
 
             return record
 
+        LoggingInstrumentor._our_factory = record_factory
         logging.setLogRecordFactory(record_factory)
 
         # Here we need to handle 3 scenarios:
@@ -262,9 +264,22 @@ class LoggingInstrumentor(BaseInstrumentor):  # pylint: disable=empty-docstring
             LoggingInstrumentor._logging_handler = handler
 
     def _uninstrument(self, **kwargs):
-        if LoggingInstrumentor._old_factory:
-            logging.setLogRecordFactory(LoggingInstrumentor._old_factory)
-            LoggingInstrumentor._old_factory = None
+        if LoggingInstrumentor._our_factory is not None:
+            current = logging.getLogRecordFactory()
+            if current is LoggingInstrumentor._our_factory:
+                # We are still at the head of the factory chain — safe to
+                # restore the previous factory without dropping any other
+                # custom factories that may have been registered before us.
+                logging.setLogRecordFactory(LoggingInstrumentor._old_factory)
+            # If current is NOT our factory another library installed its own
+            # factory on top of ours after we instrumented.  Blindly restoring
+            # _old_factory would cut that library's factory out of the chain
+            # (the bug described in #3808).  In that case we leave the chain
+            # intact and accept that our factory remains in it; removing a
+            # node from the middle of the linked list is not safely possible
+            # without cooperation from the other factories.
+        LoggingInstrumentor._old_factory = None
+        LoggingInstrumentor._our_factory = None
 
         if LoggingInstrumentor._logging_handler:
             logging.getLogger().removeHandler(


xrmx (commented):
Could you please add a test case for this? And a changelog entry.

PR #4653, opened 2026-06-03
fix(instrumentation-botocore): do not mark HTTP 3xx responses as span errors

### LINKED ISSUE #4633: Botocore instrumentation incorrectly marks S3 304 Not Modified as span error
### Describe your environment

OS: Ubuntu
Python version: Python `3.11.10`
Package version: `opentelemetry-instrumentation-botocore==0.62b1`

### What happened?

When using conditional S3 requests (eg using `IfNoneMatch` or `IfModifiedSince` parameters), a `304 Not Modified` response is returned by botocore as a `ClientError` exception. The botocore instrumentor treats all `ClientError` exceptions as errors and sets `StatusCode.ERROR` on the span, even though a 304 is a normal expected
response and not a true error.

This causes `S3.GetObject` spans (and potentially other operations that return 3xx responses) to be spuriously marked as errors in tracing backends, polluting error rates and alerting.

### Steps to Reproduce

```py
import boto3

s3 = boto3.client("s3")
etag = s3.head_object(Bucket="my-bucket", Key="my-key")["ETag"]

# This raises ClientError with a 304 status - the span is incorrectly marked as error.
s3.get_object(Bucket="my-bucket", Key="my-key", IfNoneMatch=etag)
```

### Expected Result

Spans for responses with HTTP 3xx status codes should be marked `OK`, not `ERROR`.

### Actual Result

Any `ClientError`, regardless of HTTP status code, results in `StatusCode.ERROR` on the span.

### Additional context

_No response_

Yes

### Tip

stark256-spec:
## Description

Fixes #4633

### What changed and why

botocore raises `ClientError` for HTTP 3xx responses such as `304 Not Modified` (returned when using conditional S3 requests with `IfNoneMatch` or `IfModifiedSince`). The instrumentor was calling `extension.on_error` for every `ClientError` regardless of HTTP status, and the span context manager's default `set_status_on_exception` behaviour then set `StatusCode.ERROR` as the exception propagated out.

`304 Not Modified` and other 3xx responses are normal, expected outcomes — not errors. Marking them as errors pollutes error-rate dashboards and triggers false alerts.

**Fix:** check `HTTPStatusCode` from the error response metadata before invoking `on_error`. For HTTP `< 400`, explicitly set `StatusCode.OK` on the span (which the OTel spec guarantees cannot be overridden to `ERROR` by the context manager) and skip `on_error`. For 4xx/5xx the existing behaviour is unchanged.

Applied to both `BotocoreInstrumentor` and `AiobotocoreInstrumentor`.

### Testing

- [ ] Existing botocore unit tests pass
- [ ] New test covering 304 response not marked as error (to be added — happy to add if maintainers confirm the approach)


xrmx (changes requested):
Please drop the anthropic changes and add a test case for the change.

stark256-spec:
Thanks for the review! The Anthropic changes were an accidental inclusion — I was actively contributing to a separate project at the same time and the commits got mixed up. Good catch. I've dropped those and added test cases covering the 304 behaviour for both the sync (`BotocoreInstrumentor`) and async (`AiobotocoreInstrumentor`) paths.

PR #4656, opened 2026-06-03
fix(asgi): exclude background task duration from HTTP server metrics

### LINKED ISSUE #3591: Opentelemetrymiddleware: Background task duration is recorded in metrics
### Describe your environment

OS: Ubuntu
Python version: 3.12
Package version: [v0.55b1]

### What happened?

I use otel metrics. We have routes which use background tasks and when looking at the http_server_duration_milliseconds_bucket metrics we see that the time of the background task is included.
I am just interested in user facing time and not in the background task stuff.

### Steps to Reproduce

Create a fastapi route

```
from fastapi import APIRouter, Request,BackgroundTasks
import asyncio
@router.get("/healthz-with-background-task")
async def healthz_with_background_task(background_tasks: BackgroundTasks) -> dict[str, str]:
    background_tasks.add_task(sleep_background_task)
    return {"status": "ok", "message": "Background task started"}
async def sleep_background_task() -> None:
    """Background task that sleeps for 10 seconds asynchronously."""
    await asyncio.sleep(10)
    # You could add logging here to confirm the task completed
    # logger.info("Background sleep task completed")

```

Then  at this line add a print:
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/c4347e027c745ab7e1b9468ad75c37bed535ca26/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L767

print(f"duration_s: {duration_attrs_new} {duration_s}")

interestingly you get two prints:
The first one with the real duration but without attributes! duration_s: {} 0.0012163339997641742
The second one with attributes but with the duration of the background task included: duration_s: {'http.route': '/healthz-with-background-task'} 10.00230166700203

### Expected Result

I should get the user facing duration and not background task including. 

### Actual Result

I get the duration with background task including and get one empty!

### Additional context

This is a severe bug which is a blocker from using the metrics implementation. with this we will need to stick with prometheus client sdk.

None

grvmishra788:
# Description

Duration metrics include background task execution time because the ASGI middleware measures duration in the `finally` block, which blocks until background tasks complete.

This PR captures a timestamp when the final response body is sent and uses that for duration calculation. If no response is sent, it falls back to `default_timer()`.

Fixes #3591

## Type of change

# How Has This Been Tested?

- Ran ASGI instrumentation tests
- Added tests to verify background task time is excluded

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
```diff
@@ -702,6 +702,78 @@ def add_body_and_trailer_span(expected: list):
...
+        self.seed_app(app)
+        await self.send_default_request()
+        await self.get_all_output()
+
+        metrics = self.get_sorted_metrics(SCOPE)
+        duration_found = False
+        for metric in metrics:
+            if metric.name == "http.server.duration":
+                data_points = list(metric.data.data_points)
+                for point in data_points:
```
herin049: This logic of finding the duration data point appears to be duplicated in each of the three tests cases, could we create a small helper function to clean up these tests a bit?

herin049 (approved):

xrmx:
@grvmishra788 requesting a review again from a person that already approved will drop their approval which I don't think is beneficial to your cause. Especially so if the comments where about code duplication.

instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -816,7 +818,11 @@ async def __call__(
                         query,
                         self._sem_conv_opt_in_mode,
                     )
-                duration_s = default_timer() - start
+                duration_s = (
+                    response_complete_time[0] - start
```
xrmx, line 848: Maybe a silly question: why do we keep a list if we care only about the first element?

PR #4698, opened 2026-06-14
docs: document OTEL_SEMCONV_STABILITY_OPT_IN for HTTP instrumentation migration

### LINKED ISSUE #4202: Document OTEL_SEMCONV_STABILITY_OPT_IN

rajdhruvsingh:
## Description
Adds a section to `instrumentation/README.md` documenting the
`OTEL_SEMCONV_STABILITY_OPT_IN` environment variable supported
by HTTP instrumentations with `migration` semconv status.

This follows the approach suggested by @lzchen in PR #4640 —
a single excerpt in the instrumentation README rather than
individual per-instrumentation documentation.

## Changes
- `instrumentation/README.md`: added "HTTP Semantic Convention
  Migration" section above the instrumentation table explaining
  the three modes of OTEL_SEMCONV_STABILITY_OPT_IN with a link
  to the migration guide

## Related
Fixes #4202

### DIFF
--- a/.changelog/4698.fixed
+++ b/.changelog/4698.fixed
@@ -1,0 +1,1 @@
+`instrumentation`: document `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable for HTTP instrumentation migration
--- a/instrumentation/README.md
+++ b/instrumentation/README.md
@@ -1,3 +1,23 @@
+## HTTP Semantic Convention Migration
+
+Several HTTP instrumentations (those with `migration` in the Semconv status
+column) support the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable to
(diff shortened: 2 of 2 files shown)


herin049:
We already #4640 which addresses this

PR #4700, opened 2026-06-15
fastapi: fix AttributeError on partially matched routes with FastAPI 0.137 included routers

### LINKED ISSUE #4699: FastAPI v0.137.0 breaks opentelemetry-instrumentation-fastapi
### Describe your environment

OS: Windows 11
Python version: Python 3.13.7
Package version: opentelemetry-instrumentation-fastapi   0.61b0

### What happened?

A breaking change to FastAPI in v0.137.0 has affected the fastapi instrumentor. This is the relevant part of the FastAPI [release notes](https://github.com/fastapi/fastapi/releases/tag/0.137.0):

> Specific Breaking Changes
>
> Now router.routes is no longer a plain list of APIRoute objects, it can contain these intermediate objects that can contain additional routers, forming a tree.
>
> Any logic that depended on iterating on the router.routes directly would be affected, that logic cannot expect to be able to extract data from a plain list of routes, as it's no longer a plain list but a tree.
>
> Additionally, any logic that iterated on router.routes to modify them would now also see these new objects, and would not see all the routes in the app.
>
> router.routes should be considered an internal implementation detail, only passed around to the FastAPI functions that need it.

This causes an exception in `_get_route_details()` on line 490:

    File "***\site-packages\opentelemetry\instrumentation\fastapi\__init__.py", line 490, in _get_route_details
      route = starlette_route.path
              ^^^^^^^^^^^^^^^^^^^^
    AttributeError: '_IncludedRouter' object has no attribute 'path'

### Steps to Reproduce

```python
import uvicorn
from azure.monitor.opentelemetry import configure_azure_monitor
from fastapi import FastAPI, APIRouter
from fastapi.middleware.cors import CORSMiddleware

APPLICATION_INSIGHTS_CONNECTION_STRING = "..."

APP = FastAPI()
configure_azure_monitor(connection_string=APPLICATION_INSIGHTS_CONNECTION_STRING)
FastAPIInstrumentor.instrument_app(APP)

APP.add_middleware(
    CORSMiddleware, allow_origins=["*"], allow_methods=["*"], allow_headers=["*"]
)

FOO_ROUTER = APIRouter(prefix="/foo")
@FOO_ROUTER.get("")
async def foo():
    return "foo"

APP.include_router(FOO_ROUTER)
uvicorn.run(APP, host="0.0.0.0", port=8080)
```

```
import requests
requests.options("http://localhost:8080/foo")
```

### Expected Result

Response to OPTIONS

### Actual Result

Internal server error caused by AttributeError: '_IncludedRouter' object has no attribute 'path'

### Additional context

At present the maintainer of FastAPI has not to my knowledge proposed a method for external code to iterate through all routes of a FastAPI application.

None

### Tip

kshitizj03:
# Description

FastAPI 0.137.0 changed `app.routes` to include intermediate `_IncludedRouter`
objects (for routers added via `include_router`) that don't expose a `.path`
attribute. In `_get_route_details()`, the `Match.FULL` branch already guards
`starlette_route.path` with `try/except AttributeError`, but the `Match.PARTIAL`
branch does not — so a request that only partially matches an included route
(e.g. an `OPTIONS` request whose path matches but whose method doesn't) raises
`AttributeError: '_IncludedRouter' object has no attribute 'path'` (HTTP 500).
This mirrors the existing `FULL`-branch fallback onto the `PARTIAL` branch.

Recovering the templated route through `_IncludedRouter` needs FastAPI private
API and is left as a follow-up.

Fixes #4699

## Type of change

# How Has This Been Tested?

- Added regression test `test_included_router_partial_match` (returns 500 before
  the fix, passes after).
- Full fastapi suite green on CI-pinned `fastapi==0.126.0`; reproduced and
  validated against `fastapi==0.137.0`.

- [x] No.

# Checklist:


xrmx (changes requested):
I'm not sure this will test the new fastapi if we don't bump the version in test requirements

bhanuwealthy:
Patch LGTM.

kshitizj03:
You're right that with the pin at fastapi==0.126.0 the new test doesn't really exercise the fix, since _IncludedRouter only exists in 0.137+. I tried bumping the test requirement to fastapi==0.137.0 locally. The fix holds there (the AttributeError is gone and the regression test passes), but the bump also makes ~20 existing tests fail, and they all come from a single fixture, CustomRoute.matches(), which does assert "nonstandard_field" in scope. FastAPI 0.137's new _IncludedRouter._match re-invokes matches() with a re-derived scope that no longer carries that field, so the assertion trips across every test using the fixture. That looks like a pre-existing test-suite incompatibility with 0.137, separate from this crash fix.
Would you prefer I keep this PR focused on the crash fix and open a separate issue/PR for the FastAPI 0.137 test-suite migration (fixture update + version bump), or should I do the bump and fixture migration in this PR?

xrmx:
> You're right that with the pin at fastapi==0.126.0 the new test doesn't really exercise the fix, since _IncludedRouter only exists in 0.137+. I tried bumping the test requirement to fastapi==0.137.0 locally. The fix holds there (the AttributeError is gone and the regression test passes), but the bump also makes ~20 existing tests fail, and they all come from a single fixture, CustomRoute.matches(), which does assert "nonstandard_field" in scope. FastAPI 0.137's new _IncludedRouter._match re-invokes matches() with a re-derived scope that no longer carries that field, so the assertion trips across every test using the fixture. That looks like a pre-existing test-suite incompatibility with 0.137, separate from this crash fix. Would you prefer I keep this PR focused on the crash fix and open a separate issue/PR for the FastAPI 0.137 test-suite migration (fixture update + version bump), or should I do the bump and fixture migration in this PR?

I would say to fix all the 0.137 issues here so that the CI would be green, just do the changes in different commits please to make the review easier. Thanks.

kshitizj03:
I've pushed the full 0.137 fix, split into 3 commits as requested: fix, test, and version bump. The fix detects _IncludedRouter and uses effective_route_contexts() to match the flattened routes, so it recovers the real templated route instead of just avoiding the crash. Matching the flattened contexts (instead of the _IncludedRouter itself) is also what stops the ~20 CustomRoute failures, since the scope is no longer re-derived, so CI is green on 0.137 without touching that fixture. I've also bumped test-requirements.txt to fastapi 0.137.1 and left pyproject.toml at ~= 0.92, since that already allows 0.137. Tested locally: 105 passed on 0.137.1.

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -479,6 +479,20 @@ def _get_route_details(scope):
...
     for starlette_route in app.routes:
+        # FastAPI >= 0.137 nests routes added via include_router() under
+        # _IncludedRouter tree nodes, which expose no ``path`` attribute.
+        # Flatten them into their effective route contexts, each of which
+        # provides matches() and the full templated path. Matching these
+        # directly avoids _IncludedRouter.matches() re-deriving the scope.
+        if hasattr(starlette_route, "effective_route_contexts"):
+            for route_context in starlette_route.effective_route_contexts():
+                match, _ = route_context.matches(scope)
+                if match == Match.FULL:
```
herin049: Can we refactor this so we aren't duplicating the same code on lines 501-509

kshitizj03: Good catch, refactored into a `_flatten_routes` helper so the match logic isn't duplicated; pushed.

Giuzzilla:
FYI FastAPI just released [`0.137.2`](https://github.com/fastapi/fastapi/releases/tag/0.137.2) which introduces `iter_route_contexts()` as a replacement for the previous `router.routes`. Might be relevant/cleaner for the implementation.

[Related PR](https://github.com/fastapi/fastapi/pull/15785)

kshitizj03:
> FYI FastAPI just released [`0.137.2`](https://github.com/fastapi/fastapi/releases/tag/0.137.2) which introduces `iter_route_contexts()` as a replacement for the previous `router.routes`. Might be relevant/cleaner for the implementation.
> 
> [Related PR](https://github.com/fastapi/fastapi/pull/15785)

@Giuzzilla thanks for the heads up. I've adopted `iter_route_contexts()`; kept a fallback to the private API for `0.137.0/0.137.1` and plain routes for older FastAPI; bumped the test pin to 0.137.2; verified the suite passes on 0.137.2, 0.137.1, and 0.126.0.

instrumentation/opentelemetry-instrumentation-fastapi/test-requirements.txt
```diff
@@ -1,11 +1,11 @@
 annotated-types==0.7.0
-anyio==4.12.1
+anyio==4.14.0
 asgiref==3.8.1
 certifi==2024.7.4
 charset-normalizer==3.3.2
 Deprecated==1.2.14
 exceptiongroup==1.2.0
-fastapi==0.126.0
+fastapi==0.137.2
```
xrmx: How do others feel about testing only latest?

xrmx: I've pushed a commit adding support for testing multiple version, the oldest version is not really the oldest but it's older than the previous one :sweat_smile:

xrmx (approved):

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
```diff
@@ -462,6 +470,30 @@ def __init__(self, *args: Any, **kwargs: Any):
...
+    FastAPI >= 0.137.2 provides the public iter_route_contexts() for this, which
+    also wraps plain routes uniformly. On older versions fall back to the
+    private _IncludedRouter.effective_route_contexts(), and on FastAPI < 0.137
+    (no _IncludedRouter) the routes are already matchable as-is.
+    """
+    if iter_route_contexts is not None:
+        yield from iter_route_contexts(routes)
+        return
+    for starlette_route in routes:
+        if hasattr(starlette_route, "effective_route_contexts"):
```
herin049, line 491: Where does `effective_route_contexts` come from?

herin049 (approved):

instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
kshitizj03: It is the private `_IncludedRouter.effective_route_contexts()` from FastAPI 0.137.0/0.137.1 which is detected via `hasattr` so we don't import a private class. The public `iter_route_contexts()` is used on 0.137.2+.

emdneto (approved):

bhanuwealthy:
Hi @xrmx, when is this fix going to be available on PIP ?  Eagerly waiting to released.

PR #4707, opened 2026-06-16
fix(fastapi): handle _IncludedRouter in _get_route_details for FastAPI >= 0.137.0

### LINKED ISSUE #4699: FastAPI v0.137.0 breaks opentelemetry-instrumentation-fastapi

rajdhruvsingh:
## Description

Fixes #4699

FastAPI 0.137.0 introduced a breaking change where `router.routes` is no longer a flat list of route objects but a tree structure that can contain `_IncludedRouter` nodes. These nodes have no `.path` attribute, causing an `AttributeError` crash in `_get_route_details()` whenever a request is made to a route registered via `app.include_router()`.

### Root Cause

In `_get_route_details()`, the existing code iterates `app.routes` and accesses `.path` directly on each entry:

```python
    ...
    if match == Match.FULL:
        route = starlette_route.path  # 💥 AttributeError on _IncludedRouter
```

`_IncludedRouter` objects have no `.path` — they are tree nodes that wrap an `APIRouter` and expose an `effective_route_contexts()` method which recursively flattens the tree into matchable route contexts.

### Fix

When iterating `app.routes`, detect `_IncludedRouter` objects via `hasattr(starlette_route, "effective_route_contexts")` (duck typing avoids importing the private class) and use `effective_route_contexts()` to iterate the flattened route contexts, each of which has both `.matches()` and `.path`.

Also updated the `fastapi` version pin in `pyproject.toml` from `~= 0.92` to `>= 0.92` to allow FastAPI 0.137+.

### Testing

Verified manually with a FastAPI app using `app.include_router()` with FastAPI 0.137.1:
- ✅ `http.route` is correctly captured for routes behind `_IncludedRouter`
- ✅ Nested routers (router within router) work correctly
- ✅ Direct routes on `app` (no router) still work
- ✅ 404 routes still work
- ✅ 99 existing tests pass (2 pre-existing failures unrelated to this change)


xrmx:
Closing as dup of #4700. @rajdhruvsingh I've seen your comment in the issue but we have already another PR open for this sorry! I would appreciate if you can instead review the other one instead, thanks!

PR #4723, opened 2026-06-20
opentelemetry-instrumentation-django: fix spans ending prematurely for StreamingHttpResponse

### LINKED ISSUE #4681: Django Instrumentation: Spans close prematurely for StreamingHttpResponse

The current Django instrumentation uses a Middleware to open and close spans. While this works perfectly for standard HTTP responses, it falls short for StreamingHttpResponse.

Because streaming responses continue to process and send data after the middleware chain has finished executing, the request lifecycle is not properly traced, leading to incomplete or inaccurate span durations.

A way to solve this is to hook into the response's `close()` method, which Django guarantees will be called after the stream finishes. We can wrap the original close method to ensure the span is ended at the very end of the response lifecycle.

This approach would gracefully handle both regular and StreamingHttpResponse objects.

Here is a conceptual implementation:

```python
original_close = response.close

def close_span():
    try:
        original_close()
    finally:
        # TODO: End the span here

response.close = close_span
```

Note: I haven't fully tested this with async responses under ASGI yet, where `response.close` might be executed via `sync_to_async`.

Another option would be listening to Django's built-in `request_finished` signal. However, this signal doesn't always carry the exact context of the specific request/response object as cleanly as wrapping the response itself, and might make correlation trickier in concurrent environments.

### Additional Context

_No response_

Yes

### Tip

SurbhiD404:
# Description

Fixes #4681

The Django instrumentation middleware ends the request span synchronously inside `process_response()`, right after `get_response()` returns. For a normal `HttpResponse` this is correct, since the body is fully rendered by that point.

For `StreamingHttpResponse`, however, `get_response()` returns before the stream is actually consumed. Django (and WSGI servers) only iterate `response.streaming_content` afterward, once the response is being sent to the client. As a result, the span's `end_time` is set before the streamed content has finished sending, producing inaccurate span durations for streaming endpoints.

For streaming responses (`response.streaming`), span completion is now deferred until `response.close()` is called, which occurs after the stream has been consumed. Existing behavior for non-streaming responses and exception paths is preserved.

Scope: this change covers the synchronous (WSGI) path only. Async/ASGI behavior remains unchanged.

## Type of change

* [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Added `TestMiddlewareStreamingResponse` in `tests/test_middleware.py`.

The test records the order of:

* streaming content consumption
* span completion (`Span.end`)

and verifies that the span is not completed before the streaming response is consumed.

Tested locally with:

`tox -e py313-test-instrumentation-django-3`

Result:

`78 passed`

* [x] No.

# Checklist:

* [x] Followed the style guidelines of this project
* [x] Changelogs have been updated
* [x] Unit tests have been added
* [ ] Documentation has been updated

### DIFF
--- a/.changelog/4723.fixed
+++ b/.changelog/4723.fixed
@@ -1,0 +1,1 @@
+﻿`opentelemetry-instrumentation-django`: fix spans ending prematurely for StreamingHttpResponse
--- a/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
+++ b/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
@@ -419,7 +419,17 @@ class _DjangoMiddleware:
         self._active_request_counter.add(-1, active_requests_count_attrs)
 
         if activation and span:
-            if exception:
+            if response.streaming and not exception:
+                original_close = response.close
+
+                def _end_span_on_close(*args, **kwargs):
+                    try:
+                        original_close(*args, **kwargs)
+                    finally:
+                        activation.__exit__(None, None, None)
+
+                response.close = _end_span_on_close
+            elif exception:
                 activation.__exit__(
                     type(exception),
                     exception,
--- a/instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
+++ b/instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py
@@ -58,6 +58,7 @@ from .views import (
     excluded_noarg2,
     response_with_custom_header,
     route_span_name,
+    streaming_view,
     traced,
     traced_template,
 )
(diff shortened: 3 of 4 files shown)


SurbhiD404:
Hi @xrmx,

When you get a chance, I'd appreciate a review of this PR. It fixes premature span completion for Django `StreamingHttpResponse` and includes a regression test and changelog fragment.

Happy to address any feedback. Thanks!

chuenchen309:
This has been open since June 20 with no human review and the stale bot has marked it twice, so here's an independent check of it against `main` (`ac3673d2c`). The fix holds up, including on the oldest and newest Django in the tox matrix.

**The test discriminates.** Applying only the two test files, without the middleware change, fails with exactly the bug being described:

```
AssertionError: 1 != 0 : Span should not be finished yet: stream not yet consumed
```

With the middleware change, `pytest instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py` is **42 passed** on both `test-requirements-0.txt` (Django 2.2.28) and `test-requirements-3.txt` (Django 5.2.11) — the two ends of the matrix in `tox.ini`.

**The ordering is right, not just the assertion.** Tracing `Span.end` and `HttpResponseBase.close` through one request:

```
--- view returned ---
yield-1
yield-2
gen-exhausted
HttpResponseBase.close
SPAN.end
```

The span ends inside `close()`, and `close()` lands after the generator is exhausted — which is the guarantee PEP 3333 gives for the real server case.

**Two edge cases I went looking for, both fine:**

- **`close()` called more than once.** The wrapper is an instance attribute, so a second `close()` re-enters `activation.__exit__`. I expected a "Calling end() on an ended span" warning; there is none, and only one span is exported. Django's test client already calls `close()` once on its own, so this path is exercised whether or not the caller does it.
- **`FileResponse`**, which is the `StreamingHttpResponse` most people actually hit, and whose `close()` also releases the file handle. Wrapping it doesn't break that: span ends at the right point *and* `file.closed` is `True` afterwards, because `original_close` still runs first in the `try`.

Non-streaming responses take the untouched branch, as intended.

**One thing that isn't a regression but is now newly fixable.** A generator that raises partway through leaves the span looking successful:

```
span status = UNSET, description = None
events on span = []
http.status_code = 200
```

Before this change the span had already ended by the time the generator ran, so this was unreachable; now the span is still open when the failure happens, and `activation.__exit__(None, None, None)` throws that information away. Something along the lines of wrapping the iteration, or at minimum passing real `exc_info` through from `close()`, would turn a silently-200 span into a useful one. Not a blocker for what this PR sets out to do — worth saying because this change is what makes it reachable at all.

@xrmx — the stale bot will close this on its next pass. The change is small (+12 in the middleware), carries a regression test and a changelog fragment, and is green across the matrix ends, so it seems a shame to lose it to the timer.

*AI-assisted review; the runs, the event ordering and the two Django versions above are from actually executing them, not inferred. I have no stake in this PR.*

SurbhiD404:
Hi @xrmx,

Whenever you get a chance, could you please take a look at this PR? I’d really appreciate your review.

Happy to address any feedback. Thank you!

PR #4724, opened 2026-06-20
fix(fastapi): preserve OTel context from sync route handlers to BackgroundTasks (#3586)

### LINKED ISSUE #3586: FastAPI: Context change in sync route does not appear in background tasks
### Describe your environment

OS: Ubuntu
Python version: 3.11.0

Package version:
opentelemetry-api==1.34.1
opentelemetry-sdk==1.34.1
opentelemetry-instrumentation-fastapi==0.55b1

### What happened?

When having a sync route in FastAPI and launching a background task, the changes done to the context do not appear in the context of the background task.

### Steps to Reproduce

```python
from fastapi import BackgroundTasks, FastAPI
from opentelemetry import context, trace
from opentelemetry.context import set_value
from opentelemetry.sdk.resources import SERVICE_NAME, Resource

app = FastAPI()

# =========================================================================== #
# Optional: Bug is reproducible with or without setuping the resource, tracer
# provider and FastAPI instrumentor,
resource = Resource.create(attributes={SERVICE_NAME: "pricing-recommendation-engine"})

provider = TracerProvider(resource=resource)

# provider.add_span_processor(BatchSpanProcessor(ConsoleSpanExporter()))

FastAPIInstrumentor().instrument_app(app)
# =========================================================================== #


def my_background_task():
    print(context.get_current())


def add_test_value_to_context():
    ctx = context.get_current()
    ctx = set_value("test_key", "test_value", context=ctx)
    context.attach(ctx)


@app.get("/sync-endpoint")
def sync_endpoint(background_tasks: BackgroundTasks):
    print(context.get_current())
    add_test_value_to_context()
    print(context.get_current())
    background_tasks.add_task(my_background_task)
    return "Hello, World!"


@app.get("/async-endpoint")
async def async_endpoint(background_tasks: BackgroundTasks):
    print(context.get_current())
    add_test_value_to_context()
    print(context.get_current())
    background_tasks.add_task(my_background_task)
    return "Hello, World!"
```

```sh
uvicorn fastapi_context_bug:app
```

On endpoint `/sync-endpoint`, we get this:

```
{'current-span-b13884d6-004f-41b3-84b0-8704e8205ff6': _Span(name="GET /sync-endpoint", context=SpanContext(trace_id=0x3d1bdba09cbb92564db952045e92ab94, span_id=0x93ec8565cd311ed4, trace_flags=0x01, trace_state=[], is_remote=False))}
{'current-span-b13884d6-004f-41b3-84b0-8704e8205ff6': _Span(name="GET /sync-endpoint", context=SpanContext(trace_id=0x3d1bdba09cbb92564db952045e92ab94, span_id=0x93ec8565cd311ed4, trace_flags=0x01, trace_state=[], is_remote=False)), 'test_key': 'test_value'}
INFO:     127.0.0.1:45994 - "GET /sync-endpoint HTTP/1.1" 200 OK
{'current-span-b13884d6-004f-41b3-84b0-8704e8205ff6': _Span(name="GET /sync-endpoint", context=SpanContext(trace_id=0x3d1bdba09cbb92564db952045e92ab94, span_id=0x93ec8565cd311ed4, trace_flags=0x01, trace_state=[], is_remote=False))}
```

On endpoint `/async-endpoint`, we get this:
```
{'current-span-b13884d6-004f-41b3-84b0-8704e8205ff6': _Span(name="GET /async-endpoint", context=SpanContext(trace_id=0xf1009be126c6edf441c7f465fcc05e7e, span_id=0xfa1e818754376b97, trace_flags=0x01, trace_state=[], is_remote=False))}
{'current-span-b13884d6-004f-41b3-84b0-8704e8205ff6': _Span(name="GET /async-endpoint", context=SpanContext(trace_id=0xf1009be126c6edf441c7f465fcc05e7e, span_id=0xfa1e818754376b97, trace_flags=0x01, trace_state=[], is_remote=False)), 'test_key': 'test_value'}
INFO:     127.0.0.1:48510 - "GET /async-endpoint HTTP/1.1" 200 OK
{'current-span-b13884d6-004f-41b3-84b0-8704e8205ff6': _Span(name="GET /async-endpoint", context=SpanContext(trace_id=0xf1009be126c6edf441c7f465fcc05e7e, span_id=0xfa1e818754376b97, trace_flags=0x01, trace_state=[], is_remote=False)), 'test_key': 'test_value'}
```

### Expected Result

The sync route should behave as the async route.

### Actual Result

See above.

### Additional context

_No response_

None

rajdhruvsingh:
# Description

Fixes the context propagation asymmetry described in #3586: when a *synchronous*
FastAPI route handler attaches a value to the OpenTelemetry context (e.g. via
`context.attach(set_value(...))`) and then schedules a `BackgroundTasks` task,
that value was silently lost by the time the background task ran. The same
code in an *async* route handler worked correctly.

## Root cause

Sync route handlers run inside a worker thread via `anyio.to_thread.run_sync`
(used internally by Starlette/FastAPI). This copies the current context *into*
the thread, but does not propagate any mutations made inside that thread back
out once it returns. By the time `BackgroundTasks.__call__` executes, the
context is already back to its pre-handler state — the mutation is gone before
this instrumentation's existing `BackgroundTask.__call__` hook (added for
#4251, background-task span parenting) ever runs, so there's no way to recover
it after the fact.

## Fix

`BackgroundTasks.add_task` is patched to snapshot the current OpenTelemetry
context at the moment a task is scheduled (this happens inside the same
thread as the route handler, so it sees the mutated context). When the task
actually executes, that captured context is merged into the *then-current*
ambient context — keys already present in the ambient context (notably the
BackgroundTask span this instrumentation starts around each task, from
#4251) take precedence, so existing span-parenting behavior is unaffected.
Only context set during the route handler that would otherwise be lost is
backfilled.

This works identically for sync and async background task functions.

Fixes #3586

## Type of change

# How Has This Been Tested?

- Added `test_background_task_inherits_sync_route_context`: reproduces the
  exact repro from #3586 (sync and async routes, each attaching a context
  value before scheduling a background task) and asserts both now see the
  value, where previously only the async case did.
- Added `test_background_task_context_does_not_clobber_span_parenting`: a
  sync-route variant of the existing #4251 regression test, confirming the
  preserved context does not override the dedicated `BackgroundTask` span or
  its parent/child span relationships.
- Verified against the original issue's exact reproduction script
  (TestClient, sync and async endpoints) with the fix applied (both pass)
  and with the fix reverted via `git stash` (sync fails exactly as reported
  in the issue, async still passes) — confirming the fix addresses the real
  reported behavior rather than an artifact of the new tests.
- Full existing test suite for `opentelemetry-instrumentation-fastapi` run
  locally: 101 passed, 6 skipped, 0 failures (2 unrelated pre-existing
  tests deselected — `test_entry_point_exists` and
  `test_instruments_with_fastapi_installed`, which fail in any local dev
  `.venv` with multiple instrumentation packages installed, due to
  `entry_points(group="opentelemetry_instrumentor")` returning every
  registered instrumentor rather than just this one; this is a known
  environment-isolation issue unrelated to this change and not present
  under the project's isolated `tox` CI environments).

- [x] `pytest instrumentation/opentelemetry-instrumentation-fastapi/tests/`

- [ ] Yes. - Link to PR:
- [x] No.

# Checklist:

### DIFF
--- a/.changelog/4724.fixed
+++ b/.changelog/4724.fixed
@@ -1,0 +1,1 @@
+`opentelemetry-instrumentation-fastapi`: Fix OpenTelemetry context (e.g. values set via `context.attach`) not propagating from a synchronous route handler to its `BackgroundTasks`, since sync handlers run in a worker thread and the resulting context mutation was previously lost once the thread returned.
--- a/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
@@ -172,6 +172,7 @@ API
 from __future__ import annotations
 
 import functools
+import inspect
 import logging
 import types
 from typing import Any, Collection, Literal
@@ -179,11 +180,13 @@ from weakref import WeakSet as _WeakSet
 
 import fastapi
 from starlette.applications import Starlette
-from starlette.background import BackgroundTask
+from starlette.background import BackgroundTask, BackgroundTasks
 from starlette.middleware.errors import ServerErrorMiddleware
 from starlette.routing import Match, Route
 from starlette.types import ASGIApp, Receive, Scope, Send
 
+from opentelemetry import context as otel_context
+from opentelemetry.context import set_value
 from opentelemetry.instrumentation._semconv import (
     _get_schema_url,
     _OpenTelemetrySemanticConventionStability,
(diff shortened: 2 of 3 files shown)


aabmass:
@rajdhruvsingh I left a comment on https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3586#issuecomment-4837429637

PR #4727, opened 2026-06-22
feat(kafka): add messaging.kafka.cluster.id to producer/consumer spans

### LINKED ISSUE #4809: feat(kafka): add messaging.kafka.cluster.id to Kafka producer/consumer spans
## Problem

Kafka spans (aiokafka, kafka-python, confluent-kafka) do not include `messaging.kafka.cluster.id`. When the same topic name exists across multiple clusters or environments, there is no way to distinguish which cluster a span came from.

## Expected behavior

`messaging.kafka.cluster.id` ([semconv](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/kafka.md)) should be set on Kafka producer and consumer spans. It is a `Recommended` attribute in the Kafka messaging semantic conventions, so it is emitted **by default** (no opt-in flag), consistent with the other Recommended Kafka attributes these instrumentations already produce. The id is read from the client's own already-resolved metadata, so no extra broker connection is required.

## Context

Part of a cross-language effort to add this attribute consistently across OTel instrumentations. See [opentelemetry-java-instrumentation#18977](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/18977) for the cross-language tracking issue.

shashank-reddy-nr:
Fixes #4809

## Description

Adds `messaging.kafka.cluster.id` ([semconv](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/kafka.md)) to Kafka producer and consumer spans across the `aiokafka`, `kafka-python`, and `confluent-kafka` instrumentations.

It is a `Recommended` semantic-convention attribute, so it is emitted **by default** (no opt-in flag) — consistent with the other Recommended Kafka attributes these instrumentations already produce. The id is read from each client's own metadata with no extra broker connections:

- **kafka-python**: read from the client's `ClusterMetadata` (`cluster.cluster_id`), captured from the broker `MetadataResponse` via `update_metadata`. Compatible with kafka-python 2.0.x (which does not persist `cluster_id` on `ClusterMetadata`) as well as 2.1+.
- **aiokafka**: sends a single `MetadataRequest_v5` after `start()` completes and caches the result on the client object. All subsequent spans read from the cached attribute.
- **confluent-kafka** (librdkafka):
  - *Producer*: `list_topics(timeout=0)` returns immediately from librdkafka's in-process metadata cache (no network I/O). The result is cached keyed by bootstrap servers address; all subsequent calls are a dict lookup.
  - *Consumer*: reads from the same bootstrap-servers cache populated by producers. `list_topics` is never called on consumers — calling it during a consumer group rebalance can trigger a use-after-free inside librdkafka (issue #4214).

The attribute may be absent on the first span if metadata has not yet been resolved; it self-heals on subsequent spans.

## Type of change

## How Has This Been Tested?

Unit tests added covering:
- the attribute is set when the cluster id is available
- the attribute is absent when broker metadata has not yet been received
- (confluent-kafka) producer result is cached after the first call; subsequent calls never invoke `list_topics`
- (confluent-kafka) consumers read from the bootstrap cache and never call `list_topics`

- [x] No.

## Checklist:


shashank-reddy-nr:
Hi @xrmx Can you please review this PR, whenever you have time?

instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/utils.py
```diff
@@ -178,11 +220,13 @@ def _enrich_span(
 
     if operation:
         span.set_attribute(MESSAGING_OPERATION, operation.value)
-    else:
-        span.set_attribute(SpanAttributes.MESSAGING_TEMP_DESTINATION, True)
```
emdneto, line 177: Why remove this?

instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/utils.py
```diff
@@ -24,6 +26,47 @@
...
+
+
+def _extract_cluster_id(
+    instance: Any, bootstrap_servers: Optional[str] = None
+) -> Optional[str]:
+    if instance is None:
+        return None
+    if hasattr(instance, "flush"):
+        try:
+            cluster_metadata = instance.list_topics(timeout=0)
```
emdneto, line 48: so for producers it will list_topics in every call? any overhead increase here for this hot path?

instrumentation/opentelemetry-instrumentation-aiokafka/src/opentelemetry/instrumentation/aiokafka/__init__.py
```diff
@@ -131,6 +132,74 @@ class UninstrumentKwargs(TypedDict, total=False):
...
+
+
+async def _start_producer_wrapper(
+    func: Callable[..., Awaitable[None]],
+    instance: aiokafka.AIOKafkaProducer,
+    args: tuple[Any, ...],
+    kwargs: dict[str, Any],
+) -> None:
+    await func(*args, **kwargs)
+    await _fetch_and_cache_cluster_id(instance.client)
```
emdneto: Something is odd with aiokafka tests after wrapping start.

instrumentation/opentelemetry-instrumentation-aiokafka/src/opentelemetry/instrumentation/aiokafka/utils.py
```diff
@@ -79,6 +79,10 @@ async def __call__(
 
 _LOG = getLogger(__name__)
 
+# TODO(semconv #3819): once generated in opentelemetry-semantic-conventions,
+# use messaging_attributes.MESSAGING_KAFKA_CLUSTER_ID instead of this literal.
+_MESSAGING_KAFKA_CLUSTER_ID = "messaging.kafka.cluster.id"
```
emdneto, line 74: I understand that the attribute has already been added to the semantic conventions, but it hasn’t been released yet. Could we wait for the release before using it here? I’ll raise this at the next SIG meeting to get guidance on whether instrumentations should mix attributes from different semantic convention versions without the opt-in gate.

emdneto (commented):
Thanks. Left some comments.

instrumentation/opentelemetry-instrumentation-aiokafka/src/opentelemetry/instrumentation/aiokafka/utils.py
shashank-reddy-nr: Yes, we can wait until the next semantic-conventions release. I recently learned that if an attribute hasn't been added to semantic-conventions yet, we should add an opt-in experimental attribute instead. Once the attribute is present and released in sem-conv, we can directly replace the current TODO with messaging_attributes.MESSAGING_KAFKA_CLUSTER_ID.

instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/utils.py
shashank-reddy-nr: The MESSAGING_TEMP_DESTINATION attribute is deprecated in sem-conv 1.21. So, there are a bunch of places this attribute is present so reverting this change so that all those cleanup can be addressed in new PR in future.

instrumentation/opentelemetry-instrumentation-aiokafka/src/opentelemetry/instrumentation/aiokafka/utils.py
emdneto: In the meantime: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4974 @shashank-reddy-nr

instrumentation/opentelemetry-instrumentation-aiokafka/src/opentelemetry/instrumentation/aiokafka/__init__.py
shashank-reddy-nr: Addressed this

instrumentation/opentelemetry-instrumentation-aiokafka/src/opentelemetry/instrumentation/aiokafka/utils.py
shashank-reddy-nr: Hi @emdneto, when do you expect this to be completed? Do you recommend waiting until the PR above is merged?

emdneto: @shashank-reddy-nr PR needs review. If you have time, can you review?

Shriprasad-P (commented):
Thanks for adding `messaging.kafka.cluster.id` on producer/consumer spans.

A few notes:

1. **Attribute source** — please confirm where `cluster.id` is read from (client config vs broker metadata). If it can be absent on some clients, assert the attribute is omitted rather than set to empty string.
2. **Cardinality** — cluster id is usually low-cardinality, which is good for span attributes; a short note in the PR description linking the semconv requirement would help reviewers.
3. **Tests** — ensure both producer and consumer paths are covered when cluster id is present and when it is missing.

Happy to take another pass after those cases are explicit in tests.

PR #4746, opened 2026-06-26
opentelemetry-instrumentation-boto3sqs: avoid KeyError when all messages in send_message_batch fail

### LINKED ISSUE #4744: [boto3sqs] KeyError: 'Successful' in send_batch_wrapper when all SendMessageBatch entries fail
**Describe your environment**

- `opentelemetry-instrumentation-boto3sqs` 0.62b1 (also present on `main`)
- Python 3.13, boto3 SQS resource/client
- AWS Lambda, auto-instrumentation

**What happened?**

`send_batch_wrapper` accesses `retval["Successful"]` directly:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py#L270

```python
retval = wrapped(*args, **kwargs)
for successful_messages in retval["Successful"]:
    ...
```

The SQS `SendMessageBatch` response **omits the `Successful` key entirely when every entry in the batch fails** (the response then contains only `Failed`). botocore parses this into a dict with no `Successful` key, so the wrapper raises:

```
KeyError: 'Successful'
  File ".../opentelemetry/instrumentation/boto3sqs/__init__.py", line 270, in send_batch_wrapper
    for successful_messages in retval["Successful"]:
```

This is the same class of bug as the `MessageAttributes` `KeyError` reported for this instrumentation — assuming an optional/conditional response key is always present.

**Impact:** the instrumentation turns a normally-recoverable all-failed batch into an unhandled exception, and — worse — it masks the real failure: it is raised *before* the application's own handling of the `Failed` list runs, so the underlying SQS failure codes never surface.

**Expected:** a batch where all entries fail should not raise; the wrapper should simply produce no message-id attributes (there are no successful messages to enrich).

**How to reproduce?**

Call `send_message_batch` (instrumented) where the SQS response contains only `Failed` and no `Successful` key.

**Fix**

Use `retval.get("Successful", [])`. PR incoming.

ferencjuhasz:
## Description

`Boto3SQSInstrumentor`'s `send_message_batch` wrapper unconditionally indexes `retval["Successful"]`:

```python
retval = wrapped(*args, **kwargs)
for successful_messages in retval["Successful"]:
```

When **every** entry in a batch fails, the SQS `SendMessageBatch` response contains only `Failed` and omits the `Successful` key entirely. The wrapper then raises `KeyError: 'Successful'`. Because this happens inside the instrumentation wrapper, the exception propagates out of `client.send_message_batch` and breaks the caller — even though the failure is an ordinary per-message batch failure the application is expected to handle from the response.

```
File ".../boto3/resources/action.py", line 88, in __call__
File ".../opentelemetry/instrumentation/boto3sqs/__init__.py", line 270, in send_batch_wrapper
    for successful_messages in retval["Successful"]:
KeyError: 'Successful'
```

Fixes #4744

## Fix

Iterate over `retval.get("Successful", [])`. Spans for failed entries are still created and ended by the existing loop, so tracing behavior is otherwise unchanged.

## Tests

The `send_message_batch` wrapper previously had no test coverage. This PR adds:

- `test_send_message_batch` — happy path: asserts the producer span per entry, full span attributes, and context propagation injection.
- `test_send_message_batch_all_failed` — regression test reproducing the `KeyError` when the response has no `Successful` key.


xrmx (approved):

xrmx:
@ferencjuhasz no need to rebase the changes, we'll take care of that at merge time, thanks!

ferencjuhasz:
sorry about that, thanks!

emdneto (approved):
I found this related issue #4744 that was closed without a PR. Just reopned and added it to the description of this PR.

PR #4751, opened 2026-06-27
openai-v2: await AsyncStream.close on default async streaming path

### LINKED ISSUE #4710: openai-v2: BaseStreamWrapper.close() never awaits AsyncStream.close() — async streaming leaks connections/memory (OOM)
### Describe your environment

- `opentelemetry-instrumentation-openai-v2==2.4b0` (latest release on PyPI; also applies to the earlier `2.x` betas that ship the single sync-`close` `StreamWrapper`)
- `openai` async client (`AsyncOpenAI` / `AsyncAzureOpenAI`), streaming chat completions
- Python 3.13

### What happened?

`BaseStreamWrapper.close()` is a **synchronous** method that calls `self.stream.close()` without awaiting it:

```python
# src/opentelemetry/instrumentation/openai_v2/patch.py  (2.4b0, line 655)
def close(self):
    self.stream.close()
    self.cleanup()
```

For the **async** streaming path, `async_chat_completions_create` returns `ChatStreamWrapper(BaseStreamWrapper)` (patch.py:243), so `self.stream` is an `openai.AsyncStream`, whose `close()` is a coroutine:

```python
# openai/_streaming.py
class AsyncStream(Generic[_T]):
    async def close(self) -> None:   # coroutine
        await self.response.aclose()
        ...
```

Calling `self.stream.close()` without `await` therefore:

1. emits `RuntimeWarning: coroutine 'AsyncStream.close' was never awaited`, and
2. **never closes the underlying httpx response/connection**, so the buffered response + connection are leaked on every streamed async turn.

This also defeats application-side cleanup: code that does `await wrapper.close()` gets `None` back (the sync `close()` returns `None` after creating-and-dropping the coroutine), so even a correct caller cannot close the real stream. Over a long-lived async process this accumulates as a connection/memory leak.

### Steps to Reproduce

```python
import asyncio
from openai import AsyncOpenAI
from opentelemetry.instrumentation.openai_v2 import OpenAIInstrumentor

OpenAIInstrumentor().instrument()
client = AsyncOpenAI()

async def main():
    for _ in range(1000):
        stream = await client.chat.completions.create(
            model="gpt-4o-mini",
            messages=[{"role": "user", "content": "hi"}],
            stream=True,
        )
        # consume a little, then close early (mirrors cancellation / early-exit)
        async for _ in stream:
            break
        await stream.close()  # wrapper.close() is sync -> AsyncStream.close() never awaited

asyncio.run(main())
# observe: RuntimeWarning: coroutine 'AsyncStream.close' was never awaited
# and growing httpx connections / RSS
```

### Expected Result

Closing an async streaming wrapper closes the underlying `AsyncStream` (awaits `AsyncStream.close()`), releasing the httpx response/connection. No `coroutine ... was never awaited` warning, no leak.

### Actual Result

The `AsyncStream.close()` coroutine is created but never awaited; the underlying connection/response is leaked.

### Additional context

This appears to already be **fixed on `main`** by #4500 (merged 2026-05-12), which introduced `AsyncChatStreamWrapper` extending `util-genai`'s `AsyncStreamWrapper`, whose `close()` is now `async` and awaits the underlying stream (`util/opentelemetry-util-genai/.../stream.py`, `async def close(self): ... await self._self_stream.close()`).

However, **no released version contains the fix** — `2.4b0` is the latest on PyPI (published 2026-05-01) and predates #4500. Could a release be cut that includes the async-close fix? Until then, every released `openai-v2` instrumentation version leaks `AsyncStream`s for async streaming callers.

yulin-li:
# Description

`BaseStreamWrapper.close()` is synchronous and calls `self.stream.close()` without awaiting it. On the default (non-experimental) async streaming path, `async_chat_completions_create_v_old` returns a `LegacyChatStreamWrapper` wrapping an `openai.AsyncStream`, whose `close()` is a coroutine. Calling it without `await` therefore:

1. emits `RuntimeWarning: coroutine 'AsyncStream.close' was never awaited`, and
2. never closes the underlying httpx response/connection, leaking it on every early-closed async stream.

This also means a correct caller doing `await wrapper.close()` got `None` back, so they could not close the real stream either.

The experimental path (`gen_ai_latest_experimental` opt-in) was already fixed by #4500 via util-genai's `AsyncStreamWrapper`, but the default path remained affected.

This PR adds `AsyncLegacyChatStreamWrapper(LegacyChatStreamWrapper)` with an awaitable `close()` that awaits the underlying `AsyncStream.close()`, and uses it for async streaming chat completions on the legacy path. This mirrors the existing sync/async split already present in `response_wrappers.py` (`AsyncResponseStreamWrapper`).

Fixes #4710

## Type of change

# How Has This Been Tested?

- Added `test_async_chat_completion_streaming_close_awaits_underlying_stream`, which asserts the legacy async wrapper's `close()` awaits the underlying `AsyncStream.close()` and finalizes the span.
- Updated `test_async_chat_completion_streaming_not_complete` to `await response.close()` on the legacy path.
- Ran the openai-v2 suite with `-W error::RuntimeWarning`; the `coroutine ... was never awaited` warning is gone and all tests pass.

- [ ] Yes.
- [x] No.

# Checklist:


instrumentation-genai/opentelemetry-instrumentation-openai-v2/.changelog/4751.fixed
```diff
@@ -0,0 +1 @@
+`opentelemetry-instrumentation-openai-v2`: await `AsyncStream.close` when closing an async streaming chat completion on the default semantic convention path so the underlying httpx response/connection is released instead of leaked
```
xrmx, line 1: This should go inside the `.changelog` in the instrumentation directory

xrmx (approved):
I was going to cut a new release for openai-v2 thinking this was fixed in main, but isn't. So this LGTM and I think is small enough to merge even if the work has been moved to the genai repo.

instrumentation-genai/opentelemetry-instrumentation-openai-v2/.changelog/4751.fixed
yulin-li: Moved the fragment into the openai-v2 package's own `.changelog/` directory (`instrumentation-genai/opentelemetry-instrumentation-openai-v2/.changelog/4751.fixed`), since openai-v2 is now released independently. Thanks!


_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

eternalcuriouslearner (approved):

lmolkova (approved):

yulin-li:
Hi, I don't have permission to merge, could you please help to merge it?

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/patch.py
```diff
@@ -806,3 +806,20 @@ def cleanup(self, error: Optional[BaseException] = None):
         else:
             self.span.end()
         self._started = False
+
+
+class AsyncLegacyChatStreamWrapper(LegacyChatStreamWrapper):
+    """Async variant of :class:`LegacyChatStreamWrapper`.
+
+    Wraps an ``openai.AsyncStream`` whose ``close`` is a coroutine. ``close``
```
xrmx, line 814: A copilot review also spotted that we are not awaiting close in `__aexit__` when the openai api is used as async context manager, e.g.

```
    with vcr.use_cassette(
        "test_async_chat_completion_streaming_not_complete.yaml"
    ):
        response = await async_openai_client.chat.completions.create(**kwargs)

        underlying_close = AsyncMock()
        response.stream.close = underlying_close

        async with response:
            async for _ in response:
                # consume one chunk, then exit early
                break

    underlying_close.assert_awaited_once()
```

Mind to fix this as well?

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/patch.py
```diff
@@ -806,3 +806,20 @@ def cleanup(self, error: Optional[BaseException] = None):
...
+    ``AsyncStream.close`` is awaited, releasing the httpx response/connection
+    instead of leaking it (and avoiding the ``coroutine ... was never awaited``
+    warning).
+    """
+
+    stream: AsyncStream
+
+    async def close(self) -> None:
+        await self.stream.close()
+        self.cleanup()
```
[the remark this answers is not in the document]
yulin-li: Wrapped the await in `try`/`finally` so `cleanup()` (span finalization) always runs even if `AsyncStream.close()` raises, and the original exception propagates unmodified.


_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

instrumentation-genai/opentelemetry-instrumentation-openai-v2/tests/test_async_chat_completions.py
```diff
@@ -1013,6 +1015,48 @@ async def test_async_chat_completion_streaming_not_complete(
...
+
+        async for _ in response:
+            # consume one chunk, then exit early (mirrors cancellation)
+            break
+
+        underlying_close = AsyncMock()
+        response.stream.close = underlying_close
+
+        # close() must be awaitable and must await the underlying stream close.
+        await response.close()
```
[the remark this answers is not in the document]
yulin-li: Switched to `AsyncMock(wraps=response.stream.close)` so the real close still runs (releasing the connection) while the test asserts it was awaited.


_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/patch.py
yulin-li: Good catch. Added an `__aexit__` override on `AsyncLegacyChatStreamWrapper` that awaits `AsyncStream.close`, so exiting the response via `async with` releases the httpx connection too. Also added a test covering the context-manager exit path using your suggested approach.


_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

PR #4754, opened 2026-06-28
docs: add instrumentation naming convention to contributing guide

### LINKED ISSUE #910: Add note to contributing guide about instrumentation name
A followup to https://github.com/open-telemetry/opentelemetry-python/issues/2457, instrumentation names should use a canonical identifier used anywhere the instrumentation needs to obtain a tracer.

krrishapatel:
# Description

Adds a note to the instrumentation guidelines section of CONTRIBUTING.md documenting that instrumentations should use their top-level package name (e.g. `"opentelemetry.instrumentation.requests"`) as a fixed string when obtaining a tracer or meter, rather than `__name__`.

Fixes #910

## Type of change

# How Has This Been Tested?

Documentation-only change.

- [x] No.

Assisted-by: Claude Opus 4.6

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -371,6 +371,8 @@ Below is a checklist of things to be mindful of when implementing a new instrume
   - Contributions related to outdated HTTP semantic conventions (conventions prior to becoming [stable](https://github.com/open-telemetry/semantic-conventions/tree/v1.23.0)) will likely be discouraged, as they increase complexity and the potential for misconceptions.
 - Contains a name that is not already claimed in [Pypi](https://pypi.org/). Contact a maintainer, bring the issue up in the weekly Python SIG or create a ticket in Pypi if a desired name has already been taken.
 - Extends from [BaseInstrumentor](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/2518a4ac07cb62ad6587dd8f6cbb5f8663a7e179/opentelemetry-instrumentation/src/opentelemetry/instrumentation/instrumentor.py#L35)
+- Uses a canonical instrumentation name when obtaining a tracer or meter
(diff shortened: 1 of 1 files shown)


xrmx (changes requested):
This contains unrelated changes

krrishapatel:
@xrmx Rebased to remove the unrelated change — this now only touches CONTRIBUTING.md. Mind taking another look?

krrishapatel:
Closing this myself — the guidance as written is wrong, and I should have checked it against the codebase before proposing it.

I went back through the shipped instrumentations, and 35 of them pass `__name__` to `get_tracer`, including `requests`, `django`, and `flask`. So this PR was telling new contributors to avoid the convention that the repo overwhelmingly follows.

The reason `__name__` is correct there is that these calls live in the package's top-level `__init__.py`, where `__name__` already evaluates to the canonical name, e.g. in `opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py` it is exactly `"opentelemetry.instrumentation.requests"`. There is nothing to fix in that case.

The narrower thing I was actually reacting to is real but rare: when `get_tracer` is called from a *submodule*, `__name__` picks up the submodule path instead of the package name. There are two such spots in shipped code:

- `opentelemetry-instrumentation-pika/src/opentelemetry/instrumentation/pika/pika_instrumentor.py:202`
- `opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py:434` (this one is in an `__init__.py`, so it is fine)

That is a two-line observation about one package, not a contributing-guide checklist item, so it does not justify this PR. If maintainers think the pika case is worth normalizing I am happy to open a small targeted PR for it, but I would rather ask first than add another unsolicited one.

Apologies for the noise on this and on #4753. @xrmx your "unrelated changes" review was the correct call on the original version, and I should have validated the underlying claim rather than just trimming the diff.

PR #4757, opened 2026-06-29
fix(openai-agents-v2): default content capture to NO_CONTENT when env var is unset

### LINKED ISSUE #4756: instrumentation-openai-agents-v2 captures prompts by default
### Describe your environment

_No response_

### What happened?

When `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` is unset, `opentelemetry-instrumentation-openai-agents-v2` captures prompts, system instructions, model responses, and tool arguments and results.

### Steps to Reproduce

Instrument an OpenAI agent without setting `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`, then run it with a prompt and inspect the exported spans.

### Expected Result

The exported spans do not contain prompts or model responses.

### Actual Result

The exported spans contain prompts and model responses.

### Additional context

_No response_

_No response_

### Tip

_No response_

rajdhruvsingh:
## Description

Fixes #4756

When `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` is not set,
`_resolve_content_mode` was returning `ContentCaptureMode.SPAN_AND_EVENT`,
causing prompts, system instructions, model responses, and tool arguments
to be captured in exported spans by default.

Per the GenAI semantic conventions, content capture should be **opt-in**.

### Root cause

Three return paths in `_resolve_content_mode` defaulted to `SPAN_AND_EVENT`:
- `value is None` — env var not set
- `value == ""` — env var set but empty
- `mapping.get(text, ...)` fallback — unrecognized value

All three now return `NO_CONTENT`.

## Type of change

## How Has This Been Tested?

- Updated two existing unit tests in `test_zz_coverage_improvements.py`
  that were asserting the old buggy behavior (`None` and `""` → `SPAN_AND_EVENT`),
  now correctly assert `NO_CONTENT`.
- Updated `test_agent_span_collects_child_messages` and
  `test_response_span_records_response_attributes` in `test_tracer.py` to
  explicitly pass `capture_message_content="span_only"` — these tests check
  content is captured when opted in, so they must opt in explicitly rather
  than relying on the default.
- Manually verified all code paths in `_resolve_content_mode` including:
  `None`, `""`, unknown string, `"true"`, `"false"`, `"span_only"`, `"event_only"`.
- Full test suite: **103 passed, 0 failed**.

- [x] No.

## Checklist

### DIFF
--- a/.changelog/4757.fixed
+++ b/.changelog/4757.fixed
@@ -1,0 +1,1 @@
+`opentelemetry-instrumentation-openai-agents-v2`: fix default content capture mode to NO_CONTENT when env var is unset
--- a/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/__init__.py
+++ b/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2/src/opentelemetry/instrumentation/openai_agents/__init__.py
@@ -81,11 +81,11 @@ def _resolve_content_mode(value: Any) -> ContentCaptureMode:
         )
 
     if value is None:
-        return ContentCaptureMode.SPAN_AND_EVENT
+        return ContentCaptureMode.NO_CONTENT
 
     text = str(value).strip().lower()
     if not text:
-        return ContentCaptureMode.SPAN_AND_EVENT
+        return ContentCaptureMode.NO_CONTENT
 
     mapping = {
         "span_only": ContentCaptureMode.SPAN_ONLY,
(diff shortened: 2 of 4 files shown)


herin049:
I believe all gen-AI packages have moved to https://github.com/open-telemetry/opentelemetry-python-genai

rajdhruvsingh:
> I believe all gen-AI packages have moved to https://github.com/open-telemetry/opentelemetry-python-genai

Thanks @herin049  for the clarification! I wasn't aware of the migration. Should I close this PR and open the same change in the opentelemetry-python-genai repository instead?

PR #4769, opened 2026-07-01
instrumentation: don't override an existing error span status in _set…

### LINKED ISSUE #3713: `_set_status` semconv helper can override span status if already defined

WatchTree-19:
# Description

`_set_status` (the shared HTTP semconv helper) ends by unconditionally calling `span.set_status(Status(status))`. The SDK already ignores an `UNSET` status and keeps an existing `OK`, but it still replaces an existing `ERROR` status, dropping any description the application or another instrumentation set. On a 4xx/5xx response this overwrote a user-set error description to `None`, which is the behaviour reported in #3713.

This change reads the span's current status and skips the `set_status` call when the span already carries an `ERROR`, so the more specific status is preserved. Successful/redirect responses (which map to `UNSET`) and spans that do not expose a readable status are unaffected, so behaviour is unchanged for the common case.

Picking this up after it sat idle for a few months. Thanks @RiyaChaturvedi37 for flagging the approach back in April; happy to hand it back if you are still on it.

Fixes #3713

## Type of change

# How Has This Been Tested?

Added unit tests in `opentelemetry-instrumentation/tests/test_semconv.py`:

- `test_set_status_preserves_existing_error_status`: an ERROR status already set on the span is not overwritten on a 500.
- `test_set_status_sets_error_when_status_unset`: an ERROR is still recorded when the span status is UNSET.
- `test_set_status_sets_error_when_no_status_attribute`: spans without a readable status (`getattr` -> `None`) are still updated.

The full `test_semconv.py` suite passes (64 tests).

- [x] No.

# Checklist:


opentelemetry-instrumentation/tests/test_semconv.py
```diff
@@ -521,6 +521,61 @@ def test_non_recording_span(self):
         self.assertIn("http.status_code", metrics_attributes)
         self.assertIn("http.response.status_code", metrics_attributes)
 
+    def test_set_status_preserves_existing_error_status(self):
+        # An ERROR status already set on the span (e.g. by the application or
+        # another instrumentation) must not be overwritten by _set_status, which
+        # would drop its description. See issue #3713.
+        span = Mock()
```
herin049: All of these tests should be rewritten to use real spans and not `Mock` objects.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -684,7 +684,17 @@ def _set_status(
                     span.set_attribute(ERROR_TYPE, status_code_str)
                 metrics_attributes[ERROR_TYPE] = status_code_str
         if span.is_recording():
-            span.set_status(Status(status))
+            # Do not override a status that has already been set on the span.
+            # The SDK ignores an UNSET status and keeps an existing OK, but it
+            # will replace an existing ERROR, dropping any description set by
+            # the application or another instrumentation. Skip in that case so
+            # the more specific status is preserved (see issue #3713).
+            current_status = getattr(span, "status", None)
```
herin049, line 672: This doesn't look correct to me, look at the `Span` implementation in the SDK.

herin049 (changes requested):

WatchTree-19:
thanks @herin049.

tests: rewritten to use real recording SDK spans instead of Mocks (e0a5df5), asserting on the resulting span.status. covers the #3713 case (an app-set ERROR + description survives a 5xx), plus fresh -> ERROR on 5xx, 2xx leaving the status UNSET, and an app-set OK preserved.

on the fix, what looks off to you? the SDK Span.set_status already ignores an UNSET status and keeps an existing OK, but for an existing ERROR it just does self._status = status, which replaces it and drops the description. that's the #3713 report: an app-set ERROR + description overwritten to None by the instrumentation's Status(status). the guard only skips that one case, and the new real-span tests cover the behaviour across UNSET/OK/ERROR.

reading span.status is the only way to see the current status here, since the trace API doesn't expose a getter on Span (only the SDK ReadableSpan does), and we're already inside is_recording(). if you'd rather not read span.status from this helper, happy to do it another way, what did you have in mind?

herin049:
> thanks @herin049.
> 
> tests: rewritten to use real recording SDK spans instead of Mocks ([e0a5df5](https://github.com/open-telemetry/opentelemetry-python-contrib/commit/e0a5df5eaf3ff1ff046f9b28e82a199db33c71fe)), asserting on the resulting span.status. covers the #3713 case (an app-set ERROR + description survives a 5xx), plus fresh -> ERROR on 5xx, 2xx leaving the status UNSET, and an app-set OK preserved.
> 
> on the fix, what looks off to you? the SDK Span.set_status already ignores an UNSET status and keeps an existing OK, but for an existing ERROR it just does self._status = status, which replaces it and drops the description. that's the #3713 report: an app-set ERROR + description overwritten to None by the instrumentation's Status(status). the guard only skips that one case, and the new real-span tests cover the behaviour across UNSET/OK/ERROR.
> 
> reading span.status is the only way to see the current status here, since the trace API doesn't expose a getter on Span (only the SDK ReadableSpan does), and we're already inside is_recording(). if you'd rather not read span.status from this helper, happy to do it another way, what did you have in mind?

It looks like the `Span` `set_status` method will be a no-op if the status is already set to `OK`. It's probably best to change the check to explicitly check if the status is `UNSET`.

WatchTree-19:
good call - switched to an explicit UNSET check (3d1ba13). the SDK already treats OK as final and ignores a later set_status, so guarding on UNSET is clearer than skipping only ERROR, and it still preserves an app-set ERROR + description on a later 5xx (the #3713 case). real-span tests unchanged, 65 pass.

WatchTree-19:
Merged current main in (`35da7cbf`) — the branch was showing `behind`. Net diff is unchanged: still the three files, and the full `test_semconv.py` suite is **65 passed** against the merged tree.

@herin049 — you asked for the explicit UNSET check and I pushed that in `3d1ba13` about two weeks ago, so I think this is waiting on you rather than on me. The guard now reads:

```python
current_status = getattr(span, "status", None)
if current_status is None or current_status.status_code is StatusCode.UNSET:
    span.set_status(Status(status))
```

which is the version you suggested — the SDK already treats OK as final, so guarding on UNSET is clearer than special-casing ERROR, and it still preserves an app-set ERROR + description on a later 5xx (the #3713 case). The four `_set_status` tests cover UNSET → ERROR, 2xx leaving UNSET, an app-set OK preserved, and the #3713 regression itself.

No rush, but the `CHANGES_REQUESTED` can't clear itself — happy to take another round if anything still looks off.

WatchTree-19:
before the stale bot gets this one. @herin049 i reworked per your review (commit landed after the change request), the instrumentation no longer overrides an already-set error span status. re-review when you have a moment?

WatchTree-19:
@herin049 you're right on the `_semconv.py` point, and having gone and read `Span.set_status` in the SDK I don't think this change belongs here at all.

Two problems with what I wrote:

**It reads state the API doesn't define.** `getattr(span, "status", None)` only resolves on `ReadableSpan` in the SDK. The API `Span` ABC has `set_status` but no readable `status` property, so for any other implementation the `getattr` silently returns `None` and the guard is a no-op. The behaviour would then differ by span implementation, which is worse than the bug I was trying to fix.

**It duplicates precedence logic that already lives in the SDK.** `set_status` already encodes the rules:

```python
if (
    self._status
    and self._status.status_code is StatusCode.OK
    or status.status_code is StatusCode.UNSET
):
    return
```

OK is final, UNSET is ignored. Adding a second, different rule in instrumentation code means two places decide precedence and they don't agree.

So the real gap in #3713 is narrower than I framed it: `set_status` will let a bare `Status(ERROR)` replace an existing `Status(ERROR, "some description")`, and the description is lost. That's a rule about status precedence, which means it belongs next to the rules already there, in `opentelemetry-python`'s `Span.set_status`, not in `_semconv`.

I'd rather close this and open it against the SDK than patch around it here. Want me to do that, or would you prefer it stays in contrib with a different approach — for example threading the description through so nothing is dropped, without introspecting span state?

On the tests: agreed on dropping the `Mock` objects, but I'd rather not rewrite them until the above is settled, since they're testing behaviour that may not survive.

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
WatchTree-19: Addressed in 3d1ba13 — switched to the explicit UNSET check you suggested, so the SDK's own "OK is final" behaviour is relied on rather than special-cased. Resolving, but happy to reopen if the shape still looks wrong to you.

WatchTree-19:
Closing this, and @herin049 your objection is the reason.

You pointed out that the trace API does not expose a readable status on `Span` at all, only the SDK's `ReadableSpan` does, so a helper that reads `span.status` before setting it is relying on an implementation detail. On top of that it put precedence logic in two places, here and in `set_status`, which is exactly the kind of pair that drifts apart and then disagrees.

The narrower gap underneath all this is real, though, and it is in the SDK. `set_status` returns early when the span is already `OK` and when the incoming status is `UNSET`, but nothing stops a bare `Status(ERROR)` replacing an existing `Status(ERROR, "...")`, so a specific message is silently swapped for none. I have raised that as open-telemetry/opentelemetry-python#5661 with a reproduction, and the fix goes in `set_status`, where precedence already lives, rather than in contrib.

Thanks for the review. Steering it out of the instrumentation helper was the right call and it took the problem somewhere it can actually be fixed once.

PR #4741, opened 2026-06-25
Fix duplicate Pyramid traversal subscribers

sjh9714:
# Description

Fixes #4663.

`opentelemetry-instrumentation-pyramid` can register duplicate `_before_traversal` subscribers when a Pyramid include calls `config.commit()` and a later include creates another child configurator. This change records that the callbacks have already been registered in Pyramid settings, so later includes re-enable tracing without adding a duplicate subscriber or tween.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [x] `uvx --from tox --with tox-uv tox -e py314-test-instrumentation-pyramid`
- [x] `uvx --from tox --with tox-uv tox -e lint-instrumentation-pyramid`
- [x] `uvx --from tox --with tox-uv tox -e ruff`

# Does This PR Require a Core Repo Change?

- [x] No.

# Checklist:

- [x] Followed the style guidelines of this project
- [x] Changelogs have been updated
- [x] Unit tests have been added
- [ ] Documentation has been updated

PR #4773, opened 2026-07-08
fix(pyramid): prevent double BeforeTraversal subscriber after config.commit()

krrishapatel:


xrmx:
Duplicated of #4741, not cool to waste reviewers time with duplicated PRs.

Issue #4787, opened 2026-07-13
Implement `db.query.parameter.<key>` across database instrumentations


ocelotl:
## Summary

The [database semantic conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/),
and in particular [`db.query.parameter.<key>`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-query-parameter),
are implemented in very few of our database instrumentations, and where
parameters are captured today it is done with a non-standard attribute.
Prepared statement parameter values are important for end users debugging
queries, and we should support them consistently across the board.

This is a tracking issue. The actual work is split into one sub-issue per
package so each can be reviewed, merged and reverted independently (one PR per
sub-issue). This issue is closed once all sub-issues are closed.

## Background

Grepping the repo for `db.query.parameter` / `DB_QUERY_PARAMETER` returns
nothing — no instrumentation emits the standard attribute. The instrumentations
that capture parameters at all use a custom, non-spec attribute
`db.statement.parameters`, a single stringified blob of the whole parameter
sequence.

Problems with the current behavior:

1. `db.statement.parameters` is not a semantic convention attribute. The
   spec-defined attribute is `db.query.parameter.<key>`, one attribute per
   parameter.
2. Named vs. positional parameters are not distinguished. Named parameters
   should be keyed by name (`db.query.parameter.userName`), positional
   parameters by 0-based index (`db.query.parameter.0`).
3. The legacy blob is emitted in all semconv modes, including the new/stable
   database semconv where it does not belong.
4. Parameters are captured even on batch operations (`executemany`), which the
   spec says they SHOULD NOT be.

## Target behavior

Every in-scope instrumentation should, when parameter capture is opted in:

- Emit one `db.query.parameter.<key>` attribute per parameter — keyed by name
  for named parameters and by 0-based index for positional parameters, values
  captured as strings.
- Keep parameter capture behind an opt-in (values may contain sensitive data).
- Emit `db.query.parameter.<key>` only under the new/stable database semconv.
  Keep `db.statement.parameters` for the old semconv (both under `database/dup`)
  so existing users are not broken during the migration.
- Not capture parameters on batch operations (`executemany`).

The shared `dbapi` base and a small helper in
`opentelemetry.instrumentation._semconv` do the bulk of the work and are reused
by every SQL instrumentation.

## Scope

In scope: SQL / prepared-statement instrumentations. Out of scope: NoSQL /
[...]

PR #4798, opened 2026-07-13
Implement `db.query.parameter.<key>` in the dbapi base

### LINKED ISSUE #4788: `opentelemetry-instrumentation-dbapi`: capture `db.query.parameter.<key>`
Part of #4787 (global `db.query.parameter.<key>` tracking issue).

## Background

`dbapi` is the shared base used by most SQL instrumentations (psycopg, psycopg2,
mysql, mysqlclient, pymysql, pymssql, sqlite3). Today `CursorTracer._populate_span`
captures parameters as a single non-standard `db.statement.parameters` blob when
`capture_parameters=True`, in all semconv modes and even on batch operations.

This sub-issue is foundational: it also adds the shared helper the other
sub-issues reuse.

## What to do

1. Add a shared helper `_set_db_query_parameters(result, parameters, sem_conv_opt_in_mode)`
   in `opentelemetry.instrumentation._semconv` that:
   - only emits under the new database semconv (`_report_new`);
   - keys named parameters (mapping) by name and positional parameters
     (sequence) by 0-based index;
   - treats `str`/`bytes` as a single scalar (does not iterate characters);
   - stringifies each value, capturing falsy values too (`0`, `False`, `""`);
   - uses the `DB_QUERY_PARAMETER_TEMPLATE` constant from the semconv package
     (no hardcoded strings).
2. In `CursorTracer._populate_span`, when `capture_parameters` is enabled:
   - emit `db.statement.parameters` only under the old semconv (`_report_old`);
   - emit `db.query.parameter.<key>` under the new semconv via the helper;
   - skip `db.query.parameter.<key>` for batch operations (`executemany`),
     detected from the wrapped cursor method.

## Acceptance criteria

- Default (old) semconv: `db.statement.parameters` unchanged; no
  `db.query.parameter.*`.
- `database` (stable) semconv: `db.query.parameter.<key>` emitted; no
  `db.statement.parameters`.
- `database/dup` semconv: both present.
- Positional params keyed by index, named params keyed by name, values are
  strings.
- `executemany` never emits `db.query.parameter.*`.

## Tests

- Unit tests for the helper (positional, named, string-value type, scalar
  string, `None`, old-mode no-op).
- Integration tests in `test_dbapi_integration.py` for old / new / dup modes,
  named parameters and the batch-suppression case.

ocelotl:
# Description

Adds a shared `_set_db_query_parameters` helper and wires it into the DB-API `CursorTracer` so that, when `capture_parameters` is enabled, prepared statement parameters are recorded as `db.query.parameter.<key>` (keyed by name for named params, 0-based index for positional). The legacy `db.statement.parameters` blob is now emitted only under the old semconv; the standard attribute is emitted under the new database semconv. Parameters are not captured on batch operations (`executemany`). This is the foundation every DB-API-based instrumentation inherits.

Part of #4787. Fixes #4788.

## Type of change

# How Has This Been Tested?

- [x] Unit tests: dbapi suite 62 passed; `_semconv` suite 67 passed.

- [x] No.

# Checklist:


opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -599,6 +601,32 @@ def _set_db_statement(
...
+    # or bytes value is a single scalar parameter, not a sequence of parameters.
+    if isinstance(parameters, Mapping):
+        items = parameters.items()
+    elif isinstance(parameters, (str, bytes, bytearray)):
+        return
+    elif isinstance(parameters, Sequence):
+        items = enumerate(parameters)
+    else:
+        return
+    for key, value in items:
```
herin049, line 613: Should we maybe perform some validation on `value` here? Perhaps limit to just basic types (e.g. int, float, bool, bytes and str) and maybe maps/lists?

herin049 (approved):
Can we add some Docker tests for this?

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
herin049, line 857: We may want to consider also pulling from `**kwargs`

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
ocelotl: Hmm, not sure, the semconv defines db.query.parameter.<key> as a string representation of the parameter value, so str() here matches the spec. Restricting to a fixed set of basic types would silently drop common DB parameter types like datetime, Decimal and UUID, which are exactly the values you'd want when debugging, so I'd rather keep str(). If unbounded size is the concern we could add a length cap, but that feels like a separate change and the SDK's attribute value length limit alrea

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
ocelotl: Yes, but that's a bit out of scope for this PR, I would rather do this in other PR specially considering that the keyword name is driver-specific.

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -851,6 +862,9 @@ def traced_execution(
             return query_method(*args, **kwargs)
 
         operation_name = self.get_operation_name(cursor, args)
+        # Query parameters must not be captured for batch operations, which are
+        # executed through the cursor's executemany method.
+        is_batch = getattr(query_method, "__name__", None) == "executemany"
```
tammy-baylis-swi, line 865: I think we should also consider [execute_batch](https://www.psycopg.org/docs/extras.html#psycopg2.extras.execute_batch) for psycopg2.

tammy-baylis-swi:
Makes sense to me to expose this with `OPT_IN` for usability even if the `db.statement.parameters` wasn't an old standard.

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -773,7 +776,15 @@ def _populate_span(
             span.set_attribute(attribute_key, attribute_value)
 
         if self._db_api_integration.capture_parameters and len(args) > 1:
-            span.set_attribute("db.statement.parameters", str(args[1]))
+            parameters = args[1]
+            if _report_old(sem_conv_mode):
+                span.set_attribute("db.statement.parameters", str(parameters))
```
tammy-baylis-swi: It might be more uniform to have both the `_report_old` vs `_report_new` checked in `_set_db_query_parameters` like the other semconv utils.

ocelotl:
> Can we add some Docker tests for this?

right, added them

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
ocelotl: Not sure... execute_batch renders each row with `cur.mogrify(sql, args)` and then calls `cur.execute(b";".join(sqls))`. That call has a single positional argument, and our capture is guarded by `len(args) > 1`, so nothing is captured for it

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
ocelotl: Right, fixed :v:

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
tammy-baylis-swi: Ah ok. THanks for checking!

tammy-baylis-swi (approved):
Thanks for revisions! lgtm

PR #4807, opened 2026-07-13
sqlalchemy: add capture_parameters option for db.query.parameter.<key>

### LINKED ISSUE #4797: `opentelemetry-instrumentation-sqlalchemy`: capture `db.query.parameter.<key>`
Part of #4787 (global `db.query.parameter.<key>` tracking issue).

## Background

`sqlalchemy` uses its own `EngineTracer` and does not capture query parameters
at all today, even though it already uses the semconv stability mechanism.
SQLAlchemy exposes the bound parameters of a statement through its event
system (e.g. the `before_cursor_execute` hook receives `parameters` /
`executemany`).

## What to do

- Add a `capture_parameters` option to the sqlalchemy instrumentor (off by
  default, since parameter values may be sensitive).
- When enabled, emit `db.query.parameter.<key>` under the new semconv using the
  shared `_set_db_query_parameters` helper from
  `opentelemetry.instrumentation._semconv`.
  - Key named parameters (dict) by name and positional parameters (sequence) by
    0-based index.
- Do not capture parameters when SQLAlchemy reports the execution as
  `executemany` (batch).
- Follow the migration rule: new attribute under the new semconv only. If a
  legacy `db.statement.parameters` behavior is added for the old semconv, keep
  it behind `_report_old`; otherwise emitting only the new attribute under the
  new semconv is acceptable since there is nothing to preserve.

## Acceptance criteria

- A user can enable parameter capture via `capture_parameters`.
- New semconv: `db.query.parameter.<key>` emitted with the correct keys.
- Batch (`executemany`) executions do not emit `db.query.parameter.*`.
- Values captured as strings.

## Tests

- Cover positional and named parameters, and the batch case.
- Verify exact attribute names and string value types against the spec.
- Test against the oldest and latest supported SQLAlchemy versions.

ocelotl:
# Description

Adds a `capture_parameters` option to the SQLAlchemy instrumentor and records parameters as `db.query.parameter.<key>` under the new database semconv, not on `executemany` batches.

Part of #4787. Fixes #4797.

> Stacked on #4788 (the shared dbapi base + `_set_db_query_parameters` helper). Until #4788 merges, this PR's diff is cumulative and includes those base changes; it will be rebased onto `main` afterwards.

## Type of change

# How Has This Been Tested?

- [x] Unit tests: 64 passed, 7 skipped (real in-memory sqlite).

- [x] No.

# Checklist:


ocelotl:
do not close

ocelotl:
WIP

PR #4833, opened 2026-07-18
opentelemetry-instrumentation-httpx: support strict typing

### LINKED ISSUE #4818: opentelemetry-instrumentor-httpx: support strict typing
This is a followup to https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4730#discussion_r3472283062

We should see if we can improve the typing story here. In service of that, we should add opentelemetry-instrumentation-httpx here...

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/tox.ini#L1186

and do a pass over the instrumentor's typing as a whole.

henry3260:
# Description

Add `opentelemetry-instrumentation-httpx` to the pyright strict typecheck
environment (`tox.ini` typecheck deps + pyright `include` in `pyproject.toml`)
and fix the typing across the instrumentation. Follow-up to the review
discussion in #4730.

Notable changes, none of which are intended to change runtime behavior:

- Restructure the `isinstance` dispatch in `_extract_response` (check `tuple`
  first) so pyright narrowing works soundly in both branches; this removes the
  previously flagged unnecessary `typing.cast` calls.
- Declare the transport `handle_request`/`handle_async_request` methods as
  returning `httpx.Response` to be compatible with the `httpx.BaseTransport`
  override; the httpx < 0.20 tuple form is still handled at runtime.
- Dispatch sync/async hooks in `instrument_client` via `TypeIs` guards
  (pyright does not narrow callable unions in the negative direction, so a
  pair of positive guards is used).
- Type span/metric attribute dicts as `dict[str, AttributeValue]` and make the
  wrapper histogram/hook parameters optional to match call sites.
- Widen `RequestInfo`/`ResponseInfo` `url` and `extensions` field annotations
  to match the raw values the transport layer actually passes.

Remaining suppressions: two `typing.cast` calls at the transport returns
(control-flow invariant pyright cannot see, documented inline) and two
targeted `pyright: ignore` comments for untyped `wrapt` and the deprecated
`MetricInstruments` name required by the old semconv mode — both follow
existing precedent in already-typechecked packages.

Package tests remain excluded from typecheck for now, consistent with the
other packages in the include list.

Fixes #4818

## Type of change

# How Has This Been Tested?

- `tox -e typecheck` — 0 errors for the package
- `tox -e lint-instrumentation-httpx` — 10.00/10
- `tox -e py312-test-instrumentation-httpx-{0,1,2,3}-wrapt2`,
  `py312-test-instrumentation-httpx-1-wrapt1`, and
  `py310-test-instrumentation-httpx-1-wrapt2` — all pass; the `-0` factor
  (httpx 0.18) exercises the legacy tuple code path touched by the
  `_extract_response` restructuring

- [x] No.

# Checklist:

- [ ] Unit tests have been added (typing-only change; covered by the existing test suite)
- [ ] Documentation has been updated (not applicable)

### DIFF
--- a/.changelog/4833.added
+++ b/.changelog/4833.added
@@ -1,0 +1,1 @@
+`opentelemetry-instrumentation-httpx`: support strict typing
--- a/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-httpx/src/opentelemetry/instrumentation/httpx/__init__.py
@@ -438,7 +438,13 @@ if typing.TYPE_CHECKING:
         Request: type[httpx.Request]
         Response: type[httpx.Response]
         Headers: type[httpx.Headers]
-        URL: type[httpx.URL]
+        # Declared as a callable rather than type[httpx.URL] because the
+        # URL constructor of httpx < 0.20.0 additionally accepts the raw
+        # URL tuple its transport API passes around.
+        URL: typing.Callable[
+            [str | httpx.URL | tuple[bytes, bytes, int | None, bytes]],
+            httpx.URL,
+        ]
         BaseTransport: type[httpx.BaseTransport]
         AsyncBaseTransport: type[httpx.AsyncBaseTransport]
         HTTPTransport: type[httpx.HTTPTransport]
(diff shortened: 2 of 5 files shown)


henry3260:
Gentle ping @aabmass 
Could you review this PR when you got a chance? Thanks!

tammy-baylis-swi (approved):
Lgtm; maintainers will also take a look.

emdneto (approved):

PR #4840, opened 2026-07-19
opentelemetry-instrumentation-django: add code attributes to spans

### LINKED ISSUE #1074: Add code information in span based on semantic conventions in Django instrumentation
Not a direct problem, but it would make it easier to trace back information to the source code.

The view function information is already available in the 'process_view' middleware function. 
Adding that information using the `code.namespace` and `code.function` semantic convention attributes would greatly ease debugging traces later on.

Using additional middleware

manduinca:
# Description

Adds `code.function.name`, `code.file.path` and `code.line.number` span attributes to the Django instrumentation. The attributes are set in `process_view`, where the resolved view callable is available.

The issue originally asked for `code.namespace` and `code.function`, but those are deprecated now, so this uses the current stable code attributes from `opentelemetry.semconv.attributes.code_attributes`.

The extraction is best effort and never raises: it handles plain view functions, class based views, `functools.partial` and decorated views. For class based views the line number is omitted on purpose, since classes have no `__code__` and resolving the line would mean parsing source on every request.

Fixes #1074

## Type of change

# How Has This Been Tested?

- [x] New tests in `test_middleware.py` (function view, class based view, partial, decorated view) and `test_middleware_asgi.py` (async view), asserting the exact attribute names and values against the view's own `__code__` metadata
- [x] `tox -e py312-test-instrumentation-django-{0,1,2,3}` (Django 2.x through 5.2), 82 passed on each
- [x] `tox -e lint-instrumentation-django`, ruff 0.14.1 check and format, codespell

Note on the async test: it needs `override_settings(ROOT_URLCONF=...)` because `settings.configure()` from `setUpClass` is not visible to the URL resolver when another test module configured settings first in the same process. The existing ASGI tests don't hit this because both urlconfs mirror the same routes.

- [x] No.

# Checklist:


tammy-baylis-swi:
Thanks for contributing @manduinca ! I [commented on the linked issue](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1074#issuecomment-5320575625) and will copy here for visibility:

> This is an older issue so I just wanted to gauge current use cases.
> 
> Currently the LoggingInstrumenter writes `code.file.path` etc ([semconv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/code/)) which can be opted into with manual instrumentation or `OTEL_PYTHON_LOG_AUTO_INSTRUMENTATION=true`. Is this enough for current users, with consideration for codebase upkeep?

manduinca:
Thanks @tammy-baylis-swi. The logging path answers a different question than this PR.

The `code.*` attributes on log records tell you where a log line was emitted, and they only exist if the app opted into log auto-instrumentation and something actually logged during the request. If the view delegates to a helper and the helper logs, they point at the helper. This PR puts the attributes on the server span, so they identify the view that handled the request even when nothing was logged, which is the common case: you take a slow or failed span in the trace view and jump straight to the handler, or aggregate latency and errors per handler in the backend. `http.route` doesn't fully cover that since routes and views are not 1:1 in Django (one CBV or view function can be wired to several routes).

On upkeep, the change is one helper plus a single call in `process_view` behind `span.is_recording()`, uses the stable semconv constants, and is tested across the four supported Django versions. If the concern is default span size, I can gate it behind an opt-in env var like the logging one.

xrmx:
I understand the issue but I'm not sure the http semconv suggests to use code attributes. Let me ask to the semconv people.

xrmx:
> I understand the issue but I'm not sure the http semconv suggests to use code attributes. Let me ask to the semconv people.

I was able to bring this to the semantic conventions SIG, got some feedback:
- could make sense as opt-in, conformance tests may flag this as violation
- Some java instrumentations may create a span at controller level (as in MVC) and store these code attributes there
- Ok if first span does this, not ok for followup spans as it would like a poor version of profiling

@manduinca Any chance you can take a look at the java instrumentations report for any prior art?

manduinca:
Thanks for taking it to the SIG.

On prior art: yes, there's a fair amount of it. `CodeAttributesExtractor` lives in `instrumentation-api-incubator` under `semconv/code/`, and it sets `code.namespace` and `code.function` as locally-declared keys plus `code.function.name` imported from `io.opentelemetry.semconv.CodeAttributes` — so Java is already emitting the stable attribute this PR uses, alongside the deprecated pair.

It's consumed through `CodeAttributesGetter` by roughly a dozen instrumentations: jaxrs (1.0 and common), jaxws, grails, finatra, jfinal, mybatis, apache-elasticjob, jdbc's `DataSourceCodeAttributesGetter`, `methods`, and the `@WithSpan` annotation ones.

That also covers the controller point: jaxrs, grails and finatra are exactly that case, and they put the attributes on the controller span, which is nested under the HTTP server span.

On the first-span concern, this PR already satisfies it, though by a different route than Java. `process_view` doesn't create a span — it writes onto the server span `process_request` already put in `request.META`. One request means one set of `code.*` attributes, on the root span, with no extra span involved. There's no path here that decorates followup spans.

Opt-in is fine by me. The Django instrumentation already has an `environment_variables.py` with `OTEL_PYTHON_DJANGO_INSTRUMENT`, so I'd add `OTEL_PYTHON_DJANGO_CODE_ATTRIBUTES` there, defaulting to off, and gate the `process_view` block on it. Say if you'd rather have a different name or a package-wide variable instead of a Django-specific one, and I'll push that.

PR #4843, opened 2026-07-20
Add host.id resource detector

### LINKED ISSUE #4830: Add host.id resource attribute detector

I would like the `host.id` resource attribute emitted ([semconv ref](https://opentelemetry.io/docs/specs/semconv/resource/host/)) 

In the java agent we have a [host id resource detector](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/resources/library/src/main/java/io/opentelemetry/instrumentation/resources/HostIdResource.java) that does the machin-id lookup. I would like to add an equivalent resource module in this [project](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/resource) called `opentelemetry-resource-detector-host` that provides the same functionality

the collector can stamp the host.id, but that doesn't help in all deployment types where there might not be a collector in the position to do that.

### Additional Context

_No response_

Yes

### Tip

jaydeluca:
# Description

This PR adds a `host.id` resource detector following [the semconv spec](https://opentelemetry.io/docs/specs/semconv/resource/host/), and using the [java agent implementation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/resources/library/src/main/java/io/opentelemetry/instrumentation/resources/HostIdResource.java) as a reference. 

Fixes #4830

## Type of change

# How Has This Been Tested?

I wrote a small test app,  and added

```python
...
from opentelemetry.resource.detector.host import HostIdResourceDetector
...

    resource = get_aggregated_resources(
        detectors=[HostIdResourceDetector()],
        initial_resource=Resource.create(
            {ResourceAttributes.SERVICE_NAME: "host-detector-manual-test"}
        ),
    )
```

Ran the app (on macOS) and observed the attribute

image

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


resource/opentelemetry-resource-detector-host/tests/test_host.py
```diff
@@ -0,0 +1,173 @@
...
+    "\r\nHKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Cryptography\r\n"
+    "    MachineGuid    REG_SZ    12345678-90ab-cdef-1234-567890abcdef\r\n"
+)
+
+
+def _completed(stdout: str = "", return_code: int = 0) -> SimpleNamespace:
+    return SimpleNamespace(stdout=stdout, stderr="", returncode=return_code)
+
+
+class HostIdResourceDetectorTest(TestBase):
```
herin049: Can we update these tests so that they test as much of the functionality end-to-end as possible? Almost everything is being patched here, which doesn't make these tests that useful. Our CI runs on both Linux and Windows so you should be able to at least test these systems end-to-end.

resource/opentelemetry-resource-detector-hostid/src/opentelemetry/resource/detector/hostid/__init__.py
```diff
@@ -0,0 +1,148 @@
+# Copyright The OpenTelemetry Authors
```
herin049, line 1: Can we add a `py.typed` marker file and add this to the list of packages to type check?

resource/opentelemetry-resource-detector-host/src/opentelemetry/resource/detector/host/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+            # Line looks like: `    MachineGuid    REG_SZ    <value>`
+            parts = line.split()
+            if len(parts) == 3:
+                return parts[2]
+    return None
+
+
+def _get_host_id() -> str | None:
+    system = platform.system()
+    if system == "Linux":
```
herin049: Small nit, but I'd prefer to use a `match` statement here.

resource/opentelemetry-resource-detector-host/src/opentelemetry/resource/detector/host/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+    machine id of the host and returns it in a Resource
+    """
+
+    def detect(self) -> Resource:
+        try:
+            host_id = _get_host_id()
+            resource = Resource.get_empty()
+            if host_id:
+                resource = resource.merge(Resource({HOST_ID: host_id}))
+            return resource
```
herin049:
```suggestion
            if host_id := _get_host_id():
                return Resource({HOST_ID: host_id})
```

resource/opentelemetry-resource-detector-host/src/opentelemetry/resource/detector/host/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+        # pylint: disable=broad-except
+        except Exception as exception:
+            logger.warning(
+                "%s Resource Detection failed silently: %s",
+                self.__class__.__name__,
+                exception,
+            )
+            if self.raise_on_error:
+                raise exception
+            return Resource.get_empty()
```
herin049:
```suggestion
        return Resource.get_empty()
```

herin049 (changes requested):

resource/opentelemetry-resource-detector-host/src/opentelemetry/resource/detector/host/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+)
+
+logger = getLogger(__name__)
+
+# Non-privileged machine-id sources per the semantic conventions:
+# https://opentelemetry.io/docs/specs/semconv/resource/host/#non-privileged-machine-id-lookup
+_LINUX_MACHINE_ID_PATHS = ("/etc/machine-id", "/var/lib/dbus/machine-id")
+_BSD_HOSTID_PATH = "/etc/hostid"
+_BSD_KENV_COMMAND = ("kenv", "-q", "smbios.system.uuid")
+_MACOS_IOREG_COMMAND = ("ioreg", "-rd1", "-c", "IOPlatformExpertDevice")
```
martincostello: You might want to update the implementation for this based on the change proposed in https://github.com/open-telemetry/semantic-conventions/pull/3896.

Similar considerations might be needed for `kenv` and `reg` too.

.github/workflows/generate_workflows.py
```diff
@@ -11,7 +11,14 @@
 tox_ini_path = Path(__file__).parent.parent.parent.joinpath("tox.ini")
 workflows_directory_path = Path(__file__).parent
 
-generate_test_workflow(tox_ini_path, workflows_directory_path, "ubuntu-latest")
+generate_test_workflow(
+    tox_ini_path,
+    workflows_directory_path,
+    "ubuntu-latest",
+    additional_jobs={
+        "py312-test-resource-detector-host-latest": ("windows-latest",),
```
jaydeluca: not sure if there's a better way for me to run this on windows but scope it to just this test suite?

herin049: You shouldn't need to touch these files at all, you should be able to just have pytest skip the platform specific tests if they don't match

herin049: Look at `@pytest.mark.skipif` or `@unittest.skipIf`

jaydeluca: i am using something similar: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4843/changes#diff-d44fe2506f2682e80ab459c5ef5c010e9bef8f4ff0af3965d47be85c97e00595R58-R61

But i dont see any of the jobs using a windows image?

jaydeluca: if you'd prefer, i can remove all of this to keep it simple for now and followup with a change scoped to just introducing the change to run this on a windows runner. the windows specific tests just won't run for now

herin049: Sorry, I was mistaken, we only run Windows on CI for the main `opentelemetry-python` repo. We can think about adding a follow-up for testing on Windows.

resource/opentelemetry-resource-detector-host/src/opentelemetry/resource/detector/host/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+    """Detects the ``host.id`` resource attribute from the non-privileged
+    machine id of the host and returns it in a Resource
+    """
+
+    def detect(self) -> Resource:
+        try:
+            if host_id := _get_host_id():
+                return Resource({HOST_ID: host_id})
+            return Resource.get_empty()
+
```
herin049:
```suggestion
```

resource/opentelemetry-resource-detector-host/src/opentelemetry/resource/detector/host/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+        case "Linux":
+            return _get_linux_machine_id()
+        case "Darwin":
+            return _get_macos_machine_id()
+        case "Windows":
+            return _get_windows_machine_id()
+        case _ if system == "DragonFly" or system.endswith("BSD"):
+            return _get_bsd_machine_id()
+        case _:
+            logger.debug(
```
herin049: I'd maybe switch this log, and some of the others used here from `debug` to `warning`.

resource/opentelemetry-resource-detector-host/tests/test_host.py
```diff
@@ -0,0 +1,254 @@
...
+        attributes = resource.attributes
+        self.assertEqual(dict(attributes), {HOST_ID: expected})
+        self.assertIsInstance(attributes[HOST_ID], str)
+
+    def _assert_real_host_id(self) -> None:
+        resource = HostIdResourceDetector().detect()
+        self.assertIn(HOST_ID, resource.attributes)
+        host_id = resource.attributes[HOST_ID]
+        self.assertIsInstance(host_id, str)
+        self.assertTrue(host_id)
```
herin049: Can we make a stronger assertion here: that this is a 32 character lowercase hexadecimal string, or better yet read the contents of `/etc/machine-id` or `/var/lib/dbus/machine-id`

docs/index.rst
```diff
@@ -53,6 +53,7 @@ install <https://pip.pypa.io/en/stable/reference/pip_install/#editable-installs>
     pip install -e ./instrumentation-genai/opentelemetry-instrumentation-openai-v2
     pip install -e ./sdk-extension/opentelemetry-sdk-extension-aws
     pip install -e ./resource/opentelemetry-resource-detector-containerid
+    pip install -e ./resource/opentelemetry-resource-detector-host
```
emdneto: I think the package name like this can create confusion with the default [HostResourceDetector](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/resources/__init__.py#L492C7-L492C28) in the SDK. Maybe rename it to hostid, host-id?

Maybe I'm being conservative here from the user perspective, but just sharing some thoughts.

jaydeluca: i agree, i'm going to update it, thanks

emdneto (approved):

lmolkova:
@jaydeluca do you know why Java has hostid separate from host detectors? I posted this in https://cloud-native.slack.com/archives/C014L2KCTE3/p1789059294013659. They are part of the same resource and same entity and having another package in python does not seem necessary or right. 
Also for Java, it's the same binary, for python it's literally a new artifact - a lot of toil which would be gone if host.id stayed in the existing detector.

jaydeluca:
I believe the reasoning in java is around the SPI mechanics, and we are already merging them and enabling them by default in declarative configuration. So I think you are right that this is not the right solution for python, I'll look into moving this into the existing detector in the core repo.

PR #4858, opened 2026-07-23
fix(kafka-python): record actual producer partition from send() future

### LINKED ISSUE #4625: Incorrect partition in opentelemetry-instrumentation-kafka-python

bourbonkk:
# Description

Fixes #4625

The kafka-python producer instrumentation recorded `messaging.kafka.partition` from a value it computed *before* `send()`, by calling `instance._partition()` itself. With the default partitioner and no key, `_partition()` selects a partition at random; `send()` then calls `_partition()` again independently and picks a *different* partition, so the span attribute frequently did not match where the message was actually delivered.

This reads the partition back from the `FutureRecordMetadata` returned by `send()` (`_produce_future.topic_partition`), which is the partition `send()` actually assigned. It is accurate for all cases (explicit partition, key-based, and random keyless), and removes the speculative `_wait_on_metadata()` / `_serialize()` / `_partition()` calls the instrumentation previously made. If the attribute can't be read it is omitted rather than recording a wrong value.

The producer span now wraps the `send()` call so the partition can be read from its result before the span ends.

> Note: an earlier revision of this PR took a narrower approach (only record the partition when deterministic). It was replaced with reading the real partition from the future — see commit history.

## Type of change

# How Has This Been Tested?

- Unit tests in `tests/test_utils.py`: partition read from the future (incl. `0`), a check against a real kafka-python `FutureRecordMetadata`, graceful `None` when internals are unavailable, and `_enrich_span` recording/omitting the attribute.
- Verified `_produce_future.topic_partition` exists across the supported range (kafka-python 2.0.3 … 3.0.8).

- [x] No.

# Checklist:


bourbonkk:
Verified against a live Kafka broker (single-node, topic with 12 partitions, 40 messages sent with no key and no explicit partition, so the DefaultPartitioner assigns randomly):

- New approach (this PR): the partition recorded on the span matched the actual delivered partition (read from future.get().partition) for all 40/40 messages.
- Old approach (estimating via _partition() before send()): the recorded partition was wrong for 35/40 messages (~88%), e.g. estimate=9 → actual=3, estimate=0 → actual=8.

Also worth noting: the old code's _wait_on_metadata(topic, max_block_ms / 1000.0) call raises KafkaTimeoutError on kafka-python 2.3.2, since that argument is now milliseconds — so the pre-send estimation path was already broken on newer versions. This PR removes that call entirely.

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
```diff
@@ -126,12 +115,18 @@ def set(self, carrier: textmap.CarrierT, key: str, value: str) -> None:
 
 
 def _enrich_span(
-    span, bootstrap_servers: List[str], topic: str, partition: int
+    span,
+    bootstrap_servers: List[str],
+    topic: str,
+    partition: Optional[int],
```
herin049: nit:
```suggestion
    partition: int | None,
```

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
```diff
@@ -126,12 +115,18 @@ def set(self, carrier: textmap.CarrierT, key: str, value: str) -> None:
 
 
 def _enrich_span(
-    span, bootstrap_servers: List[str], topic: str, partition: int
+    span,
+    bootstrap_servers: List[str],
```
herin049: nit:
```suggestion
    bootstrap_servers: list[str],
```

herin049 (approved):

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
```diff
@@ -56,36 +56,25 @@ def extract_send_headers(args, kwargs):
...
+        partitioner and no key that runs a random choice, which ``send()`` then
+        redoes, so the recorded value frequently did not match where the message
+        actually landed. See
+        https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4625
+        """
         try:
-            topic = KafkaPropertiesExtractor.extract_send_topic(args, kwargs)
-            key = KafkaPropertiesExtractor.extract_send_key(args, kwargs)
-            value = KafkaPropertiesExtractor.extract_send_value(args, kwargs)
-            partition = KafkaPropertiesExtractor._extract_argument(
```
aabmass: Are these functions dead code now?

bourbonkk: Partly. Breaking it down:

- `extract_send_key` / `extract_send_value` — no callers left after this
  change.
- `extract_send_topic` — still used at utils.py:146 for the span name and
  `messaging.destination`.
- `extract_send_headers` — still used at utils.py:141 to inject the
  propagation headers.
- `_extract_argument` — still backs the two above.

I left the two unused ones alone since `KafkaPropertiesExtractor` and its
methods are public names, so dropping them is technically a breaking change,
and I wanted to keep this PR to the partition fix.

Happy to remove them here, or in a separate PR with a CHANGELOG entry —
which would you prefer?

bourbonkk: @aabmass WDYT? I'm happy to go either way.

emdneto: I'm fine with keeping them for now with a # TODO comment to deprecate/remove those functions. @aabmass do you agree? This PR fixes 2 open issues, and I'd like to get this in.

emdneto (approved):

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
xrmx: `extract_send_partition` changed its signature so if anyone is using this on their own we already broke the interface. I would remove the unused helpers at the same time.

xrmx (changes requested):
Can we please have a kafka python test added to the docker tests so we don't only rely on mocks for this? To check that we get something else than an exception for that code path.

instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
bourbonkk: @aabmass @emdneto @xrmx 
Removed Unused Helpers (Added Some Change Log Entries)
Thank you for your feedback.

bourbonkk:
@xrmx 
 Added a kafka docker test that produces to a real broker and asserts the span partition matches the delivered partition from future.get(). Unused helpers are removed as well.

xrmx (approved):

PR #4861, opened 2026-07-23
[`opentelemetry-resourcedetector-gcp`] Fix directory structure and version number

### LINKED ISSUE #4855: opentelemetry-resourcedetector-gcp 1.13.0: unmarked breaking import path rename breaks gcp-trace exporter
## Summary

`opentelemetry-resourcedetector-gcp==1.13.0` (released today via [#4749](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4749) / [#4846](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4846)) changed the public import path:

- **before (≤1.12.0a0):** `opentelemetry.resourcedetector.gcp_resource_detector`
- **after (1.13.0):** `opentelemetry.resource.detector.gcp`

The 1.13.0 changelog only says the package was “moved” and [#4749](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4749) was labeled as a non-breaking change. There is no compatibility re-export of the old module path.

## Impact

Downstream packages that still import the old path break on fresh installs. In particular, `opentelemetry-exporter-gcp-trace` (GoogleCloudPlatform/opentelemetry-operations-python, including `1.12.0`) does:

```python
from opentelemetry.resourcedetector.gcp_resource_detector import ...
```

and depends on `opentelemetry-resourcedetector-gcp ==1.*,>=1.5.0dev0`. Pip resolves that to stable `1.13.0`, which no longer provides `opentelemetry.resourcedetector`, causing:

```text
ModuleNotFoundError: No module named 'opentelemetry.resourcedetector'
```

when importing `opentelemetry.exporter.cloud_trace`.

## Ask

Please consider one of:

1. A compatibility shim / re-export of `opentelemetry.resourcedetector.gcp_resource_detector` (with deprecation) so existing consumers keep working, and/or
2. Explicitly documenting this as a breaking change and coordinating with [GoogleCloudPlatform/opentelemetry-operations-python#546](https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/issues/546) so the exporter is updated or pinned before the new path is the only option.

## Workaround

Pin `opentelemetry-resourcedetector-gcp==1.12.0a0`.

## Related

- https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/issues/546
- https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/issues/384

DylanRussell:
# Description

Moved source package from `src/opentelemetry/resource/detector/gcp/` to `src/opentelemetry/resourcedetector/gcp_resource_detector/`.

Updated package namespace from `opentelemetry.resource.detector.gcp` to `opentelemetry.resourcedetector.gcp_resource_detector`.

This matches the directory structure and namespace in https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/tree/main/opentelemetry-resourcedetector-gcp , which we want to do to avoid breaking people..

I didn't mean to change the directory structure in the first place. My agent did it and I didn't notice..

If it's important to match the directory structure used in this repo we could also leave it as it is now (revert this change), but add a compatibility shim (re-import `GoogleCloudResourceDetector`) under the existing path: `src/opentelemetry/resourcedetector/gcp_resource_detector/` since `GoogleCloudResourceDetector` is the only public object in this package.

Let me know what y'all think...

Fixes  https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4855

# How Has This Been Tested?

Unit tests

- [x] No.

# Checklist:


tammy-baylis-swi:
Good to change it back in this PR. Can discuss namespacing and major release later

resource/opentelemetry-resourcedetector-gcp/.changelog/4861.changed
```diff
@@ -0,0 +1 @@
+Update package namespace to ``opentelemetry.resourcedetector.gcp_resource_detector``.
```
xrmx:
```suggestion
Update package namespace to `opentelemetry.resourcedetector.gcp_resource_detector`.
```

This is markdown 😅

xrmx (commented):
Given that people will need to install a different package in order to use this and that people can still install the old package along with this I'm not sure using the old namespace to avoid breakage is such a strong argument. Said that I've no strong opinion.

DylanRussell:
We migrated the package name over so that they don't need to install a different package -- we didn't make any name change to the package name

DylanRussell:
If we were to change the top level folder to match the existing pattern (`resource-detector` instead of `resourcedetector`) that would alter the PyPi package distribution name.. So we could consider doing that, and maybe at the same time change the directory structure to match the existing pattern...

But lets consider that separately.. for now lets just merge this and do a new release of the existing package since people are currently having issues (https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4855) after my yanked release

resource/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/version.py
```diff
@@ -1,4 +1,4 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-__version__ = "1.14.0a0.dev"
```
aabmass, line 4: Do we want to change the version number back? We can always revisit it and make another release.

Given we might want to change the package name anyway, i think it makes sense.

aabmass (approved):

xrmx:
> We migrated the package name over so that they don't need to install a different package -- we didn't make any name change to the package name

Ah forgot that sorry, sure then it's fine for me.

xrmx (approved):

resource/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/version.py
DylanRussell: Sadly now that we did the 1 non-pre-release version `uv` is failing to install the trace exporter unless the `pre` flag is set, so I think we have to leave it like this to fix this

PR #4872, opened 2026-07-27
fix(sqlalchemy): safely handle garbage-collected target during event listener removal

### LINKED ISSUE #4817: Flaky sqlalchemy tests on Pypy
```
________ TestSqlalchemyInstrumentation.test_create_async_engine_wrapper ________

self = <tests.test_sqlalchemy.TestSqlalchemyInstrumentation testMethod=test_create_async_engine_wrapper>

    def tearDown(self):
        super().tearDown()
>       SQLAlchemyInstrumentor().uninstrument()

instrumentation/opentelemetry-instrumentation-sqlalchemy/tests/test_sqlalchemy.py:60: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
opentelemetry-instrumentation/src/opentelemetry/instrumentation/instrumentor.py:128: in uninstrument
    result = self._uninstrument(**kwargs)
instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py:304: in _uninstrument
    EngineTracer.remove_all_event_listeners()
instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py:319: in remove_all_event_listeners
    remove(weak_ref_target(), identifier, func)
.tox/pypy3-test-instrumentation-sqlalchemy-1/lib/pypy3.10/site-packages/sqlalchemy/event/api.py:213: in remove
    _event_key(target, identifier, fn).remove()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

target = None, identifier = 'checkout'
fn = <bound method EngineTracer._pool_checkout of <opentelemetry.instrumentation.sqlalchemy.engine.EngineTracer object at 0x000000000a8c4598>>

    def _event_key(target, identifier, fn):
        for evt_cls in _registrars[identifier]:
            tgt = evt_cls._accept_with(target)
            if tgt is not None:
                return _EventKey(target, identifier, fn, tgt)
        else:
>           raise exc.InvalidRequestError(
                "No such event '%s' for target '%s'" % (identifier, target)
            )
E           sqlalchemy.exc.InvalidRequestError: No such event 'checkout' for target 'None'

.tox/pypy3-test-instrumentation-sqlalchemy-1/lib/pypy3.10/site-packages/sqlalchemy/event/api.py:29: InvalidRequestError
_ TestSqlalchemyInstrumentation.test_create_async_engine_wrapper_enable_commenter _

self = <tests.test_sqlalchemy.TestSqlalchemyInstrumentation testMethod=test_create_async_engine_wrapper_enable_commenter>

    @pytest.mark.skipif(
        not sqlalchemy.__version__.startswith("1.4"),
        reason="only run async tests for 1.4",
    )
    def test_create_async_engine_wrapper_enable_commenter(self):
        async def run():
            logging.getLogger("sqlalchemy.engine").setLevel(logging.INFO)
            SQLAlchemyInstrumentor().instrument(
                enable_commenter=True,
                commenter_options={
                    "db_framework": False,
                },
            )
            from sqlalchemy.ext.asyncio import (  # pylint: disable-all  # noqa: PLC0415
                create_async_engine,
            )
    
            engine = create_async_engine("sqlite+aiosqlite:///:memory:")
            async with engine.connect() as cnx:
                await cnx.execute(text("SELECT  1;"))
            # sqlcommenter
            self.assertRegex(
                self.caplog.records[1].getMessage(),
                r"SELECT  1 /\*db_driver='(.*)',traceparent='\d{1,2}-[a-zA-Z0-9_]{32}-[a-zA-Z0-9_]{16}-\d{1,2}'\*/;",
            )
            # first span is connection to db
            self.assertEqual(spans[0].name, "connect")
            # second span is query itself
            query_span = spans[1]
            self.assertEqual(
                query_span.attributes[DB_STATEMENT],
                "SELECT  1;",
            )
    
>       asyncio.get_event_loop().run_until_complete(run())

instrumentation/opentelemetry-instrumentation-sqlalchemy/tests/test_sqlalchemy.py:442: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/PyPy/3.10.16/x64/lib/pypy3.10/asyncio/base_events.py:649: in run_until_complete
    return future.result()
/opt/hostedtoolcache/PyPy/3.10.16/x64/lib/pypy3.10/asyncio/futures.py:201: in result
    raise self._exception.with_traceback(self._exception_tb)
/opt/hostedtoolcache/PyPy/3.10.16/x64/lib/pypy3.10/asyncio/tasks.py:232: in __step
    result = coro.send(None)
instrumentation/opentelemetry-instrumentation-sqlalchemy/tests/test_sqlalchemy.py:427: in run
    self.assertRegex(
E   AssertionError: Regex didn't match: "SELECT  1 /\\*db_driver='(.*)',traceparent='\\d{1,2}-[a-zA-Z0-9_]{32}-[a-zA-Z0-9_]{16}-\\d{1,2}'\\*/;" not found in 'BEGIN (implicit)'
------------------------------ Captured log call -------------------------------
WARNING  opentelemetry.instrumentation.instrumentor:instrumentor.py:92 Attempting to instrument while already instrumented
INFO     sqlalchemy.engine.Engine:base.py:1032 BEGIN (implicit)
INFO     sqlalchemy.engine.Engine:base.py:1863 SELECT  1;
INFO     sqlalchemy.engine.Engine:base.py:1868 [generated in 0.00030s] ()
INFO     sqlalchemy.engine.Engine:base.py:1060 ROLLBACK
=========================== short test summary info ============================
FAILED instrumentation/opentelemetry-instrumentation-sqlalchemy/tests/test_sqlalchemy.py::TestSqlalchemyInstrumentation::test_create_async_engine_wrapper
FAILED instrumentation/opentelemetry-instrumentation-sqlalchemy/tests/test_sqlalchemy.py::TestSqlalchemyInstrumentation::test_create_async_engine_wrapper_enable_commenter
========================= 2 failed, 64 passed in 2.67s =========================
```

aditya-cyberverse:
Fixes #4817

### Description
During test teardown or `uninstrument()`, event listener cleanup in `EngineTracer.remove_all_event_listeners` can fail if the weakref target resolves to `None` or raises `InvalidRequestError` when SQLAlchemy tries to unregister the event.

### Changes
- Safely resolves the weakref target and checks for `None` before removal.
- Catches `InvalidRequestError` during `event.remove()` to prevent flaky teardown crashes.


chuenchen309:
This has been sitting without review for ten days, and the test workflows here are still waiting on first-contributor approval — so nothing automated has actually exercised the change yet. I ran it locally, and I think there's a blocker worth catching before a maintainer looks at it.

`event.remove(target, identifier, fn)` raises `NameError`: the name `event` is never bound in this module. The existing import brings in the functions directly:

```python
from sqlalchemy.event import (  # pylint: disable=no-name-in-module
    listen,
    remove,
)
```

Instrumenting, opening one connection, then calling `uninstrument()` on this branch (`dbef557`, SQLAlchemy 2.0.51):

```
listeners recorded: 7
--- uninstrument() ---
  File ".../opentelemetry/instrumentation/sqlalchemy/__init__.py", line 304, in _uninstrument
    EngineTracer.remove_all_event_listeners()
  File ".../opentelemetry/instrumentation/sqlalchemy/engine.py", line 317, in remove_all_event_listeners
    event.remove(target, identifier, fn)
    ^^^^^
NameError: name 'event' is not defined
```

Same script, same venv, with only `engine.py` swapped for the one on `main`:

```
listeners recorded: 7
--- uninstrument() ---
✅ no exception
```

So it fires on every `uninstrument()` where any weakref is still alive, which turns the intermittent teardown failure in #4817 into a deterministic one. Two ways out: call `remove(...)`, which is already imported, or add `from sqlalchemy import event` — with the second, the imported `remove` becomes unused (`grep -nE "(^|[^.\w])remove\("` has no hits on the branch), and #4926 is currently rolling ruff's default rules across the repo, so that would get flagged.

Two smaller things while you're in there:

- `from sqlalchemy.exc import InvalidRequestError` landed between `import weakref` and `import sqlalchemy`, i.e. in the stdlib block; the other third-party imports are in the group below. Same ruff/isort caveat as above.
- Swallowing `InvalidRequestError` is a reasonable call in teardown, but it also covers cases other than "this listener was never registered". A one-line comment saying why it's safe here would save the next reader from having to work out whether it was deliberate.

The other half of the change looks right to me — `target = weak_ref_target()` with an early `continue` is equivalent to the old `if weak_ref_target() is not None`, minus a second weakref call, so the only behavioural change is the try/except.

*AI-assisted; the traceback and the control run above are output from actually running it, not inferred.*

Issue #4871, opened 2026-07-25
kafka-python instrumentation: span lifetime does not fully encompass the instrumented operation, making synchronous failures invisible


SurbhiD404:
## Summary

While investigating the kafka-python instrumentation, I found two cases where
the span lifetime does not fully encompass the operation being instrumented.
Although they affect different code paths (producer and consumer), both stem
from the same structural pattern: the instrumented operation executes outside
the lifetime of the span intended to represent it. This causes synchronous
failures to be missing from tracing.

I verified this against current `main` using the real OpenTelemetry SDK,
wrote regression tests reproducing the behavior, compared the implementation
with sibling Kafka instrumentations, and reviewed git history to determine
whether this was intentional or introduced later.

---

### Current behavior

In `utils.py` (`_wrap_send`), the producer span is created, propagation
headers are injected, and the span exits **before** the wrapped `send()`
operation is executed.

```python
with tracer.start_as_current_span(...):
    propagate.inject(...)
    ...

return func(*args, **kwargs)
```

If the underlying `send()` raises synchronously (for example,
`ConnectionError` or a serialization error), the exception occurs after the
producer span has already ended.

### Reproducer

Using the real OpenTelemetry SDK:

```text
ConnectionError propagated.
Exported spans: 1
StatusCode.UNSET
events=[]
```

The operation failed, but the exported producer span completed successfully
(`StatusCode.UNSET`) and contains no recorded exception.

### Proposed behavior

The producer span should remain active while the synchronous `send()` call
executes so that failures are recorded on the span.

As a local experiment, I temporarily moved:

```python
return func(*args, **kwargs)
```

inside the span's `with` block.

Without any other changes, the regression test changed from failing to
passing, suggesting that the current span lifetime is responsible for the
observed behavior.

---

### Current behavior

In `utils.py` (`_wrap_next`), the wrapped receive operation executes before
any consumer span is created.

```python
record = func(*args, **kwargs)

if record:
    _create_consumer_span(record)

return record
```

If the wrapped receive call raises, execution exits immediately and no
consumer span is created.

### Reproducer

Using the real OpenTelemetry SDK:

```text
ValueError propagated.
Exported spans: 0
```

No receive span is created for the failed operation.

### Question for maintainers

Is this the intended behavior for exceptions raised during receive?

Receive/poll spans represent the receive operation itself, and OpenTelemetry's
general guidance recommends recording exceptions from instrumented
[...]

PR #4875, opened 2026-07-28
fix(kafka-python): keep producer spans around send

tomatotomata:
## Summary\n- execute the synchronous producer send() call inside the producer span\n- ensure serialization and connection failures are recorded by the span\n- add a regression test that verifies operation/span lifetime ordering\n\nFixes #4871\n\nValidation: python -m py_compile instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py instrumentation/opentelemetry-instrumentation-kafka-python/tests/test_utils.py, git diff --check\n\nThe full targeted pytest could not run locally because the repository's OpenTelemetry package dependencies are not installed in this environment.\n\n


emdneto:
Closing as spam.

PR #4878, opened 2026-07-28
fix(django): finish streaming spans on response close

### LINKED ISSUE #4681: Django Instrumentation: Spans close prematurely for StreamingHttpResponse

tomatotomata:
## What does this PR do?

Fixes #4681 by keeping Django server spans active for `StreamingHttpResponse` objects until Django invokes the response's guaranteed `close()` lifecycle hook.

The existing response status/header instrumentation and response hook still run in `process_response`; duration metrics, active-request accounting, context detachment, and span finalization are deferred only for streaming responses. The close wrapper is idempotent and finalizes telemetry even when the original close method raises.

## Checklist

- [x] Added regression coverage for deferred span completion and idempotent close handling.
- [x] Ran the Django middleware test suite.
- [x] Ran `git diff --check` and Python compilation.

Test result: `42 passed` for `instrumentation/opentelemetry-instrumentation-django/tests/test_middleware.py`.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
+++ b/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware/otel_middleware.py
@@ -324,6 +324,8 @@ class _DjangoMiddleware:
         if not _is_asgi_supported and is_asgi_request:
             return response
 
+        is_streaming_response = bool(getattr(response, "streaming", False))
+
         activation = request.META.pop(self._environ_activation_key, None)
         span = request.META.pop(self._environ_span_key, None)
         active_requests_count_attrs = request.META.pop(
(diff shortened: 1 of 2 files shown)


emdneto:
Closing as spam.

PR #4881, opened 2026-07-28
test(grpc): use latest and oldest dependency sets

### LINKED ISSUE #4201: grpc-instrumentation: simplify tests-requirements here to use latest-oldest testing pattern
> grpc-instrumentation: simplify tests-requirements here to use latest-oldest testing pattern 

 _Originally posted by @emdneto in [#4193](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4193/changes#r2804447832)_

tomatotomata:
## Description

Simplify the gRPC instrumentation test matrix to the repository's standard latest/oldest dependency pattern.

- Replace the six duplicated grpcio/wrapt requirement files with one `test-requirements.in` and uv-generated `latest` / `oldest` lock files.
- Exercise Python 3.10?3.14 against both dependency sets.
- Keep the supported wrapt range in the input so the oldest resolution covers the lower bound rather than maintaining separate wrapt environments.

Fixes #4201

## Validation

- `uv pip compile` completed for both latest and lowest-direct requirements.
- `git diff --check` passes.
- Ruff/format checks are not applicable to the requirement files or tox-only changes.
- The local Windows run is environment-limited: 127 tests passed, while Unix-domain-socket tests cannot bind on this host and the installed grpc/protobuf combination causes additional pre-existing compatibility failures. CI's Linux matrix will exercise the generated environments.


emdneto:
Closing as spam.

PR #4883, opened 2026-07-28
Remove `packaging` as a runtime dependency of opentelemetry-instrumentation

### LINKED ISSUE #4882: Remove `packaging` as a runtime dependency of `opentelemetry-instrumentation`
# Remove `packaging` as a runtime dependency of `opentelemetry-instrumentation`

## Summary

`opentelemetry-instrumentation` declares [`packaging`](https://pypi.org/project/packaging/)
(`packaging >= 18.0`) as a **runtime** dependency. It is used only for two small,
well-defined tasks — comparing PEP 440 version strings and parsing PEP 508
requirement specifiers — both of which can be satisfied by a tiny internal
implementation, letting us drop the external dependency.

This issue is scoped to the core `opentelemetry-instrumentation` package. The
same removal for the four instrumentations that also depend on `packaging` is
tracked separately and depends on this one (they will reuse the internal
implementation added here):

- #4884 — `opentelemetry-instrumentation-falcon`
- #4885 — `opentelemetry-instrumentation-flask`
- #4886 — `opentelemetry-instrumentation-pika`
- #4887 — `opentelemetry-instrumentation-sqlalchemy`

## Motivation: the auto-instrumentation injector

The main driver is the **OpenTelemetry auto-instrumentation injector** (the Python
auto-instrumentation shipped by the OpenTelemetry Operator). The injector bundles
the instrumentation packages **and all of their runtime dependencies** into an
init-container image, copies that directory into a volume shared with the target
application's container, and prepends it to the application's `PYTHONPATH`.

As a result, every runtime dependency is injected **into the user's application
process**, alongside the application's own dependencies:

- **Version conflicts / shadowing.** `packaging` is an extremely common library
  that many applications already import. Injecting our copy onto `PYTHONPATH`
  risks shadowing the version the application expects, or being shadowed — the
  class of hard-to-debug failures the injector is designed to avoid.
- **Payload size.** Everything shipped must be baked into the init-container
  image and copied at pod startup.
- **Blast radius.** Fewer third-party packages forced into a customer's runtime
  means a smaller surface for incompatibilities.

Because `packaging` is used only for a narrow, stable subset of functionality,
the cost of carrying it as an injected runtime dependency outweighs the cost of a
small internal implementation.

## Where it is used in `opentelemetry-instrumentation`

- `pyproject.toml`: `packaging >= 18.0` in `dependencies`.

### PEP 440 version comparison (`packaging.version`)

- `src/opentelemetry/instrumentation/_semconv.py:745` — compares semantic
  convention **schema version** strings (`Version(a) > Version(b)`) to select the
  highest schema URL across signal types.

### PEP 508 requirement parsing / PEP 440 specifier matching (`packaging.requirements`, and transitively `packaging.specifiers` / `packaging.markers`)

- `src/opentelemetry/instrumentation/dependencies.py:9` — the dependency-conflict
  machinery (`get_dist_dependency_conflicts`, `get_dependency_conflicts`) parses
  each instrumentation's declared `instruments` / `instruments-any` dependency
  strings into `Requirement` objects, evaluates their environment markers
  (`extra == "instruments"`), and checks the installed distribution version
  against the requirement's specifier (`req.specifier.contains(dist_version)`).
- `src/opentelemetry/instrumentation/bootstrap.py:90` — `Requirement(req)` when
  computing which instrumentation packages to install/print.

## Proposed resolution

Add a small internal implementation under
`opentelemetry.instrumentation._packaging` covering just the required subset:

- PEP 440 `Version` parsing, normalization and ordering.
- PEP 508 `Requirement` parsing plus PEP 440 specifier containment and marker
  evaluation.

Point `_semconv.py`, `dependencies.py` and `bootstrap.py` at it, and remove
`packaging` from `opentelemetry-instrumentation`'s `dependencies`. The internal
implementation must match `packaging`'s behavior on the inputs these sites see,
verified against `packaging` across a broad version/specifier corpus.

ocelotl:
# Description

Removes `packaging` as a runtime dependency of **`opentelemetry-instrumentation`**,
replacing its use with a small internal implementation under
`opentelemetry.instrumentation._packaging`:

- PEP 440 `Version` (parsing, normalization, ordering) for the schema-version
  comparison in `_semconv.py`.
- PEP 508 `Requirement` parsing plus PEP 440 specifier containment and marker
  evaluation for the dependency-conflict machinery in `dependencies.py` and
  `bootstrap.py`.

`packaging` is dropped from `opentelemetry-instrumentation`'s `dependencies`.

Motivation: the auto-instrumentation injector ships instrumentation and its
runtime dependencies onto the target application's `PYTHONPATH`, so every runtime
dependency (like `packaging`) is injected into the user's process and risks
version shadowing/conflicts. See the issue for details.

The same removal for the four instrumentations that depend on `packaging`
(falcon #4884, flask #4885, pika #4886, sqlalchemy #4887) is handled in separate
PRs that reuse the internal implementation introduced here.

Fixes #4882

## Type of change

# How Has This Been Tested?

- New unit tests for the internal version, specifiers, requirements and markers
  modules (`tests/test_packaging_*.py`).
- The internal implementation was validated against `packaging` across a broad
  corpus of PEP 440 versions and specifiers (parsing, normalization, attributes
  and ordering) with no divergence.
- Existing `test_dependencies.py` continues to pass against the internal
  implementation.

- [x] No.

# Checklist:

### DIFF
--- a/.changelog/4883.removed
+++ b/.changelog/4883.removed
@@ -1,0 +1,1 @@
+`opentelemetry-instrumentation`: remove the runtime `packaging` dependency in favor of an internal PEP 440/508 implementation
--- a/opentelemetry-instrumentation/pyproject.toml
+++ b/opentelemetry-instrumentation/pyproject.toml
@@ -28,7 +28,6 @@ dependencies = [
   "opentelemetry-api ~= 1.4",
   "opentelemetry-semantic-conventions == 0.66b0.dev",
   "wrapt >= 1.0.0, < 3.0.0",
-  "packaging >= 18.0",
 ]
 
 [project.scripts]
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_packaging/__init__.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_packaging/__init__.py
@@ -1,0 +1,12 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Internal, dependency-free replacements for the subset of ``packaging`` used
+by OpenTelemetry instrumentation.
+
+This package exists so that ``opentelemetry-instrumentation`` and the
+instrumentations that build on it do not need ``packaging`` as a runtime
+dependency. Only the behaviour actually relied on by this repository is
+implemented, following PEP 440 (versions and specifiers) and PEP 508
+(requirements and environment markers).
+"""
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_packaging/markers.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_packaging/markers.py
@@ -1,0 +1,368 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""A minimal PEP 508 environment marker implementation.
+
+This provides the subset of :mod:`packaging.markers` that OpenTelemetry
+instrumentation relies on: parsing a marker expression such as
+``extra == "instruments"`` (optionally combined with ``and``/``or`` and other
+environment markers) and evaluating it against an environment.
+"""
+
+from __future__ import annotations
+
+from operator import eq, ne
+from os import name as os_name
+from platform import (
+    machine,
+    python_implementation,
+    python_version,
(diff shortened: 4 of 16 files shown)


ocelotl:
still working on this one

PR #4892, opened 2026-07-29
docs: fill FastAPI instrumentation setup gaps

### LINKED ISSUE #3301: Doc holes in FastAPI instrumentation
Mentioning FastAPI specifically as that's what I'm using but I wouldn't be surprised if it held true for the others. So coming to add opentelemetry instrumentation as someone primarily not using python but using other language ecosystems I don't want to rely on assumptions on what's provided as not every language is equally fleshed out in opentelemetry IME.

So in these docs https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/fastapi/fastapi.html#usage the missing information for me is:

1. What environment variables can set things up? Do I have ones like `OTEL_EXPORTER_OTLP_ENDPOINT` `OTEL_EXPORTER_OLTP_PROTOCOL` `OTEL_SERVICE_NAME` or does this have to be setup separately
2. Does trace propagation work or do I need manual intervention
3. If it does work does it also work for websocket requests?
4. How do I integrate logs so they show up as events
5. There should be a link to the opentelemetry-instrumentation package I had to search pypi for it because it's not a very google friendly name (a lot of things related to general python, languages I use more often and opentelemetry in general)

tomatotomata:
Closes #3301

## Summary
- document `opentelemetry-instrument` setup and link the launcher package
- describe standard environment variables and automatic HTTP propagation
- clarify WebSocket route coverage and the distinction between request spans and message-level events
- document logging correlation as a separate signal with a concrete setup example
- add a changelog fragment

## Validation
- `git diff --check`
- reviewed the FastAPI ASGI middleware and existing WebSocket integration tests to keep the documentation aligned with implementation
- the package test suite cannot collect in this Windows checkout because the editable instrumentation package is not installed (`ModuleNotFoundError`)
- RST parser tooling is unavailable (`docutils` is not installed)

### DIFF
--- a/.changelog/3301.changed
+++ b/.changelog/3301.changed
@@ -1,0 +1,3 @@
+`opentelemetry-instrumentation-fastapi`: document automatic setup, standard
+OpenTelemetry environment variables, propagation, WebSocket coverage, and
+logging correlation.
--- a/instrumentation/opentelemetry-instrumentation-fastapi/README.rst
+++ b/instrumentation/opentelemetry-instrumentation-fastapi/README.rst
@@ -10,7 +10,9 @@ OpenTelemetry FastAPI Instrumentation
 This library provides automatic and manual instrumentation of FastAPI web frameworks,
 instrumenting http requests served by applications utilizing the framework.
 
-auto-instrumentation using the opentelemetry-instrumentation package is also supported.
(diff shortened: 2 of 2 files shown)


emdneto:
Closing as spam.

Issue #4639, opened 2026-05-30
Provide Documentation to Set Up Shared `-contrib` and `sdk` Workspace


SoulofAkuma:
### What problem do you want to solve?

I've recently investigated #3895 which potentially spans both the `opentelemetry-python-contrib` and the `opentelemetry-python` repositories. Both of these repositories have great configuration to set up `uv` workspaces for everything that is contained within the respective repository, but there is no documentation on how to setup a shared environment for both repositories to troubleshoot or develop across the boundary between the `contrib` and non-`contrib` repo.

### Describe the solution you'd like

I'd like to add documentation and potentially configuration to this repository on how to setup local workspaces that include local versions of both `opentelemetry-python-contrib` and `opentelemetry-python`, instead of having separate workspaces that, in case of the `opentelemetry-python-contrib` repo, always install `opentelemetry-python` as a packaged dependency instead of utilizing a locally cloned version.

### Would you like to implement a fix?

Yes

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #4895, opened 2026-07-29
docs: explain shared core and contrib workspace

tomatotomata:
## Summary\n\n- document an issue-backed workflow for testing contrib against a sibling local opentelemetry-python checkout\n- install the core packages editable after the normal uv sync so changes are immediately visible\n- include PowerShell path and verification guidance\n\nCloses #4639\n\n## Validation\n\n- git diff --check\n\nThe uv CLI was not available in this environment, so the installation commands were not executed locally.


emdneto:
Closing as spam.

PR #4901, opened 2026-07-30
Fix asyncio `to_thread` instrumentation to measure actual function execution

### LINKED ISSUE #4900: opentelemetry-instrumentation-asyncio: to_thread instrumentation does not measure function execution (span/metric duration is always ~0, state is always "finished")
### Describe your environment

OS: any (reproduced on macOS 15 / Darwin 25.5.0) 
Python version: Python 3.12 (applies to all supported versions) 
Package version: 0.66b0.dev (current main), present since the instrumentation was introduced

### What happened?

`AsyncioInstrumentor.trace_to_thread` is supposed to wrap the function passed to
`asyncio.to_thread()` so that its execution is traced and measured. However, it
returns the original function unmodified and records the span and metrics at
*wrap time* instead of around the actual execution:

```python
def trace_to_thread(self, func: callable):
    ...
    start = default_timer()
    ...
    span = self._tracer.start_span(...) if ... else None
    attr = {"type": "to_thread", "name": func_name}
    exception = None
    try:
        attr["state"] = "finished"
        return func          # <-- returns func without wrapping it
    except Exception:        # <-- can never fire; `return func` doesn't call func
        attr["state"] = "exception"
        raise
    finally:
        self.record_process(start, attr, span, exception)  # <-- records immediately
```

Consequences:

1. The `asyncio to_thread-<name>` span ends before the function even starts
   running in the worker thread, so its duration is always a few microseconds.
2. The `asyncio.process.duration` histogram records the wrapping overhead, not
   the function execution time.
3. `state` is always `"finished"` — exceptions raised by the function in the
   worker thread are never observed, so the `exception` state and
   `span.record_exception` are dead code for `to_thread`.

The existing tests (`tests/test_asyncio_to_thread.py`) only assert span names
and metric attributes, not durations or the error path, which is why this was
never caught.

### Steps to Reproduce

```python
import asyncio, os, time

os.environ["OTEL_PYTHON_ASYNCIO_TO_THREAD_FUNCTION_NAMES_TO_TRACE"] = "slow_func"

from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
from opentelemetry.instrumentation.asyncio import AsyncioInstrumentor

exporter = InMemorySpanExporter()
tp = TracerProvider()
tp.add_span_processor(SimpleSpanProcessor(exporter))
trace.set_tracer_provider(tp)

AsyncioInstrumentor().instrument()

def slow_func():
    time.sleep(0.5)

async def main():
    await asyncio.to_thread(slow_func)

asyncio.run(main())

span = exporter.get_finished_spans()[0]
print(span.name, (span.end_time - span.start_time) / 1e9)
```

### Expected Result

The span (and the `asyncio.process.duration` histogram) should cover the
function execution, i.e. a duration of roughly 0.5s:

```
asyncio to_thread-slow_func 0.5001...
```

If the function raises, the metric should be recorded with
`state = "exception"` and the span should record the exception with an ERROR
status.

### Actual Result

```
asyncio to_thread-slow_func 0.000068
```

The span duration is ~68µs and the histogram records ~35µs. Exceptions from the
function are never reflected in the span or metrics.

### Additional context

The fix seems straightforward: instead of returning `func` directly,
`trace_to_thread` should return a wrapper (e.g. via `functools.wraps`) that
starts the timer/span, calls `func(*args, **kwargs)` inside try/except/finally,
and records the metrics and span around the actual call. That also makes the
existing `exception` handling reachable. I'm happy to submit a PR if the
direction sounds right.

Yes

### Tip

bourbonkk:
# Description

`trace_to_thread` returned the target function unwrapped and recorded the span and metrics at wrap time: the `asyncio to_thread-<name>` span ended before the function started running in the worker thread, `asyncio.process.duration` recorded ~0, and exceptions raised by the function were never observed (`state` was always `"finished"`).

This change makes `trace_to_thread` return a `functools.wraps` wrapper that starts the timer/span and records metrics around the actual call, making the existing exception handling reachable.

It also drops the `_is_instrumented` guard in this path: the wrapper is created per `to_thread()` call so there is no re-wrapping risk, and marking the user's function meant every call after the first was not instrumented at all.

Fixes #4900

## Type of change

# How Has This Been Tested?

- New unit tests: span/histogram duration covers a `time.sleep(0.1)` execution, exception path (span ERROR status + exception event, metric `state="exception"`), repeated calls each instrumented.
- Full `instrumentation-asyncio` test suite passes on Python 3.10, 3.12, and 3.14.

- [x] No.

# Checklist:


chuenchen309:
Ran this against `main` (`7704aab5`) and this branch (`1b81303f`) with the same probe script, tracing
`slow,multiply,boom,timeouty,cancelly,parented` via
`OTEL_PYTHON_ASYNCIO_TO_THREAD_FUNCTION_NAMES_TO_TRACE`. The diagnosis holds and the fix does what it
says:

| probe | `main` | this branch |
|---|---|---|
| `slow()` sleeps 100 ms | span 0.3 ms, `duration` sum 0.000 s | span **100.5 ms**, sum **0.100 s** |
| `multiply()` twice, same function object | **1** span, `created` = 1 | **2** spans, `created` = **2** |
| worker raises `ValueError` | status `UNSET`, no events, `state=finished` | status **`ERROR`**, event `exception`, `state=exception` |
| `to_thread` inside an active span | child of `outer` | child of `outer` (unchanged) |

The second row is the one I'd call out in the changelog as well. `_is_instrumented()` isn't a
predicate — it *marks* on first call (`instrumentation_state.py`, `_instrumented_tasks.add(obj)` on
the miss path), so on `main` the first `to_thread(multiply, …)` registered `multiply` itself and every
later call returned early. Dropping the guard is right, and it isn't reachable from anywhere else:
`trace_to_thread` is only called from `wrap_to_thread`, and the wrapper it now returns is never handed
back to user code, so there's nothing to re-wrap.

I also checked the thing `functools.wraps` usually breaks, since the wrapped object here is whatever
the caller passed. Plain function, `functools.partial`, callable instance, bound method, `lambda`, and
a builtin (`len`) all return the right result on both `main` and this branch — no change. And the
parent row above holds because `asyncio.to_thread` does `ctx = contextvars.copy_context()` before
handing off, so creating the span inside the worker still sees the caller's context.

Two things I think are worth changing, both in the same three lines.

**1. `state` no longer agrees with the other two paths.** `determine_state()` maps `CancelledError` →
`cancelled` and `TimeoutError` → `timeout`, and both `trace_coroutine` and `trace_future` use it. The
new `to_thread` wrapper hardcodes `"exception"` instead. Same exception, two answers:

```
worker raises TimeoutError   -> {'type': 'to_thread', 'name': 'timeouty',    'state': 'exception'}
coroutine raises TimeoutError -> {'type': 'coroutine', 'name': 'coro_timeout', 'state': 'timeout'}
```

`TimeoutError` is close to the defining payload for `to_thread` — a blocking socket or file call is
what gets pushed to a worker in the first place — so this isn't an edge case, and it means
`asyncio.process.duration{state="timeout"}` silently under-counts.

**2. A `BaseException` leaves `state` off the attribute set entirely.** `except Exception` doesn't
catch `CancelledError` (a `BaseException` since 3.8), but `finally` still records. On this branch:

```
created point value=1 attrs={'type': 'to_thread', 'name': 'cancelly'}          # no 'state' key
```

Every point on `main` carried `state`, so this introduces a new series that appears only when
something escapes, and the span for that call is left `UNSET` with no `exception` event — the failure
is invisible in both signals. Catching `BaseException` and calling `determine_state(exc)` covers this
and (1) together, and matches how `trace_coroutine` handles its `CancelledError` case.

The suite is green either way (`22 passed`), and `test_to_thread_repeated_calls` is a good pin for the
guard removal — the two above are the cases the new tests don't reach yet.

*AI-assisted; every table row and attribute set above is output from the two runs described, not inferred.*

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_to_thread.py
```diff
@@ -53,6 +54,115 @@ async def to_thread():
...
+            assert result == 6
+
+        asyncio.run(to_thread())
+
+        spans = self.memory_exporter.get_finished_spans()
+        self.assertEqual(len(spans), 1)
+        span = spans[0]
+        self.assertGreaterEqual(span.end_time - span.start_time, 0.1 * 10**9)
+
+        for metric in self.memory_metrics_reader.get_metrics_data().resource_metrics[0].scope_metrics[0].metrics:
```
xrmx:
```suggestion
        for metric in self.get_sorted_metrics():
```

This is a bit easier on the eyes :) There's more occurencies later.

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_to_thread.py
```diff
@@ -53,6 +54,115 @@ async def to_thread():
...
+            if metric.name == "asyncio.process.duration":
```
xrmx: Also can we make tests fail if we don't find the metric we expect?

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_to_thread.py
bourbonkk: Thanks, applied — and to the other 6 occurrences in the file too. I used `get_sorted_metrics(SCOPE)` with the instrumentation scope, to match what `test_asyncio_future_cancellation.py` already does.

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_to_thread.py
bourbonkk: Good catch — the old for ... if metric.name == ... shape passed silently when the metric was never emitted. Now each test asserts both metrics are present via a small helper and checks the data points directly. Verified by patching record_process to skip recording: all 7 tests fail instead of passing. 3ae08526

xrmx (approved):

instrumentation/opentelemetry-instrumentation-asyncio/src/opentelemetry/instrumentation/asyncio/__init__.py
```diff
@@ -216,30 +216,35 @@ def wrap_taskgroup_create_task(method, instance, args, kwargs) -> None:
 
     def trace_to_thread(self, func: callable):
         """
-        Trace a function, but if already instrumented, skip double-wrapping.
+        Wrap a function so that its execution in the worker thread is
```
aabmass, line 219: nit would it be appropriate to update this function's name to capture the changed purpose?

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_to_thread.py
```diff
@@ -43,15 +53,119 @@ async def to_thread():
...
+        created, duration = self.get_created_and_duration_metrics()
+        for metric in (created, duration):
+            self.assertEqual(len(metric.data.data_points), 1)
+            point = metric.data.data_points[0]
+            self.assertEqual(point.attributes["type"], "to_thread")
+            self.assertEqual(point.attributes["name"], "multiply")
+
+    def test_to_thread_duration_covers_execution(self):
+        def multiply(x, y):
+            time.sleep(0.1)
```
aabmass: would it be not-flaky still if we use shorter sleep? it would be good to keep the tests speedy as possible.

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_to_thread.py
```diff
@@ -43,15 +53,119 @@ async def to_thread():
...
+            return x * y
+
+        async def to_thread():
+            result = await asyncio.to_thread(multiply, 2, 3)
+            assert result == 6
+
+        asyncio.run(to_thread())
```
aabmass: can do without the extra function? or alternatively, make these test cases async and just await directly in the test body.

```suggestion
        result = asyncio.run(asyncio.to_thread(multiply, 2, 3))
        assert result == 6
```

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_to_thread.py
```diff
@@ -43,15 +53,119 @@ async def to_thread():
...
+            self.assertEqual(point.attributes["name"], "multiply")
+
+    def test_to_thread_duration_covers_execution(self):
+        def multiply(x, y):
+            time.sleep(0.1)
+            return x * y
+
+        async def to_thread():
+            result = await asyncio.to_thread(multiply, 2, 3)
+            assert result == 6
```
aabmass: nit i don't think the assertion is too meaningful for this test
```suggestion
```

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_to_thread.py
```diff
@@ -43,15 +53,119 @@ async def to_thread():
...
+
+        asyncio.run(to_thread())
+
+        spans = self.memory_exporter.get_finished_spans()
+        self.assertEqual(len(spans), 1)
+        span = spans[0]
+        self.assertGreaterEqual(span.end_time - span.start_time, 0.1 * 10**9)
```
aabmass: I think [assertAlmostEqual](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual) would be better/more readable, passing places or delta param by name.

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_to_thread.py
```diff
@@ -43,15 +53,119 @@ async def to_thread():
...
+
+        _, duration = self.get_created_and_duration_metrics()
+        self.assertEqual(len(duration.data.data_points), 1)
+        self.assertGreaterEqual(duration.data.data_points[0].sum, 0.1)
```
aabmass: assertAlmostEqual

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_to_thread.py
```diff
@@ -43,15 +53,119 @@ async def to_thread():
...
+
+    def test_to_thread_exception(self):
+        def multiply(x, y):
+            raise ValueError("fail")
+
+        async def to_thread():
+            await asyncio.to_thread(multiply, 2, 3)
```
aabmass: same comment on unnecessary wrapper

instrumentation/opentelemetry-instrumentation-asyncio/tests/test_asyncio_to_thread.py
```diff
@@ -43,15 +53,119 @@ async def to_thread():
...
+        created, duration = self.get_created_and_duration_metrics()
+        for metric in (created, duration):
+            self.assertEqual(len(metric.data.data_points), 1)
+            self.assertEqual(metric.data.data_points[0].attributes["state"], "timeout")
+
+    def test_to_thread_cancelled_state(self):
+        def multiply(x, y):
+            raise asyncio.CancelledError()
+
+        async def to_thread():
```
aabmass: same comment

aabmass (approved):
Thank you just a few nits

bourbonkk:
@aabmass 
All addressed in df4051b8. Full asyncio suite passes.
Thank you for your feedback.

instrumentation/opentelemetry-instrumentation-asyncio/src/opentelemetry/instrumentation/asyncio/__init__.py
```diff
@@ -215,31 +215,39 @@ def wrap_taskgroup_create_task(method, instance, args, kwargs) -> None:
         )
 
     def trace_to_thread(self, func: callable):
+        return self.wrap_to_thread_func(func)
+
+    def wrap_to_thread_func(self, func: callable):
```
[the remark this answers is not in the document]
xrmx: @bourbonkk Isn't this correct though?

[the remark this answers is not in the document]
bourbonkk: Yes, agreed. i missed it. — fixed in 6986e3e8.

PR #4905, opened 2026-07-31
fix(logging): do not unlink log record factories chained after ours

### LINKED ISSUE #3808: opentelemetry-instrumentation-logging: LoggingInstrumentor._uninstrument CORRUPTS the log factory linked list

UTKARSH698:
Fixes #3808

## Problem

`logging.setLogRecordFactory` is a single global slot. The convention is to chain: read the current factory, close over it, install a wrapper. `LoggingInstrumentor._instrument` does exactly that.

`_uninstrument` did not:

```python
```

Restoring `_old_factory` unconditionally assumes we are still the head of the chain. If an application or another library installed a factory after us, that factory is silently unlinked:

```
before:  APP ---> OTEL ---> ORIGINAL
after:   APP      OTEL ---> ORIGINAL     # APP is now unreachable
```

Reproduced on `main`, instrumenting and then chaining an app factory on top:

```
before uninstrument -> app_field: present
after  uninstrument -> app_field: MISSING
factory is app_factory?  False
factory is original?     True
```

The app's factory is gone, and every enrichment it performed is silently lost. As the issue notes, this also bites via `instrument()`, which calls `_uninstrument()` internally.

## Change

Track the factory we install as `_our_factory`, and only restore `_old_factory` while we are still the head of the chain. If something has been chained on top, leave the chain intact and warn — a node cannot be removed from the middle of the chain without the cooperation of the factory that wrapped it, so the honest options are "leave it" or "corrupt it".

This means OTel attributes may persist on records after `uninstrument()` in the chained case. That is a deliberate trade: the alternative is dropping someone else's factory. The warning explains it.

The unchained case — by far the common one, and the one the existing tests cover — is unchanged: the original factory is fully restored.

## Tests

Two tests, both in `TestLoggingInstrumentor`:

- `test_uninstrument_keeps_factories_chained_after_ours` — installs an app factory on top, uninstruments, and asserts the app factory is still the head and still enriches records. **Fails on unpatched `main`** (verified by stashing only the source change: `1 failed, 1 passed`).
- `test_uninstrument_restores_factory_when_nothing_chained` — regression guard for the common path; passes before and after.

`pytest tests/test_logging.py` goes from `6 failed, 19 passed` to `6 failed, 21 passed`. The 6 failures are pre-existing on a clean checkout of `main` in my environment (I had to install released `opentelemetry-api`/`sdk` rather than the unreleased `0.66b0.dev` core, so there is some version skew locally); they are untouched by this change.

`ruff check` and `ruff format --check` clean on both files with the pinned `ruff==0.14.1`.

## Credit

The approach here — tracking `_our_factory` and only restoring when still at the head — is the one proposed by @stark256-spec in #4651. That PR was reviewed by @xrmx, who asked for a test case and a changelog entry, and was then closed by stale-bot before the author could follow up. This PR reimplements that fix and supplies both missing pieces.


chuenchen309:
The diagnosis holds up and the choice not to touch the chain is the right one — a node genuinely cannot be
removed from the middle of a `setLogRecordFactory` chain without the cooperation of whoever wrapped it. I
reproduced both branches against the head of this PR (`20fa1bfd`), instrumenting with
`set_logging_format=True`, chaining an app factory on top, then uninstrumenting:

| | app factory survives | chain head still the app factory | OTel fields still injected |
|---|---|---|---|
| this PR as pushed | yes | yes | `otelServiceName`, `otelSpanID`, `otelTraceID`, `otelTraceSampled` |

So the unlink is fixed. What is left is the other half of the same problem: after `uninstrument()` the
instrumentation is still running. Your warning says so honestly, but it does not have to be true — the
wrapper can disable itself, and `_our_factory` is already the state needed for it. Inside `record_factory`,
right after `record = old_factory(*args, **kwargs)`:

```python
if LoggingInstrumentor._our_factory is not record_factory:
    return record
```

`old_factory` is a closure variable, so it survives `_uninstrument` clearing the class attributes, and
re-instrumenting later rebinds `_our_factory` to the new wrapper, which disables any orphan left in
someone else's chain. With that in place the same experiment gives: app factory survives, chain head
unchanged, **no** OTel fields injected.

**But it should not land on its own, because it uncovers something already broken.** `_instrument` with
`set_logging_format=True` calls `logging.basicConfig(format=...)` with a format string that interpolates
`%(otelTraceID)s` and friends, and `_uninstrument` never restores it. On `main` today, with nothing chained
after the instrumentor:

```
root handler fmt after instrument  : ... [trace_id=%(otelTraceID)s span_id=%(otelSpanID)s ...] - %(message)s
root handler fmt after uninstrument: ... [trace_id=%(otelTraceID)s span_id=%(otelSpanID)s ...] - %(message)s
factory restored to: LogRecord
emit raised a formatting error: True
  -> ValueError: Formatting field not found in record: 'otelTraceID'
```

Every record logged after `uninstrument()` fails to format. This PR does not cause it and does not need to
fix it, but it explains why the "leave the chain alone" branch currently looks harmless: it keeps injecting,
which keeps the leftover format string satisfied. Making the wrapper a genuine no-op would expose the
formatting error on that path too, so the two belong together — probably as a separate issue for the format
not being restored, with the self-disabling wrapper on top of it.

Two smaller notes on the patch as it stands:

- The warning fires on every `uninstrument()` where anything was chained after us, which for a test suite
  that instruments and uninstruments per test is once per test. Worth considering `_logger.debug`, or at
  least wording it as a one-off condition.
- `test_uninstrument_restores_factory_when_nothing_chained` takes its expected value from the
  implementation's own `LoggingInstrumentor._old_factory`, so it can pass vacuously: if `_old_factory` were
  ever `None` at that point, `setLogRecordFactory(None)` followed by `assertIs(None, None)` is green while
  the logging module is left unusable. Capturing `logging.getLogRecordFactory()` before `instrument()` and
  asserting against that would be independent of the code under test.

*AI-assisted; both tables are output from running the two variants and `main`, not inferred.*

UTKARSH698:
Both good catches -- I've pushed fixes for each.

**The vacuous test.** You're right. `test_uninstrument_restores_factory_when_nothing_chained` sourced its expected value from `LoggingInstrumentor._old_factory`, i.e. from the code under test, so it passed even when that attribute was `None` -- I confirmed the old assertion still passes in that state while logging is left completely broken (`TypeError: 'NoneType' object is not callable` on the next record). It now captures the real pre-instrumentation factory by observation and asserts against that.

**The warning level.** You were right and my first instinct was wrong. I was going to argue for keeping `warning` because the orphan keeps injecting -- but that only holds when context injection is on. With a bare `instrument()` the wrapper short-circuits on `if not inject_context and not callable(_log_hook)` and returns the record untouched, so the warning was firing with no observable consequence, which is precisely the per-test noise you hit. It now fires only when the orphan will actually keep mutating records. Tests cover both branches.

chuenchen309:
Both hold up on `df93b34f`, checked by running rather than reading.

The test no longer passes vacuously — it takes the pre-instrumentation factory by observation, and the
comment you left on it says why, which is the part that keeps it from drifting back. Full suite is 75
passed here.

The warning change is right too, and the branch you distinguished is the one that matters:

| | orphan still mutates records? | warns on uninstrument |
|---|---|---|
| `instrument()` | no — short-circuits | **no** |
| `instrument(set_logging_format=True)` | yes | yes |
| `instrument(inject_trace_context=True)` | yes | yes |

So the per-test noise is gone without losing the case where an operator genuinely needs to know.

I also filed the format-restore half I mentioned as #4949, so it stops living in this thread: after
`uninstrument()`, the `basicConfig` format string still interpolates `%(otelTraceID)s` and friends while
the records no longer carry them, so every subsequent record fails to format and `handleError` swallows
it. It reproduces on `main` with nothing chained, so it is independent of this PR.

Worth flagging the ordering, though, because the two do interact: this PR leaves the orphaned wrapper
injecting fields, and that is currently what keeps the stale format string satisfied. If a later change
makes the orphan a true no-op — which is the direction the self-disabling idea points — it would expose
#4949 on that path too. Nothing to do about it here; just better that it is written down somewhere
before someone lands the no-op and sees logging break for reasons that look unrelated.

*AI-assisted with agentic coding tools; all changes were reviewed and tested by me. The table is output
from running the three cases against this branch.*

UTKARSH698:
/dashboard route:reviewers

The flagged thread was answered on 08-11 and @chuenchen309 confirmed both fixes on 08-12; the remaining half was split out as #4949. Nothing outstanding on my side.

pmcollins:
Hi, thanks for the submission. I added a question about the motivating use case to #3808.

Looks like the proposed solution keeps the instrumentation's log-record factory in the chain. As a result, it will continue injecting trace context and calling the log hook after `LoggingInstrumentor.uninstrument()`.

I understand leaving the factory in the chain because there is no clean way to remove it from the middle. Have you considered effectively deactivating it so that uninstrument() turns it into a pass-through?

UTKARSH698:
Thanks — yes, and I think you're right that it should. Pushed in `12b1490`.

`record_factory` now checks the class-level `_our_factory` before it touches the record:

```python
record = old_factory(*args, **kwargs)

if LoggingInstrumentor._our_factory is not record_factory:
    return record
```

`old_factory` is a closure variable, so a stranded node still forwards the chain correctly after `_uninstrument` clears the class attributes — it just stops mutating. Re-instrumenting rebinds `_our_factory` to the new wrapper, so an orphan left by an earlier cycle stays disabled rather than double-injecting.

Behaviour after `uninstrument()` with an app factory chained on top, measured rather than reasoned about:

| instrument mode | app factory survives | OTel attrs before | OTel attrs after |
|---|---|---|---|
| `instrument()` | yes | 0/4 | 0/4 |
| `instrument(inject_trace_context=True)` | yes | 4/4 | **0/4** |
| `instrument(set_logging_format=True)` | yes | 4/4 | **0/4** |

This also removes the warning the previous revision emitted, along with the `_injects_context` attribute that existed only to decide whether to emit it — with the wrapper inert there is nothing left to report.

**One interaction worth flagging before this lands,** since it's the reason I'd originally stopped short of the no-op. @chuenchen309 [raised it here](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4905#issuecomment-5248347679) and [filed it as #4949](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4905#issuecomment-5267089134): `_instrument(set_logging_format=True)` calls `logging.basicConfig()` with a format string interpolating `%(otelTraceID)s`, and `_uninstrument` never restores it. While the orphan kept injecting, that stale format string stayed satisfied on the chained path. Now it doesn't:

```
chained_after=False -> fmt still references otel: True | ValueError: Formatting field not found in record: 'otelTraceID'
chained_after=True  -> fmt still references otel: True | ValueError: Formatting field not found in record: 'otelTraceID'
```

So the chained path now fails exactly the way the unchained path already fails on `main` today — this PR makes the two consistent rather than introducing a new failure, and #4949 is the fix for both. Happy to take that one next if you'd like them to land together, or to hold this until it's in; your call on ordering.

Tests: the two warning tests are replaced by three covering context injection stopping, the log hook stopping (a different branch of `record_factory`), and re-instrumentation keeping an earlier orphan disabled. I checked none of them pass vacuously by deleting the guard — all three fail without it. Package suite is 76 passed.

On the motivating use case — I'll follow up on #3808.

PR #4926, opened 2026-08-05
Apply `ruff` default rules to repo, update `uv`

DylanRussell:
# Description

Similar to https://github.com/open-telemetry/opentelemetry-python/pull/5491 -- use the default ruleset for ruff 0.16.1 (removing explicit rule selection so Ruff defaults apply)

Also update uv to latest version (`0.12.1`) and update the lock file.

## Type of change

# How Has This Been Tested?

Unit tests

- [x] No.

# Checklist:

- [] Changelogs have been updated


pyproject.toml
```diff
@@ -153,28 +153,25 @@ members = [
 
 [tool.ruff]
 # https://docs.astral.sh/ruff/configuration/
-target-version = "py310"
```
emdneto, line 156: Why remove this?

DylanRussell: Ruff automatically infers the minimum target Python version (py310) from requires-python = ">=3.10" in [project] of pyproject.toml. So we don't need it here..

emdneto: ah, got it. fine then

emdneto (approved):

emdneto (changes requested):
From SIG call: Let's use extend-select to add back the ones inadvertently removed

aabmass (approved):

emdneto (approved):

PR #4932, opened 2026-08-06
opentelemetry-instrumentation-kafka-python: keep the producer span active while send executes

henry3260:
# Description

`_wrap_send` created the producer span, injected the propagation headers, and then exited the `with` block before calling the wrapped `KafkaProducer.send`. Because `start_as_current_span` ends the span on exit, the instrumented operation ran entirely outside the span it represents: a synchronous failure in `send` was raised after the span had already ended, so the exported span carried `StatusCode.UNSET` and no exception event, and its duration only covered attribute setting and header injection.

This moves `return func(*args, **kwargs)` inside the `with` block so the span stays active while `send` executes. Failures are then recorded by the context manager's default `record_exception` / `set_status_on_exception` behaviour and the original exception is re-raised unmodified. This matches the existing confluent-kafka instrumentation, which already calls the wrapped `produce` inside the span.

Note that `KafkaProducer.send` reports `BrokerResponseError` through the returned future rather than raising, so those failures are still not visible on the span; this change covers the synchronous failure path only.

Part of #4871

## Type of change

# How Has This Been Tested?

Added `TestWrapSendSpanLifetime` in `instrumentation/opentelemetry-instrumentation-kafka-python/tests/test_utils.py`, using the real SDK via `TestBase` rather than a mock tracer:

- [x] `test_send_runs_inside_the_producer_span` — captures `trace.get_current_span()` from inside the wrapped call and asserts its `span_id` matches the exported span, confirming the producer span is the active span while `send` runs.
- [x] `test_send_exception_is_recorded_and_reraised` — asserts the original exception object is re-raised, and that the exported span has `SpanKind.PRODUCER`, `StatusCode.ERROR`, and an `exception` event with the expected `exception.type`.

Both new tests fail on `main` and pass with this change. The existing tests in the package are unmodified and continue to pass.

### DIFF
--- a/.changelog/4931.fixed
+++ b/.changelog/4931.fixed
@@ -1,0 +1,1 @@
+`opentelemetry-instrumentation-kafka-python`: keep the producer span active while the wrapped `send` executes so synchronous failures are recorded on it
--- a/instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
+++ b/instrumentation/opentelemetry-instrumentation-kafka-python/src/opentelemetry/instrumentation/kafka/utils.py
@@ -171,7 +171,7 @@ def _wrap_send(tracer: Tracer, produce_hook: ProduceHookT) -> Callable:
             except Exception as hook_exception:  # pylint: disable=W0703
                 _LOG.exception(hook_exception)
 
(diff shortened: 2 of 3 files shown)


henry3260:
dup with #4858

Issue #2643, opened 2024-06-28
Psycopg: Exception when running `conn.execute("")`


Holi0317:
### Describe your environment

OS: MacOS
Python version: Python v3.12.4
Package version: 0.46b0

### What happened?

When running `conn.execute("")`, opentelementry hook raise exception and fail the query.

### Steps to Reproduce

```py
with psycopg.connect("dbname=test user=postgres") as conn:
    conn.execute("")
```

### Expected Result

The execution should work, at least send the query to postgres.

### Actual Result

```
  File "/Volumes/Projects/xxx/.venv/lib/python3.12/site-packages/opentelemetry/instrumentation/psycopg/__init__.py", line 274, in get_operation_name
    return self._leading_comment_remover.sub("", statement).split()[0]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

### Additional context

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/58a6d3a79c2ab33b80786216e8601e880858b9ab/instrumentation/opentelemetry-instrumentation-psycopg/src/opentelemetry/instrumentation/psycopg/__init__.py#L274

This line is causing the exception.

I also think executing an empty string is weird. But turns out you can actually do that. Psycopg itself is doing this in the reference implementation for connection pool check.

https://github.com/psycopg/psycopg/blob/f2c83579e2523a3ba43caa61a59876f0b9e0b266/psycopg_pool/psycopg_pool/pool.py#L528-L545

### Would you like to implement a fix?

None

PR #4934, opened 2026-08-06
fix(dbapi): avoid IndexError in get_operation_name for comment/whites…

Atishyy27:
# Description

`get_operation_name` in the DB-API instrumentation strips a leading SQL comment and
takes the first token as the operation name. If a statement is truthy but has no tokens
left after stripping (a comment-only or whitespace-only statement), `.split()[0]` raises
`IndexError`, which propagates out of the instrumented `execute()` call and breaks it.

This guards the result the same way the adjacent `_Template` branch already does
(`tokens[0] if tokens else ""`) and returns an empty operation name instead. It extends the
empty-string handling added in #2643 to the comment-only / whitespace-only case.

No open issue for this one — found while reading the operation-name parsing. Related: #2643.

## Type of change

# How Has This Been Tested?

Added `test_operation_name_no_tokens_after_comment_strip` in `test_dbapi_integration.py`,
which executes a comment-only (`/* comment only */`) and a whitespace-only (`"   "`)
statement and asserts instrumentation does not raise. Verified locally that it passes and
that the existing `test_span_succeeded` still passes.

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-dbapi/tests/test_dbapi_integration.py
```diff
@@ -124,6 +124,31 @@ def test_span_succeeded(self):
...
+            )
+            mock_connection = db_integration.wrapped_connection(
+                mock_connect, {}, connection_props
+            )
+            cursor = mock_connection.cursor()
+            # Must not raise IndexError.
+            cursor.execute(query)
+        spans_list = self.memory_exporter.get_finished_spans()
+        self.assertEqual(len(spans_list), 2)
+        for span in spans_list:
```
henry3260: Could we wrap the loop body in self.subTest(query=query)? Then the failure output names which query broke, and one failing input doesn't stop the other from running.

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
```diff
@@ -834,7 +834,8 @@ def get_operation_name(self, cursor: CursorT, args: Sequence[Any]) -> str:  # py
         # `query` can be an empty string. See #2643
         if query and isinstance(query, str):
             # Strip leading comments so we get the operation name.
-            return self._leading_comment_remover.sub("", query).split()[0]
+            tokens = self._leading_comment_remover.sub("", query).split()
```
henry3260, line 790: Same shape exists in psycopg2's `get_operation_name` override — it has no guard at all, so even conn.execute("") from #2643 still raises.

henry3260 (commented):
Thanks for the fix! Just some nits.
I think we should add a changelog fragment under .changelog/ instead of editing CHANGELOG.md directly.

instrumentation/opentelemetry-instrumentation-dbapi/src/opentelemetry/instrumentation/dbapi/__init__.py
Atishyy27: good catch. yeah, psycopg2's get_operation_name override has the same unguarded .split()[0], so conn.execute("") from #2643 still raises there. i'll fix psycopg2 (and check psycopg + sqlalchemy for the same shape) in a separate PR to keep this one a single logical change, rather than bundling it here.

instrumentation/opentelemetry-instrumentation-dbapi/tests/test_dbapi_integration.py
Atishyy27: done in bddd827 
wrapped the loop body in self.subTest(query=query), so a failing input names its query and one bad case doesn't stop the others. thanks!

Atishyy27:
also moved the changelog to a .changelog/4934.fixed fragment per your note.

chuenchen309:
Ran both this and #4936 against `main` at `038eb07ba`. The failure reproduces, and the repo has already decided this case once, in a way that argues for both patches.

**Reproduced.** Calling `get_operation_name` directly, with a control group so the harness is honest:

```
dbapi.CursorTracer.get_operation_name (str branch)
  '/* comment only */'  -> IndexError: list index out of range
  '   '                 -> IndexError: list index out of range
  ''                    -> ''            (the #2643 guard)
  'SELECT 1'            -> 'SELECT'
  '/* c */ SELECT 1'    -> 'SELECT'

psycopg.CursorTracer.get_operation_name (Composable branch, #4936)
  same six inputs, same six results
```

`get_operation_name` is called unconditionally from `_traced_execution`, so this comes out of the instrumented `execute()` rather than being contained.

**There is already a precedent in-tree, and it is the same fix.** `asyncpg` hits the identical expression in three places (`_do_execute`, `_do_cursor_execute`, `_do_prepared_execute`) and each one is wrapped:

```python
try:
    name = self._leading_comment_remover.sub("", name).split()[0]
except IndexError:
    name = ""
```

So "no tokens left after stripping ⇒ empty operation name" is already this repo's answer to this exact input; dbapi and psycopg are the packages that never got it.

**Two sites are still uncovered after both PRs.** Same probe, same clean checkout:

| site | on `main` |
|---|---|
| `psycopg2/__init__.py:335` — `get_operation_name`, str branch | raises on both inputs |
| `sqlalchemy/engine.py:319` — `_operation_name`, `parts.append(...)` | raises on both inputs |
| `asyncpg` ×3 | guarded already |

`psycopg2` is the same override shape #4936 fixes, one package over. Not asking you to widen these PRs — you already asked for one logical change each — but if you are doing them one at a time, that is the rest of the list.

**One thing neither PR changes, mentioned so it is not mistaken for a gap.** The regex is `^/\*.*?\*/`: anchored, no `re.DOTALL`, and it does not know about `--` comments. So `'-- just a note\n'` does not raise — it returns `'--'`, which becomes the span name. Degenerate output rather than a crash, so it is a different question from this PR, but "empty after stripping" is not the only way that parse ends badly.

*AI-assisted; every result above is output from running the code on `038eb07ba`, not read off the source.*

instrumentation/opentelemetry-instrumentation-dbapi/tests/test_dbapi_integration.py
xrmx: The subTest should wrap the assertion, if you add a `self.memory_exporter.clear()` after each iteration I guess you use one loop and assert in the same loop

xrmx (changes requested):

xrmx (approved):

Atishyy27:
Addressed the review feedback and added the related fixes for psycopg, psycopg2, and SQLAlchemy. The PR is ready for maintainer review/merge.

xrmx:
@Atishyy27 Please fix pre-commit

PR #4936, opened 2026-08-08
fix(psycopg): avoid IndexError in get_operation_name for comment/whitespace-only statements

Atishyy27:
Same class as #4934, in the psycopg instrumentation's `get_operation_name` override: a comment-only / whitespace-only `Composed` statement is truthy but has no tokens left after leading-comment stripping, so `.split()[0]` raises `IndexError`. This guards it the same way the dbapi base does (`tokens[0] if tokens else ""`).

Follow-up to #4934, kept as a separate PR per the single-logical-change discussion there.

**How Has This Been Tested?**
Added `test_instrument_connection_comment_only_composed_query` (a comment-only `Composed` no longer raises); the existing `test_instrument_connection_composed_query` still passes. Verified locally.


instrumentation/opentelemetry-instrumentation-psycopg/tests/test_psycopg_integration.py
```diff
@@ -419,6 +419,21 @@ def test_instrument_connection_composed_query(self):
...
+        cnx = psycopg.connect(database="test")
+        query: Composed = SQL("/* comment only */").format()
+
+        cnx = PsycopgInstrumentor().instrument_connection(cnx)
+        cursor = cnx.cursor()
+        # Must not raise IndexError.
+        cursor.execute(query)
+
+        spans_list = self.memory_exporter.get_finished_spans()
+        self.assertEqual(len(spans_list), 1)
```
xrmx, line 435: same as `psycopg2` please assert the operation name. Next time I would wait for the first PR to get merged so that the following can take previous feedback into account.

xrmx (changes requested):

xrmx (approved):

tammy-baylis-swi (approved):

PR #4938, opened 2026-08-08
Pass the response message to the aio gRPC client response hook

### LINKED ISSUE #3490: gRPC AIO Client Instrumentation Unary-Unary Response Hook Bad Arguments
### Describe your environment

OS: Ubuntu 24.04.2
Python version: Python 3.12.3
Package version: 0.54b0

### What happened?

The gRPC AIO client instrumentation response hook for unary-unary receives as arguments the span and the details of the call, but it should receive the span and the response, just like the non-AIO version.

### Steps to Reproduce

`requirements.txt`:
```
grpcio~=1.71
protobuf~=5.29
grpcio-tools~=1.71

opentelemetry-instrumentation-grpc~=0.54b0
opentelemetry-api~=1.33
opentelemetry-sdk~=1.33
```

```shell
pip install -r ./requirements.txt
```

`greeter.proto`:
```proto
syntax = "proto3";

service Greeter {
  rpc SayHello(HelloRequest) returns (HelloReply);
}

message HelloRequest {
  string name = 1;
}

message HelloReply {
  string message = 1;
}
```

```shell
python -m grpc_tools.protoc -I ./ --python_out ./ --grpc_python_out ./ --pyi_out ./ ./greeter.proto
```

`async.py`:
```python
import asyncio

import grpc.aio
from opentelemetry.instrumentation.grpc import GrpcAioInstrumentorClient
from opentelemetry.sdk.trace.export import SimpleSpanProcessor, ConsoleSpanExporter

from greeter_pb2 import HelloReply, HelloRequest
from greeter_pb2_grpc import GreeterServicer, add_GreeterServicer_to_server, GreeterStub


class GreeterServicerImpl(GreeterServicer):
	async def SayHello(self, request, context):
		return HelloReply(message=f"Hello, {request.name}!")


tracer_provider.add_span_processor(SimpleSpanProcessor(ConsoleSpanExporter()))


def request_hook(span, request):
	print(f"Request hook called with a {type(span)} and a {type(request)}.")


def response_hook(span, response):
	print(f"Response hook called with a {type(span)} and a {type(response)}.")


GrpcAioInstrumentorClient().instrument(request_hook=request_hook, response_hook=response_hook)


async def amain():
	server = grpc.aio.server()
	add_GreeterServicer_to_server(GreeterServicerImpl(), server)
	server.add_insecure_port('localhost:50051')
	await server.start()
	
	async with grpc.aio.insecure_channel('localhost:50051') as channel:
		stub = GreeterStub(channel)
		print((await stub.SayHello(HelloRequest(name='Alireza'))).message)
	
	await server.stop(None)

asyncio.run(amain())
```

`sync.py`:
```python
import concurrent.futures

import grpc
from opentelemetry.instrumentation.grpc import GrpcInstrumentorClient
from opentelemetry.sdk.trace.export import SimpleSpanProcessor, ConsoleSpanExporter

from greeter_pb2 import HelloReply, HelloRequest
from greeter_pb2_grpc import GreeterServicer, add_GreeterServicer_to_server, GreeterStub


class GreeterServicerImpl(GreeterServicer):
	def SayHello(self, request, context):
		return HelloReply(message=f"Hello, {request.name}!")


tracer_provider.add_span_processor(SimpleSpanProcessor(ConsoleSpanExporter()))


def request_hook(span, request):
	print(f"Request hook called with a {type(span)} and a {type(request)}.")


def response_hook(span, response):
	print(f"Response hook called with a {type(span)} and a {type(response)}.")


GrpcInstrumentorClient().instrument(request_hook=request_hook, response_hook=response_hook)

server = grpc.server(concurrent.futures.ThreadPoolExecutor(max_workers=1))
add_GreeterServicer_to_server(GreeterServicerImpl(), server)
server.add_insecure_port('localhost:50051')
server.start()

with grpc.insecure_channel('localhost:50051') as channel:
	stub = GreeterStub(channel)
	print(stub.SayHello(HelloRequest(name='Alireza')).message)

server.stop(None)
```

### Expected Result

`async.py`:
```text
Request hook called with a <class 'opentelemetry.sdk.trace._Span'> and a <class 'greeter_pb2.HelloRequest'>.
Response hook called with a <class 'opentelemetry.sdk.trace._Span'> and a <class 'greeter_pb2.HelloReply'>.
...
```

`sync.py`:
```text
Request hook called with a <class 'opentelemetry.sdk.trace._Span'> and a <class 'greeter_pb2.HelloRequest'>.
Response hook called with a <class 'opentelemetry.sdk.trace._Span'> and a <class 'greeter_pb2.HelloReply'>.
...
```

### Actual Result

`async.py`:
```text
Request hook called with a <class 'opentelemetry.sdk.trace._Span'> and a <class 'greeter_pb2.HelloRequest'>.
Response hook called with a <class 'opentelemetry.sdk.trace._Span'> and a <class 'str'>.
...
```

`sync.py`:
```text
Request hook called with a <class 'opentelemetry.sdk.trace._Span'> and a <class 'greeter_pb2.HelloRequest'>.
Response hook called with a <class 'opentelemetry.sdk.trace._Span'> and a <class 'greeter_pb2.HelloReply'>.
...
```

### Additional context

These are the solutions that came into my mind.

## First Solution: Quick One-liner Fix

One way to fix this is to change this line
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/4a1e0ce94183a7f57990ab283a3ccc0dbf445ccd/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_aio_client.py#L47

into 

```python
response_hook(span, call._call_response.result())
```

but that is by accessing a protected member, which should be normally accessed via an `await call` in an async function; also, the result (`.result()` call) is probably sitting there as it is a done callback, but there is no documented guarantees.

## Second Solution: Avoid Done Callback

Another possible fix that came to mind is to change these lines
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/4a1e0ce94183a7f57990ab283a3ccc0dbf445ccd/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_aio_client.py#L104-L124

into

```python
    async def _wrap_unary_response(self, continuation, span):
        try:
            call = await continuation()

            code = await call.code()
            details = await call.details()

            span.set_attribute(SpanAttributes.RPC_GRPC_STATUS_CODE, code.value[0])
            if code != grpc.StatusCode.OK:
                span.set_status(Status(status_code=StatusCode.ERROR, description=details))

            response = await call
            self._call_response_hook(span, response)

            return call
        except grpc.aio.AioRpcError as exc:
            self.add_error_details_to_span(span, exc)
            raise exc
```

The same idea is for the stream response:
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/4a1e0ce94183a7f57990ab283a3ccc0dbf445ccd/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_aio_client.py#L126-L136

Yet I am not a expert in how gRPC interceptors work, and I am worried that it might mess-up the interceptors stack; it seems promising as the same approach is used for the stream response case.

Yes

alliasgher:
Fixes #3490

### Description

The aio client's unary `response_hook` receives the gRPC status **detail string** instead of the response message, so it is `''` on a successful call.

`add_done_callback` cannot take a coroutine, so `_wrap_unary_response` pre-fetches `code` and `details` and passes them into the callback factory. The callback then does `response_hook(span, details)` — `details` was never the response, it just happened to be in scope.

This is an internal inconsistency rather than a design question. Both of the other paths already pass the deserialized message:

- `_client.py:112-113` (sync): `if self._response_hook: self._call_response_hook(span, response)`
- `_aio_client.py` `_wrap_stream_response`: `self._call_response_hook(span, response)`

Only the aio unary path differs. It affects unary-unary and stream-unary.

### Fix

`await call` to get the response before registering the callback. That is free latency-wise — `await call.code()` above it already blocks until the RPC has completed — and `grpc.aio` caches the unary result, so the caller's own `await` still returns the response. The extra await is skipped entirely when no hook is registered.

### Two judgement calls I want to flag rather than have you find

1. **The hook no longer fires on error.** Today it fires with `''` on a non-OK status; now it fires only on OK. That matches the sync client, which never reaches its hook when the call raises, but it *is* a behaviour change for anyone relying on the empty-string call.

   I deliberately did not `await call` on the error path. Doing so raises `AioRpcError` inside the `try`, which the `except grpc.aio.AioRpcError` catches and re-raises from the interceptor rather than from the caller's own `await` — and `add_done_callback` is then never registered, so `span.end()` never runs and the span leaks.

2. The hooks have no docstrings specifying their arguments, so the case for "response message" rests on parity with `_client.py` and `_wrap_stream_response` rather than on documented contract. Happy to be told the aio unary hook was meant to be different.

### Tests

The existing test asserted the bug:

```python
def response_hook(span, response):
    span.set_attribute("response_data", response)
...
    self.assertEqual(span.attributes["response_data"], "")
```

Passing a protobuf message to `set_attribute` is silently dropped, so it could not simply be left alone. The hook now mirrors the sync test's `response.response_data`, and on current `main` that fails with `AttributeError: 'str' object has no attribute 'response_data'`.

Added two cases: stream-unary (same bug, separate call path) and one asserting the hook does not fire on a failed RPC.

`pytest tests/` in the grpc package: 142 passed. There are 10 pre-existing failures in `TestOpenTelemetryServerInterceptorUnix` in my environment (Unix domain sockets on macOS) — I diffed the failing-test sets with and without this change via junit XML and they are identical, so nothing here is a regression. Ruff check and format are clean.

I ran against grpcio 1.75.1 (the pinned test requirement). I did not separately run against the oldest supported grpcio in `test-requirements-0.txt`; `add_done_callback` and awaiting a completed unary call are long-stable, but flagging that I did not verify it.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_aio_client.py
+++ b/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_aio_client.py
@@ -104,8 +104,8 @@ class _BaseAioClientInterceptor(OpenTelemetryClientInterceptor):
             code = await call.code()
             details = await call.details()
 
-            # The response message is what the hook is documented to receive,
-            # and it is what the sync client and the streaming path both pass.
+            # The sync client and the aio streaming path both pass the response
+            # message to the hook; this path passed the status details instead.
             # grpc.aio caches the unary result, so awaiting the call here does
             # not consume it -- the caller's own await still yields it.
             response = None
--- a/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_aio_client.py
+++ b/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_aio_client.py
@@ -21,7 +21,7 @@ from opentelemetry.trace.status import Status, StatusCode
 logger = logging.getLogger(__name__)
 
 
-def _unary_done_callback(span, code, details, response_hook):
+def _unary_done_callback(span, code, details, response, response_hook):
     def callback(call):
         try:
             span.set_attribute(
@@ -35,7 +35,8 @@ def _unary_done_callback(span, code, details, response_hook):
                         description=details,
                     )
                 )
-            response_hook(span, details)
+            else:
+                response_hook(span, response)
 
         finally:
             span.end()
@@ -103,7 +104,15 @@ class _BaseAioClientInterceptor(OpenTelemetryClientInterceptor):
             code = await call.code()
             details = await call.details()
 
-            callback = _unary_done_callback(span, code, details, self._call_response_hook)
+            # The response message is what the hook is documented to receive,
+            # and it is what the sync client and the streaming path both pass.
+            # grpc.aio caches the unary result, so awaiting the call here does
+            # not consume it -- the caller's own await still yields it.
+            response = None
+            if code == grpc.StatusCode.OK and self._response_hook:
+                response = await call
+
+            callback = _unary_done_callback(span, code, details, response, self._call_response_hook)
             try:
                 call.add_done_callback(callback)
             except NotImplementedError:
--- a/instrumentation/opentelemetry-instrumentation-grpc/tests/test_aio_client_interceptor_hooks.py
+++ b/instrumentation/opentelemetry-instrumentation-grpc/tests/test_aio_client_interceptor_hooks.py
@@ -7,7 +7,7 @@ import grpc
 from opentelemetry.instrumentation.grpc import GrpcAioInstrumentorClient
 from opentelemetry.test.test_base import TestBase
 
-from ._aio_client import simple_method
+from ._aio_client import client_streaming_method, simple_method
 from ._server import create_test_server
 from .protobuf import test_server_pb2_grpc  # pylint: disable=no-name-in-module
 
@@ -17,7 +17,7 @@ def request_hook(span, request):
 
 
 def response_hook(span, response):
-    span.set_attribute("response_data", response)
+    span.set_attribute("response_data", response.response_data)
 
 
 def request_hook_with_exception(_span, _request):
@@ -63,7 +63,48 @@ class TestAioClientInterceptorWithHooks(TestBase, IsolatedAsyncioTestCase):
             self.assertEqual(span.attributes["request_data"], "data")
 
             self.assertIn("response_data", span.attributes)
-            self.assertEqual(span.attributes["response_data"], "")
+            self.assertEqual(span.attributes["response_data"], "data")
+        finally:
+            instrumentor.uninstrument()
+
+    async def test_response_hook_stream_unary(self):
+        instrumentor = GrpcAioInstrumentorClient()
+
+        try:
+            instrumentor.instrument(response_hook=response_hook)
+
+            channel = grpc.aio.insecure_channel("localhost:25565")
+            stub = test_server_pb2_grpc.GRPCTestServerStub(channel)
+
+            response = await client_streaming_method(stub)
+            assert response.response_data == "data"
+
+            spans = self.memory_exporter.get_finished_spans()
+            self.assertEqual(len(spans), 1)
+            span = spans[0]
+
+            self.assertIn("response_data", span.attributes)
+            self.assertEqual(span.attributes["response_data"], "data")
+        finally:
+            instrumentor.uninstrument()
+
+    async def test_response_hook_not_called_on_error(self):
+        """There is no response message to hand the hook when the RPC fails."""
+        instrumentor = GrpcAioInstrumentorClient()
+
+        try:
+            instrumentor.instrument(response_hook=response_hook)
+
+            channel = grpc.aio.insecure_channel("localhost:25565")
+            stub = test_server_pb2_grpc.GRPCTestServerStub(channel)
+
+            with self.assertRaises(grpc.RpcError):
+                await simple_method(stub, error=True)
+
+            spans = self.memory_exporter.get_finished_spans()
+            self.assertEqual(len(spans), 1)
+
+            self.assertNotIn("response_data", spans[0].attributes)
         finally:
             instrumentor.uninstrument()
 
--- a/instrumentation/opentelemetry-instrumentation-grpc/tests/test_aio_client_interceptor_hooks.py
+++ b/instrumentation/opentelemetry-instrumentation-grpc/tests/test_aio_client_interceptor_hooks.py
@@ -91,9 +91,13 @@ class TestAioClientInterceptorWithHooks(TestBase, IsolatedAsyncioTestCase):
     async def test_response_hook_not_called_on_error(self):
         """There is no response message to hand the hook when the RPC fails."""
         instrumentor = GrpcAioInstrumentorClient()
+        calls = []
+
+        def recording_hook(span, response):
+            calls.append(response)
 
         try:
-            instrumentor.instrument(response_hook=response_hook)
+            instrumentor.instrument(response_hook=recording_hook)
 
             channel = grpc.aio.insecure_channel("localhost:25565")
             stub = test_server_pb2_grpc.GRPCTestServerStub(channel)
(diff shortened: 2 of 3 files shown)


instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_aio_client.py
```diff
@@ -103,7 +104,15 @@ async def _wrap_unary_response(self, continuation, span):
             code = await call.code()
             details = await call.details()
 
-            callback = _unary_done_callback(span, code, details, self._call_response_hook)
+            # The response message is what the hook is documented to receive,
+            # and it is what the sync client and the streaming path both pass.
```
henry3260:
```suggestion
            # The sync client and the aio streaming path both pass the response
            # message to the hook; this path passed the status details instead.
```

instrumentation/opentelemetry-instrumentation-grpc/tests/test_aio_client_interceptor_hooks.py
```diff
@@ -63,7 +63,48 @@ async def test_request_and_response_hooks(self):
...
+            channel = grpc.aio.insecure_channel("localhost:25565")
+            stub = test_server_pb2_grpc.GRPCTestServerStub(channel)
+
+            with self.assertRaises(grpc.RpcError):
+                await simple_method(stub, error=True)
+
+            spans = self.memory_exporter.get_finished_spans()
+            self.assertEqual(len(spans), 1)
+
+            self.assertNotIn("response_data", spans[0].attributes)
```
henry3260:
```suggestion
    async def test_response_hook_not_called_on_error(self):
        """There is no response message to hand the hook when the RPC fails."""
        instrumentor = GrpcAioInstrumentorClient()
        calls = []
    
        def recording_hook(span, response):
            calls.append(response)
    
        try:
            instrumentor.instrument(response_hook=recording_hook)
    
            channel = grpc.aio.insecure_channel("localhost:25565")
            stub = test_server_pb2_grpc.GRPCTestServerStub(channel)
    
            with self.assertRaises(grpc.RpcError):
                await simple_method(stub, error=True)
    
            self.assertEqual(calls, [])
```

henry3260 (commented):
lgtm, just two nits, `test_response_hook_not_called_on_error` will be green in main, so we should assert that the hook is not called

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_aio_client.py
alliasgher: Applied, thanks.

instrumentation/opentelemetry-instrumentation-grpc/tests/test_aio_client_interceptor_hooks.py
alliasgher: Good catch, and you're right that it was green on main. The reason is a bit sneaky: on main the hook *is* called, with the details string `''`, but the shared `response_hook` does `response.response_data`, which raises `AttributeError` on a str and gets swallowed by `_safe_invoke`. So the attribute was absent either way and the assertion held regardless of the fix.

Switched to your recording hook. Verified by restoring `_aio_client.py` to upstream `3a96d1cf` with the new test in place, and it now fails properly:

```
E   AssertionError: Lists differ: [''] != []
```

I kept one extra assertion, that the span is still ended, since not awaiting a failed call is what keeps `add_done_callback` registered, and a regression there would leak the span rather than just skip the hook.

Full grpc suite is green here now: 152 passed. Worth noting that supersedes the caveat in my PR description about 10 pre-existing `TestOpenTelemetryServerInterceptorUnix` failures; those were an artifact of my earlier local env and pass in a clean one.

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_aio_client.py
```diff
@@ -35,7 +35,8 @@ def callback(call):
                         description=details,
                     )
                 )
-            response_hook(span, details)
+            else:
+                response_hook(span, response)
```
ocelotl, line 39: Before this change the hook runs when RPC fails, after this change the hook will not. Since that is user visible, the changelog should warn about this change.

PR #4942, opened 2026-08-10
fix(psycopg2): avoid IndexError in get_operation_name for comment/whitespace-only statements

Atishyy27:
Same class as #4934, in the psycopg2 instrumentation's `get_operation_name`: a comment-only / whitespace-only statement is truthy but has no tokens after leading-comment stripping, so `.split()[0]` raises `IndexError`. Guards it like the dbapi base does. Extends the empty-string handling from #2643 to this case.

**How Has This Been Tested?** Added `test_span_name_comment_or_whitespace_only` (comment-only + whitespace-only statements no longer raise); existing tests still pass. Verified locally.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-psycopg2/tests/test_psycopg2_integration.py
+++ b/instrumentation/opentelemetry-instrumentation-psycopg2/tests/test_psycopg2_integration.py
@@ -130,6 +130,18 @@ class TestPostgresqlIntegration(TestBase):
         self.assertEqual(spans_list[4].name, "query")
         self.assertEqual(spans_list[5].name, "query")
 
+    def test_span_name_comment_or_whitespace_only(self):
+        # Regression: a comment-only or whitespace-only statement is truthy but has
+        # no tokens after leading-comment stripping; get_operation_name must not
+        # raise IndexError.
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry-instrumentation-psycopg2/tests/test_psycopg2_integration.py
```diff
@@ -130,6 +130,18 @@ def test_span_name(self):
...
+        # Regression: a comment-only or whitespace-only statement is truthy but has
+        # no tokens after leading-comment stripping; get_operation_name must not
+        # raise IndexError.
+        Psycopg2Instrumentor().instrument()
+        cnx = psycopg2.connect(database="test")
+        cursor = cnx.cursor()
+        cursor.execute("/* comment only */")
+        cursor.execute("   ")
+        spans_list = self.memory_exporter.get_finished_spans()
+        self.assertEqual(len(spans_list), 2)
```
xrmx, line 143: Can we assert operation name values?

xrmx (changes requested):

xrmx (approved):

tammy-baylis-swi (approved):

PR #4943, opened 2026-08-10
fix(sqlalchemy): avoid IndexError in _operation_name for comment/whitespace-only statements

Atishyy27:
Same class as #4934, in the SQLAlchemy instrumentation's `_operation_name`: a comment-only / whitespace-only statement is truthy but has no tokens after leading-comment stripping, so `.split()[0]` raises `IndexError`. Guards it so such a statement falls back to the db name / vendor instead of crashing.

**How Has This Been Tested?** Added `test_operation_name_comment_or_whitespace_only` (unit test on `_operation_name`); existing tests still pass. Verified locally.


instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -315,8 +315,12 @@ def _operation_name(self, db_name, statement):
...
             # For some very special cases it might not record the correct statement if the SQL
             # dialect is too weird but in any case it shouldn't break anything.
-            # Strip leading comments so we get the operation name.
-            parts.append(self._leading_comment_remover.sub("", statement).split()[0])
+            # Strip leading comments so we get the operation name. A statement that
+            # is truthy but has no tokens left (comment-only or whitespace-only) must
+            # not raise IndexError; skip adding an operation part in that case.
+            tokens = self._leading_comment_remover.sub("", statement).split()
+            if tokens:
+                parts.append(tokens[0])
```
henry3260, line 323: Falling back to the db name for the span name makes sense, but the same value is also used as `db.operation.name`, so that field ends up holding the database name (or the vendor, e.g. sqlite, when there is no db name). The conflation isn't introduced by this PR  `_operation_name` has fed `db.operation.name` since #4110 , just noting it here. Might be worth a follow-up.

instrumentation/opentelemetry-instrumentation-sqlalchemy/tests/test_sqlalchemy.py
```diff
@@ -59,6 +59,20 @@ def tearDown(self):
         super().tearDown()
         SQLAlchemyInstrumentor().uninstrument()
 
+    def test_operation_name_comment_or_whitespace_only(self):
```
henry3260, line 62: Constructing EngineTracer registers event listeners on the engine that the test never removes, and the tearDown uninstrument() logs Attempting to uninstrument while already uninstrumented. EngineTracer.remove_all_event_listeners() would keep it clean.

henry3260 (commented):
Looks very nice! just leave some nits

instrumentation/opentelemetry-instrumentation-sqlalchemy/tests/test_sqlalchemy.py
```diff
@@ -59,6 +59,25 @@ def tearDown(self):
...
+        # Regression: a comment-only or whitespace-only statement is truthy but has
+        # no tokens after leading-comment stripping; _operation_name must not raise
+        # IndexError and should fall back to the db name / vendor.
+        engine = create_engine("sqlite:///:memory:")
+        tracer = self.tracer_provider.get_tracer(__name__)
+        engine_tracer = EngineTracer(tracer, engine, mock.Mock())
+        try:
+            self.assertEqual(
+                engine_tracer._operation_name("mydb", "/* comment only */"),
```
xrmx: Can we test this instrumenting things and calling execute instead of an internal method?

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
xrmx: Yeah, care to file a bug so we can track it?

xrmx (approved):

Atishyy27:
Filed https://github.com/open-telemetry/opentelemetry-python-contrib/issues/5042 for this, as you asked.

instrumentation/opentelemetry-instrumentation-sqlalchemy/tests/test_sqlalchemy.py
```diff
@@ -59,6 +59,27 @@ def tearDown(self):
...
+        # Regression: a comment-only or whitespace-only statement is truthy but has
+        # no tokens after leading-comment stripping; executing it must not raise
+        # IndexError and the resulting span must not crash on naming either.
+        engine = create_engine("sqlite:///:memory:")
+        SQLAlchemyInstrumentor().instrument(
+            engine=engine,
+            tracer_provider=self.tracer_provider,
+        )
+        cnx = engine.connect()
+        # Must not raise IndexError.
```
ocelotl: If what is being tested is that an error is not raised:

```
try:
    something()
except Exception as exception:
    self.fail(f"An unexpected exception was raised {exception}")
```

Letting an exception raise by itself will fail the test case but not with a proper `AssertionError` that shows the real intention of the test case.

ocelotl (approved):
Just a minor comment

instrumentation/opentelemetry-instrumentation-sqlalchemy/tests/test_sqlalchemy.py
Atishyy27: Done, both executes are wrapped now, so a regression fails as an assertion with the reason instead of a raw error. Thanks for the review.

PR #4944, opened 2026-08-10
opentelemetry-util-http: add redact_query_string and apply it in the wsgi instrumentation

henry3260:
# Description

`opentelemetry-instrumentation-wsgi` writes the raw query string into `http.target` and `url.query` without redaction, so presigned-URL credentials such as `Signature` and `AWSAccessKeyId` reach the trace backend. `redact_url` is only called on the `else` branch of `collect_request_attributes`, which is taken when neither `RAW_URI` nor `REQUEST_URI` is set. gunicorn sets `RAW_URI` and uWSGI sets `REQUEST_URI`, so in production that branch is never reached; the werkzeug development server is what exercises it, which is why the gap does not show up locally.

The semantic conventions say query string values for `AWSAccessKeyId`, `Signature`, `sig` and `X-Goog-Signature` SHOULD be redacted by default and replaced by `REDACTED`, with case-sensitive matching. `tornado` and `aiohttp-server` already redact before calling `_set_http_target`; this brings `wsgi` in line.

The existing `redact_query_parameters` is not usable here: it calls `urlparse` on what is an attacker-controlled target, returning it unchanged when parsing fails, and its `parse_qs`/`urlencode` round trip drops valueless parameters and rewrites `%20` as `+`. This PR adds `redact_query_string` to `opentelemetry-util-http`, which takes a bare query string, splits on `&` and `=` without parsing a URL, percent-decodes parameter names before matching (so `%53ignature` is redacted like `Signature`), and leaves everything it does not match byte for byte. `wsgi` redacts the target's own query and `QUERY_STRING` separately, since PEP 3333 does not require a server to supply both. `url.path` still comes from `PATH_INFO` and is unchanged.

## Type of change

# How Has This Been Tested?

- [x] `uv run tox -e py312-test-util-http` — 74 passed. 15 new tests for `redact_query_string` covering each parameter in `PARAMS_TO_REDACT`, multiple matches in one query, case-sensitive matching, percent-encoded parameter names (`%53ignature`, `%53%69gnature`), parity with `redact_query_parameters` on a percent-encoded name, repeated parameters, valueless and blank parameters, a value containing `[` that would break `urlparse`, and byte-for-byte round-tripping when nothing matches.
- [x] `uv run tox -e py312-test-instrumentation-wsgi` — 61 passed. 9 new tests covering `default`, `http` and `http/dup` semconv modes, `RAW_URI` and `REQUEST_URI` as the target source, a target that is not a parseable URL, an environ with no `QUERY_STRING`, a percent-encoded parameter name, `url.path` still reporting the `PATH_INFO` value, and a query with no sensitive parameters round-tripping unchanged. Each redaction test also asserts that no secret value appears in any attribute of the returned dict.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


util/opentelemetry-util-http/tests/test_redact_query_string.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def test_match_is_case_sensitive(self):
+        # Query parameter names are case sensitive and the semantic
+        # conventions list the exact keys to redact.
+        self.assertEqual(redact_query_string("SIGNATURE=secret"), "SIGNATURE=secret")
+
+    def test_does_not_parse_the_value_as_a_url(self):
+        # A value that would break urlparse must still be handled.
+        self.assertEqual(
+            redact_query_string("next=//exa[mple&Signature=s"),
```
xrmx, line 91: Speaking of invalid urls, how does it behave with more than one `?` in the path, e.g. `?foo?bar&baz` ?

henry3260: > Speaking of invalid urls, how does it behave with more than one `?` in the path, e.g. `?foo?bar&baz` ?

Good edge case!
`?` has no special meaning to `redact_query_string` — it only splits on `&` and `=`, never
parses a URL. So `foo?bar&baz` splits into `foo?bar` and `baz`; neither has an `=`, so there
is no value to redact and the query round-trips verbatim. `_redact_target` partitions on the
first `?` only (later ones are valid query characters per RFC 3986), so `/p?foo?bar&baz`
behaves the same.

Redaction elsewhere in the query still works: `foo?bar&Signature=s` → `foo?bar&Signature=REDACTED`.

The case worth calling out is a `?` right before a sensitive key: `foo?Signature=s` is not
redacted, because the parameter name genuinely is `foo?Signature`, not `Signature`. `parse_qs`
reads it as one key too, so this matches what `redact_query_parameters` does — the new helper
isn't weaker on any of these.

Added tests for it: several `?` in a query, a `?` inside a redacted value, and a parity check
against `redact_query_parameters`.

henry3260:
@xrmx Could you take a look when you got a chance? thanks

PR #4924, opened 2026-08-04
Remove CONTRIBUTING guidelines for moved GenAI instrumentations

tammy-baylis-swi:
# Description

Python GenAI framework instrumentations have been moved to https://github.com/open-telemetry/opentelemetry-python-genai/ and we are deprecating the duplicates that are in this repo: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4918. This `CONTRIBUTING.md` update removes the outdated guidelines and points to the modern repo. IMO [the modern guidelines](https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/CONTRIBUTING.md) are nice and simple, and I don't want to port over what was here. But LMK your thoughts!

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update

# How Has This Been Tested?

Not tested.

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [X] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [X] Followed the style guidelines of this project
- [ ] Changelogs have been updated
- [ ] Unit tests have been added
- [X] Documentation has been updated

PR #4954, opened 2026-08-12
Fix server.address extraction for openai 3.x (httpx2)

### LINKED ISSUE #4953: `openai` 3.x: `server.address` / `server.port` are never recorded (`isinstance` check misses `httpx2.URL`)
## Describe your environment

- OS: macOS 15 (arm64)
- Python: 3.13.13
- Package version: `opentelemetry-instrumentation-openai-v2` 2.4b0
- Related: `openai` 3.0.0, `httpx` 0.28.1, `httpx2` 2.10.0

## What happened?

Spans produced for OpenAI SDK calls no longer carry the `server.address` (or `server.port`) attribute
when the installed `openai` is 3.x. The same code against `openai` 2.54.0 records them normally.

The cause is in `instrumentation/openai_v2/utils.py`:

```python
from httpx import URL          # line 23
...
def get_server_address_and_port(client_instance):     # line 98
    base_client = getattr(client_instance, "_client", None)
    base_url = getattr(base_client, "base_url", None)
    if not base_url:
        return None, None
    address = None
    port = None
    if isinstance(base_url, URL):        # httpx.URL
        address = base_url.host
        port = base_url.port
    elif isinstance(base_url, str):
        url = urlparse(base_url)
        address = url.hostname
        port = url.port
    ...
```

`openai` 3.x builds on **`httpx2`**, so `client._client.base_url` is an **`httpx2.URL`**. It is neither an
`httpx.URL` nor a `str`, so both branches are skipped and the function returns `(None, None)`. Because the
values are only set when truthy, the attributes are dropped silently, with no warning and no error.

## Steps to reproduce

```python
from openai import OpenAI
from opentelemetry.instrumentation.openai_v2.utils import get_server_address_and_port

client = OpenAI(base_url="https://openrouter.ai/api/v1", api_key="x")
print(type(client._client.base_url))                      # <class 'httpx2.URL'>
print(get_server_address_and_port(client.chat.completions))
# openai 2.54.0 -> ('openrouter.ai', None)
# openai 3.0.0  -> (None, None)
```

End to end, with a real request through the instrumentor, the emitted span attributes are:

```
openai 2.54.0: gen_ai.operation.name, gen_ai.provider.name, gen_ai.request.model,
               gen_ai.response.finish_reasons, gen_ai.response.id, gen_ai.response.model,
               gen_ai.usage.input_tokens, gen_ai.usage.output_tokens, server.address
openai 3.0.0 : (identical, minus server.address)
```

## Expected result

`server.address` and `server.port` are recorded regardless of which URL type the SDK uses internally, as
required by the semantic conventions for client spans.

## Actual result

Both attributes are absent on `openai` 3.x.

## Why it matters beyond a missing attribute

`gen_ai.provider.name` names the **SDK**, not the operator of the endpoint. For anything reached through
the OpenAI-compatible protocol, such as a router or gateway, an Azure OpenAI deployment, or a self-hosted
vLLM or Ollama server, `server.address` is the only attribute distinguishing them. Without it, all of these are
indistinguishable from a direct `api.openai.com` call, so consumers that attribute traffic per provider or
per region will silently misattribute it. It fails quietly, which makes it easy to miss in a version bump.

## Suggested fix

Avoid the type check and read the URL structurally, which also covers any future URL implementation:

```python
def get_server_address_and_port(client_instance):
    base_client = getattr(client_instance, "_client", None)
    base_url = getattr(base_client, "base_url", None)
    if not base_url:
        return None, None

    address = getattr(base_url, "host", None)
    port = getattr(base_url, "port", None)
    if not address:                       # plain string base_url
        parsed = urlparse(str(base_url))
        address, port = parsed.hostname, parsed.port

    if port == 443:
        port = None
    return address, port
```

Both `httpx.URL` and `httpx2.URL` expose `.host` / `.port`, so this keeps existing behaviour and fixes
3.x. A regression test asserting `server.address` is present for a non-default `base_url` would catch the
next recurrence. The current tests pass with the attribute missing.

I'm happy to open a PR with the fix and the test if that's useful.

ari-abb:
Fixes #4953

## Description

`get_server_address_and_port` decided how to read `base_url` with
`isinstance(base_url, httpx.URL)`. `openai` 3.x builds on `httpx2`, so `client._client.base_url` is an
`httpx2.URL`: neither that branch nor the `str` branch matches, the function returns `(None, None)`, and
`server.address` and `server.port` are silently dropped from every span.

This reads the URL structurally instead, which covers `httpx.URL`, `httpx2.URL`, any other URL-like
object and plain strings, so extraction no longer depends on which HTTP client the SDK uses internally.

Removing the `isinstance` check also removes the only use of the `httpx` import, which matters
separately: this package declares no dependency on `httpx`, and `openai` 3.x depends on `httpx2`
instead, so on a clean install of `openai==3.0.0` plus this package `httpx` can be absent entirely and
the module-scope `from httpx import URL` made the package unimportable rather than merely degraded.

This looks like a spot missed by the recent httpx2 migration (#4730, #4948) rather than a separate
problem.

Behaviour is preserved on every path that works today:

| `base_url` | before | after |
|---|---|---|
| `httpx.URL` | `.host` / `.port` | unchanged |
| plain string | `urlparse` | unchanged |
| explicit `:443` | port omitted | unchanged |
| missing or falsy | `(None, None)` | unchanged |
| `httpx2.URL` | `(None, None)` | `.host` / `.port` |
| other URL-like | `(None, None)` | `.host` / `.port` |

## Type of change

## How Has This Been Tested?

New `tests/test_server_address.py` covering `httpx.URL`, `httpx2.URL`, a plain string, an explicit
`:443` (asserting the port is still omitted), a missing `base_url`, and that the package imports with
`httpx` uninstalled. The last one is skipped on `openai` 2.x, which depends on `httpx` itself, so the
situation cannot arise there.

Verified in two environments:

- the pinned test environment (`openai==2.26.0`, `httpx==0.27.2`): full package suite **193 passed, 28
  skipped**, no change in behaviour;
- `openai==3.0.0` with `httpx` genuinely absent: the package imports and the new tests pass. Reverting
  the fix in that environment makes the test module fail at import with
  `ModuleNotFoundError: No module named 'httpx'`, which is the regression this guards.

Worth noting: `tests/requirements.latest.txt` pins `openai==2.26.0`, so nothing in CI currently
exercises `openai` 3.x. Happy to add `httpx2` there, or to bump the pin in a separate PR, if you would
like the new cases to run in CI rather than skip.

- [ ] Yes.
- [x] No.

## Checklist:

- [x] Changelog fragment added (`.changelog/4954.fixed`)
- [x] Documentation has been updated (not applicable: no user-facing behaviour or API change)


aabmass:
Hi @ari-abb can thanks for the PR, but development on this has moved to https://github.com/open-telemetry/opentelemetry-python-genai (see https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4924). The version in this repo will be deleted from this repo once we make a final release with the deprecation notice.

I'm happy to transfer the issue over if you can reopen the PR?

PR #4918, opened 2026-08-02
Deprecate google-genai and vertexai instrumentations, deprecate the genai folder

lmolkova:
- update docs for google-genai and vertex
- update instructions, contrib, agents.md to point to new repo for all genai instrumentations
- clean up remaining mentions of genai instrumentations

PR #4904, opened 2026-07-30
Deprecate openai-v2 and openai-agents-v2 instrumentations

lmolkova:
These instrumentations moved to `opentelemetry-instrumentation-genai-openai` and `opentelemetry-instrumentation-genai-openai-agents` in the https://github.com/open-telemetry/opentelemetry-python-genai repository.

- Mark the instrumentor classes as deprecated via `typing_extensions.deprecated` with a docstring note pointing to the new packages
- Add deprecation warnings to the package READMEs (new package name, breaking changes, changelog link)
- Point the instrumentation-genai README table rows to the new packages

PR #4955, opened 2026-08-14
RELEASING.md: sync the per-package major/minor release list with the Prepare release workflow

henry3260:
# Description

Follow-up to #4918.

`RELEASING.md` documents which packages the [`[Package] Prepare release`](./.github/workflows/package-prepare-release.yml) workflow supports, but the list drifted out of sync with the workflow's `options:` choices. #4904 dropped `opentelemetry-instrumentation-openai-v2` and `opentelemetry-instrumentation-openai-agents-v2` from the workflow, and #4918 dropped `opentelemetry-instrumentation-vertexai` and `opentelemetry-instrumentation-google-genai`, but neither PR updated the doc. The doc still listed all four, so a release manager reading it would look for choices the workflow no longer offers.

This updates the list to match the workflow (6 packages), and adds a short note explaining why the deprecated GenAI instrumentations are absent here while still being present in the [Releasing individual package](#releasing-individual-package) list further down. That asymmetry is intentional and matches the "security patches only" policy from #4918: those packages no longer get major or minor releases from this repository, but can still get patch releases from their existing `package-release/` branches, so they remain in `package-prepare-patch-release.yml` and `package-release.yml`.

No workflow behaviour changes. `eachdist.ini`'s `[exclude_release]` list is deliberately left alone: the deprecated packages must stay excluded from the general release for as long as they live in this repository.

## Type of change

# How Has This Been Tested?

Documentation-only change, so no test suite applies. Verified by comparing the `options:` list in each per-package release workflow against the corresponding list in `RELEASING.md`:

- [x] `RELEASING.md` "Preparing a major or minor release for individual package" list now matches `.github/workflows/package-prepare-release.yml` exactly.
- [x] `RELEASING.md` "Releasing individual package" list already matched `.github/workflows/package-release.yml` and is unchanged.
- [x] The `#releasing-individual-package` anchor in the new note resolves to the existing `### Releasing individual package` heading.
- [x] `uv run pre-commit run --files RELEASING.md` passes.

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

- [ ] Changelogs have been updated — n/a, needs the **Skip Changelog** label (see note above)


henry3260:
Could a maintainer add the `Skip Changelog` label? This is a docs-only change to the release runbook, same as #4918 and #4924. Thanks!

RELEASING.md
```diff
@@ -17,19 +17,19 @@
...
 > [!NOTE]
-> Per-package release is supported for the following packages only:
+> Per-package major or minor release is supported for the following packages only:
 > - opentelemetry-opamp-client
 > - opentelemetry-propagator-aws-xray
 > - opentelemetry-resource-detector-azure
 > - opentelemetry-resourcedetector-gcp
 > - opentelemetry-sdk-extension-aws
-> - opentelemetry-instrumentation-openai-v2
-> - opentelemetry-instrumentation-openai-agents-v2
```
xrmx, line 27: I think these should still be releasable though, in 7ca24747ccffc23a80165bdba0020540c2a42b9c we said that these only receives security patches and for that I guess we need them listed in the workflow :)

cc @lmolkova @lzchen @aabmass

xrmx (changes requested):

RELEASING.md
henry3260: > I think these should still be releasable though, in [7ca2474](https://github.com/open-telemetry/opentelemetry-python-contrib/commit/7ca24747ccffc23a80165bdba0020540c2a42b9c) we said that these only receives security patches and for that I guess we need them listed in the workflow :)
> 
> cc @lmolkova @lzchen @aabmass

The patch path still works I think — google-genai already did exactly this in May: #4597 backported the fix onto
`package-release/opentelemetry-instrumentation-google-genai/v0.7bx`, then #4598 cut v0.7b1 from it. `package-prepare-patch-release.yml` and `package-release.yml` both still list all four packages, so that route is unchanged.

What #4904 and #4918 did remove is cutting a *new* minor from main. If you'd rather have that back as well, happy to flip this PR into restoring the four entries in `package-prepare-release.yml` and drop the doc change instead.

lmolkova: Thanks for the explanation, it makes sense. We can still release patch versions from this repo and that's the only thing we need to keep.

lmolkova (approved):

emdneto (approved):

xrmx (approved):

PR #4966, opened 2026-08-17
Add semantic-convention conformance test for the Flask instrumentation

ocelotl:
## Description

Adds a semantic-convention conformance test for the Flask instrumentation. It runs as an ordinary `pytest` test in the Flask instrumentation's existing test suite (no separate tox environment or workflow): the test drives the Flask instrumentation over a defined set of HTTP requests and checks that the emitted telemetry conforms to the HTTP semantic conventions. A convention violation is a normal test failure.

The check reuses the tooling from `open-telemetry/semantic-conventions-conformance` (the generic runner, the HTTP domain runner, the mock server, and the HTTP test client) via its pytest plugin, which collects the `conformance.yaml` and turns each scenario into a pytest item. The telemetry is validated with Weaver against a pinned semantic-conventions registry.

This is the first of a planned series: the intent is to add semantic-convention conformance tests to the Python instrumentations where this kind of testing applies, one at a time, with Flask first.

### What it adds

- `instrumentation/opentelemetry-instrumentation-flask/tests/semconv_conformance/`: a shared Flask app, a scenario that enables `FlaskInstrumentor`, a `conformance.yaml`, and a `conftest.py` that provisions the Weaver binary and the pinned registry on demand (and skips with a clear reason if they cannot be obtained, for example offline).
- The four conformance tooling packages added to the Flask instrumentation's `test-requirements-3.txt` (and `pytest>=8`, which the tooling requires). Older Flask test environments do not install the plugin and simply do not collect the file.

### Draft status and open points

This is a draft for discussion. Known points:

- The conformance tooling is not yet published to a package index, so it is currently installed from a pinned Git ref of the conformance repository. This would move to a released dependency once the tooling is published.
- The test currently runs in the newest Flask test environment only.

Feedback on placement and on the overall approach is welcome.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-flask/test-requirements-3.txt
+++ b/instrumentation/opentelemetry-instrumentation-flask/test-requirements-3.txt
@@ -10,7 +10,10 @@ MarkupSafe==2.1.2
 packaging==24.0
 pluggy==1.6.0
 py-cpuinfo==9.0.0
-pytest==7.4.4
+# The semantic-convention conformance tooling below requires pytest >= 8; this
+# env runs the conformance test case as a normal pytest test, so pytest is
+# bumped here (the older flask envs keep 7.4.4 and simply do not collect it).
+pytest>=8
 tomli==2.0.1
 typing_extensions==4.12.2
 Werkzeug>=3.1.0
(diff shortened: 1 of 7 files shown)


ocelotl:
WIP

PR #4967, opened 2026-08-17
Add semantic-convention conformance test for the requests instrumentation

ocelotl:
## Description

Adds a semantic-convention conformance test for the `requests` instrumentation, the HTTP-client counterpart of the Flask (server) conformance test in #4966. It runs as an ordinary `pytest` test in the `requests` instrumentation's existing test suite (no separate tox environment or workflow): the test drives the `requests` instrumentation over a defined set of HTTP requests against a mock server and checks that the emitted telemetry conforms to the HTTP semantic conventions. A convention violation is a normal test failure.

The check reuses the tooling from `open-telemetry/semantic-conventions-conformance` (the generic runner, the HTTP domain runner, the mock server, and the HTTP test client) via its pytest plugin, which collects the `conformance.yaml` and turns each scenario into a pytest item. The telemetry is validated with Weaver against a pinned semantic-conventions registry.

This is the second instrumentation in a planned series (Flask was first, in #4966): the intent is to add semantic-convention conformance tests to the Python instrumentations where this kind of testing applies, one at a time.

### Shared harness / code reuse

The Weaver-and-registry provisioning `conftest.py` is a shared harness kept byte-for-byte identical across instrumentations (the same file as in #4966). Only the per-instrumentation parts differ: the scenario (which turns on `RequestsInstrumentor`), the `conformance.yaml` (client shape: it starts the HTTP mock server and runs a client scenario), the client workload, and the pinned tooling lines added to this package's `test-requirements.txt`.

### What it adds

- `instrumentation/opentelemetry-instrumentation-requests/tests/semconv_conformance/`: a client workload that drives the shared request contract with a `requests.Session`, a scenario that enables `RequestsInstrumentor`, a `conformance.yaml` (client scenario plus a `server:` block that runs the mock server), and the shared `conftest.py`.
- The four conformance tooling packages added to the `requests` instrumentation's `test-requirements.txt` (and `pytest>=8`, which the tooling requires).

### Draft status and open points

This is a draft for discussion. Known points, same as #4966:

- The conformance tooling is not yet published to a package index, so it is currently installed from a pinned Git ref of the conformance repository. This would move to a released dependency once the tooling is published.
- The long-term way to remove the duplicated `conftest.py` is to move the provisioning upstream into the conformance tooling itself.

Feedback on the approach is welcome.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-requests/test-requirements.txt
+++ b/instrumentation/opentelemetry-instrumentation-requests/test-requirements.txt
@@ -8,7 +8,10 @@ iniconfig==2.0.0
 packaging==24.0
 pluggy==1.6.0
 py-cpuinfo==9.0.0
-pytest==7.4.4
+# The semantic-convention conformance tooling below requires pytest >= 8; the
+# requests test suite runs the conformance test case as a normal pytest test,
+# so pytest is bumped here.
+pytest>=8
 requests==2.33.1
 tomli==2.0.1
 typing_extensions==4.12.2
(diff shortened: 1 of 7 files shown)


ocelotl:
WIP

PR #4974, opened 2026-08-18
prepare semconv utilities to support messaging stability opt-in

emdneto:
# Description
As discussed in last week's SIG:

This PR adds helper utilities to support the stability opt-in migration for the Messaging Semantic Conventions. I intentionally pinned the helpers to the latest semantic convention version (1.44.0). Since the messaging conventions are not yet stable, starting with the most recent version seems appropriate to me, but let me know if you don't agree with that.

Contributors can now use these helpers to migrate individual messaging instrumentations. 

Please take thisinto consideration: **Library-specific attributes, such as those for Kafka and SQS, have intentionally been left out and can be added as needed when each instrumentation is migrated.**


opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -613,6 +629,78 @@ def _set_db_redis_database_index(
...
+
+def _set_messaging_operation(
+    result: MutableMapping[str, AttributeValue],
+    operation: str,
+    sem_conv_opt_in_mode: _StabilityMode,
+) -> None:
+    if _report_old(sem_conv_opt_in_mode):
+        set_string_attribute(result, SpanAttributes.MESSAGING_OPERATION, operation)
+    if _report_new(sem_conv_opt_in_mode):
+        set_string_attribute(result, messaging_attributes.MESSAGING_OPERATION_TYPE, operation)
```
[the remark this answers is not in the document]
emdneto: i think it's better to adapt on the instrumentation side to use the correct messaging operation type.

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
```diff
@@ -245,6 +245,7 @@ async def background_execution_trailers_asgi(scope, receive, send):
                 "more_trailers": False,
             }
         )
+        scope["background_task_start_time_ns"] = time.time_ns()
```
emdneto: Had to do this fix because CI was failing. See #5010 

Thx @xrmx for the help on this

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -613,6 +629,78 @@ def _set_db_redis_database_index(
...
+
+def _set_messaging_system(
+    result: MutableMapping[str, AttributeValue],
+    system: str,
+    sem_conv_opt_in_mode: _StabilityMode,
+) -> None:
+    if _report_old(sem_conv_opt_in_mode):
+        set_string_attribute(result, SpanAttributes.MESSAGING_SYSTEM, system)
+    if _report_new(sem_conv_opt_in_mode):
+        set_string_attribute(result, messaging_attributes.MESSAGING_SYSTEM, system)
```
xrmx: This is the same attribute so not sure we need the conditionals. Or even the helper

```suggestion
    set_string_attribute(result, messaging_attributes.MESSAGING_SYSTEM, system)
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -613,6 +629,78 @@ def _set_db_redis_database_index(
...
+
+def _set_messaging_message_id(
+    result: MutableMapping[str, AttributeValue],
+    message_id: str,
+    sem_conv_opt_in_mode: _StabilityMode,
+) -> None:
+    if _report_old(sem_conv_opt_in_mode):
+        set_string_attribute(result, SpanAttributes.MESSAGING_MESSAGE_ID, message_id)
+    if _report_new(sem_conv_opt_in_mode):
+        set_string_attribute(result, messaging_attributes.MESSAGING_MESSAGE_ID, message_id)
```
xrmx:
```suggestion
    set_string_attribute(result, messaging_attributes.MESSAGING_MESSAGE_ID, message_id)
```

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
emdneto: Agreed

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -689,6 +777,7 @@ def _get_schema_version_for_opt_in_mode(
     signal_versions = {
         _OpenTelemetryStabilitySignalType.HTTP: Schemas.V1_21_0.value,
         _OpenTelemetryStabilitySignalType.DATABASE: Schemas.V1_25_0.value,
+        _OpenTelemetryStabilitySignalType.MESSAGING: Schemas.V1_44_0.value,
```
emdneto, line 758: I'm using this version since there's no stable version yet

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -613,6 +629,72 @@ def _set_db_redis_database_index(
...
 
+# Messaging
+
+
+def _set_messaging_system(
+    result: MutableMapping[str, AttributeValue],
+    system: str,
+    sem_conv_opt_in_mode: _StabilityMode,
+) -> None:
+    set_string_attribute(result, messaging_attributes.MESSAGING_SYSTEM, system)
```
xrmx: Looks like we can drop this and the other helper after
https://github.com/open-telemetry/opentelemetry-python-contrib/pull/5019

Since there won't be confusion for the instrumentations.

emdneto: @xrmx dropped the helpers

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -613,6 +629,56 @@ def _set_db_redis_database_index(
     # No new attribute - db.redis.database_index was removed with no replacement in semconv 1.38.0
 
 
+# Messaging
+
+
+def _set_messaging_operation(
```
henry3260, line 635: The old and new conventions don't just rename the key the allowed values differ too. MessagingOperationValues (old) has `publish/receive/process`, while MessagingOperationTypeValues (new) uses `create/send/receive/process/settle,` with `publish` deprecated in favor of `send`.

Since _set_messaging_operation writes the same string to both keys, in messaging/dup mode one side will always carry a non-compliant value whenever the caller records a send operation: passing "publish" puts a deprecated value in messaging.operation.type, and passing "send" puts a value into messaging.operation that the old convention never defined (breaking existing dashboards filtering on publish). The tests only cover RECEIVE, which happens to be identical in both enums, so this case isn't exercised.

Would it make sense to have the helper accept the old value and translate when emitting the new key? Something like:

```
_MESSAGING_OPERATION_OLD_TO_NEW = {
    MessagingOperationValues.PUBLISH.value: MessagingOperationTypeValues.SEND.value,
}

def _set_messaging_operation(result, operation, sem_conv_opt_in_mode):
    if _report_old(sem_conv_opt_in_mode):
        set_string_attribute(result, SpanAttributes.MESSAGING_OPERATION, operation)
    if _report_new(sem_conv_opt_in_mode):
        new_operation = _MESSAGING_OPERATION_OLD_TO_NEW.get(operation, operation)
        set_string_attribute(result, messaging_attributes.MESSAGING_OPERATION_TYPE, new_operation)
```

That way each instrumentation keeps passing the value it already emits today, and the old-vs-new discrepancy is handled once here instead of in every messaging instrumentation. Alternatively, if the intent is that callers pass whatever value matches their target convention, a docstring note making that contract explicit would also help. wdyt?

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
```diff
@@ -613,6 +629,56 @@ def _set_db_redis_database_index(
...
+def _set_messaging_temp_destination(
+    result: MutableMapping[str, AttributeValue],
+    temporary: bool,
+    sem_conv_opt_in_mode: _StabilityMode,
+) -> None:
+    if temporary is None:
+        return
+
+    if _report_old(sem_conv_opt_in_mode):
+        result[SpanAttributes.MESSAGING_TEMP_DESTINATION] = temporary
```
henry3260, line 655: I think it is worth noting why we use `SpanAttributes` here. IIUC it's because the old messaging attributes `(MESSAGING_DESTINATION, MESSAGING_TEMP_DESTINATION, MESSAGING_CONVERSATION_ID) `have already been removed from` _incubating.attributes.messaging_attributes`, so `SpanAttributes` is the only remaining source

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
emdneto: hmmm. I was thinking in passing that on the consumer side, but I missed the `dup` scenario. I'll make this change

shashank-reddy-nr: +1 on @henry3260's suggestion. This isn't just theoretical: confluent-kafka's producer code already passes `MessagingOperationTypeValues.PUBLISH` (`confluent_kafka/__init__.py:370`), so once Kafka instrumentations wire into _set_messaging_operation in messaging/dup mode, this would emit the deprecated "publish" value under `messaging.operation.type` instead of "send". My PR (#4727) is waiting on this helper landing, so I'd rather see the translation handled once here than have every messaging instrumentation, Kafka included, reimplement it.

One more thing worth flagging: `test_operation_mapping` currently uses RECEIVE for both old and new values, and those happen to be identical in both enums ("receive"), so the test passes whether or not the `publish/send `mapping gets fixed. Might be worth adding a case with a divergent value like `publish` once the fix lands, so the test actually guards against regressing this.

emdneto: Yup I agree. Pushed the change. @shashank-reddy-nr @henry3260 PTAL again. Thank you all

opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py
emdneto: Exactly

lzchen (approved):

henry3260 (approved):
Nice!

PR #4982, opened 2026-08-21
fix(logging): restore handlers and formatters on uninstrument

### LINKED ISSUE #4949: LoggingInstrumentor: uninstrument() leaves the basicConfig format string asking for otel fields, so every later record fails to format
### Describe your environment

OS: Ubuntu (WSL2)
Python version: 3.14.3
Package version: `opentelemetry-instrumentation-logging` 0.66b0.dev0, `opentelemetry-api` 1.45.0.dev0,
built from `main` at `7704aab58`

### What happened?

`LoggingInstrumentor().instrument(set_logging_format=True)` calls
`logging.basicConfig(format=..., level=...)` with a format string that interpolates `%(otelTraceID)s`,
`%(otelSpanID)s`, `%(otelServiceName)s` and `%(otelTraceSampled)s`. `_uninstrument()` restores the log
record factory but leaves that format string in place on the root handler:

```python
```

Once the factory is restored, records no longer carry those attributes — but the handler still asks for
them. **Every record logged after `uninstrument()` fails to format.** `logging` routes the failure
through `Handler.handleError`, so it is not raised to the caller: the application keeps running and its
log output silently stops, with a traceback per record on stderr.

This is distinct from #3808 / #4905, which are about the factory *chain* being cut. This one survives
even when nothing else is chained — it is the `basicConfig` half of `_instrument` having no counterpart
in `_uninstrument`.

### Steps to Reproduce

```python
import logging
from opentelemetry.instrumentation.logging import LoggingInstrumentor

LoggingInstrumentor().instrument(set_logging_format=True)
logging.getLogger("demo").warning("while instrumented")   # fine

LoggingInstrumentor().uninstrument()
logging.getLogger("demo").warning("after uninstrument")   # never reaches the stream
```

### Expected Result

After `uninstrument()`, logging is left as it was found: the record emits normally.

### Actual Result

```
=== after instrument(set_logging_format=True) ===
root handler fmt: ['%(asctime)s %(levelname)s [%(name)s] [%(filename)s:%(lineno)d] [trace_id=%(otelTraceID)s span_id=%(otelSpanID)s resource.service.name=%(otelServiceName)s trace_sampled=%(otelTraceSampled)s] - %(message)s', None]
record factory  : record_factory

=== after uninstrument() ===
root handler fmt: ['%(asctime)s %(levelname)s [%(name)s] [%(filename)s:%(lineno)d] [trace_id=%(otelTraceID)s span_id=%(otelSpanID)s resource.service.name=%(otelServiceName)s trace_sampled=%(otelTraceSampled)s] - %(message)s']
record factory  : LogRecord

--- Logging error ---
KeyError: 'otelTraceID'
ValueError: Formatting field not found in record: 'otelTraceID'
```

Formatting the record directly against the surviving handler gives the same
`ValueError: Formatting field not found in record: 'otelTraceID'`.

### Additional context

The restore half was written once already: #4204 ("Saves/restores old state for clean
uninstrumentation") was closed by its author as *"Closing pending #4210"*. #4210 landed — it moved the
handlers into this package and even wraps `logging.basicConfig` — but it did not carry the
save/restore over, so the gap outlived the PR that was going to close it.

On reach: this only bites code that actually calls `uninstrument()`, so it is not a
production-hot-path bug. The two places it shows up are test suites that instrument and uninstrument
per test, and applications that toggle instrumentation at runtime. In both, the symptom is confusing
out of proportion to the cause, because the logging stops rather than raising.

Worth noting it interacts with #4905: making the orphaned wrapper a genuine no-op (which that PR is
heading toward) would *expose* this on the path where the wrapper currently keeps injecting the fields
and thereby keeps the stale format string satisfied. So the ordering probably matters — this one
first, or the two together.

Yes — happy to, if the direction is agreed. The narrow version is to capture the root handler
formatters and level in `_instrument` when `set_logging_format` applied, and put them back in
`_uninstrument`.

*AI-assisted with agentic coding tools; all changes were reviewed and tested by me. The output above is
from running the snippet against `main` at `7704aab58`, not reconstructed.*

sidsri14:
# Description

Fixes #4949.

When `LoggingInstrumentor().instrument(set_logging_format=True)` is called, `logging.basicConfig()` configures or attaches root handlers with a format string interpolating OpenTelemetry trace context fields (`%(otelTraceID)s`, `%(otelSpanID)s`, etc.).

During `_uninstrument()`, the log record factory was restored to `_old_factory`, but the format string remained active on the root logger's handlers. Because newly created `LogRecord` instances no longer carried `otelTraceID`, any record logged after `uninstrument()` failed to format, causing silent log dropping and `ValueError: Formatting field not found in record: 'otelTraceID'` via `Handler.handleError`.

## Changes

1. In `_instrument()`: When `set_logging_format` is active, capture the pre-instrumentation root logger state (`level`, `handlers`, and existing `formatters`).
2. In `_uninstrument()`: Restore the root logger level, restore previous formatters on pre-existing handlers, and remove any handlers added during instrumentation.
3. In `tests/test_logging.py`: Added unit tests verifying:
   - Logging after `uninstrument()` succeeds without `KeyError` / `ValueError` on formatting.
   - Pre-existing handler formatters are accurately restored after an instrument/uninstrument cycle.

## Type of change

## How Has This Been Tested?

- Ran `pytest instrumentation/opentelemetry-instrumentation-logging/tests`: all 73 tests passed.
- Added regression tests covering both freshly configured and pre-configured logging handlers across instrument/uninstrument cycles.

## Checklist:

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/__init__.py
@@ -162,13 +162,15 @@ class LoggingInstrumentor(BaseInstrumentor):  # pylint: disable=empty-docstring
             root = logging.getLogger()
             old_level = root.level
             old_handlers = list(root.handlers)
-            old_formatters = {h: h.formatter for h in old_handlers}
+            logging.basicConfig(format=log_format, level=log_level)
+            # Record only the handlers basicConfig actually added so that
+            # uninstrument removes exactly those and does not drop handlers the
+            # application added while instrumented.
+            added_handlers = [h for h in root.handlers if h not in old_handlers]
             LoggingInstrumentor._old_handlers_state = {
                 "old_level": old_level,
-                "old_handlers": old_handlers,
-                "old_formatters": old_formatters,
+                "added_handlers": added_handlers,
             }
-            logging.basicConfig(format=log_format, level=log_level)
 
         inject_context = set_logging_format or kwargs.get("inject_trace_context", False)
 
@@ -266,7 +268,5 @@ class LoggingInstrumentor(BaseInstrumentor):  # pylint: disable=empty-docstring
             root.setLevel(state["old_level"])
 
             for h in list(root.handlers):
-                if h not in state["old_handlers"]:
+                if h in state["added_handlers"]:
                     root.removeHandler(h)
-                elif h in state["old_formatters"]:
-                    h.setFormatter(state["old_formatters"][h])
--- a/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/__init__.py
@@ -136,6 +136,7 @@ class LoggingInstrumentor(BaseInstrumentor):  # pylint: disable=empty-docstring
     _old_factory = None
     _log_hook = None
     _logging_handler = None
+    _old_handlers_state = None
 
     def instrumentation_dependencies(self) -> Collection[str]:
         return _instruments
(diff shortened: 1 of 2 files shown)


HarperZ9:
Nice, this is a real gap (#4949). Two things I noticed:

The cleanup removes more than it should — `old_handlers` is snapshotted before `basicConfig`, and uninstrument drops everything not in that set. So if the app adds its own handler after `instrument()`, uninstrument silently removes it too (on a handler-less root `old_handlers` is empty, so both the otel handler and the user's get dropped). Better to record just what `basicConfig` actually added (diff the root's handlers right after the `basicConfig` call) and remove only those.

The formatter-restore test also doesn't exercise the restore: `test_uninstrument_restores_preexisting_handler_formatter` adds the handler before `instrument()`, and `basicConfig` no-ops when the root already has handlers, so the formatter is never changed — the assertion passes even with the restore branch deleted. Starting from a handler-less root (so `basicConfig` runs), then checking the added handler is gone and `root.level` is restored, would actually test it.

PR #4987, opened 2026-08-22
util-genai: include reasoning tokens in the output token metric

venkat-uk:
# Description

`InferenceInvocation` treats reasoning tokens as part of the output token count on the
span, but not on the token usage metric.

`_get_attributes()` sets `gen_ai.usage.output_tokens` to `output_tokens + thinking_tokens`.
`_get_metric_token_counts()` recorded `output_tokens` alone, so the
`gen_ai.client.token.usage` histogram and the span disagreed for any response that carried
reasoning tokens. When only `thinking_tokens` was set the histogram recorded no output
point at all, while the span still reported the count.

This matters for cost: providers bill reasoning tokens as output tokens, and the GenAI
semantic conventions say `gen_ai.usage.reasoning.output_tokens` SHOULD be included in
`gen_ai.usage.output_tokens`. A dashboard built on the metric under-reported spend against
the same call the span reported correctly.

I moved the sum into `_get_total_output_tokens()` and used it in both places, so the two
paths cannot drift apart again. The comment on `self.output_tokens` already stated this
intent: "Output tokens will ultimately be the sum of normal output tokens and thinking
tokens."

Only `InferenceInvocation` carries `thinking_tokens`, so nothing else changes. Inputs are
untouched, and an invocation with no reasoning tokens records exactly what it recorded
before.

Fixes # (no issue)

## Type of change

# How Has This Been Tested?

- [x] Two new cases in `util/opentelemetry-util-genai/tests/test_handler_metrics.py`, each
  asserting the span attribute and the metric point together so the two can be compared
  directly:
  - `test_stop_llm_token_metric_counts_reasoning_tokens_as_output`: `output_tokens=7`,
    `thinking_tokens=3`. Before the change the span read 10 and the metric read 7. Both
    read 10 now.
  - `test_stop_llm_token_metric_records_reasoning_tokens_alone`: `thinking_tokens=4` with
    no `output_tokens`. Before the change the span read 4 and `gen_ai.client.token.usage`
    was never emitted. Both read 4 now.

Both tests fail on `466ae4c` without the source change.

Commands run locally on Python 3.12:

    python -m pytest util/opentelemetry-util-genai/tests
    233 passed (231 before, plus the 2 new cases)

    ruff 0.16.0 check + format --check on the two changed files    # clean
    pylint 4.0.5 --rcfile .pylintrc opentelemetry-util-genai       # 9.98/10, no message on the changed files
    pyright 1.1.404 util/opentelemetry-util-genai                  # same result before and after the change

- [ ] Yes. - Link to PR:
- [x] No.

# Checklist:


lmolkova (changes requested):
Thanks for the contribution!

This code has moved to https://github.com/open-telemetry/opentelemetry-python-genai/tree/main/util/opentelemetry-util-genai. Please check if it's still a gap and if so, consider sending a PR there.

venkat-uk:
Thanks for the pointer. I checked, and it is not a gap any more. Closing this rather than re-sending it.

The reason is worth recording for anyone who finds this later. #283 in `opentelemetry-python-genai` fixed the same problem, and fixed it the other way round from this PR:

> `candidates_token_count` from the google-genai response excludes thought tokens, so the reasoning tokens were missing from the `gen_ai.client.token_usage` metric (the span attribute happened to include them only because util summed `thinking_tokens` into output). Make the instrumentation set `output_tokens` to the full count (candidates + thoughts) and have util emit `LLMInvocation.output_tokens` directly, so the span attribute and the metric agree.

So responsibility now sits with the instrumentation, and util emits what it is given. In current `main` both `_get_attributes` and `_get_metric_token_counts` read `self.output_tokens` plainly, and they agree.

That also makes this PR the wrong change for the new repository. Summing in util would double count for any instrumentation that already reports the full completion count, which is what the OpenAI API does: `reasoning_tokens` sits inside `completion_tokens_details` and is already part of `completion_tokens`. #283's own changelog notes that the old util double-counted for exactly this reason.

Sorry for the noise, and thanks for taking the time to redirect me instead of just closing it.

PR #4989, opened 2026-08-24
feat: add opentelemetry-instrumentation-oracledb package

herin049:
# Description

Adds support for instrumenting the OracleDB Python client. This package is being relocated from [herin049/opentelemetry-instrumentation-oracledb](https://github.com/herin049/opentelemetry-instrumentation-oracledb)

## Type of change

# How Has This Been Tested?

```
uv run tox -e py314-test-instrumentation-oracledb-latest
```

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-oracledb/src/opentelemetry/instrumentation/oracledb/__init__.py
```diff
@@ -0,0 +1,321 @@
...
+
+
+class _AsyncTracedCursorProxy:
+    def __init__(self, cursor: Any, db_api_integration: DatabaseApiIntegration) -> None:
+        self.__wrapped__ = cursor
+        self._self_cursor_tracer = CursorTracer[Any](db_api_integration)
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self.__wrapped__, name)
+
```
pmcollins, line 130: I integration tested this instrumentation locally against an Oracle db, and everything I tested worked fine except for using `with` instead of `async with` on an async cursor. This works without instrumentation but with an instrumented cursor raises a type error:

    async def use_cursor_context_manager() -> None:
        async with oracledb.connect_async(...) as connection:
            with connection.cursor():
                pass

    asyncio.run(use_cursor_context_manager())

Looks like adding support for synchronous `__enter__` and `__exit__`  here should fix it.

pmcollins (commented):
Thanks for doing this. LGTM but for one use case.

instrumentation/opentelemetry-instrumentation-oracledb/src/opentelemetry/instrumentation/oracledb/__init__.py
herin049: Gotcha, correct me if I'm wrong but the example you have above does work, correct? It's just the synchronous case that has a gap? I'll can fix this and add a regression test case, thanks for verifying! 

By the way, I'll be added Docker tests in a later PR to keep the size of this PR manageable.

pmcollins: It works when not instrumented, but when instrumented it produces:

> TypeError: 'opentelemetry.instrumentation.oracledb._AsyncTracedCursorProxy' object does not support the context manager protocol (missed __exit__ method) but it supports the asynchronous context manager protocol. Did you mean to use 'async with'?

Looks like the line that triggers the error is `with connection.cursor():`. It works if you instead do `async with connection.cursor():` but my understanding is that the synchronous version should still be supported.

herin049: Thanks, I've resolved this issue.

instrumentation/opentelemetry-instrumentation-oracledb/src/opentelemetry/instrumentation/oracledb/__init__.py
```diff
@@ -0,0 +1,329 @@
...
+            self.__wrapped__, self.__wrapped__.executemany, *args, **kwargs
+        )
+
+    async def callproc(self, *args: Any, **kwargs: Any) -> Any:
+        return await self._self_cursor_tracer.traced_execution_async(
+            self.__wrapped__, self.__wrapped__.callproc, *args, **kwargs
+        )
+
+
+class _AsyncTracedConnectionProxy:
```
pmcollins: Looks like the OTel async proxy does not pass attribute writes to the underlying connection. So if you try to update the connection with something like `connection.autocommit = True`, the async proxy will hold the value, but the underlying Oracle connection won't be affected.

instrumentation/opentelemetry-instrumentation-oracledb/src/opentelemetry/instrumentation/oracledb/__init__.py
```diff
@@ -0,0 +1,329 @@
...
+            connection_attributes=connection_attributes,
+            version=version,
+            tracer_provider=tracer_provider,
+            enable_commenter=enable_commenter,
+            commenter_options=commenter_options,
+            connect_module=connect_module,
+            enable_attribute_commenter=enable_attribute_commenter,
+        )
+        connection = wrapped(*args, **kwargs)
+        return _AsyncTracedConnectionProxy(connection, integration)
```
pmcollins, line 265: This works without instrumentation but looks like it fails after enabling instrumentation:

```
connection = await pool.acquire()
await pool.release(connection)
```

`pool.acquire()` returns `_AsyncTracedConnectionProxy`, but `AsyncConnectionPool.release()` expects an `oracledb.AsyncConnection`. Raises TypeError: connection must be an instance of oracledb.AsyncConnection.

instrumentation/opentelemetry-instrumentation-oracledb/src/opentelemetry/instrumentation/oracledb/__init__.py
```diff
@@ -0,0 +1,329 @@
...
+class _OracleDatabaseApiIntegration(DatabaseApiIntegration):
+    def get_connection_attributes(self, connection: object) -> None:
+        super().get_connection_attributes(connection)
+        for attribute_name, connection_attribute in _ORACLE_CONNECTION_ATTRIBUTES.items():
+            value = getattr(connection, connection_attribute, None)
+            if isinstance(value, str) and value:
+                self.span_attributes[attribute_name] = value
+
+
+class _AsyncTracedCursorProxy:
```
pmcollins: Looks like this has a similar problem to that of the connection proxy where attribute writes are not passed through. e.g.

```
cursor = connection.cursor()
cursor.arraysize = 200
```

sets the property on the proxy but not on the Oracle cursor. The solution may be to use `wrapt.ObjectProxy`.

instrumentation/opentelemetry-instrumentation-oracledb/src/opentelemetry/instrumentation/oracledb/__init__.py
```diff
@@ -0,0 +1,329 @@
...
+    ORACLE_DB_SERVICE,
+)
+
+if TYPE_CHECKING:
+    from opentelemetry.instrumentation.dbapi import TracedConnectionProxy
+    from opentelemetry.trace import TracerProvider
+
+_logger = logging.getLogger(__name__)
+
+_DATABASE_SYSTEM = "oracle"
```
pmcollins: Looks like we need to handle the legacy and stable values for this? With `OTEL_SEMCONV_STABILITY_OPT_IN=database`, we emit `db.system.name=oracle`, but looks like the stable value is `oracle.db`.

https://opentelemetry.io/docs/specs/semconv/db/oracledb/

instrumentation/opentelemetry-instrumentation-oracledb/src/opentelemetry/instrumentation/oracledb/__init__.py
```diff
@@ -0,0 +1,329 @@
...
+
+
+class OracleDBInstrumentor(BaseInstrumentor):
+    """Instrument synchronous and asynchronous python-oracledb connections."""
+
+    def instrumentation_dependencies(self) -> Collection[str]:
+        return _instruments
+
+    def _instrument(self, **kwargs: Any) -> None:
+        tracer_provider = kwargs.get("tracer_provider")
```
pmcollins, line 287: Should this also read and pass through meter_provider? Looks like the dbapi instrumentation creates database metrics.

pmcollins (commented):
Hi, thanks for the changes so far -- manually integration tested this using an Oracle DB and noted a few findings

instrumentation/opentelemetry-instrumentation-oracledb/src/opentelemetry/instrumentation/oracledb/__init__.py
herin049: This is a new library, so I think we can just have this adhere to new semantic conventions and exclude the opt-in.

pmcollins:
Looking great. Did another round of AI assisted integration testing and just found that there are convenience methods on AsyncConnection that bypass tracing: execute(),
executemany(), fetchall(), fetchmany(), fetchone(), callproc(), and callfunc()

Also, looks like this emits legacy attributes by default and stable if you set OTEL_SEMCONV_STABILITY_OPT_IN=database. Did we want to just emit stable?

herin049:
> Looking great. Did another round of AI assisted integration testing and just found that there are convenience methods on AsyncConnection that bypass tracing: execute(), executemany(), fetchall(), fetchmany(), fetchone(), callproc(), and callfunc()
> 
> Also, looks like this emits legacy attributes by default and stable if you set OTEL_SEMCONV_STABILITY_OPT_IN=database. Did we want to just emit stable?

I originally planned on just emitting stable conventions, but realized that the DB API uses the `_OpenTelemetrySemanticConventionStability` helper which is a singleton. This means I can't directly configure the DB API to specifically use stable semantic conventions, and I also can't just configure `_OpenTelemetrySemanticConventionStability` to be set to stable without potentially impacting other libraries. 

It's easy enough to just support both for now, we can create a follow-up to refactor the DB API so that it can be configured independently of `_OpenTelemetrySemanticConventionStability`.

pmcollins:
LGTM overall -- found a possible straggler. If you set `OTEL_SEMCONV_STABILITY_OPT_IN=database/dup` my understanding is that you're supposed to get

    db.system:      oracle
    db.system.name: oracle.db

but what's emitted is:

    db.system:      oracle.db
    db.system.name: oracle.db

For reference: https://opentelemetry.io/docs/specs/semconv/non-normative/db-migration/

herin049:
> LGTM overall -- found a possible straggler. If you set `OTEL_SEMCONV_STABILITY_OPT_IN=database/dup` my understanding is that you're supposed to get
> 
> ```
> db.system:      oracle
> db.system.name: oracle.db
> ```
> 
> but what's emitted is:
> 
> ```
> db.system:      oracle.db
> db.system.name: oracle.db
> ```
> 
> For reference: https://opentelemetry.io/docs/specs/semconv/non-normative/db-migration/

@pmcollins Yes, I found this minor bug as well. Unfortunately, it's also a little tricky to fix and requires refactoring the DB API because the DB API library uses the following logic:
```python
_set_db_system(
    span_attrs,
    self._db_api_integration.database_system,
    sem_conv_mode,
)
```
Hence, this is a bug present in all of the DB instrumentors. 

My preference is to create an issue for this, or just focus on switching to stable semantic conventions.

PR #4997, opened 2026-08-25
fix(botocore): tag Bedrock embedding InvokeModel calls as embeddings

### LINKED ISSUE #4996: Bedrock embedding calls are tagged with `chat` instead of `embeddings`
The Bedrock botocore extension tags embedding calls as `chat` calls, so anything downstream that
sorts spans by `gen_ai.operation.name` misclassifies them.

## What happens

`extensions/bedrock.py` assigns an LLM-style operation name to every `InvokeModel` call, with no
detection of embedding models. On `main` (line 167):

```python
model_id = self._call_context.params.get(_MODEL_ID_KEY)
if model_id:
    attributes[GEN_AI_REQUEST_MODEL] = model_id
    attributes[GEN_AI_OPERATION_NAME] = GenAiOperationNameValues.CHAT.value
```

A later substring test at line 199 refines Titan to `text_completion`, but nothing ever assigns
`embeddings`. `grep -i embed` returns zero hits in the file.

Observed with `opentelemetry-instrumentation-botocore` 0.60b1, SDK 1.39.1, on a Cohere embedding
model:

```
rpc.method             = "InvokeModel"
gen_ai.system          = "aws.bedrock"
gen_ai.request.model   = "cohere.embed-v4:0"
gen_ai.operation.name  = "chat"
```

Present on `main` and on the released `v0.60b1`, so not a recent regression.

## Expected

`gen_ai.operation.name = "embeddings"`, which the GenAI semantic conventions define for exactly
this case (`open-telemetry/semantic-conventions-genai`, `model/gen-ai/registry.yaml`). This is a
consistency gap, not a new feature: `opentelemetry-instrumentation-openai-v2` already sets
`GenAiOperationNameValues.EMBEDDINGS.value` for its embedding calls (added in #3461).

## Why it matters

Downstream tools resolve span kind from `gen_ai.operation.name`. Datadog's [Agent Observability product](https://www.datadoghq.com/products/ai/agent-observability/), for example, bills only `llm` spans and treats `embedding` spans (among others) as free. Mislabeling an embedding call as `chat` moves it into the billed category, so this isn't just a metadata quibble, it misattributes real cost for anyone running Cohere or similar embedding models on Bedrock.

## Notes for whoever picks this up

The defect exists in two places independently: `extract_attributes` and
`_extract_metrics_attributes` (lines 138-156) each assign the operation name, so a fix touching
only one leaves `gen_ai.client.operation.duration` and `gen_ai.client.token.usage` still labeled
`chat`. The span name (line 393, `f"{operation_name} {request_model}"`) inherits whichever value
wins.

Related: #3210 (bedrock tracking issue), #3461 (openai-v2 embeddings precedent).

BetterAndBetterII:
# Description

`extract_attributes` and `_extract_metrics_attributes` both labeled every Bedrock `InvokeModel` call as `chat` (or Titan `text_completion` when the `modelId` contained `amazon.titan`). Embedding model IDs such as `cohere.embed-v4:0` and `amazon.titan-embed-*` were therefore recorded with `gen_ai.operation.name=chat`, and the span name plus duration/token metrics inherited that value.

Detect embedding `modelId`s (substring `embed`) first in both extract paths and set `gen_ai.operation.name` to `embeddings`. Titan text-completion and chat models are unchanged.

Fixes #4996

## Type of change

# How Has This Been Tested?

- [x] Unit tests in `test_botocore_bedrock.py` covering both extract sites, with and without a request body, for Cohere and Titan embedding model IDs, plus unchanged Titan text-completion and chat models

- [x] No.

# Checklist:


JoeyPatricio:
Not an approver, but I checked this out and ran it on Python 3.13.

test_botocore_bedrock.py is 91/91 green on the branch. Reverting only bedrock.py to main while keeping the new tests fails all 16 embedding cases, so the tests cover the behavior they claim. GenAiOperationNameValues.EMBEDDINGS.value is "embeddings" in the semconv package this code imports, matching the spec's embeddings span definition.

The pre-fix failures show the bug had two shapes. Titan embed models returned chat with no body and text_completion with a body, so the same model produced different operation names depending on call shape. Worth adding to the description.

1. In the third call site the Titan branch calls _extract_titan_attributes, while the embedding branch sets the operation name and returns. Is this deliberate? The semconv defines gen_ai.embeddings.dimension.count, and Titan embed takes a dimensions parameter. The tests assert only model and operation name, so nothing pins the behavior either way.

2. The with_body=True parametrization passes {"inputText": "hello"} for every model, including cohere.embed-v4:0. Cohere Embed on Bedrock takes texts/input_type (AWS docs); inputText is Titan's schema. It passes because the embedding path never reads the body, but it will mislead whoever adds body extraction.

3. Substring matching on embed beats a prefix check here, since it catches cross-region inference profile IDs like global.cohere.embed-v4:0. Any false-positive case you considered?

instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_bedrock.py
```diff
@@ -38,6 +46,97 @@ def filter_message_keys(message, keys):
...
+    "cohere.embed-v4:0",
+    "amazon.titan-embed-text-v1",
+    "amazon.titan-embed-text-v2:0",
+    "amazon.titan-embed-image-v1",
+)
+
+
+@pytest.mark.parametrize("model_id", EMBEDDING_MODEL_IDS)
+@pytest.mark.parametrize("with_body", [False, True])
+def test_extract_attributes_embedding_operation_name(model_id, with_body):
```
xrmx: I would prefer to have recorded tests instead of unit testing the helpers. I don't think we have any embedding  test already, so introducing something like a parametrized `test_invoke_model_with_embeddings_model`. Also I think one model entry per model family is enough.

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/bedrock.py
```diff
@@ -100,6 +100,10 @@
 _MODEL_ID_KEY: str = "modelId"
 
 
+def _is_embedding_model(model_id: str) -> bool:
+    return "embed" in model_id.lower()
```
xrmx: Are model names case insensitive? Anyway it looks like this will work from what I can see in https://docs.aws.amazon.com/bedrock/latest/userguide/models-api-compatibility.html

xrmx (changes requested):

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/bedrock.py
xrmx: As @JoeyPatricio spotted maybe we should split the model_id by `.` and look only in the last part?

instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_bedrock.py
BetterAndBetterII: Replaced the helper unit tests with parametrized recorded `test_invoke_model_with_embeddings_model` (one model per family).

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/bedrock.py
BetterAndBetterII: Detection now uses only the last dotted segment of modelId.

BetterAndBetterII: AWS docs list model IDs in lowercase; the last-segment check still lowercases so mixed-case IDs match.

instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/extensions/bedrock.py
```diff
@@ -474,6 +487,8 @@ def _invoke_model_on_success(
             result["body"] = StreamingBody(new_stream, len(body_content))
 
             response_body = json.loads(body_content.decode("utf-8"))
+            if _is_embedding_model(model_id):
```
henry3260, line 490: IIUC this makes `gen_ai.usage.input_tokens` disappear for Titan embed models on main the response reaches `_handle_amazon_titan_response`, which sets it from `inputTextTokenCount`.

instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_bedrock.py
```diff
@@ -1269,6 +1269,68 @@ def get_model_name_from_family(llm_model):
...
+    response_body = json.loads(response["body"].read())
+    assert response_body
+    if model_family == "amazon.titan":
+        assert "embedding" in response_body
+    elif model_family == "cohere.embed":
+        assert "embeddings" in response_body
+    else:
+        pytest.xfail(f"model family not handled: {model_family}")
+
+    (span,) = span_exporter.get_finished_spans()
```
henry3260, line 1323: The new test also pins this: `assert_stream_completion_attributes` leaves `input_tokens` as None

henry3260 (commented):
Non-blocking thought: this fix had to touch 5 places because the model_id dispatch is repeated across 4 functions (_extract_metrics_attributes, extract_attributes, _get_request_messages, _invoke_model_on_success). Would it be worth centralizing the family detection?

xrmx:
> Non-blocking thought: this fix had to touch 5 places because the model_id dispatch is repeated across 4 functions (_extract_metrics_attributes, extract_attributes, _get_request_messages, _invoke_model_on_success). Would it be worth centralizing the family detection?

This code will be replaced by another instrumentation in the python-genai repository so won't worry much about code cleanups.

PR #5001, opened 2026-08-26
opentelemetry-instrumentation-tornado: do not unpatch a handler class that only inherits the patched marker

### LINKED ISSUE #3072: AttributeError in unpatch_handler_class when child class inherits from instrumented parent handler in opentelemetry-instrumentation-tornado
### Describe your environment

OS: Mac (Sonoma 14.6.1)
Python version: Python 3.12
Package version: 
- opentelemetry-instrumentation-tornado: 0.49b2
- tornado: 6.4.2

### What happened?

OpenTelemetry Tornado instrumentation throws `AttributeError` when `unpatch_handler_class` is called on a child class after its parent class has been instrumented. This occurs due to inconsistent behavior between `getattr` and `delattr` when handling class variables in inheritance.

### Steps to Reproduce

1. Install dependencies:
```
python>=3.12
opentelemetry-instrumentation-tornado>=0.49b2
tornado>=6.4.2
```

2. Create the following test code:
```python
import tornado.web
import tornado.httputil
import unittest.mock
import opentelemetry.instrumentation.tornado


class Handler1(tornado.web.RequestHandler):
    pass


class Handler2(Handler1):
    def initialize(self):
        opentelemetry.instrumentation.tornado.unpatch_handler_class(type(self))


def main():
    opentelemetry.instrumentation.tornado.TornadoInstrumentor().instrument()
    app = tornado.web.Application()
    req = tornado.httputil.HTTPServerRequest(
        connection=tornado.http1connection.HTTP1Connection(
            unittest.mock.MagicMock(), False
        )
    )
    Handler1(app, req)  # Parent handler instantiation
    Handler2(app, req)  # Child handler instantiation
```

3. Run the code

### Expected Result

The code should execute without errors, properly handling the unpatching of the child handler class.

### Actual Result

An `AttributeError` is raised:
```
AttributeError: type object 'Handler2' has no attribute '_otel_patched_key'
```

### Additional context

The issue stems from inconsistent behavior in Python's attribute handling:

1. `getattr(cls, '_otel_patched_key')` traverses the inheritance chain
2. `delattr(cls, '_otel_patched_key')` only operates on the immediate class

The sequence causing the error:
1. Parent class (Handler1) gets instrumented with `_otel_patched_key`
2. Child class (Handler2) calls `unpatch_handler_class`
3. `getattr` check passes (finds attribute in parent)
4. `delattr` fails (can't find attribute in child)

None

ocelotl:
Fixes #3072

`unpatch_handler_class` decided whether a class was patched using
`getattr(cls, _OTEL_PATCHED_KEY, False)`, which walks the inheritance chain.
When a handler class was patched and a subclass of it was not, that check
returned `True` for the subclass, the function continued, and
`delattr(cls, _OTEL_PATCHED_KEY)` raised `AttributeError`, because the marker
attribute lives on the base class and not on the subclass.

The check now looks at `cls.__dict__` directly, so a class that only inherits
the marker returns early and is left alone.

`patch_handler_class` deliberately keeps using `getattr`. Skipping a subclass
whose base class is already patched is correct there, because the subclass
inherits the wrapped methods and has nothing of its own to patch.

Added a test that subclasses a patched handler, unpatches the subclass, and
asserts that no exception is raised and that the base class stays patched. It
fails with `AttributeError` without the fix.

### DIFF
--- a/.changelog/5001.fixed
+++ b/.changelog/5001.fixed
@@ -1,0 +1,1 @@
+`opentelemetry-instrumentation-tornado`: stop `unpatch_handler_class` raising `AttributeError` on a subclass of a patched handler
--- a/instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-tornado/src/opentelemetry/instrumentation/tornado/__init__.py
@@ -527,7 +527,11 @@ def patch_handler_class(
 
 
 def unpatch_handler_class(cls):
-    if not getattr(cls, _OTEL_PATCHED_KEY, False):
+    # Look at the class' own __dict__ instead of using getattr, which would
+    # also find the marker on a base class. A subclass of a patched class
+    # inherits the marker but owns neither it nor the wrapped methods, so
+    # there is nothing to undo on it.
+    if _OTEL_PATCHED_KEY not in cls.__dict__:
         return
 
     unwrap(cls, "prepare")
--- a/instrumentation/opentelemetry-instrumentation-tornado/tests/test_instrumentation.py
+++ b/instrumentation/opentelemetry-instrumentation-tornado/tests/test_instrumentation.py
@@ -17,6 +17,7 @@ from opentelemetry.instrumentation.propagators import (
     set_global_response_propagator,
 )
 from opentelemetry.instrumentation.tornado import (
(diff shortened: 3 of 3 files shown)


ocelotl:
I gambled a bit and predicted the PR number for the changelog fragment, just in case

ocelotl:
Fixed the pending issues :v:

Issue #653, opened 2021-09-02
Fix and re-enable the flaky celery tests


owais:
Some celery tests were disabled a long time ago because of inconsistency. 

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/tests/opentelemetry-docker-tests/tests/celery/test_celery_functional.py#L36
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/tests/opentelemetry-docker-tests/tests/celery/test_celery_functional.py#L146
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/tests/opentelemetry-docker-tests/tests/celery/test_celery_functional.py#L193
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/tests/opentelemetry-docker-tests/tests/celery/test_celery_functional.py#L209
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/tests/opentelemetry-docker-tests/tests/celery/test_celery_functional.py#L479

All of the above tests either use `.delay()` or `.apply_async()` to trigger the Celery tasks. This means a span is generated for the producer, then a message is put on a queue/broker. A celery worker then picks up the message and executes the relavant task which generates the consumer span. This is tricky as there can be some delay between producing and consuming. It looks like our tests do not take this delay into account and expect both producer and consumer spans to be available immediately after calling delay/apply_async which obviously fails a lot of times.

Solution here would be to wait till the expected number of spans arrive to memory exporter or wait on a condition fulfilled by the task to be executed. Whichever solution we pick should have a timeout with a assertion error so tests don't get stuck forever in case the message never arrives.

PR #5015, opened 2026-08-30
celery: wait for exported spans instead of result.ready() in tests

### LINKED ISSUE #5009: Flaky celery test
This is with Python 3.14:

```
__________________ TestCeleryInstrumentation.test_task_raises __________________

self = <tests.test_tasks.TestCeleryInstrumentation testMethod=test_task_raises>

    def test_task_raises(self):
        CeleryInstrumentor().instrument()
    
        result = task_raises.delay()
    
        timeout = time.time() + 60 * 1  # 1 minutes from now
        while not result.ready():
            if time.time() > timeout:
                break
            time.sleep(0.05)
    
        spans = self.sorted_spans(self.memory_exporter.get_finished_spans())
>       self.assertEqual(len(spans), 2)
E       AssertionError: 1 != 2

instrumentation/opentelemetry-instrumentation-celery/tests/test_tasks.py:123: AssertionError
----------------------------- Captured stdout call -----------------------------
 
 -------------- celery@runnervmgx7h7 v5.3.6 (emerald-rush)
--- ***** ----- 
-- ******* ---- Linux-6.17.0-1022-azure-x86_64-with-glibc2.39 2026-08-28 09:04:22
- *** --- * --- 
- ** ---------- [config]
- ** ---------- .> app:         __main__:0x7f29ba8b6510
- ** ---------- .> transport:   memory://localhost//
- ** ---------- .> results:     rpc://
- *** --- * --- .> concurrency: 1 (solo)
-- ******* ---- .> task events: OFF (enable -E to monitor tasks in this worker)
--- ***** ----- 
 -------------- [queues]
                .> celery           exchange=celery(direct) key=celery
                

----------------------------- Captured stderr call -----------------------------
[2026-08-28 09:04:22,731: WARNING/MainProcess] No hostname was supplied. Reverting to default 'localhost'
[2026-08-28 09:04:22,734: WARNING/MainProcess] No hostname was supplied. Reverting to default 'localhost'
[2026-08-28 09:04:23,742: ERROR/MainProcess] Task tests.celery_test_tasks.task_raises[b0d8851d-0ffb-467f-8721-c75b3d3703e9] raised unexpected: CustomError('The task failed!')
  File "/home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/.tox/py314-test-instrumentation-celery/lib/python3.14/site-packages/celery/app/trace.py", line 477, in trace_task
    R = retval = fun(*args, **kwargs)
                 ~~~^^^^^^^^^^^^^^^^^
  File "/home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/.tox/py314-test-instrumentation-celery/lib/python3.14/site-packages/celery/app/trace.py", line 760, in __protected_call__
    return self.run(*args, **kwargs)
           ~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-celery/tests/celery_test_tasks.py", line 29, in task_raises
    raise CustomError("The task failed!")
tests.celery_test_tasks.CustomError: The task failed!
[2026-08-28 09:04:24,736: WARNING/MainProcess] Attempting to uninstrument while already uninstrumented
------------------------------ Captured log call -------------------------------
WARNING  kombu.connection:connection.py:669 No hostname was supplied. Reverting to default 'localhost'
WARNING  kombu.connection:connection.py:669 No hostname was supplied. Reverting to default 'localhost'
ERROR    celery.app.trace:trace.py:270 Task tests.celery_test_tasks.task_raises[b0d8851d-0ffb-467f-8721-c75b3d3703e9] raised unexpected: CustomError('The task failed!')
  File "/home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/.tox/py314-test-instrumentation-celery/lib/python3.14/site-packages/celery/app/trace.py", line 477, in trace_task
    R = retval = fun(*args, **kwargs)
                 ~~~^^^^^^^^^^^^^^^^^
  File "/home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/.tox/py314-test-instrumentation-celery/lib/python3.14/site-packages/celery/app/trace.py", line 760, in __protected_call__
    return self.run(*args, **kwargs)
           ~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/runner/work/opentelemetry-python-contrib/opentelemetry-python-contrib/instrumentation/opentelemetry-instrumentation-celery/tests/celery_test_tasks.py", line 29, in task_raises
    raise CustomError("The task failed!")
tests.celery_test_tasks.CustomError: The task failed!
WARNING  opentelemetry.instrumentation.instrumentor:instrumentor.py:130 Attempting to uninstrument while already uninstrumented

```

aniketwaghh:
# Description

The celery tests wait on `result.ready()` before asserting on spans, but Celery stores the task result before it dispatches `task_postrun`, and `CeleryInstrumentor` ends the run span from its `task_postrun` receiver. So the result can be ready while the run span has not been ended and exported yet, and the assertion runs against an incomplete exporter. That is the `1 != 2` in the issue.

These now wait for the expected number of exported spans instead. `test_task_clears_start_time_cache` also waits for the start-time cache to drain, because that pop happens a few statements after the span is ended in the same receiver. `test_baggage` and `test_uninstrument` keep `result.ready()`, which is the right thing for them to wait on.

Test-only change, no instrumentation code touched.

Two things worth flagging:

- The old loops `break` on timeout and fall through into the assertion, so a genuine hang showed up as a confusing span-count mismatch. These raise with the reason instead.
- `test_metrics.py` waits the same way and has the same race. I left it alone to keep this to one logical change, but it is the same fix if you want it. It already carries a `skipif(PyPy, reason="Fails randomly in pypy")`, which looks like this race showing up there.

Prior analysis of this race in the docker suite is in #4964 by @Pissinatti-py, for #653. Different file and different issue, so no overlap with this change, but they described the cause first and it is worth reading alongside.

Fixes #5009

## Type of change

# How Has This Been Tested?

Ordering measured on macOS arm64 against `f1b9368aa`, showing the result landing before the span ends:

```
   34.304 ms  task_prerun
   35.232 ms  backend.mark_as_failure ENTER
   35.302 ms  task_failure
   35.446 ms  task_postrun          <- run span ends here
   36.029 ms  result.ready() == True
```

```
AT result.ready(): ['apply_async/celery_test_tasks.task_raises']
ONE SECOND LATER : ['apply_async/...', 'run/celery_test_tasks.task_raises']
```

To make it deterministic rather than waiting for a loaded runner, I connected a 0.5s `task_postrun` receiver ahead of the instrumentor's, which widens the real window without touching any source:

```python
from celery import signals
signals.task_postrun.connect(lambda *a, **k: time.sleep(0.5), weak=False)
```

With that loaded, on `main`:

```
8 failed, 22 passed
FAILED test_tasks.py::TestCeleryInstrumentation::test_task
FAILED test_tasks.py::TestCeleryInstrumentation::test_task_clears_start_time_cache
FAILED test_tasks.py::TestCeleryInstrumentation::test_task_not_instrumented_does_not_raise
FAILED test_tasks.py::TestCeleryInstrumentation::test_task_raises
FAILED test_tasks.py::TestCeleryInstrumentation::test_task_use_span_links
FAILED test_tasks.py::TestCelerySignatureTask::test_hidden_task
FAILED test_metrics.py::TestMetrics::test_basic_metric
FAILED test_metrics.py::TestMetrics::test_metric_uninstrument
```

`test_task_raises` fails with `AssertionError: 1 != 2` at `test_tasks.py:123`, the error and line from the CI traceback in the issue.

With this change, same widened window:

```
2 failed, 28 passed
```

The six in `test_tasks.py` are gone. The two remaining are `test_metrics.py`, which this PR does not touch.

Unmodified suite:

```
uvx --from tox --with tox-uv tox -e py312-test-instrumentation-celery   ->  30 passed
uvx --from tox --with tox-uv tox -e lint-instrumentation-celery         ->  10.00/10
ruff check / ruff format --check                                        ->  clean
```

- [x] Reproduced the race on `main` and confirmed the fix removes it
- [x] Full celery suite green on py3.12 and py3.14

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

- [ ] Changelogs have been updated - test-only, so I believe this needs the `Skip Changelog` label rather than a fragment. Happy to add one if you would rather have it.
- [x] Unit tests have been added - existing tests made deterministic
- [ ] Documentation has been updated - not applicable

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-celery/tests/test_tasks.py
+++ b/instrumentation/opentelemetry-instrumentation-celery/tests/test_tasks.py
@@ -27,6 +27,30 @@ from .celery_test_tasks import (
 )
 
 
+def wait_for(predicate, message, timeout_s=60):
+    """Poll ``predicate`` until it is true, failing the test if it never is."""
+    deadline = time.time() + timeout_s
+    while not predicate():
+        if time.time() > deadline:
+            raise AssertionError(f"timed out after {timeout_s}s waiting for {message}")
+        time.sleep(0.05)
+
+
+def wait_for_spans(exporter, span_count, timeout_s=60):
+    """Wait until ``span_count`` spans have been exported.
+
+    Celery stores the task result before it dispatches ``task_postrun``, and the
+    instrumentation ends the run span from ``task_postrun``. So an ``AsyncResult``
+    can be ready while the run span has not been ended and exported yet, and
+    waiting on ``result.ready()`` alone makes the span assertions racy.
+    """
+    wait_for(
+        lambda: len(exporter.get_finished_spans()) >= span_count,
+        f"{span_count} spans to be exported",
+        timeout_s,
+    )
+
+
 class TestCeleryInstrumentation(TestBase):
     def setUp(self):
         super().setUp()
@@ -45,13 +69,9 @@ class TestCeleryInstrumentation(TestBase):
     def test_task(self):
         CeleryInstrumentor().instrument()
 
-        result = task_add.delay(1, 2)
+        task_add.delay(1, 2)
 
-        timeout = time.time() + 60 * 1  # 1 minutes from now
-        while not result.ready():
-            if time.time() > timeout:
-                break
-            time.sleep(0.05)
+        wait_for_spans(self.memory_exporter, 2)
 
         spans = self.sorted_spans(self.memory_exporter.get_finished_spans())
         self.assertEqual(len(spans), 2)
@@ -98,11 +118,13 @@ class TestCeleryInstrumentation(TestBase):
 
         result = task_add.delay(1, 2)
 
-        timeout = time.time() + 60 * 1  # 1 minutes from now
-        while not result.ready():
-            if time.time() > timeout:
-                break
-            time.sleep(0.05)
+        wait_for_spans(self.memory_exporter, 2)
+        # the cache is cleared a few statements after the run span is ended, in
+        # the same task_postrun receiver, so it needs its own wait
+        wait_for(
+            lambda: not instrumentor.task_id_to_start_time,
+            "the task start time cache to be cleared",
+        )
 
         self.assertTrue(result.ready())
         self.assertEqual(result.result, 3)
@@ -111,13 +133,9 @@ class TestCeleryInstrumentation(TestBase):
     def test_task_raises(self):
         CeleryInstrumentor().instrument()
 
-        result = task_raises.delay()
+        task_raises.delay()
 
-        timeout = time.time() + 60 * 1  # 1 minutes from now
-        while not result.ready():
-            if time.time() > timeout:
-                break
-            time.sleep(0.05)
+        wait_for_spans(self.memory_exporter, 2)
 
         spans = self.sorted_spans(self.memory_exporter.get_finished_spans())
         self.assertEqual(len(spans), 2)
@@ -218,11 +236,7 @@ class TestCeleryInstrumentation(TestBase):
 
         result = task_add.delay(1, 2)
 
-        timeout = time.time() + 60 * 1  # 1 minutes from now
-        while not result.ready():
-            if time.time() > timeout:
-                break
-            time.sleep(0.05)
+        wait_for_spans(self.memory_exporter, 2)
 
         spans = self.sorted_spans(self.memory_exporter.get_finished_spans())
         self.assertEqual(len(spans), 2)
@@ -235,13 +249,9 @@ class TestCeleryInstrumentation(TestBase):
     def test_task_use_span_links(self):
         CeleryInstrumentor().instrument(use_span_links=True)
 
-        result = task_add.delay(1, 2)
+        task_add.delay(1, 2)
 
-        timeout = time.time() + 60 * 1  # 1 minute from now
-        while not result.ready():
-            if time.time() > timeout:
-                break
-            time.sleep(0.05)
+        wait_for_spans(self.memory_exporter, 2)
 
         spans = self.sorted_spans(self.memory_exporter.get_finished_spans())
         self.assertEqual(len(spans), 2)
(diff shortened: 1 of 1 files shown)


instrumentation/opentelemetry-instrumentation-celery/tests/test_tasks.py
henry3260: One thing while you're in this test: `unwrap(utils, "retrieve_context")` is the last statement, so it's skipped whenever anything above it raises. That was already the case with assertEqual, and `wait_for_spans` adds one more path. When it leaks, retrieve_context stays patched for the rest of the process, so `test_task_raises`, `test_task_use_span_links` and `test_uninstrument` all run against the patched version and fail for reasons that have nothing to do with them.

`self.addCleanup(unwrap, utils, "retrieve_context") `right after the wrap would make it unconditional.

henry3260 (approved):
Thanks for the fix!

aniketwaghh:
Done in 80655de — `addCleanup` right after the wrap, and I dropped the trailing `unwrap` since it is dead once the cleanup is unconditional (`unwrap` guards on `isinstance(func, BaseObjectProxy)`, so it was idempotent anyway).

One correction worth recording: I tried to reproduce the knock-on failures and could not. Injecting a failure before the old `unwrap` does leak the patch, but `test_task_raises`, `test_task_use_span_links` and `test_uninstrument` all still pass (1 failed, 29 passed). The wrapper only nulls the token, and that path logs rather than raises. Your ordering is right — alphabetical collection does run those three after this test — the leak just is not fatal to them today. Still worth making unconditional, since that is luck rather than design.

30 passed, lint 10.00/10.

aniketwaghh:
@xrmx this one is test-only, so there is no behaviour change to describe. Going by CONTRIBUTING ("if there is any impact on behavior") that reads as `Skip Changelog` rather than a fragment, same as #5018 and #4956. Could you add the label?

Happy to add a fragment instead if you would rather have one.

aniketwaghh:
Rebased onto main to clear the stale changelog failure — the label landed after that run finished, so it never re-evaluated. That worked, but the push also reset the fork workflow approval, so CI is sitting at `action_required` and needs a maintainer to kick it off. Sorry for the extra step.

PR #5016, opened 2026-08-31
celery: wait for the recorded metric instead of result.ready() in the metrics tests

henry3260:
# Description

`test_metrics.py` has the same race that #5015 fixes in `test_tasks.py`. `get_metrics()` waits on `result.ready()` before reading the metrics, but Celery stores the task result before it dispatches `task_postrun`, and `CeleryInstrumentor` records the runtime histogram from its `task_postrun` receiver. So the result can be ready before the histogram has been recorded, and the assertion runs against an empty reader.

`get_metrics()` now waits for the recorded run count instead. It takes the expected count so that the second call in `test_metric_uninstrument`, which asserts the histogram has accumulated two runs, waits for the second run rather than returning on the first. The third call there asserts that nothing is recorded after `uninstrument()`; there is no metric to wait for, so it waits on `result.ready()` to keep the assertion from passing vacuously.

With the race gone, `test_metric_uninstrument`'s `skipif(PyPy, reason="Fails randomly in pypy")` is no longer needed, so it is removed and the test runs on PyPy again.

Test-only change, no instrumentation code touched.

One note: `wait_for` duplicates the helper #5015 adds to `test_tasks.py`. I kept this standalone against `main` so the two can merge independently — different files, no conflict. Happy to rebase onto #5015 and share one helper if you prefer.

Related: #5015.

## Type of change

# How Has This Been Tested?

The window is too narrow to hit reliably on an idle machine, so I widened it with a temporary `conftest.py` that connects a slow `task_postrun` receiver ahead of the instrumentor's. No source touched:

```python
signals.task_postrun.connect(lambda *a, **k: time.sleep(0.5), weak=False)
```

With that loaded, on `main`, `test_basic_metric` fails with `AssertionError: 0 != 1` and `test_metric_uninstrument` with `IndexError: list index out of range`, on both CPython and PyPy. With this change both pass; the failures left in that run are all in `test_tasks.py`, which this PR does not touch.

Unmodified suite, macOS arm64 against `f1b9368aa`:

```
tox -e py312-test-instrumentation-celery   ->  30 passed
tox -e py314-test-instrumentation-celery   ->  30 passed
tox -e pypy3-test-instrumentation-celery   ->  30 passed (29 passed, 1 skipped before)
tox -e lint-instrumentation-celery         ->  10.00/10
ruff check / ruff format --check           ->  clean
```

- [x] Reproduced the race on `main` and confirmed the fix removes it
- [x] Full celery suite green on py3.12, py3.14 and pypy3

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

- [x] Unit tests have been added - existing tests made deterministic
- [ ] Documentation has been updated - not applicable

### DIFF
--- a/.changelog/5016.fixed
+++ b/.changelog/5016.fixed
@@ -1,0 +1,1 @@
+`opentelemetry-instrumentation-celery`: fix flaky metrics tests and restore PyPy test coverage
--- a/instrumentation/opentelemetry-instrumentation-celery/tests/test_metrics.py
+++ b/instrumentation/opentelemetry-instrumentation-celery/tests/test_metrics.py
@@ -3,10 +3,7 @@
 
 import threading
 import time
-from platform import python_implementation
 from timeit import default_timer
-
-from pytest import mark
 
 from opentelemetry.instrumentation.celery import CeleryInstrumentor
 from opentelemetry.test.test_base import TestBase
@@ -95,7 +92,6 @@ class TestMetrics(TestBase):
             est_value_delta=200,
         )
 
-    @mark.skipif(python_implementation() == "PyPy", reason="Fails randomly in pypy")
     def test_metric_uninstrument(self):
         CeleryInstrumentor().instrument()
 
--- a/instrumentation/opentelemetry-instrumentation-celery/tests/test_metrics.py
+++ b/instrumentation/opentelemetry-instrumentation-celery/tests/test_metrics.py
@@ -16,6 +16,21 @@ from .celery_test_tasks import app, task_add
 SCOPE = "opentelemetry.instrumentation.celery"
 
 
+def wait_for(predicate, message, timeout_s=60):
+    """Poll ``predicate`` until it is true, failing the test if it never is."""
+    deadline = time.time() + timeout_s
+    while not predicate():
+        if time.time() > deadline:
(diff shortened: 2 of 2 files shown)


Shriprasad-P (approved):
Reviewed the Celery task/result ordering and the instrumentation path.

The race is real: Celery can persist the task result before `task_postrun`, while the runtime histogram is recorded only from the post-run receiver. That means `result.ready()` can become true before the metric exists, so waiting on the recorded run count is the right synchronization point.

Using expected counts 1 then 2 correctly prevents the second assertion from returning on the first sample. After `uninstrument()`, waiting for task completion is appropriate because the signal handler has already been disconnected and there is no new metric to wait for.

Removing the PyPy skip also looks justified given that this race explains the previous flake and the PyPy CI coverage is green.

One minor future-proofing note: `recorded_run_count()` currently relies on `metrics[0]`, which is safe while Celery exposes only this metric. Selecting `flower.task.runtime.seconds` by name would be slightly more robust if additional metrics are added later.

LGTM.

emdneto (approved):

Shriprasad-P (commented):
## Review

Good flakiness fix. Waiting on `AsyncResult.ready()` races `task_postrun` metric recording; polling the recorded run count is the right synchronization point.

Also nice to drop the PyPy skip once the race is gone. LGTM for the test change.

PR #5018, opened 2026-09-01
opentelemetry-instrumentation-asgi: make trailers test deterministic

### LINKED ISSUE #5010: asgi: flaky test
opentelemetry-instrumentation-asgi test seems flaky here https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/33184817795/job/98895055406?pr=4974

```
______________________ TestAsgiApplication.test_trailers _______________________

self = <tests.test_asgi_middleware.TestAsgiApplication testMethod=test_trailers>

    async def test_trailers(self):
        """Test that trailers are emitted as expected and that the server span is ended
        BEFORE the background task is finished."""
        app = otel_asgi.OpenTelemetryMiddleware(background_execution_trailers_asgi)
        self.seed_app(app)
        await self.send_default_request()
        outputs = await self.get_all_output()
    
        def add_body_and_trailer_span(expected: list):
            body_span = {
                "name": "GET / http send",
                "kind": trace_api.SpanKind.INTERNAL,
                "attributes": {"asgi.event.type": "http.response.body"},
            }
            trailer_span = {
                "name": "GET / http send",
                "kind": trace_api.SpanKind.INTERNAL,
                "attributes": {"asgi.event.type": "http.response.trailers"},
            }
            expected[2:2] = [body_span]
            expected[4:4] = [trailer_span] * 2
            return expected
    
        self.validate_outputs(outputs, modifiers=[add_body_and_trailer_span])
        span_list = self.get_finished_spans()
        server_span = span_list[-1]
        assert server_span.kind == SpanKind.SERVER
        span_duration_nanos = server_span.end_time - server_span.start_time
>       self.assertLessEqual(
            span_duration_nanos,
            _SIMULATED_BACKGROUND_TASK_EXECUTION_TIME_S * 10**9,
        )
E       AssertionError: 19942814 not less than or equal to 10000000.0
```

Shriprasad-P:
## Description

Fixes #5010.

`TestAsgiApplication.test_trailers` previously compared the total server span duration against the simulated 10 ms background-task duration.

That made the test sensitive to CI scheduling and instrumentation overhead even when the server span correctly ended before the background task.

The test now records when background execution begins and directly verifies that the server span ended before that point.

Production ASGI middleware behavior is unchanged.

## Type of change

# How Has This Been Tested?

- [x] Targeted `test_trailers` test passes
- [x] `test_asgi_middleware.py`: 66 passed
- [x] ASGI instrumentation tox environment: 93 passed
- [x] ASGI pylint: 10.00/10
- [x] Ruff check/format on changed file passes
- [x] `test_trailers` repeated 20 times with 0 failures

- [x] No.

# Checklist:

This is a test-only change with no user-facing behavior change, so no changelog fragment was added. Please apply the **Skip Changelog** label if required.

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
@@ -204,6 +204,8 @@ async def background_execution_asgi(scope, receive, send):
                 "body": b"*",
             }
         )
+        # Record when the background task starts; the server span must already be ended by now.
+        scope["background_task_start_ns"] = time.time_ns()
         time.sleep(_SIMULATED_BACKGROUND_TASK_EXECUTION_TIME_S)
 
 
@@ -245,7 +247,7 @@ async def background_execution_trailers_asgi(scope, receive, send):
                 "more_trailers": False,
             }
         )
-        # Wrapped send() ends the SERVER span before returning; record that instant.
+        # Record when the background task starts; the server span must already be ended by now.
         scope["background_task_start_ns"] = time.time_ns()
         time.sleep(_SIMULATED_BACKGROUND_TASK_EXECUTION_TIME_S)
 
@@ -617,10 +619,9 @@ class TestAsgiApplication(AsyncAsgiTestBase):
         span_list = self.get_finished_spans()
         server_span = span_list[-1]
         assert server_span.kind == SpanKind.SERVER
-        span_duration_nanos = server_span.end_time - server_span.start_time
         self.assertLessEqual(
-            span_duration_nanos,
-            _SIMULATED_BACKGROUND_TASK_EXECUTION_TIME_S * 10**9,
+            server_span.end_time,
+            self.scope["background_task_start_ns"],
         )
 
     async def test_exclude_internal_spans(self):
@@ -672,7 +673,6 @@ class TestAsgiApplication(AsyncAsgiTestBase):
         span_list = self.get_finished_spans()
         server_span = span_list[-1]
         assert server_span.kind == SpanKind.SERVER
-        self.assertIsNotNone(server_span.end_time)
         self.assertLessEqual(
             server_span.end_time,
             self.scope["background_task_start_ns"],
--- a/instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
+++ b/instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
@@ -366,6 +366,7 @@ class TestAsgiApplication(AsyncAsgiTestBase):
         # Ensure modifiers is a list
         modifiers = modifiers or []
         # Check for expected outputs
+        self.assertTrue(outputs, "ASGI application produced no response messages")
         response_start = outputs[0]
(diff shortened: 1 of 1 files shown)


instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
```diff
@@ -670,10 +672,10 @@ def add_body_and_trailer_span(expected: list):
         span_list = self.get_finished_spans()
         server_span = span_list[-1]
         assert server_span.kind == SpanKind.SERVER
-        span_duration_nanos = server_span.end_time - server_span.start_time
+        self.assertIsNotNone(server_span.end_time)
```
henry3260: nit: I think this assertion is redundant. `get_finished_spans() `only returns spans that already went through `Span.end()`, and end() sets `_end_time` before handing the span to the processor/exporter, so `end_time` can't be None here.

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
```diff
@@ -245,6 +245,8 @@ async def background_execution_trailers_asgi(scope, receive, send):
                 "more_trailers": False,
             }
         )
+        # Wrapped send() ends the SERVER span before returning; record that instant.
```
henry3260:
```suggestion
        # Record when the background task starts; the server span must already be ended by now.
```
the comment describes the middleware internals rather than what this line records and this timestamp is the start of the background task, not the span's end time

henry3260 (commented):
Thanks for your fix :)

`test_background_execution` above has the exact same assertion pattern and the same root cause, it just hasn't been unlucky on CI yet. Could we fix it in this PR too? thanks!

aabmass (approved):
@Shriprasad-P do you have time to address the feedback today? Otherwise I'll probably push the changes as most PRs are failing CI

Shriprasad-P:
Thanks Aaron, yes, I can address it now. I'll update
test_background_execution to use the same deterministic boundary check,
remove the redundant end_time assertion, and incorporate Henry's comment
suggestion. I'll run the ASGI tests and push the changes shortly.

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
Shriprasad-P: Removed the redundant `end_time` assertion as suggested. Since the span comes from `get_finished_spans()`, the ordering assertion can use `end_time` directly.

instrumentation/opentelemetry-instrumentation-asgi/tests/test_asgi_middleware.py
Shriprasad-P: Updated the comment to describe the timestamp itself as the start of background execution.

Shriprasad-P:
Updated `test_background_execution` as well. It now records the start of background execution and verifies that the server span ended before that boundary, matching the deterministic approach used for the trailers case.

emdneto (approved):

lzchen (approved):

aabmass:
@Shriprasad-P the test is unfortunately still failing here

PR #5020, opened 2026-09-01
exceptions: use the `exception` event name per semantic conventions

### LINKED ISSUE #4972: exceptions: review reported semantic conventions
I was reading this https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/ and it looks like semconv suggests to send the exceptions data as attributes instead of body as we are doing now in the exceptions instrumentation.

We can add the attributes, not sure we should be removing the body to not break any user.

cc @iblancasa

iblancasa:
# Description

`opentelemetry-instrumentation-exceptions` now emits the `exception` event name on every record, instead of deriving it from the exception type (`sys`/`threading` hooks) or the asyncio context message.

Fixes #4972

## Type of change

# How Has This Been Tested?

- [x] Updated and ran the package unit tests

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-exceptions/src/opentelemetry/instrumentation/exceptions/__init__.py
```diff
@@ -169,12 +187,15 @@ def _wrap_asyncio_call_exception_handler(
         if context:
             exc = context.get("exception")
             if isinstance(exc, BaseException):
+                # Exceptions raised without arguments stringify to "", so the
+                # event loop message is the only description left to record.
                 message = context.get("message")
+                body = str(exc) or (str(message) if message else "")
```
henry3260: Was the body fallback intentional? It's not part of #4972 and it makes the body inconsistent: `ValueError("boom")` drops the loop message entirely, while `ValueError()` puts the loop message in the body, where it no longer matches `exception.message`. I'd keep `body=str(exc)` here and handle the loop message separately if it's worth keeping.

iblancasa: I guess you are right. I've removed it, so body=str(exc) unconditionally again and the asyncio path no longer touches the body at all.

I aded this because the loop message used to be the event name on the asyncio path, so switching to the constant exception leaves it with nowhere to go. It's now dropped entirely... and I was trying to avoid that.

instrumentation/opentelemetry-instrumentation-exceptions/src/opentelemetry/instrumentation/exceptions/__init__.py
```diff
@@ -17,6 +17,17 @@
 This instrumentation captures uncaught process exceptions, uncaught thread
 exceptions, and unhandled asyncio task exceptions and emits them as
 OpenTelemetry logs.
+
+Emitted records follow the `exception log semantic conventions
+<https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/>`_:
+the exception instance is handed to the logs API, which records
```
emdneto, line 23:
```suggestion

Emitted records follow the `exception log semantic conventions
<https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/>`_:
the exception instance is passed to the Logs API, and the configured OpenTelemetry SDK derives
```

emdneto (commented):
one small comment to make it clear. We assume users have configured either the upstream SDK or another SDK implementation that derives them in the same way.

PR #5025, opened 2026-09-02
Clarify error message when there is a `DependencyConflict`

DylanRussell:
# Description

Updates the log messages that occur when the `instrument` method fails due to `DependencyConflict` to be more clear..

Also raised the severity of the error message that occurs on the auto instrumentation path when the library is installed with an incompatible version

## Type of change

# How Has This Been Tested?

Unit tests

- [x] No.

# Checklist:

### DIFF
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
@@ -88,7 +88,7 @@ def _load_instrumentors(distro):
             entry_point_dist = entry_point_finder.dist_for(entry_point)
             conflict = get_dist_dependency_conflicts(entry_point_dist)
             if conflict:
-                _logger.debug(
+                _logger.error(
                     "Skipping instrumentation %s: %s",
                     entry_point.name,
                     conflict,
@@ -103,7 +103,7 @@ def _load_instrumentors(distro):
             # returning a DependencyConflict. Keeping this error handling in case custom
             # distro and instrumentor behavior raises a DependencyConflictError later.
             # See https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3610
-            _logger.debug(
+            _logger.error(
                 "Skipping instrumentation %s: %s",
                 entry_point.name,
                 exc.conflict,
--- a/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
+++ b/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
@@ -88,11 +88,10 @@ def _load_instrumentors(distro):
             entry_point_dist = entry_point_finder.dist_for(entry_point)
             conflict = get_dist_dependency_conflicts(entry_point_dist)
             if conflict:
-                _logger.error(
-                    "Skipping instrumentation %s: %s",
-                    entry_point.name,
-                    conflict,
-                )
+                if conflict.is_version_conflict:
+                    _logger.error(conflict.format_message(entry_point.name))
+                else:
+                    _logger.debug(conflict.format_message(entry_point.name))
                 continue
 
             # tell instrumentation to not run dep checks again as we already did it above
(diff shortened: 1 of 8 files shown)


opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -65,6 +68,13 @@ def _load_distro() -> BaseDistro:
     return DefaultDistro()
 
 
+def _log_conflict(entry_point_name: str, conflict: DependencyConflict) -> None:
+    if conflict.is_version_conflict:
+        _logger.error(conflict.format_message(entry_point_name))
+    else:
+        _logger.debug(conflict.format_message(entry_point_name))
```
aabmass, line 75: This branch is not conflicted, but we log out a message still? I guess I'm confused how a DependencyConflict object can be not conflicted

opentelemetry-instrumentation/src/opentelemetry/instrumentation/dependencies.py
```diff
@@ -53,6 +53,35 @@ def __init__(
...
+                return (
+                    f'{instrumentor_name} instruments any of "{self.required_any}", '
+                    f'but currently installed version(s) ("{self.found_any}") fall outside of that range, '
+                    f"so nothing can be instrumented."
+                )
+            return (
+                f'{instrumentor_name} requires any of "{self.required_any}", '
+                f"but none are installed, so nothing can be instrumented."
+            )
+        return str(self)
```
aabmass, line 83: Do we need both `__str__` and `format_message()` or they could be one thing?

aabmass (approved):

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
herin049: I think the idea is that a `DependencyConflict` can be created and not have version conflicts if it doesn't have all of the required packages installed (or any of the "any" requirements).

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
```diff
@@ -65,6 +68,13 @@ def _load_distro() -> BaseDistro:
     return DefaultDistro()
 
 
+def _log_conflict(entry_point_name: str, conflict: DependencyConflict) -> None:
```
herin049, line 71: Can we just make this a method on `DependencyConflict` itself that accepts a `Logger` instance?

DylanRussell: yeah good idea.. done

opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py
DylanRussell: correct.. basically this branch is for when the instrumentation is installed but not the package it's instrumenting

opentelemetry-instrumentation/src/opentelemetry/instrumentation/dependencies.py
DylanRussell: yeah switched to just use format_message

opentelemetry-instrumentation/src/opentelemetry/instrumentation/dependencies.py
```diff
@@ -53,10 +57,44 @@ def __init__(
         self.required_any = required_any
         self.found_any = found_any
 
+    @property
+    def is_version_conflict(self) -> bool:
+        """False when the dependency is not installed at all, rather than installed at an incompatible version."""
+        return bool(self.found or self.found_any)
+
+    def log(self, logger: Logger, instrumentor_name: str) -> None:
```
herin049, line 65: These methods are being brought into the public namespace. Is this intentional? These seem internal only, maybe we should make them private?

opentelemetry-instrumentation/tests/test_dependencies.py
```diff
@@ -264,5 +264,67 @@ def version_side_effect(package_name):
         self.assertTrue(isinstance(conflict, DependencyConflict))
         self.assertEqual(
             str(conflict),
-            '''DependencyConflict: requested any of the following: "['bar~=2.0; extra == "instruments-any"', 'baz~=3.0; extra == "instruments-any"']" but found: "[]"''',
+            """This instrumentation requires any of "bar~=2.0; extra == "instruments-any", baz~=3.0; extra == "instruments-any"", but none are installed, so nothing can be instrumented.""",
         )
+
+    def test_dependency_conflict_format_message_required_version_mismatch(self):
```
herin049, line 270: Nit: We could maybe make these tests a bit more concise by parameterizing them.

herin049 (approved):
Lgtm, just two small comments

opentelemetry-instrumentation/tests/test_dependencies.py
DylanRussell: Added parameterization but not sure it's any more concise..

opentelemetry-instrumentation/src/opentelemetry/instrumentation/dependencies.py
DylanRussell: Not intentional, made them private

PR #4759, opened 2026-06-30
Drop elasticsearch instrumentation

xrmx:
# Description

The ElasticSearch Python client has a native OTel instrumentation since 8.13, released in 2014. Since January 2026 elasticsearch 7 is not supported anymore and so every supported client have OTel support and we can drop our instrumentation.

Fixes #4716
Closes #1205
Closes #2542
Closes #2021
Closes #1633
Closes #3034
 

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [x] Followed the style guidelines of this project
- [x] Changelogs have been updated
- [ ] Unit tests have been added
- [ ] Documentation has been updated

PR #5026, opened 2026-09-02
Allow per-package patch releases of the elasticsearch instrumentation

ocelotl:
# Description

The elasticsearch instrumentation was dropped from `main` by #4759, so `0.64b0` is its
last published version. That version pins `opentelemetry-instrumentation == 0.64b0` and
`opentelemetry-semantic-conventions == 0.64b0`, and those exact pins break any environment
that still installs the package:

* with nothing else constrained, the resolver silently pulls the whole OpenTelemetry
  install back to `0.64b0`, including instrumentations the user did not intend to downgrade
* with a current instrumentation pinned, for example `opentelemetry-instrumentation-requests==0.65b0`,
  resolution fails outright

Fixing that needs one more patch release, `0.64b1`, that loosens the two pins. This PR only
makes such a release possible. It adds `opentelemetry-instrumentation-elasticsearch` to the
`package` choice input of the `[Package] Prepare patch release` and `[Package] Release`
workflows, and records the arrangement in `RELEASING.md`. No package code changes here.

## Maintainer action needed

I do not have push access to this repository, so I cannot create the release branch myself.
Could a maintainer create

```
package-release/opentelemetry-instrumentation-elasticsearch/v0.64bx
```

from tag `v0.64b0`? The tag is the only possible source, because #4759 deleted the package
from `main`. Once that branch exists I will open the follow-up PR against it with the pin
change, which also has to mirror these two workflow entries onto the branch, since
`workflow_dispatch` reads its input definition from the ref the workflow runs against.

## Type of change

# How Has This Been Tested?

Both failure modes were reproduced with `uv pip compile` against PyPI, and the loosened
pins were confirmed to resolve against the current release before proposing this route.
The workflow files themselves have no test; both were checked to still parse and to list
ten packages.

- [ ] Yes. - Link to PR:
- [x] No.

# Checklist:

### DIFF
--- a/.changelog/5026.changed
+++ b/.changelog/5026.changed
@@ -1,0 +1,1 @@
+`opentelemetry-instrumentation-elasticsearch`: allow patch releases
--- a/.github/workflows/package-prepare-patch-release.yml
+++ b/.github/workflows/package-prepare-patch-release.yml
@@ -10,6 +10,7 @@ on:
         - opentelemetry-resource-detector-azure
         - opentelemetry-resourcedetector-gcp
         - opentelemetry-sdk-extension-aws
+        - opentelemetry-instrumentation-elasticsearch
         - opentelemetry-instrumentation-openai-v2
         - opentelemetry-instrumentation-openai-agents-v2
         - opentelemetry-instrumentation-vertexai
--- a/.github/workflows/package-release.yml
+++ b/.github/workflows/package-release.yml
@@ -10,6 +10,7 @@ on:
         - opentelemetry-resource-detector-azure
         - opentelemetry-resourcedetector-gcp
         - opentelemetry-sdk-extension-aws
+        - opentelemetry-instrumentation-elasticsearch
         - opentelemetry-instrumentation-openai-v2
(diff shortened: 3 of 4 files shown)


ocelotl:
@xrmx please take a look

xrmx:
@ocelotl I think we have already discussed this at the Python SIG and let met rehash that I'm against ad-hoc handling of this of any other package. When we'll have a loosen dependency story we'll have this kind of issues sorted out.

ocelotl:
@xrmx so, just to make sure I understand your objection. You are not objecting the loosening of dependencies as a whole. You are objecting treating this instrumentation separately?

lzchen:
@ocelotl 

I think the guidance in the last SIG when this issue was brought up was that we would wait until semantic conventions and `opentelemetry-instrumentation` are marked stable before loosening dependencies. If it isn't an extremely desired we don't want to set precedence of just doing one-off patch releases like this.

ocelotl:
@open-telemetry/opentelemetry-python-contrib-maintainers what should we do with this one? is there a way to put it in "standby" or do you want me to close it?

/dashboard route:reviewers

PR #5029, opened 2026-09-05
fix(aiohttp-client): create the duration histograms once, not per ClientSession

### LINKED ISSUE #4991: Possible memory leak in aiohttp-client instrumentation
### Describe your environment

OS: Microsoft Windows 11 Enterprise, 10.0.26100 Build 26100
Python version: Python 3.13.13 (tags/v3.13.13:01104ce, Apr  7 2026, 19:25:48) [MSC v.1944 64 bit (AMD64)
Package version: opentelemetry-instrumentation-aiohttp-client 0.65b0

### What happened?

Possible memory leak when instrumenting aiohttp client connections. Every new connection creates a meter at https://github.com/open-telemetry/opentelemetry-python-contrib/blob/466ae4c00de77d1f54f77a9600f1d08eba2306b5/instrumentation/opentelemetry-instrumentation-aiohttp-client/src/opentelemetry/instrumentation/aiohttp_client/__init__.py#L395 and it is never cleaned up.

### Steps to Reproduce

This setup is based on the sample code from https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/aiohttp_client/aiohttp_client.html.

```
import asyncio
import aiohttp
from opentelemetry.instrumentation.aiohttp_client import (
    AioHttpClientInstrumentor
)
import tracemalloc

# Enable instrumentation
AioHttpClientInstrumentor().instrument()


# Create a session and make an HTTP get request
async def get(url):
    async with aiohttp.ClientSession() as session:
        async with session.get(url, ssl=False) as response:
            await response.text()

async def main():
    # Enable tracemalloc
    tracemalloc.start()
    before = tracemalloc.take_snapshot()

    for i in range(1000):
        print(f"Request {i+1}")
        await get("<URL>")
        await asyncio.sleep(5)  # Add a small delay

    after = tracemalloc.take_snapshot()
    stats = after.compare_to(before, 'lineno')
    for stat in stats[:10]:
        print(stat)
    tracemalloc.stop()


if __name__ == "__main__":
    asyncio.run(main())
```

### Expected Result

The expected result is no memory leaks.

### Actual Result

This is the snapshot diff
```
\venv\Lib\site-packages\opentelemetry\metrics\_internal\__init__.py:163: size=143 KiB (+143 KiB), count=1001 (+1001), average=146 B
\venv\Lib\site-packages\opentelemetry\metrics\_internal\__init__.py:581: size=112 KiB (+112 KiB), count=1000 (+1000), average=115 B
\venv\Lib\site-packages\opentelemetry\instrumentation\_semconv.py:697: size=78.1 KiB (+78.1 KiB), count=1000 (+1000), average=80 B
\venv\Lib\site-packages\opentelemetry\metrics\_internal\__init__.py:199: size=62.5 KiB (+62.5 KiB), count=1000 (+1000), average=64 B
\venv\Lib\site-packages\opentelemetry\metrics\_internal\__init__.py:498: size=54.7 KiB (+54.7 KiB), count=1000 (+1000), average=56 B
\venv\Lib\site-packages\opentelemetry\metrics\_internal\__init__.py:200: size=54.7 KiB (+54.7 KiB), count=1000 (+1000), average=56 B
\venv\Lib\site-packages\opentelemetry\metrics\_internal\__init__.py:499: size=54.5 KiB (+54.5 KiB), count=997 (+997), average=56 B
\venv\Lib\site-packages\opentelemetry\metrics\_internal\__init__.py:584: size=31.2 KiB (+31.2 KiB), count=1000 (+1000), average=32 B
```

### Additional context

_No response_

No

### Tip

bharadwaj-pendyala:
# Description

Fixes #4991. `instrumented_init` called `create_trace_config()` for every `aiohttp.ClientSession`; `create_trace_config()` calls `get_meter()` and `create_histogram()`, retaining one meter and histogram per session for the life of the process.

Leak reproduction on `a3881e6` with the default (proxy) meter provider, opening and closing sessions with no requests:

| sessions | retained meters | before fix | after fix |
|---|---|---|---|
| 100 | `_PROXY_METER_PROVIDER._meters` | 100 | 1 |
| 200 | | 200 | 1 |
| 300 | | 300 | 1 |

`_instrument()` now builds histograms once and passes them to private `_create_trace_config()`. Public `create_trace_config()` keeps its signature, docstring and behaviour, so direct callers are unaffected. This matches `requests`, `httpx` and `aiohttp-server`.

Repeated `instrument()`/`uninstrument()` cycles still retain meters: 300 cycles retain 300 meters here and in `opentelemetry-instrumentation-requests`. `get_meter()` now runs per `instrument()` call rather than per session; this is existing repository behaviour.

## Type of change

# How Has This Been Tested?

- [x] New regression test `test_instrument_creates_one_meter_for_all_sessions`: fails on `main` with `1 != 5`, passes with the fix.
- [x] `pytest tests/test_aiohttp_client_integration.py`: 40 passed, up from 39; 36 pre-existing local-environment failures before and after (released core 1.39.0/0.60b0 against this tree's 0.66b0.dev pins).
- [x] Verified `http.client.duration count=1` with an SDK `MeterProvider` and `InMemoryMetricReader` through `AioHttpClientInstrumentor` and direct `create_trace_config()`.
- [x] `ruff check` and `ruff format --check` clean on the package.

- [ ] Yes. - Link to PR:
- [x] No.

# Checklist:

AI assistance was used for this change (Claude).

### DIFF
--- a/instrumentation/opentelemetry-instrumentation-aiohttp-client/src/opentelemetry/instrumentation/aiohttp_client/__init__.py
+++ b/instrumentation/opentelemetry-instrumentation-aiohttp-client/src/opentelemetry/instrumentation/aiohttp_client/__init__.py
@@ -233,7 +233,7 @@ from opentelemetry.instrumentation.utils import (
     is_http_instrumentation_enabled,
     unwrap,
 )
-from opentelemetry.metrics import MeterProvider, get_meter
+from opentelemetry.metrics import Histogram, MeterProvider, get_meter
 from opentelemetry.propagate import inject
 from opentelemetry.semconv.attributes.error_attributes import ERROR_TYPE
 from opentelemetry.semconv.metrics import (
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry-instrumentation-aiohttp-client/src/opentelemetry/instrumentation/aiohttp_client/__init__.py
```diff
@@ -411,6 +348,31 @@ def create_trace_config(
...
+    sem_conv_opt_in_mode: _StabilityMode = _StabilityMode.DEFAULT,
+    captured_request_headers: list[str] | None = None,
+    captured_response_headers: list[str] | None = None,
+    sensitive_headers: list[str] | None = None,
+    duration_histogram_old: Histogram | None = None,
+    duration_histogram_new: Histogram | None = None,
+) -> aiohttp.TraceConfig:
+    schema_url = _get_schema_url(sem_conv_opt_in_mode)
+
+    tracer = get_tracer(
```
herin049, line 370: Should we fix this potential leak as well?

bharadwaj-pendyala: I checked, and that one doesn't retain. `_ProxyMeterProvider.get_meter` appends every meter to `self._meters` and nothing drops them, which is what made the meter grow. `ProxyTracerProvider.get_tracer` returns a fresh `ProxyTracer` and keeps no reference, and an SDK `TracerProvider` dedupes on `InstrumentationScope`. Measured on this branch with 300 sessions: 300 live `ProxyTracer` objects, 0 once the sessions close, `len(tp._tracers) == 1` under the SDK.

Hoisting it to match `requests`/`httpx` does move provider resolution into `instrument()`. With `OTEL_PYTHON_TRACER_PROVIDER=sdk_tracer_provider` and `set_tracer_provider()` called after `instrument()`, I get 1 exported span today and 0 with the tracer hoisted, plus "Overriding of current TracerProvider is not allowed". `requests` already behaves that way, so it's defensible on its own, but it's a behaviour change rather than a leak fix.

herin049: I think it's OK if we align the HTTP clients so that they exhibit the same behavior. Feel free to add a note in the changelog.

bharadwaj-pendyala: Pushed in 5070e10. `_instrument` resolves the tracer once now and passes it in, and `create_trace_config` still resolves per call since callers invoke that one themselves. The changelog fragment records that a provider installed after `instrument()` no longer gets picked up.

CI hasn't run on this PR yet, both pushes are stuck at `action_required` waiting on workflow approval.

PR #5031, opened 2026-09-06
opentelemetry-instrumentation-asgi: set error.type when the application raises

### LINKED ISSUE #2699: opentelemetry-instrumentation-asgi: emit`error.type` attribute when having errors
### Describe your environment

instrumentation-asgi: 0.47b0.dev

### What happened?

instrumentation-asgi isn't setting `error.type` and we should set it in the new semconv as defined [here](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#http-server-semantic-conventions)

### Steps to Reproduce

Run any asgi example which raises exception before receive/send request/response

### Expected Result

>If the request fails with an error before response status code was sent or received, error.type SHOULD be set to exception type (its fully-qualified class name, if applicable) or a component-specific low cardinality error identifier.

Exception present in `error.type`

### Actual Result

No `error.type` attribute set

### Additional context

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#http-server-semantic-conventions

None

chrikrah:
Fixes #2699.

`OpenTelemetryMiddleware.__call__` wrapped the application in `try`/`finally` with no `except`, so an
exception escaping the application left neither the server span nor `http.server.request.duration` with
`error.type`. The sibling wsgi, pyramid and aiohttp-server instrumentations all set it; ASGI was the
straggler, and it is the one every FastAPI, Starlette, Quart and Django-ASGI app routes through.

One clause, one guard:

```python
if _report_new(self._sem_conv_opt_in_mode) and ERROR_TYPE not in attributes:
```

The guard exists because `attributes` is the same dict `_set_status` writes into. When a response has
already sent an error status, `_set_status` has derived `error.type` from it (`"500"`), and note [4] accepts
either that or the exception type. Keeping the status-derived value means a streaming response that sends
5xx and then fails still reports the code its consumers already query on, rather than silently switching to
`ValueError`.

Cancellation is not reported, and that falls out rather than being decided: `asyncio.CancelledError` derives
from `BaseException`, so `except Exception` never sees it, which is the same line `trace.use_span` draws.

### Two things a reviewer should weigh

**The guard is a compatibility choice, not a spec requirement.** wsgi has no precondition at all: it writes
`req_attrs[ERROR_TYPE] = type(ex).__qualname__` unconditionally and does clobber a status-derived `"500"`.
Note [4] sanctions both values, so consistency with wsgi is a legitimate alternative to what I did here.
I chose the non-clobbering option because it changes no existing signal. Say the word and I will drop it.

**A completed response that then raises records `error.type` on the metric but not on the span**, because
`otel_send` has already ended the span by then. wsgi has the identical property. Gating the metric write on
`span.is_recording()` would be worse, since it would make a metric depend on sampling.

### Deliberately not in scope

An `ExceptionGroup` reports `error.type="ExceptionGroup"` for every distinct underlying failure. Unwrapping
needs a policy answer for the mixed-leaf case, and whatever is decided applies equally to wsgi, pyramid and
aiohttp-server, so it does not belong in an ASGI-only diff. Happy to open a follow-up.

### Tests

Five in `TestAsgiApplicationRaisingError`, covering the three semconv modes and both directions of the
guard:

- `http` and `http/dup`: `error.type == "ValueError"` on the span and on `http.server.request.duration`,
  asserting the value type too, and staying off the old `http.server.duration` in dup mode.
- default mode: `http.server.request.duration` is not emitted at all, and the old duration metric carries no
  `error.type`.
- an error status already sent (500, one `more_body=True` chunk, then raise) keeps `error.type == "500"`.
- an ok status already sent (200, one chunk, then raise) still gets `error.type == "ValueError"`.

Those last two pin the guard from opposite directions: removing it fails the first, broadening it back to
status presence fails the second.

`tox -e py312-test-instrumentation-asgi`: 99 passed, up from 93 on `main` at `f778897`. pylint 10.00/10, pyright clean, ruff clean.
wsgi 52, pyramid 32, aiohttp-server 34, starlette 62, fastapi 109 passed 8 skipped, all unchanged.


henry3260 (approved):
lgtm from my side

chrikrah:
@xrmx @aabmass could one of you take a look, or point me at whoever owns this instrumentation?
`.github/component_owners.yml` has no `asgi` entry, so I had no owner to route it to. @henry3260
approved it on 7 September.

I am raising it now for one reason. The stale bot closed a previous attempt at this same fix. I would
rather not lose this one the same way.

The change is 18 added lines plus tests. `error.type` is now set on the ASGI server span and on the
duration metric when the application raises. `main` never set it. The precondition is
`ERROR_TYPE not in attributes`. So a handler that sends 5xx and then raises keeps the status-derived
`"500"`. One that sends 200 and then raises gets `"ValueError"`. That second case matches the worked
server example at `http-spans.md:967`, which carries status 201 and `error.type` together.

One thing I would rather put in front of you than leave to be found. wsgi has no such precondition. It
does overwrite a status-derived `"500"`. Note [4] at `http-spans.md:585-590` allows either value, so
this guard is a backward-compatibility choice, not a spec requirement. Ask me to match wsgi and I
will. That drops the condition, its comment, and `test_asgi_error_type_keeps_response_status_new_semconv`,
which exists only to pin it.

Proven locally with `tox -e py312-test-instrumentation-asgi`: 99 passed.

Shriprasad-P (changes requested):
I checked this guard against the HTTP semantic-convention behavior.
Mention that preserving the existing status-derived error.type for a 5xx response seems reasonable to you.
Mention that the spec allows either the status-code string or exception type in that case.
Note that this also avoids changing the telemetry ASGI already emits after a 5xx response.
State that you don't see this condition as a blocker.

Shriprasad-P (commented):
## Review

Setting `error.type` on uncaught app exceptions for the new HTTP semconv path looks correct, and preserving an already-set status-derived `error.type` for streamed 5xx is a thoughtful edge case.

Appreciate the notes on `CancelledError` / `BaseException` behavior matching `trace.use_span`. Tests look thorough from the diff size; LGTM directionally.

chrikrah (commented):
Thanks for the second look.

On the 5xx case you examined, the server-span note allows either value.
`semantic-conventions/docs/http/http-spans.md:590` (`a11c510`), under `## HTTP server`, says `error.type`
"SHOULD be set to the status code number (represented as a string), an exception type (if thrown) or a
component-specific error identifier".

One precision on the guard, since it is a single line: it writes only when `error.type` is absent, so it
never overwrites. What it turns on is an existing `error.type`, not a status having gone out.
`test_asgi_error_type_keeps_ok_response_status_new_semconv` covers the difference: the app sends 200,
then fails, and `error.type` becomes `"ValueError"`, because a 200 records no `error.type` to suppress.

@xrmx @aabmass two reviews now read this as correct. It still needs an Approver, and
`.github/component_owners.yml` has no `asgi` entry, so I have nobody to route it to.

chrikrah (commented):
Answering the last line of your first review directly: I do not see that condition as a blocker either,
and if you would rather match `wsgi` I will drop it.

The head moved after both of your reviews. The same two commits are replayed onto `main` at `f778897`,
head `f0cd99e`, and the diff is unchanged.

From your side an approving review is the one thing that supersedes the earlier `Changes requested`. If
this settles it, that clears it. If it does not, tell me what is still open.

Issue #4975, opened 2026-08-19
http server instrumentations: make it possible to sample according to captured headers


xrmx:
I would like to do sampling depending on any captured request header.

This should be doable with the Composite Rule Based Sampler but we need instrumentations to collect these headers before the span is created. AFAICS no http server instrumentation does that. The capturing of headers is opt-in so any eventual performance hit will be paid only if requested.

Instrumentations to fix:

- [ ] aiohttp-server, see `create_aiohttp_middleware`
- [ ] django, see `_DjangoMiddleware.process_request`
- [ ] asgi, see `OpenTelemetryMiddleware.__call__`
- [ ] falcon, see `_InstrumentedFalconAPI.__call__`
- [ ] flask, see `_wrapped_before_request`
- [ ] pyramid, see `_before_traversal`
- [ ] tornado, see `_start_span`
- [ ] wsgi, see `OpenTelemetryMiddleware.__call__`

yashaskn8:
Hi @xrmx, I'd like to take the ASGI part of this issue.

I traced the current flow in `OpenTelemetryMiddleware.__call__`. The regular request attributes are collected before `_start_internal_or_server_span(...)`, but the opt-in request headers from `collect_custom_headers_attributes(...)` are currently collected only after the span has already been created and are then added with `set_attributes()`. Because the sampler receives the attributes passed at span creation time, those captured headers are not currently available for sampling decisions.

My plan is to make the configured and sanitized request-header attributes available in the initial attributes used for SERVER span creation, while preserving the current opt-in behavior and avoiding any unintended change for INTERNAL spans.

I’d also add a focused regression test with a sampler that captures the attributes passed to `should_sample`, so we can verify that a configured request header is actually visible during the sampling decision.

Would a scoped PR for the ASGI instrumentation be a good first part of #4975?

PR #5032, opened 2026-09-07
opentelemetry-instrumentation-wsgi: expose captured request headers to samplers

ghdtjdwn:
# Description

Pass configured, sanitized request headers as initial WSGI SERVER span attributes so samplers can use them when deciding whether to record a request. Preserve the exclusion of captured headers from INTERNAL spans and metrics.

The existing `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST` setting opts into header capture and makes those same headers available to samplers. When unset or empty, return before constructing the sanitizer or iterating the WSGI environment to build the headers dictionary.

Related to #4975; addresses the WSGI item only.

## Type of change

# How Has This Been Tested?

- [x] WSGI tests: 55 passed on each of Python 3.12 and 3.14 with `CORE_REPO_SHA=5aa2f8f69a5b619e94ded29dd235a7285a591177 tox -e py312-test-instrumentation-wsgi,py314-test-instrumentation-wsgi`.
- [x] Tests cover sampler inputs and sample/drop decisions, root and remote-parent requests, redaction, disabled capture, INTERNAL spans, and metric attribute exclusion. The new regression verifies that unset and empty capture settings skip environment iteration; it fails before the short-circuit and passes afterward.
- [x] `tox -e lint-instrumentation-wsgi` with the same Core SHA, pre-commit checks on changed files, package Ruff checks, and `python scripts/check_license_header.py` passed.

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
```diff
@@ -1081,51 +1123,125 @@ def test_custom_request_headers_non_recording_span(self):
...
             not_expected = {
                 "http.request.header.custom_test_header_1": ("Test Value 1",),
             }
             for key, _ in not_expected.items():
+                self.assertNotIn(key, sampler.attributes)
                 self.assertNotIn(key, span.attributes)
 
+    @mock.patch.dict(
+        "os.environ",
+        {OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST: ""},
```
emdneto, line 1216: @xrmx was your initial idea to re-use the existing env var or adding a new one?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -660,6 +662,9 @@ def __call__(self, environ: WSGIEnvironment, start_response: StartResponse):
             self._sem_conv_opt_in_mode,
         )
 
+        if trace.get_current_span() is trace.INVALID_SPAN:
```
emdneto, line 669: Could we short-circuit inside `collect_custom_request_headers_attributes()` before building the headers dictionary?

Something like: 

```
captured_headers = get_custom_headers(
    OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST
)
if not captured_headers:
    return {}
```

emdneto (commented):
One small comment for now

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
ghdtjdwn: Implemented in 0173150. The helper now reads the capture setting first and returns `{}` when it is unset or empty, before constructing `SanitizeValue` or iterating the WSGI environment. The configured path reuses the parsed header list.

Added a regression that checks both unset and empty settings and verifies `environ.items()` is not called. It fails before the change and passes afterward. All 55 WSGI tests pass on each of Python 3.12 and 3.14; lint and formatting checks also pass.

instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
ghdtjdwn: This PR reuses `OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST`, based on #4975 describing sampling from headers that are already captured through opt-in configuration. The configured headers are sanitized before reaching the sampler, using the existing redaction settings. With the capture setting unset or empty, no request headers are collected.

I've made that behavior explicit in the PR description; the original intent still needs confirmation from @xrmx.

xrmx: My idea was just to move the collection of captured headers before the span creation. No other changes in behavior with the current code

ghdtjdwn: Thanks for clarifying, @xrmx. That matches the current patch at 0173150: the existing configured, sanitized request headers are passed into WSGI SERVER span creation, using the same capture setting and redaction rules. INTERNAL spans still do not receive captured headers. The early return for unset or empty capture configuration is the optimization @emdneto requested. No additional behavior change is intended.

ghdtjdwn:
Synced with `main` at 6a264578; the PR head is now 50e3d999. Re-ran all 55 WSGI tests on each of Python 3.12 and 3.14 against Core `ab2267483ae402a7de09f89d999af11fa42f613b`. Package lint, changed-file Ruff/format checks, and license headers pass.

The [CI](https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/34578269241) and [changelog](https://github.com/open-telemetry/opentelemetry-python-contrib/actions/runs/34578268301) runs for this head are waiting for maintainer approval. Could a maintainer approve those runs when available?

instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py
```diff
@@ -660,6 +666,9 @@ def __call__(self, environ: WSGIEnvironment, start_response: StartResponse):
...
+            req_attrs.update(collect_custom_request_headers_attributes(environ))
```
herin049: Should we just move this into `collect_request_attributes`?

ghdtjdwn: Done in 55d3d07. `collect_request_attributes()` now accepts `capture_custom_headers: bool = False` to include configured, sanitized request headers when starting a SERVER span, while preserving the existing behavior for callers that omit it (default `False`).

Added unit test coverage in `test_request_attributes_capture_custom_headers_only_when_enabled` to verify headers are captured only when enabled and that `environ.items()` is not called when unconfigured. All WSGI tests pass on Python 3.12.

ghdtjdwn:
Addressed @herin049's review feedback in 55d3d07: moved captured request header collection into `collect_request_attributes(..., capture_custom_headers=True)` and added regression tests. All WSGI tests pass locally on Python 3.12.

Could a maintainer please approve the CI workflows for head `55d3d07` when convenient?

ghdtjdwn:
Fixed the `misc / docs` failure in 97a292b. Sphinx could not resolve the `AnyValue` forward reference inside `AttributeValue` on the public `collect_request_attributes` / `add_response_attributes` signatures, so I restored the pre-existing annotations; the diff against `main` is now only the `capture_custom_headers` change. Verified locally by building the WSGI docs page with `-W` (the warning reproduces on the previous head and is gone now); WSGI tests and lint pass.

Could a maintainer approve the CI and changelog workflows for `97a292b`?

instrumentation/opentelemetry-instrumentation-wsgi/tests/test_wsgi_middleware.py
```diff
@@ -1036,6 +1043,41 @@ def test_mark_span_internal_in_presence_of_span_from_other_framework(self):
             self.assertEqual(parent_span.context.span_id, span_list[0].parent.span_id)
 
 
+_CUSTOM_REQUEST_HEADER = f"{HTTP_REQUEST_HEADER_TEMPLATE}.custom_test_header_1"
+
+
+class _RequestHeaderSampler(Sampler):
```
xrmx, line 1049: I've asked on another PR to have a common sampler in test utils instead of having one for each instrumentation https://github.com/open-telemetry/opentelemetry-python-contrib/pull/5073#discussion_r4061037725

ghdtjdwn: Thanks for the pointer. I opened open-telemetry/opentelemetry-python#5681 adding a shared `CapturingSampler` to `opentelemetry-test-utils`, generalized so both this WSGI test and the Django one in #5073 can use it. I have the WSGI test switched over locally (56 tests pass against that branch) and will push that here once the core PR lands, so this PR stays green against core `main` in the meantime.

ghdtjdwn:
/dashboard route:reviewers

The open thread is waiting on open-telemetry/opentelemetry-python#5681, which adds the shared CapturingSampler. I'll switch this test over as soon as it lands.

PR #5043, opened 2026-09-10
opentelemetry-instrumentation-sqlalchemy: stop appending the database name to db.operation.name

### LINKED ISSUE #5042: sqlalchemy: db.operation.name carries the database name as well as the operation
`_operation_name` in the SQLAlchemy instrumentation returns the operation token joined to the database name, and that same string is used for two different consumers.

At `engine.py:377` it becomes the span name, which is correct, since the span convention is `<operation> <target>`.

At `engine.py:348` it is passed to `_set_db_operation`, which writes it verbatim to `db.operation.name` and to the legacy `db.operation` (`_semconv.py:612`). Stable semconv defines `db.operation.name` as the name of the operation being executed, and the database name is not part of that.

The path, for a sqlite engine:

```python
engine = create_engine("sqlite:///:memory:")
SQLAlchemyInstrumentor().instrument(engine=engine, tracer_provider=tp)
engine.connect().execute(text("SELECT 1 + 1"))
# db.operation.name gets "SELECT :memory:"; semconv wants "SELECT"
```

This is on main today and is independent of #4943, which only stops an IndexError in the same function and leaves this behaviour alone. Filing it here as @xrmx asked on that PR.

henry3260:
# Description

`EngineTracer._operation_name` returns `<operation> <db_name>` (e.g. `SELECT :memory:`), which is the correct span name per the `<operation> <target>` span convention. The same string was also passed to `_set_db_operation`, so `db.operation.name` (and the legacy `db.operation`) carried the database name too. Stable semconv defines `db.operation.name` as just the name of the operation being executed.

This PR splits the operation token extraction into a new `_operation(statement)` helper, keeps `_operation_name` for the span name, and sets `db.operation.name` / `db.operation` from `_operation` only. `_set_db_client_span_attributes` no longer needs `db_name`.

Fixes #5042

## Type of change

# How Has This Been Tested?

- [x] Updated the existing `db.operation` and `db.operation.name` assertions in `test_sqlalchemy.py` from `SELECT :memory:` to `SELECT`
- [x] `uv run tox -e py312-test-instrumentation-sqlalchemy-2` (59 passed, 7 skipped)
- [x] `uv run pre-commit run ruff --files ...` on the changed files

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/engine.py
```diff
@@ -307,16 +307,27 @@ def remove_all_event_listeners(cls):
...
-        if isinstance(statement, str):
-            # otel spec recommends against parsing SQL queries. We are not trying to parse SQL
-            # but simply truncating the statement to the first word. This covers probably >95%
-            # use cases and uses the SQL statement in span name correctly as per the spec.
-            # For some very special cases it might not record the correct statement if the SQL
-            # dialect is too weird but in any case it shouldn't break anything.
-            # Strip leading comments so we get the operation name.
-            parts.append(self._leading_comment_remover.sub("", statement).split()[0])
+        operation = self._operation(statement)
+        if operation:
```
herin049: nit:
```suggestion
        if operation := self._operation(statement):
```

herin049 (approved):

PR #5045, opened 2026-09-11
Add a policy for accepting new instrumentations

### LINKED ISSUE #5028: There is no policy regarding which instrumentations to accept
In the 2026-09-03 Python SIG meeting we discussed a topic: we don't have a policy on how to decide if we want to include a new instrumentation in the contrib repo or not.

#### Definitions

`A`: an individual from the group of regular contributors, approvers, maintainers, etc. to this repo
`B`: an individual that does not belong to `A`

Contributions are very appreciated but maintaining new instrumentations can be difficult:

1. It is a growing load on maintainers
2. Maintainers may not be experts on the library the instrumentation instruments

For example, this scenario `S` may happen: `B` adds a new instrumentation `X` that instruments library `Y`, offers to maintain it and does. But after some time `B` is no longer able to maintain `X` and from that point on `A` has to maintain `X`.

Now, let's assume `B` is proposing adding `X` to this repo. What can `A` do?

`A` can only decide to accept or reject `X`.

That decision depends a lot on `X`. Since `S` may happen, `A` cannot count on `B` to maintain `X`. `A` has limited resources (time, mostly) so `A` must decide carefully to accept `X` or not. And that decision will be made depending on several factors, how critical `X` is, how many people will use it and so on.

### What do I (@ocelotl) recommend?

I only see 2 possibilities here:

1. `A` accepts `X` **_and accepts to maintain `X`_**.
2. `A` rejects `X`.

**and accepts to maintain `X`** is the critical part here: so far `A` had counted on `B` to maintain `X`. I think `A` should reconsider that now.

#### What about `A` accepting `X` and then dropping it later if needed?

That's another possibility, but:

1. `A` first needs a policy on how to drop `X`
2. `A` could save all this trouble by simply rejecting `X` first

Something important here: `A` rejecting `X` does not mean there will never be an instrumentation for `Y`. `A` rejecting `X` only means `X` will not be _in this repo_. `X` can be in another repo.

### What does Java do?

@trask joined the SIG meeting and shared with us the experience in Java, where they have 2 repos:

`G`: https://github.com/open-telemetry/opentelemetry-java-instrumentation
`H`: https://github.com/open-telemetry/opentelemetry-java-contrib

As far as I could understand there were instrumentations in both `G` and `H` and Java maintainers are commited to maintain the ones in `G` and not drop them but are ok with dropping one from `H`.

Now I am not sure I uderstood correctly because it seems to me like only `G` has instrumentations.

@trask please correct me here.

In any case, we need a policy to decide which instrumentations to accept, and this policy should be written down and define who will maintain such instrumentations. Please add your comments ✌️ 😎 

@tammy-baylis-swi 
@herin049 
@xrmx
@lzchen 
@aabmass 
@lmolkova 
@pmcollins 
@carlosalberto

ocelotl:
# Description

Adds a policy to `CONTRIBUTING.md` describing how the approvers and maintainers of this repository decide whether to accept a new instrumentation, and who maintains it afterwards.

The policy has two possible outcomes for a proposed instrumentation: it is accepted and the approvers and maintainers take on maintaining it themselves, or it is rejected. There is no outcome in which somebody outside those groups is counted on to maintain an accepted instrumentation. The section also states what a rejection does and does not mean, and asks for new instrumentations to be proposed in an issue before they are written.

The section "Expectations from contributors" contradicted this, since it said the community expects contributors to keep maintaining the instrumentations they contribute and that it is necessary to rely on their expertise. It now says that staying involved is welcome and valuable but not required, and that the responsibility lies with the approvers and maintainers alone.

This is the policy discussed in the 2026-09-03 Python SIG meeting. Wording is open to discussion, in particular whether the decision criteria should be written down, which this PR deliberately leaves out.

Fixes #5028

## Type of change

# How Has This Been Tested?

- [x] Documentation only change, no code affected. Checked that every in-page anchor link in `CONTRIBUTING.md` resolves to an existing heading, including the new index entries and cross references.

- [ ] Yes. - Link to PR:
- [x] No.

# Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -368,30 +368,31 @@ The continuous integration overrides that environment variable with as per the c
 
 ## Policy for accepting new instrumentations
 
-Contributions are very appreciated. Every proposal to add a new instrumentation
-to this repository is considered on its own merits, and the outcome can be
-acceptance or rejection. This section describes how that decision is made and
-why it is made that way.
+Every proposal to add a new instrumentation to this repository is considered on
+its own merits, and the outcome can be acceptance or rejection. This section
+describes how that decision is made and why it is made that way.
 
 ### Why this policy exists
 
-Adding an instrumentation to this repository is not a one time event, it is a
-permanent, recurring cost:
+Consider what happens after an instrumentation is added to this repository:
 
-* It is a growing load on the approvers and maintainers of this repository.
+* Every instrumentation adds load to the approvers and maintainers of this
+  repository, and that load grows with every instrumentation that is added.
 * The approvers and maintainers are usually not experts in the library that a
   given instrumentation instruments.
 * Most of the packages in this repository are released in lockstep, so a single
   instrumentation that nobody can fix holds back or breaks the release of
   everything else.
+* A contributor who is not an approver or a maintainer of this repository
+  frequently proposes a new instrumentation, offers to maintain it, and does
+  maintain it for some time. Later that person stops contributing, for
+  perfectly legitimate reasons. The instrumentation does not stop existing when
+  that happens: the approvers and maintainers of this repository inherit it,
+  including the parts of it that they do not understand.
 
-The situation that motivates this policy happens often. A contributor who is
-not an approver or a maintainer of this repository proposes a new
-instrumentation, offers to maintain it, and does maintain it for some time.
-Later that person stops contributing, for perfectly legitimate reasons. The
-instrumentation does not stop existing when that happens: the approvers and
-maintainers of this repository inherit it, including the parts of it that they
-do not understand.
+Adding an instrumentation to this repository is therefore not a one time event.
+It is a permanent, recurring cost, and it is paid by the approvers and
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -40,6 +40,11 @@ If you are using AI agents to assist with contributions, please read [AGENTS.md]
   - [Running Tests Locally](#running-tests-locally)
     - [Testing against a different Core repo branch/commit](#testing-against-a-different-core-repo-branchcommit)
   - [Style Guide](#style-guide)
+  - [Policy for accepting new instrumentations](#policy-for-accepting-new-instrumentations)
+    - [Why this policy exists](#why-this-policy-exists)
```
emdneto, line 44: I’d add another bullet here to clarify:

Before proposing new instrumentation: Contributors should first consider adding native OpenTelemetry instrumentation directly to the package or library rather than creating new instrumentation here. If that isn’t possible, they should open an issue to gauge community interest. If there is sufficient interest, the maintainers and approvers can review the proposal and discuss the next steps.

PR #5055, opened 2026-09-12
fix(grpc): invoke callable hooks that lack __name__

### LINKED ISSUE #5054: grpc: invoke callable hooks that do not expose __name__
Applications registering callable objects or ordinary functools.partial objects as gRPC client hooks silently lose all hook-provided span enrichment and receive an erroneous callback-failure log on every call. The RPC itself still runs. Named functions and bound methods are unaffected.

At `instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py:68`:

`_safe_invoke` evaluates `function_name = function.__name__` before `function(*args)` inside the same try block. A callable instance implementing __call__ but not __name__ raises AttributeError at the first statement. The handler logs that error and never invokes the hook. Both `_call_request_hook` and `_call_response_hook` explicitly accept any callable at lines 164-172, so this metadata lookup imposes an unintended extra requirement. `_BaseAioClientInterceptor` inherits these methods in _aio_client.py:46, affecting both sync and async clients. Existing hook tests supply ordinary functions, which all have __name__.

## Proposed fix

Make callback-name lookup optional in _client.py so missing __name__ does not prevent invocation, retaining the existing exception boundary and fallback label. Extend tests/test_client_interceptor_hooks.py and tests/test_aio_client_interceptor_hooks.py with callable-object and partial hooks, successful span enrichment, and throwing-callable coverage. Keep the change scoped to gRPC rather than refactoring other packages' similarly named helpers. Add a root fixed changelog fragment after contribution prerequisites are resolved.

## How to see it

Unexecuted, unambiguous Python execution trace. Define Hook with __call__ that appends its arguments to a list and no __name__ attribute. Construct OpenTelemetryClientInterceptor with this object as request_hook and response_hook, and call its two hook dispatch methods with sentinel span and message objects. callable returns True, then _safe_invoke raises AttributeError before __call__; the list stays empty instead of containing both calls. Test the same inherited dispatch on an aio interceptor. As public-path regression coverage, register callable-object and partial request hooks with GrpcInstrumentorClient and GrpcAioInstrumentorClient using the existing local test server; assert the hook's expected span attribute is present and no callback error is logged. A throwing callable should still be invoked once, logged, and prevented from breaking the RPC. These assertions fail before the fix and pass after it.

If this looks right I can push `fix/grpc-callable-hooks` on `anxkhn/opentelemetry-python-contrib` instead of opening a pull request first.

Shriprasad-P:
## Fixes
Fixes #5054

## Summary
`_safe_invoke` accessed `function.__name__` directly, so callable objects / `functools.partial` hooks raised `AttributeError` and never ran. Use `getattr(function, \"__name__\", \"<unknown>\")` so hooks without `__name__` still execute.

## Test plan
- [x] Added sync + async tests for callable objects, `functools.partial`, and exception-logging paths
- [x] Existing interceptor hook tests still pass


instrumentation/opentelemetry-instrumentation-grpc/tests/test_aio_client_interceptor_hooks.py
```diff
@@ -28,6 +29,53 @@ def response_hook_with_exception(_span, _response):
     raise Exception()  # pylint: disable=broad-exception-raised
 
 
+class CallableRequestHook:
```
herin049: These two classes are nearly identical to each other, can be consolidate to reduce duplication?

instrumentation/opentelemetry-instrumentation-grpc/tests/test_client_interceptor_hooks.py
```diff
@@ -55,6 +71,53 @@ def response_hook_with_exception(_span, _response):
     raise Exception()  # pylint: disable=broad-exception-raised
 
 
+class CallableRequestHook:
```
herin049: These is also directly code duplication here and we should try to eliminate.

instrumentation/opentelemetry-instrumentation-grpc/tests/test_aio_client_interceptor_hooks.py
herin049: Can we add some unit tests that test `_safe_invoke` in isolation?

herin049 (changes requested):

instrumentation/opentelemetry-instrumentation-grpc/tests/test_aio_client_interceptor_hooks.py
Shriprasad-P: Addressed in 8de84752f5d79a3c10a173c38a2496a5bd0631a: the nearly identical request, response, and exception callable test classes are replaced with one shared RecordingHook helper in tests/_hooks.py, used by both client test modules.

instrumentation/opentelemetry-instrumentation-grpc/tests/test_client_interceptor_hooks.py
Shriprasad-P: Addressed in 8de84752f5d79a3c10a173c38a2496a5bd0631a: sync and asyncio tests still exercise their distinct instrumentation paths explicitly, while the shared RecordingHook and set_span_attribute helpers remove the duplicated hook implementation.

instrumentation/opentelemetry-instrumentation-grpc/tests/test_aio_client_interceptor_hooks.py
Shriprasad-P: Addressed in 8de84752f5d79a3c10a173c38a2496a5bd0631a: added tests/test_safe_invoke.py with isolated coverage for named functions, callable objects without __name__, functools.partial, raising callables with invocation/logging assertions, and a broken __name__ lookup.

instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/_client.py
```diff
@@ -63,9 +63,8 @@ def callback(response_future):
 
 
 def _safe_invoke(function: Callable, *args):
-    function_name = "<unknown>"
     try:
-        function_name = function.__name__
+        function_name = getattr(function, "__name__", "<unknown>")
```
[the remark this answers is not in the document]
Shriprasad-P: Fixed in 8de84752f5d79a3c10a173c38a2496a5bd0631a: function_name is initialized to <unknown> before getattr, so a failing __name__ lookup cannot leave it unbound and mask the original hook failure with UnboundLocalError.

instrumentation/opentelemetry-instrumentation-grpc/tests/test_aio_client_interceptor_hooks.py
```diff
@@ -91,3 +139,100 @@ async def test_hooks_with_exception(self):
...
+            self.assertIn("partial_response_data", span.attributes)
+            self.assertEqual(span.attributes["partial_response_data"], "")
+        finally:
+            instrumentor.uninstrument()
+
+    async def test_callable_hook_with_exception(self):
+        """Test that callable objects that raise exceptions are still invoked once."""
+        instrumentor = GrpcAioInstrumentorClient()
+        request_hook_obj = CallableHookWithException()
+        response_hook_obj = CallableHookWithException()
```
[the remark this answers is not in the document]
Shriprasad-P: Fixed in 8de84752f5d79a3c10a173c38a2496a5bd0631a: the asyncio raising-callable regression now asserts request_hook.calls == 1 and response_hook.calls == 1, and verifies both safe fallback-name error logs.

instrumentation/opentelemetry-instrumentation-grpc/tests/test_client_interceptor_hooks.py
```diff
@@ -120,3 +169,95 @@ def test_hooks_with_exception(self):
...
+            self.assertIn("partial_response_data", span.attributes)
+            self.assertEqual(span.attributes["partial_response_data"], "data")
+        finally:
+            instrumentor.uninstrument()
+
+    def test_callable_hook_with_exception(self):
+        """Test that callable objects that raise exceptions are still invoked once."""
+        instrumentor = GrpcInstrumentorClient()
+        request_hook_obj = CallableHookWithException()
+        response_hook_obj = CallableHookWithException()
```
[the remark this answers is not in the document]
Shriprasad-P: Fixed in 8de84752f5d79a3c10a173c38a2496a5bd0631a: the synchronous raising-callable regression now asserts request_hook.calls == 1 and response_hook.calls == 1, and verifies both safe fallback-name error logs.

Shriprasad-P:
Follow-up: rewritten remaining commits under my GitHub identity for EasyCLA consistency. The earlier review items (shared RecordingHook helper, isolated `_safe_invoke` unit tests, unbound `function_name` fix, and invocation assertions on raising callables) remain in place from 8de8475 / successor rewrite. Please take another look when convenient.

Issue #1704, opened 2023-03-01
[boto3sqs] Instrumetntation might leak memory while handling processing spans


mariojonke:
The instrumentation tracks the processing spans that it creates for received SQS messages in a more or less global dictionary on the [Boto3SQSInstrumentor](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/v0.36b0/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py#L210).
As dictionary key the [receipt handle](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-queue-message-identifiers.html#receipt-handle) of the received SQS messages is used.
Spans/entries are removed from this dictionary by the instrumentation only in one of the following cases:
* the SQS message is deleted with boto3's [sqs.delete_message](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/v0.36b0/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py#L350) API
* the SQS message is deleted with boto3's [sqs.delete_message_batch](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/v0.36b0/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py#L364)
* The same SQS message is received again with boto3's [sqs.receive_message](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/v0.36b0/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py#L330)

According to the [AWS docs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-queue-message-identifiers.html#receipt-handle) the receipt handles of received messages change with every call to `sqs.receive_message`. 
This means that the instrumentation's attempt to remove the span/entry from the dictionary in the instrumenting `sqs.receive_message` function is not working as intended and will leak memory if a message isn't properly deleted, e.g. if an exception happens before the `sqs.delete_message` API can be called.

Another case where memory might get leaked is when `multiprocessing` is used. E.g. SQS messages are received in a parent process and processing + deletion of messages is done in separate child processes.
[...]

PR #5056, opened 2026-09-12
fix(boto3sqs): start processing spans lazily on message access

### LINKED ISSUE #1703: [boto3sqs] Processing spans are started to early
When calling the boto3 `sqs.receive_message` API the instrumenting function already [pre-starts the processing spans](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/v0.36b0/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py#L333) for every received message before the `receive_message` call even returns.  
Since the actual processing of the messages however happens later, somewhere in the user code, the processing spans are started too early. 

**Steps to reproduce**

Run the [test_receive_message](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/v0.36b0/instrumentation/opentelemetry-instrumentation-boto3sqs/tests/test_boto3sqs_instrumentation.py#L263) unit test and debug the [instrumenting `receive_message`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/v0.36b0/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py#L298) function. 
As can be seen there is a processing span pre-started for every received message (in the unit test these spans won't show up until when they are finished since span processor and in-memory exporter only track ended spans).

Processing spans are created when the actual processing of the SQS message happens.

All processing spans are immediately created in the instrumenting `receive_message`

**Additional context**
For the instrumentation it is probably hard, if not impossible, to determine when processing of an SQS message starts or ends as this is determined only by the user code. 

A possible heuristical approach to the problem might be to try starting and ending the processing spans while the list of received messages is iterated. However, since there is no way of knowing how the list is actually used (e.g. random access by index, used in list comprehension to create a slightly modified new list, ...) this would also only work for some certain use-cases and might produce unwanted spans otherwise.

Shriprasad-P:
## Fixes
Fixes #1703

## Summary
Start boto3sqs processing spans lazily on first message access (`__getitem__`) instead of eagerly during `receive_message`, so batch message timings reflect actual processing duration.

## Test plan
- [x] Unit tests updated/passing for span timing behavior
- [x] Existing boto3sqs instrumentation tests pass


chrikrah (commented):
Two things from reproducing this on `31cd3a2`, with a consumer that receives, iterates and deletes.

The receive-to-process parent edge is gone. On `f778897` the receive span and both process spans share one
trace, with receive as their parent. On this branch each process span is the root of its own trace, so one
poll becomes three unrelated traces. The move itself causes it: `_create_processing_span` ran at `:287`
inside `start_as_current_span(f"{queue_name} receive")`, and `_create_processing_span_on_access` now runs at
`:130` in `ContextableList.__getitem__`, after that block has exited. #1703 asks only about timing, the
description mentions only timing, and the one `trace_id` assertion in the tests is on the producer link, so
nothing catches it.

A root plus a link is defensible for a batch consumer under the messaging conventions. If that is what you
intend, it is still a separate change from #1703, so the description should say so and a test should pin it.

The retention also moves rather than ending. `_store_message_metadata` at `:246` writes
`pending_message_metadata[receipt_handle]` holding the whole message dict, and both evictions need that same
handle: `:137` on first access, `:191` in `_safe_end_processing_span`. A consumer that fails before reading a
message never reaches the first, and SQS issues a new handle on every redelivery, so nothing later matches.
Ten redelivery rounds of two unread messages leave twenty entries. Same root cause as #1704.

I have not run the wider suite.

Shriprasad-P:
Thanks for the detailed reproduction. Addressed both findings in 155715d8ec523636b7d6db08c0d46cfb81f7d294.

PROCESS spans still start lazily on first message access, but the returned ContextableList now captures the active RECEIVE context and passes it explicitly to start_span. This preserves the existing RECEIVE → PROCESS parent relationship and trace ID while retaining the producer context as a link.

I also removed the class-level pending_message_metadata cache. Temporary receive metadata is held only by the returned message list, so unread/redelivered messages with new receipt handles no longer accumulate retained message dictionaries.

Regression coverage now covers delayed start timing, RECEIVE/PROCESS parentage, sequential batch timing, unread redeliveries, delete cleanup, and uninstrument cleanup.

.changelog/5056.fixed
```diff
@@ -0,0 +1 @@
+`opentelemetry-instrumentation-boto3sqs`: fix lazy processing span lifecycle
```
chrikrah: `fix lazy processing span lifecycle` names neither the timing change nor the retention change.

```suggestion
`opentelemetry-instrumentation-boto3sqs`: start processing spans on first access and end spans of undeleted messages
```

chrikrah (commented):
Both findings are fixed on `0e1a0d1`. The receive result no longer pickles, and what I predicted on #1703 for a pooled consumer was wrong.

| case | `f778897` (main) | `0e1a0d1` (head) |
|---|---|---|
| receive, iterate, delete | 3 spans, 1 trace, `process` parent is `receive` | same, and a producer `traceparent` stays a link |
| ten rounds of two unread messages | 20 in `received_messages_spans` | 0, and nothing in the module holds them |
| same rounds, read, never deleted | 20 | 1 while the caller holds the lists, 0 once dropped, all 20 `process` spans export |
| `pickle.dumps(response["Messages"])` | OK | `TypeError: cannot pickle '_thread.lock' object` |
| `ThreadPoolExecutor.map`, four 50 ms messages, four workers | 50 to 52 ms each | 0.1 to 0.2 ms for three, 51 ms for the last |

The `TypeError`:

- Holding the state on the list instead of the class is what stops it pickling. `boto3sqs/__init__.py:136` (`_instrumentor`) and `:139` (`_receive_context`) each hold a lock. So does `:141` (`_processing_states`) once the caller has read a message.
- `copy.deepcopy`, `Process(args=(messages,))` under `spawn` or `forkserver`, and `ProcessPoolExecutor.submit` under `fork` fail the same way. `Pool.map(fn, messages)` pickles one dict at a time and works.
- `pickle` walks `__iter__` at `:184` before it raises. Every `process` span but the last exports at 0 ms, and the last stays open as the current span.
- The consumer in #1704 hits this whenever the batch reaches the child by pickle. `spawn` is the default on macOS and Windows, `forkserver` on Linux from Python 3.14, and `tox.ini:85` lists 3.14.

Two fixes, either with a `pickle.dumps(response)` assertion. (1) The `__reduce__` below, 21 passed on a copy of `0e1a0d1`. A pickled child then gets a plain list with no `process` spans, which is what `f778897` gives a `spawn` child. (2) Carry `queue_name`, `queue_url` and the receive `SpanContext`, and rebuild the list in the child. That keeps the `receive` parent across the boundary.

```python
def __reduce__(self):
    return (list, (list.copy(self),))
```

The `map` row corrects my claim on #1703 that a non-loop consumer is no worse than today. `map` submits every message from the calling thread, and `:165` ends the active span on the next handle. So a span ends at the next `submit`, not when its worker finishes. The description should say that a pooled consumer gets near-zero `process` spans for every message but the last.

`tox -e py312-test-instrumentation-boto3sqs`: 21 passed. Removing `context=receive_context` at `:296` fails `test_receive_message` and `test_processing_spans_are_isolated_between_contexts`. `test_unread_redeliveries_do_not_retain_message_metadata` pins the `pending_message_metadata` removal. I did not run the other Python versions.

.changelog/5056.fixed
Shriprasad-P: Updated in d3d52cd to:

`opentelemetry-instrumentation-boto3sqs`: start processing spans on first access and end spans of undeleted messages

PR #3743, opened 2025-09-09
[WIP]add commenter_options in trace_integration

chloe-zh97:
# Description

This PR adds support for passing `commenter_options` to the `trace_integration` function and `DatabaseApiIntegration` in the `opentelemetry-instrumentation-dbapi` package. This allows users to customize SQLCommenter behavior (for example, disabling `opentelemetry_values`) when instrumenting DB-API drivers.

Fixes #3726 

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

- Added a new unit test `test_commenter_options_propagation`.
- `tox -e lint` and  `pytest`.

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [x] Followed the style guidelines of this project
- [x] Changelogs have been updated
- [x] Unit tests have been added
- [x] Documentation has been updated

PR #5058, opened 2026-09-13
Add IBM MQ instrumentation for the ibmmq and pymqi clients

### LINKED ISSUE #5059: Instrumentation request: IBM MQ (ibmmq and pymqi clients)

There is no IBM MQ instrumentation in this repo, so Python services that talk to MQ produce
no messaging spans at all. If a request comes in over HTTP and the handler puts a message on
a queue, the trace just stops at the put. You get no producer span, no consumer span on the
other side, and no record of which queue or which queue manager was involved.

I checked this rather than assuming it. Running the stock distro with
`opentelemetry-bootstrap -a install` against a real MQ broker gives zero spans for MQ calls,
while a `requests` call in the same process is traced normally, so it is genuinely a missing
instrumentation and not a misconfiguration on my side.

There are two clients in common use and neither is covered. IBM's official `ibmmq` package
does about 1.3M downloads a month on PyPI, and the older `pymqi` does about 64k. IBM states
`ibmmq` supersedes `pymqi` and is API compatible with it.

MQ is also a bit awkward compared to Kafka or RabbitMQ in one specific way that makes the
missing coverage more annoying than usual: queue manager names are not unique across hosts,
so even if you log the queue and queue manager name yourself you cannot reliably tell two
queue managers apart in a federated setup. The queue manager identifier, which is unique, is
only reachable from inside the process holding the connection, so this is not something you
can add at the collector or work around in application code without writing your own
wrappers.

A new `opentelemetry-instrumentation-ibmmq` package that traces both clients with no
application code change.

Wrap points:

* `QueueManager.connect` and `QueueManager.connect_with_options`, to read the queue manager
  identifier once per connection
* `Queue.put` and `QueueManager.put1` for producer spans
* `Queue.get` for consumer spans
* `Queue.cb` and `QueueManager.cb` for the async MQCB callback path, which only `ibmmq` has

`put1` needs its own wrapper because it sits on `QueueManager` and does not go through
`Queue.put`, so a Queue-only wrap set misses MQPUT1 completely. `put_rfh2`, `get_rfh2` and
`get_no_jms` are deliberately left alone since they call `put()` and `get()` internally and
wrapping them too would double count.

Both clients present the same call surface so the wrappers are shared. They differ in only
two places, getting from a Queue to its QueueManager and reading a queue's name, and duck
typing covers that. `instruments-any` declares either client as sufficient, the same pattern
`opentelemetry-instrumentation-cassandra` uses for cassandra-driver and scylla-driver.

Spans carry `messaging.system`, `messaging.operation` and `messaging.destination.name`, and
are named `<queue> publish` / `<queue> receive` with PRODUCER and CONSUMER kinds. The queue
manager identifier goes on as `messaging.ibmmq.queue_manager.id`, gated behind
`OTEL_PYTHON_IBMMQ_EXPERIMENTAL_SPAN_ATTRIBUTES` and off by default while the convention is
unratified.

Two behaviours I think are worth agreeing on up front, because both are easy to get wrong:

A `get` that fails with `MQRC_NO_MSG_AVAILABLE` should produce no span. Any consumer using
`MQGMO_WAIT` gets that reason on every idle poll and treats it as routine, so spanning it
would bury a backend in empty receive spans for an idle consumer.

A browse is only an option bit on the same `get()` call, not a separate API, so a browse has
to stay distinguishable from a real consume. Otherwise the data claims an application consumes
from a queue when it only looked at it.

I have opened #5058 with the implementation, tested against a live MQ 9.4.5 broker as well as
under unit tests, so there is something concrete to look at rather than just a proposal.

A standalone package on PyPI, outside this repo. Works, but it would not be picked up by
`opentelemetry-bootstrap`, so users would have to know it exists and install it by hand. That
is most of the value of auto-instrumentation gone.

Adding the instrumentation inside IBM's `ibmmq` client itself. It already ships a small
OpenTelemetry module for trace context propagation, so there is a precedent. Two problems:
it would not cover `pymqi` users at all, and the release cadence would be IBM's rather than
this repo's.

Doing it in the collector. Not possible for the part that matters. There is no IBM MQ
receiver that generates spans, and the collector has no connection to a queue manager, so it
has no way to know which queue manager a given application's span belongs to. That value only
exists in process.

Leaving it to application code. This is what people do today, and it is the thing I am trying
to remove. It means every team writing the same wrappers around `put` and `get`, getting the
2033 and browse cases wrong in slightly different ways.

### Additional Context

Implementation is up as #5058.

On the spec question in the template header: the instrumentation itself is Python specific,
but the one new attribute is not, so I opened
open-telemetry/semantic-conventions#4081 for
`messaging.ibmmq.queue_manager.id` and an `ibmmq` value for `messaging.system`. It is still
open. Until it lands, that attribute stays behind the experimental flag, which is what the
Java and .NET agents do with the same attribute.

The same work is in flight in the sibling repos, so this is not a Python-only idea:

* Java: open-telemetry/opentelemetry-java-instrumentation#19896
* Node.js: open-telemetry/opentelemetry-js-contrib#3743, with instrumentation request
  open-telemetry/opentelemetry-js-contrib#3742

On CI cost, since new instrumentations often bring some: none needed here. The tests are pure
`unittest.mock` with fake client modules injected into `sys.modules`, matching how the other
messaging instrumentations in this repo are tested. No broker, no service container, and not
even the native MQ client library, so nothing new gets added to the test matrix.

On maintenance, I am happy to be listed as component owner for this package and to stay on it
for the bug reports and client updates. I have a live MQ environment to reproduce against,
which is usually the hard part with this library.

Yes

### Tip

Aryan16138:
# Description

This adds a new instrumentation package, `opentelemetry-instrumentation-ibmmq`, for IBM MQ.
It produces producer and consumer spans without the application having to change anything.

What gets wrapped:

* `QueueManager.connect` and `QueueManager.connect_with_options`, to read the queue manager
  identifier once per connection
* `Queue.put` and `QueueManager.put1` for producer spans
* `Queue.get` for consumer spans
* `Queue.cb` and `QueueManager.cb` for the async MQCB callback path

`put1` needs its own wrapper because it lives on `QueueManager` and does not go through
`Queue.put`, so a Queue-only wrap set misses MQPUT1 entirely. Going the other way,
`put_rfh2`, `get_rfh2` and `get_no_jms` are left alone on purpose, since they call `put()`
and `get()` internally and wrapping them too would give you a duplicate nested span for
every operation. Same reasoning for `connect_tcp_client`: both clients implement it by
calling `connect_with_options`, so wrapping both would do the identifier lookup twice per
connect.

It supports both IBM's official `ibmmq` client and the older `pymqi`. They expose the same
call surface so the wrappers are shared. There are only two places they differ, getting from
a Queue to its QueueManager and reading a queue's name, and those are handled by duck typing.
`instruments-any` declares either one as sufficient, the same way
`opentelemetry-instrumentation-cassandra` handles cassandra-driver and scylla-driver.

A few decisions I should explain, since they are not obvious from the diff:

A `get` that comes back with `MQRC_NO_MSG_AVAILABLE` produces no span. Any consumer using
`MQGMO_WAIT` gets that reason on every idle poll and treats it as routine, so emitting spans
for it would bury a backend in empty receive spans. You cannot know that has happened until
after the call returns, so the span is started before the call and just never ended on that
path, which means a span processor never sees it.

Browse bits get read off the MQGMO before the call rather than after. Both clients unpack the
returned MQGMO back over the object you passed in, so reading it afterwards is not reliable.
This matters more than it sounds: a browse is only an option bit on the same `get()`, not a
separate call, so getting it wrong means claiming an app consumes from a queue when it only
looked at it.

Spans stay current across the wrapped call. The `ibmmq` client reads the current span itself
when it attaches message trace context, so creating the span after the fact would leave that
context on the wrong span.

There is no trace context propagation in here. `ibmmq` already handles injection and
extraction on the producer, sync consumer and async callback paths, and doing it again would
double up.

`messaging.system` uses the semconv key with a literal `"ibmmq"` value, since
`MessagingSystemValues` has no IBM MQ member.

`messaging.ibmmq.queue_manager.id` and `messaging.ibmmq.browse` are not ratified conventions
yet, so neither shows up unless `OTEL_PYTHON_IBMMQ_EXPERIMENTAL_SPAN_ATTRIBUTES` is set,
which is how the Java and .NET agents handle the same attribute. Reading the queue manager
identifier is an MQINQ network round trip, so with the flag off it is skipped rather than
resolved and discarded.

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/5059

## Type of change

# How Has This Been Tested?

The unit tests need no broker. They build fake `ibmmq` and `pymqi` modules with
`unittest.mock` and put them in `sys.modules`, so no queue manager and no native MQ client
library is required to run them. That is how the other messaging instrumentations here are
tested, and it means CI needs no new service container or download step.

pytest instrumentation/opentelemetry-instrumentation-ibmmq/tests
22 passed, 5 subtests passed

The ones that matter most are the zero-spans-on-2033 case, browse detection reading the
pre-call MQGMO, the identifier being looked up exactly once per connection no matter how many
puts and gets follow, and the flag being off by default. There is also coverage for
uninstrument putting the original functions back, all three forms `put1` accepts for its
destination, the pymqi name-mangled attribute path, and a queue that was never opened
producing a span with no destination instead of blowing up.

tox entries are registered for `py3{10,11,12,13,14}` against wrapt 1.x and 2.x, plus pypy3
and a lint env.

I also ran it against a real broker: IBM MQ 9.4.5.0, three queue managers, driving an
unmodified `pymqi` sample app under stock auto-instrumentation with no changes to the app at
all. Producer and consumer spans came out named `DEV.QUEUE.1 publish` and
`DEV.QUEUE.1 receive` with the MQ field padding trimmed off the destination. Talking to three
different queue managers from one process gave three distinct identifier values, which is the
check that the value is genuinely per connection and not cached somewhere global. Sixteen
seconds of idle polling against an empty queue produced nothing. A browse `get` and a
destructive `get` on the same queue were correctly told apart, and the browse marker survived
the client overwriting the MQGMO. I confirmed the consumer span really was current while
MQGET was running. With the flag unset you still get spans, just no `messaging.ibmmq.*` on
them. After `uninstrument()` nothing is produced and no identifier lookup happens.

The gap: the async MQCB path has unit tests but I have not driven it against a live broker,
because the sample app I have does not register a callback. That is the part of this I would
most like eyes on.

`ruff check` and `ruff format` are clean and `pylint` with the repo `.pylintrc` gives 10.00/10.

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-ibmmq/src/opentelemetry/instrumentation/ibmmq/utils.py
```diff
@@ -0,0 +1,446 @@
...
+from opentelemetry.trace.span import Span
+
+_LOG = getLogger(__name__)
+
+# messaging.ibmmq.queue_manager.id is not a ratified semantic convention yet.
+# Gate it behind this env var, default off, same as the Java/.NET agents do
+# for the identical attribute.
+_EXPERIMENTAL_ENV_VAR = "OTEL_PYTHON_IBMMQ_EXPERIMENTAL_SPAN_ATTRIBUTES"
+_ATTR_QMID = "messaging.ibmmq.queue_manager.id"
+_ATTR_BROWSE = "messaging.ibmmq.browse"
```
thompson-tomo, line 22:
```suggestion
_ATTR_QMID = "ibm.mq.queue_manager.id"
_ATTR_BROWSE = "ibm.mq.browse"
```
As per semconv naming rules

instrumentation/opentelemetry-instrumentation-ibmmq/src/opentelemetry/instrumentation/ibmmq/utils.py
```diff
@@ -0,0 +1,446 @@
...
+def _enrich_span(
+    span: Span,
+    destination: str | None,
+    operation: str,
+    qmid: str | None,
+    browse: bool = False,
+) -> None:
+    # MessagingSystemValues is a closed enum with no IBM MQ member, so a
+    # literal string is intentional here.
+    span.set_attribute(messaging_attributes.MESSAGING_SYSTEM, "ibmmq")
```
thompson-tomo, line 190:
```suggestion
    span.set_attribute(messaging_attributes.MESSAGING_SYSTEM, "ibm.mq")
```
As per https://opentelemetry.io/docs/specs/semconv/general/naming/#choosing-a-system-name

instrumentation/opentelemetry-instrumentation-ibmmq/src/opentelemetry/instrumentation/ibmmq/utils.py
Aryan16138: Thanks for pointing to naming.md here, you're right that ibm.mq/ibm.mq.* is the technically compliant pairing per the root-namespace rule. Keeping messaging.ibmmq/messaging.ibmmq.* anyway: three sibling instrumentations already ship that form (Java #19896, .NET #5404, Node.js #3743), and keeping the QMID join key's shape consistent across all four languages matters more here than the naming.md technicality. Open to discussing further if that reasoning doesn't hold up for this repo.

instrumentation/opentelemetry-instrumentation-ibmmq/src/opentelemetry/instrumentation/ibmmq/utils.py
Aryan16138: Same reasoning as utils.py:22 — keeping ibmmq for messaging.system to match messaging.ibmmq.* and the other three language instrumentations.

Issue #5061, opened 2026-09-14
[boto3sqs] delete_message_batch returns inside its loop, so it handles the first entry only


chrikrah:
### Describe your environment

- `opentelemetry-instrumentation-boto3sqs` 0.65b0, and on `main` at `b4ba6ca`
- Python 3.12.3, boto3 and botocore 1.34.44
- No AWS account needed: the reproduction mocks the botocore endpoint

### What happened?

`delete_message_wrapper_batch` puts its `return` inside the `for` loop, so the loop can only ever run
one iteration and every entry after the first is unreachable:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/b4ba6ca49/instrumentation/opentelemetry-instrumentation-boto3sqs/src/opentelemetry/instrumentation/boto3sqs/__init__.py#L305-L311

Three consequences, measured on `b4ba6ca`. A consumer that polls the maximum of ten messages and
deletes them in one call sees all three:

1. Every entry after the first keeps an unended `process` span, and an unended span is never exported.
   That batch of ten exports 1 of the 10.
2. Each of those spans stays in `received_messages_spans`, which is a class attribute that
   `_uninstrument` does not clear. Eleven such polls leave 99 entries.
3. The context token that `ContextableList.__getitem__` attached is never detached, so the next span
   started in that thread is parented to a span nothing exports. In the reproduction below, an
   unrelated `start_as_current_span` after the batch delete gets the retained `rh-2` span as its
   parent. With the `return` dedented it has no parent.

### Steps to Reproduce

```
pip install opentelemetry-instrumentation-boto3sqs opentelemetry-sdk 'boto3==1.34.44'
python repro.py
```

```python
from unittest import mock

import boto3
from botocore.awsrequest import AWSResponse
from opentelemetry import trace
from opentelemetry.instrumentation.boto3sqs import Boto3SQSInstrumentor
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import SimpleSpanProcessor
from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter

exporter = InMemorySpanExporter()
provider = TracerProvider()
provider.add_span_processor(SimpleSpanProcessor(exporter))
trace.set_tracer_provider(provider)
Boto3SQSInstrumentor().instrument(tracer_provider=provider)

[...]

PR #5062, opened 2026-09-14
fix(boto3sqs): end the processing span of every entry in delete_message_batch

chrikrah:
# Description

Move the `return` in `delete_message_wrapper_batch` out of the `for` loop. It was inside it, so the loop
could only ever run one iteration and every entry after the first was unreachable. The sibling
`delete_message` wrapper (`__init__.py:295-299`) guards the same way and then calls `wrapped` once,
outside any loop, which is where this `return` belongs.

On `b4ba6ca` a consumer that polls the maximum of ten messages and deletes them in one call exports 1 of
the 10 `process` spans. The nine unended spans stay in `received_messages_spans`, a class attribute
`_uninstrument` does not clear, and the context token of the last message read is never detached, so the
next span started in that thread is parented to a span nothing exports.

Dedenting also restores the call for an empty batch: with `Entries=[]` the loop body never ran, so
`wrapped` was never called and the caller got `None` instead of botocore's return value. Uninstrumented,
the request goes out, and against the service an empty `Entries` is `EmptyBatchRequest`. So a caller
passing an empty batch today gets `None` and no request; after this it gets botocore's `ClientError`.
That is the one caller-visible change here, and it has its own changelog fragment.

The fix ends a span for every entry the caller requested. That is what this wrapper already does for the
first entry, and what `send_batch_wrapper` does for a whole send batch: neither consults `Successful`.
Reading `Successful` instead would be a behaviour change, and the `return` has to move out of the loop
under either answer.

For an entry SQS reports under `Failed`, that ends the span and drops its `received_messages_spans` key,
so the redelivery starts a fresh span under the new receipt handle instead of leaving the old one behind.
That is the leak #1704 reports, and today it happens to every entry after the first whether SQS accepted
it or not.

#5056 is open on the same file and leaves the production line this pull request moves byte-identical at
its head `d3d52cd`, so the one-line change applies either way. The test files do conflict: #5056 removes
`current_context_token`, which the token assertion below uses, so whichever lands second needs that
assertion rewritten against the new mechanism. I will do that rebase.

Refs #5061, which stays open for the unguarded `entries = kwargs.get("Entries")` at `:306`: omitting
`Entries` raises `TypeError` from inside the instrumentation where botocore raises
`ParamValidationError`. That changes an exception type, so it wants its own commit.

## Type of change

# How Has This Been Tested?

Two regression tests, both of which fail with the production change reverted and the tests kept:

- [x] `test_delete_message_batch_ends_every_entry`: two messages received, iterated, deleted in one batch
  call. Asserts both `process` spans, both message ids, an empty `received_messages_spans`, and a detached
  context token, so a span started afterwards has no parent. Reverted:
  `AssertionError: Lists differ: ['MyQueue process', 'MyQueue process'] != ['MyQueue process']`. Deleting
  the detach block instead of the dedent fails the token assertion: `AssertionError: <Token ...> is not None`.
- [x] `test_delete_message_batch_calls_the_api_for_an_empty_batch`: `Entries=[]` reaches botocore.
  Reverted: `AssertionError: unexpectedly None`.

```
tox -e 'py3{10,11,12,13,14}-test-instrumentation-boto3sqs,pypy3-test-instrumentation-boto3sqs'
    18 passed in each of the six environments, against 16 on main
tox -e lint-instrumentation-boto3sqs      rated at 10.00/10
tox -e ruff                               all five hooks passed
tox -e typecheck                          OK
```

The reproduction in #5061 prints 2, `[]` and `False` with this change; on `b4ba6ca` it prints 1,
`['rh-2']` and `True`.

Not run: `tox -e docs`, which wants system `mysql-client`, `libpq-dev` and `libsnappy-dev`.
`opentelemetry-docker-tests` has no SQS suite, so nothing there covers this change.

- [x] No.

# Checklist:

  `.changelog/5062.fixed` and `.changelog/5062.changed`, pushed just after opening because `changelog.yml`
  requires the name to match this number.

  No documented behaviour changes. The package's docs cover installation and say nothing about the
  processing span lifecycle.


JoeyPatricio (approved):
Pulled the branch and ran the boto3sqs suite. The two new tests fail on main with exactly the assertion described (only the first process span is ended, and an empty batch returns None) and pass here, 18/18 with the pinned boto3. Dedenting the return is clearly the right fix and matches the sibling delete_message wrapper. LGTM.

chrikrah:
The `CI` and `changelog` runs on `5bd9a9d` both sit at `action_required`, so no test has run on this pull request. The three checks that did are EasyCLA, `label` and Auto Assign Owners.

@xrmx, you made the last change to `boto3sqs/__init__.py`, in #5019. Either of two things would move this along.

Approving the queued workflow runs. `boto3sqs` has no entry in `.github/component_owners.yml`, so Auto Assign Owners tagged nobody to do it.

A review. The production change is one line: `delete_message_wrapper_batch` returned inside its `for` loop, so every entry after the first was unreachable. The branch is 4 files, +45 and -1, and 42 of the added lines are the two regression tests. `CONTRIBUTING.md:244` asks for two approvals from Approvers or Maintainers at different companies, and this has none of those yet.

Rebased on `3e21539`. On that base, locally:

```
$ python -m pytest instrumentation/opentelemetry-instrumentation-boto3sqs/tests/ -q
18 passed in 1.16s
```

Issue #1158, opened 2022-06-21
Metrics instrumentation dbapi


srikanthccv:
DB metrics: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/database-metrics.md

rblejter:
Can you assign this issue to me?

avzis:
Hi, can you assign this to me?
Thanks!

srikanthccv:
Feel free to ask any questions here or on slack.

tammy-baylis-swi:
I think this issue is still relevant. The db semconv for metrics is now at: https://github.com/open-telemetry/semantic-conventions/blob/main/model/db/metrics.yaml

I think DB-API is generating these metrics currently:
* `db.client.operation.duration`
* `db.client.response.returned_rows`

But does not yet generate these:

1. `db.client.connection.count`
2. `db.client.connection.idle.max`
3. `db.client.connection.idle.min`
4. `db.client.connection.max`
5. `db.client.connection.pending_requests`
6. `db.client.connection.timeouts`
7. `db.client.connection.create_time`
8. `db.client.connection.wait_time`
9. `db.client.connection.use_time`

DB-API instrumentation is a sort of core util used by other instrumentors such as SqlAlchemy and psycopg(2). I think all of the above are relevant for each "concrete" framework instrumentation but whoever works on this should double-check responsibility.

ahmed5145:
Picking this up after the recent re-triage. Thanks @tammy-baylis-swi for the breakdown. Before writing any code I want to nail down who owns what, since that responsibility question is the whole thing here.

Where `dbapi` core stands today: it already emits the two operation-level metrics.

- `db.client.operation.duration`
- `db.client.response.returned_rows`

The 9 metrics still on the list are all connection-pool metrics: `connection.count`, `idle.max`, `idle.min`, `max`, `pending_requests`, `timeouts`, `create_time`, `wait_time`, `use_time`. Raw DB-API (PEP 249) has no pool. It wraps a single `connect()` and its cursors. There is nothing in the core util that can observe a checkout, an overflow, or a timeout, so I don't think these can come from `dbapi` at all.

So my read on the split:

- `dbapi` core stays at the operation level, which is already done. The pool metrics don't belong here.
- Pool and connection metrics come from the concrete pooled instrumentations (SQLAlchemy, psycopg, and so on), where an actual pool exists to watch.

[...]

PR #5063, opened 2026-09-14
sqlalchemy: modernize db.client.connection.count metric

### LINKED ISSUE #5060: sqlalchemy: support semconv opt-in of `db.client.connection.count` metric

Noted in https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1158#issuecomment-5623559405:

> One concrete starting point: SQLAlchemy already reports connection usage, but under the deprecated `db.client.connections.usage` name (connections_usage) rather than the current `db.client.connection.count`. Migrating that to `connection.count` with `state=idle/used`, then adding the other pool metrics SQLAlchemy can see from its pool events, is a well-scoped first task.

It would be consistent if support of current/stable [metric connection count](https://github.com/open-telemetry/opentelemetry-python/blob/f3278f2043338536665c1f212bd7d72d220a5d34/opentelemetry-semantic-conventions/src/opentelemetry/semconv/_incubating/metrics/db_metrics.py#L9-L23) (away from deprecated [metric usage](https://github.com/open-telemetry/opentelemetry-python/blob/f3278f2043338536665c1f212bd7d72d220a5d34/opentelemetry-semantic-conventions/src/opentelemetry/semconv/_incubating/metrics/db_metrics.py#L252-L264)) was also an opt-in with `OTEL_SEMCONV_STABILITY_OPT_IN ` like the span attributes, introduced in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4110.

Update opentelemetry-instrumentation-sqlalchemy to emit `db.client.connections.usage` and/or `db.client.connection.count`.

_No response_

### Additional Context

_No response_

_No response_

### Tip

_No response_

ahmed5145:
## Description

Per the maintainer decision in #1158, this drops the `OTEL_SEMCONV_STABILITY_OPT_IN` approach for this metric. SQLAlchemy now always emits `db.client.connection.count` (attributes `db.client.connection.state` = idle/used, `db.client.connection.pool.name`), replacing the deprecated `db.client.connections.usage`.

The opt-in env var was only intended for span-attribute migration and doesn't apply to metrics, so there's no default-old or `database/dup` behavior: the modern metric is emitted unconditionally.

**Breaking change:** consumers of `db.client.connections.usage` will no longer receive it. This is expected per the discussion in #1158.

Fixes #5060

## Type of change

## How Has This Been Tested?

- [x] `tox -e py312-test-instrumentation-sqlalchemy-2` (61 passed, 7 skipped) on Windows / Python 3.12
- [x] `tox -e py312-test-instrumentation-sqlalchemy-1` (68 passed) on Windows / Python 3.12
- [x] Updated `test_sqlalchemy_metrics.py` to assert `db.client.connection.count` with the modern attribute names/value types, and that the deprecated `db.client.connections.usage` is no longer emitted

- [x] No.

## Checklist:


instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
```diff
@@ -42,6 +42,13 @@
 Configuration
 -------------
 
+Connection pool metrics
+***********************
+Connection pool usage is reported as the ``db.client.connection.count``
+metric, with the ``db.client.connection.pool.name`` and
+``db.client.connection.state`` (``idle``/``used``) attributes. This replaces
+the deprecated ``db.client.connections.usage`` metric.
```
tammy-baylis-swi: Thanks for consider the readthedocs!

Could this be a bit more general and so we can update when more metrics are added? Could be something like this (probably needs a lint). I'd say no need to mention the replacement since the changelog will bring us back to this PR.


```suggestion
Database client metrics
***********************
Connection pool usage is reported as the ``db.client.connection.count``
metric, with the ``db.client.connection.pool.name`` and
``db.client.connection.state`` (``idle``/``used``) attributes.

See the database semantic conventions for more information.
```

tammy-baylis-swi (approved):
Thank you @ahmed5145 , code lgtm based on discussion. I've left a comment about the doc.

This will also have to be reviewed by Maintainers.

instrumentation/opentelemetry-instrumentation-sqlalchemy/src/opentelemetry/instrumentation/sqlalchemy/__init__.py
ahmed5145: Good idea, and done

PR #5065, opened 2026-09-15
fix(tortoiseorm): avoid IndexError in _do_execute for empty/whitespace-only statements

JoeyPatricio:
# Description

Same class as #4934 / #4936 / #4942 / #4943, in the tortoiseorm instrumentation's `_do_execute`. The span name is taken as the first token of the statement with an unguarded `args[0].split()[0]`. An empty or whitespace-only statement has no tokens, so the wrapper raises `IndexError` and the user's query never reaches the driver. Without the instrumentor the same call succeeds.

Guards it the same way the dbapi base does (`tokens[0] if tokens else ""`).

Repro on `main` (tortoise-orm 1.1.8 and the pinned 0.20.0):

```python
TortoiseORMInstrumentor().instrument()
await Tortoise.init(db_url="sqlite://:memory:", modules={"models": ["__main__"]})
await Tortoise.get_connection("default").execute_query("   ")
```

```
  File ".../opentelemetry/instrumentation/tortoiseorm/__init__.py", line 268, in _do_execute
    name = args[0].split()[0]
IndexError: list index out of range
```

Only raw SQL passed through the `execute_*` client methods is affected; ORM-generated SQL is never empty.

No open issue for this one; found while checking for siblings of #4934.

## Type of change

# How Has This Been Tested?

Added `test_span_name_empty_or_whitespace_statement`. It fails on `main` with `IndexError` and passes with the fix. `tox -e py313-test-instrumentation-tortoiseorm` and `tox -e lint-instrumentation-tortoiseorm` pass.

- [x] No.

# Checklist:


instrumentation/opentelemetry-instrumentation-tortoiseorm/src/opentelemetry/instrumentation/tortoiseorm/__init__.py
```diff
@@ -265,7 +265,10 @@ async def _do_execute(self, func, instance, args, kwargs):
             return await func(*args, **kwargs)
 
         exception = None
-        name = args[0].split()[0]
+        # A statement that is empty or whitespace-only has no tokens; must not
+        # raise IndexError. Fall back to an empty operation name instead.
+        tokens = args[0].split()
+        name = tokens[0] if tokens else ""
```
henry3260, line 272: I think we should add a fallback like `dbapi` do something like

```
if not name:
    name = (
        getattr(instance, "database", None)
        or getattr(instance, "filename", None)
        or getattr(getattr(instance, "capabilities", None), "dialect", "")
    )
```

henry3260 (approved):
overall looks great just one nit

instrumentation/opentelemetry-instrumentation-tortoiseorm/src/opentelemetry/instrumentation/tortoiseorm/__init__.py
JoeyPatricio: Good call, applied it in the latest commit and updated the test to expect :memory: for the sqlite case. Thanks for the review.

Shriprasad-P (commented):
Thanks for the PR. Looks focused from a first pass — please confirm tests cover the new path and CI is green.

JoeyPatricio:
Both already in place on the current head: \test_span_name_empty_or_whitespace_statement\ exercises an empty and a whitespace-only statement and asserts each span gets the \:memory:\ fallback name instead of raising, and CI is green on c72a1594.

tammy-baylis-swi (approved):
Thanks for another one

PR #1966, opened 2023-09-22
improve semantic conventions for confluent kafka producer

oleksii-valuiskyi:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [ ] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [ ] Followed the style guidelines of this project
- [ ] Changelogs have been updated
- [ ] Unit tests have been added
- [ ] Documentation has been updated

PR #5066, opened 2026-09-15
fix(confluent-kafka): align producer and consumer spans with semantic conventions

### LINKED ISSUE #1674: ProxiedConsumer is creating spans each time it polls instead of only when it receives a message

mviamari-crexi:
# Description

Supersedes #4297, which could not be reopened after the branch was squashed and rebased.

This is a fix for some long standing issues with the ConfluentKafka span creation, particularly around `consume` and `poll`. In those cases, because the trace context isn't known until after the message is received (the context is extracted from the headers), any spans created are orphaned spans. So when the span is created for the poll operation, there is no parent trace. Then any spans created during message processing are children of that orphaned span, despite the span links, rather than part of the trace provided by the header context.

This is fixed in both cases by not creating a span for the poll or consume operations. According to the chart at the end of this semantics doc https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/kafka.md, the poll and consume operation spans would nominally be created by the client, rather than the producer or consumer. There are some semantics around what constitutes the client in this case, but the poll and consumer network operations are handled by librdkafka for the confluent-kafka producer and consumer.

On the producer side, this also fixes the `messaging.operation` value on the produce span and ensures the partition is extracted and recorded when supplied.

Attribute names are unchanged. Migrating to the current messaging semantic conventions is out of scope and can follow #4974.

Fixes #1966
Fixes #1674
Fixes #1678

## Type of change

# How Has This Been Tested?

Added unit tests

- [x] Unit Tests w/ Tox

- [x] No.

# Checklist:


Shriprasad-P (commented):
Thanks for the PR. Looks focused from a first pass — please confirm tests cover the new path and CI is green.

mviamari-crexi:
> Thanks for the PR. Looks focused from a first pass — please confirm tests cover the new path

I added three new tests.
* test_poll_no_message asserts that an empty poll produces zero spans.
* test_producer_with_partition checks the partition attribute lands on the produce span.
* test_extract_produce_partition covers the new extractor in test_utils.py.

Modifications to existing tests:
* test_poll and test_consume previously expected a recv span in the output. That's the behavior this PR modifies, so those checks were removed, and the expected process-span attribute dictionaries now include offset and partition, so those mappings are asserted directly.

* Four existing tests already cover cleanup. Additionally, the empty-poll and empty-consume tests check that no span is created and that the previous span and context token are released.

> and CI is green.

The CI and changelog workflows are pending "action required". It looks like a maintainer has to approve those workflows for CI to be fully green. The checks that do not need approval are green.

PR #5075, opened 2026-09-18
Make azure app service and azure functions resource detector process dependent

rads-1996:
# Description

Mark the Azure App Service and Azure Functions resource detectors as process-dependent so they are rerun during post-fork resource refresh, which only selects detectors where `is_process_dependent()` returns True. This ensures their Azure instance identity can retain precedence over the process-dependent `ServiceInstanceIdResourceDetector`, which otherwise generates and merges a new UUID in the child process.

## Type of change

# How Has This Been Tested?

- [ ] Unit tests have been updated, and E2E application tests have been run

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:


herin049:
If I'm using a HTTP server library like Gunicorn, won't this result in multiple worker processes having the same service instance id (assuming that the `WEBSITE_INSTANCE_ID` environment variable is identical in each worker process)?

rads-1996:
> If I'm using a HTTP server library like Gunicorn, won't this result in multiple worker processes having the same service instance id (assuming that the `WEBSITE_INSTANCE_ID` environment variable is identical in each worker process)?

Yes, for the azure resource detector we want that if a user has supplied the WEBSITE_INSTANCE_ID then that would be the the service instance id for the child as well as parent process. That was how it was working before the new `ServiceInstanceIdResourceDetector` logic was added.

herin049:
> > If I'm using a HTTP server library like Gunicorn, won't this result in multiple worker processes having the same service instance id (assuming that the `WEBSITE_INSTANCE_ID` environment variable is identical in each worker process)?
> 
> Yes, for the azure resource detector we want that if a user has supplied the WEBSITE_INSTANCE_ID then that would be the the service instance id for the child as well as parent process. That was how it was working before the new `ServiceInstanceIdResourceDetector` logic was added.

The issue with that approach is that it would violate the semantics/requirements around `service.instance.id`. Namely, that it must be unique for every `service.namespace`, `service.name` pair. Looking at existing semantic conventions, I think in this case it is more appropriate to map `WEBSITE_INSTANCE_ID` to the `host.id` attribute -- it seems to exactly capture the semantics that `WEBSITE_INSTANCE_ID` represents:

> Unique host ID. For Cloud, this must be the `instance_id` assigned by the cloud provider. 

Alternatively, I think adding a new Azure-specific attribute could also be appropriate.

rads-1996:
> > > If I'm using a HTTP server library like Gunicorn, won't this result in multiple worker processes having the same service instance id (assuming that the `WEBSITE_INSTANCE_ID` environment variable is identical in each worker process)?
> > 
> > 
> > Yes, for the azure resource detector we want that if a user has supplied the WEBSITE_INSTANCE_ID then that would be the the service instance id for the child as well as parent process. That was how it was working before the new `ServiceInstanceIdResourceDetector` logic was added.
> 
> The issue with that approach is that it would violate the semantics/requirements around `service.instance.id`. Namely, that it must be unique for every `service.namespace`, `service.name` pair. Looking at existing semantic conventions, I think in this case it is more appropriate to map `WEBSITE_INSTANCE_ID` to the `host.id` attribute -- it seems to exactly capture the semantics that `WEBSITE_INSTANCE_ID` represents:
> 
> > Unique host ID. For Cloud, this must be the `instance_id` assigned by the cloud provider.
> 
> Alternatively, I think adding a new Azure-specific attribute could also be appropriate.

So one thing to note here, this change would run the resource detectors for the child processes which would then have the same instance id as that of the parent which is desirable and the correct behavior.

For a new instance, the service.instance.id will always be different and unique from a separate instance and that is what the spec also states - 

image

jeremydvoss:
@herin049 This spec entry's requirement is about horizontal scaling: 'e.g. instances of a horizontally scaled service'. An 'instance' is a separate machine (or VM) running another instance of an app to handle separate requests and ensure redundancy. That's different than separate processes on the _same app instance_. Each subprocess should have the same instance id just like subprocesses inherit env vars. That is why https://github.com/open-telemetry/opentelemetry-python/pull/5259 broke the existing azure set up which already was spec compliant:
1. Every instance has a unique instance ID
2. No 2 instances shared the same service.* triplet.
3. Root process and subprocesses on a given instance all shared the same instance ID.

We need this PR to restore Service Instance ID functionality for multiprocess use cases like Gunicorn/Uvicorn which are very common among out customers.

UPDATE:

I see later on, the spec _recommends_ different instance ids for subprocesses:
```
For applications running behind an application server (like unicorn), we do not recommend using one identifier for all processes participating in the application. Instead, it’s recommended each division (e.g. a worker thread in unicorn) to have its own instance.id.
```
 So, I now understand the intent of the original PR. However, this is just a recommendation, not a requirement. Therefore, Azure should be able to restore the original behavior in our custom detectors. It does not affect other customers. If we add per-process instance ids in the future, it would have to be based on the existing relationship between existing App Service env vars and service instance ID rather than a random guid from the new service instance id detector.

herin049:
> @herin049 This spec entry's requirement is about horizontal scaling: 'e.g. instances of a horizontally scaled service'. An 'instance' is a separate machine (or VM) running another instance of an app to handle separate requests and ensure redundancy. That's different than separate processes on the _same app instance_. Each subprocess should have the same instance id just like subprocesses inherit env vars. That is why [open-telemetry/opentelemetry-python#5259](https://github.com/open-telemetry/opentelemetry-python/pull/5259) broke the existing azure set up which already was spec compliant:
> 
> 1. Every instance has a unique instance ID
> 2. No 2 instances shared the same service.* triplet.
> 3. Root process and subprocesses on a given instance all shared the same instance ID.
> 
> We need this PR to restore Service Instance ID functionality for multiprocess use cases like Gunicorn/Uvicorn which are very common among out customers.
> 
> UPDATE:
> 
> I see later on, the spec _recommends_ different instance ids for subprocesses:
> 
> ```
> For applications running behind an application server (like unicorn), we do not recommend using one identifier for all processes participating in the application. Instead, it’s recommended each division (e.g. a worker thread in unicorn) to have its own instance.id.
> ```
> 
> So, I now understand the intent of the original PR. However, this is just a recommendation, not a requirement. Therefore, Azure should be able to restore the original behavior in our custom detectors. It does not affect other customers. If we add per-process instance ids in the future, it would have to be based on the existing relationship between existing App Service env vars and service instance ID rather than a random guid from the new service instance id detector.

Thanks @rads-1996 and @jeremydvoss for the additional context here and yes -- https://github.com/open-telemetry/opentelemetry-python/pull/5259 was implemented specifically because the semconv recommends this approach:

> For applications running behind an application server (like unicorn), we do not recommend using one identifier for all processes participating in the application. Instead, it’s recommended each division (e.g. a worker thread in unicorn) to have its own instance.id.

The primary motivation we introduced it into Python as well is to make working with uvicorn/gunicorn slightly easier since `service.instance.id` alone is now enough to differentiate metric streams from different worker processes.

herin049 (approved):

jeremydvoss (approved):

PR #4864, opened 2026-07-24
fix(opentelemetry-instrumentation-logging): Promote `otel.event.name` to `LogRecord.event_name` in `_translate`

rads-1996:
# Description

Ported over from #4750 

`LoggingHandler._translate` never set the first-class `event_name` field on
`LogRecord`. When a user passed otel.event.name (or the deprecated event.name)
via extra={...}, it silently landed in attributes as a plain key and was
exported as a regular OTLP attribute instead of the dedicated event_name field.

This fix updates `_get_attributes` to pop both keys from the attributes dict and
return the resolved event name alongside the attributes. _translate then passes
it as event_name= to LogRecord(...).

`otel.event.name` (stable) takes precedence
`event.name` (deprecated) is used only as a fallback
Both keys are always popped so neither leaks into attributes

Fixes https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4743

Part of https://github.com/open-telemetry/community/issues/1751

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Unit tests have been added

# Does This PR Require a Core Repo Change?

- [ ] Yes. - Link to PR: 
- [x] No.

# Checklist:

See [contributing.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md) for styleguide, changelog guidelines, and more.

- [x] Followed the style guidelines of this project
- [ ] Changelogs have been updated
- [x] Unit tests have been added
- [ ] Documentation has been updated

Issue #4643, opened 2026-06-01
opentelemetry-instrumentation-logging JSON Formatter compatibility with spec


tailg8nj:
The spec for log formatting, specifically JSON has an example:

https://opentelemetry.io/docs/specs/otel/compatibility/logging_trace_context/#json-formats

Which seems to reinforce the overview of logging Trace ID and Span ID in legacy log formats:

> To summarize, the following field names should be used in legacy formats:
>
> “trace_id” for TraceId, lowercase and hex-encoded.
> “span_id” for SpanId, lowercase and hex-encoded.
> “trace_flags” for trace flags, formatted according to W3C traceflags format.

However, the log record that is created with the opentelemetry-instrumentation-logging has fields named `otelTraceID` and `otelSpanID`, etc.:

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/26c975f2aea142ef198c1a4b377d7cb71bbbbd57/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/__init__.py#L203-L205

This doesn't rear it's head with standard text formatters as the default logging format pulls the appropriate log record attributes out explicitly: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/26c975f2aea142ef198c1a4b377d7cb71bbbbd57/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/constants.py#L4

JSON formatters traditionally just take what they are given and spit them out, see reference implementation: [python-json-logger](https://github.com/nhairs/python-json-logger/blob/01cb83c1be1f26eb9f75cd3233b270fc06dd7962/src/pythonjsonlogger/core.py#L102).

I believe the attributes placed onto the log record should change to snake case, and the default text format changed to leverage them appropriately.

rajdhruvsingh:
Hi, I'm interested in working on this. Before implementing,
I wanted to clarify the migration strategy:

1. The current attribute names (otelSpanID, otelTraceID, 
   otelTraceSampled, otelServiceName) are documented and
   widely used in log format strings. Renaming them is a
   breaking change.

2. Would the preferred approach be:
   a) Add new spec-compliant names alongside existing ones
      (both work, old ones deprecated with warning)?
   b) Use OTEL_SEMCONV_STABILITY_OPT_IN style opt-in?
   c) Only change in next major version?

3. Should DEFAULT_LOGGING_FORMAT be updated to use new names?

[...]

PR #5077, opened 2026-09-19
fix(logging): stop duplicating injected trace context in log attributes

### LINKED ISSUE #5071: Logging: don't include injected trace context attributes in emitted records

When setting up opentelemetry-instrumentation-logging and using either the `set_logging_format` or `inject_trace_context` arguments, the instrumentor sets up a log record factory such that new `logging.LogRecord` objects have `otelSpanID`, `otelTraceID`, and `otelTraceSampled` attributes. The log handler set up by `logging.basicConfig()` or by a user’s on custom logging can then use those attributes when formatting messages. That’s useful, although the names are a little weird for Python (see also #4643).

However, when the [`opentelemetry.instrumentation.logging.handler.LoggingHandler` handler](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/3e21539d0ebef675630b0afce3a2401e24468c6b/instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/handler.py#L120-L124) emits an OTel `LogRecord`, it winds up including these as attributes alongside the standardized `trace_id`, `span_id`, and `flags` fields, which is both duplicative and confusing! When I tried replacing the deprecated log handling from `opentelemetry.sdk._logs` with this instrumentor, I started seeing these attributes in the monitoring service I send these logs to and thought something was broken because of their names. Eventually I realized they were just duplicative, extra information.

For example, the log data I get when exporting from this instrumentor look like this in the protobuf (using `protoc` to unpack the bytes sent to the collector):

```
resource_logs {
  resource {
    // ...a bunch of attributes describing my app...
  }
  scope_logs {
    scope {
      name: "django.request"
    }
    log_records {
      time_unix_nano: 1789677011375527168
      severity_number: SEVERITY_NUMBER_ERROR
      severity_text: "ERROR"
      flags: 3
      trace_id: "q\267\352\003\2269\326\326a\271\214\364/\325\014?"
      span_id: "\010:\336L\177\363Ix"
      observed_time_unix_nano: 1789677011378342341
      body {
        string_value: "Internal Server Error: /.test-error/"
      }
      attributes {
        key: "otelSpanID"
        value {
          string_value: "083ade4c7ff34978"                 // <-- Duplicative of span_id
        }
      }
      attributes {
        key: "otelTraceID"
        value {
          string_value: "71b7ea039639d6d661b98cf42fd50c3f" // <-- Duplicative of trace_id
        }
      }
      attributes {
        key: "otelTraceSampled"
        value {
          bool_value: true                                 // <-- Duplicative of flags (I think?)
        }
      }
      // ...other attributes my app sets...
    }
  }
}
```

It would be nice if these extra, duplicative attributes were not included in the `opentelemetry._logs.LogRecord` emitted by `opentelemetry.instrumentation.logging.handler.LoggingHandler`.

You can avoid this by setting `set_logging_format=False` and `inject_trace_context=False`, but then this information is not available for other log handlers (e.g. the stderr logger set up by `logging.basicConfig()`).

### Additional Context

_No response_

None

### Tip

RichardoMrMu:
# Description

`LoggingInstrumentor` injects `otelSpanID`, `otelTraceID`, `otelTraceSampled` and
`otelServiceName` onto every stdlib `LogRecord` when trace context injection is
enabled, so formatters can render them. `LoggingHandler._get_attributes` then
copies every non-reserved `vars(record)` entry into the emitted `LogRecord`, so
those four are exported as attributes as well — even though the emitted record
already carries the same context in its first-class `trace_id`, `span_id` and
`trace_flags` fields, and the service name in the resource.

The result is that every exported log record carries the trace context twice,
once under spec-defined fields and once under these camelCase attribute names.

This pops the four injected names in `_get_attributes`, following the convention
already established there for `event_name` in #4864: a value that belongs in a
first-class field is removed from the attribute dict in that same function.

The attributes remain on the stdlib `LogRecord`, so formatters and
`DEFAULT_LOGGING_FORMAT` are unaffected. Only the exported copy is dropped. That
distinction matters — a previous change that stopped injecting them onto the
record itself broke users who format them, as reported on #3687.

This is a narrower change than the rename discussed in #4643: it does not alter
any attribute name or anything a formatter sees.

Fixes #5071

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

# How Has This Been Tested?

- [x] `pytest instrumentation/opentelemetry-instrumentation-logging/tests/` — 75 passed, no regressions (71 pre-existing plus the 4 added here)
- [x] Added four tests to `test_handler.py`:
  - `test_injected_trace_context_not_duplicated_in_attributes` — none of the four names appear in the exported record's attributes
  - `test_first_class_trace_context_still_populated` — `trace_id` and `span_id` still match the active span, so nothing was lost
  - `test_injected_attributes_remain_on_the_stdlib_record` — the attributes are still on the stdlib record for formatters
  - `test_unrelated_extra_attributes_unaffected` — an unrelated `extra` attribute is still exported
- [x] Mutation check: reverting only the `pop` loop makes `test_injected_trace_context_not_duplicated_in_attributes` and `test_unrelated_extra_attributes_unaffected` fail, so the tests do detect the regression they describe
- [x] `ruff check` and `ruff format --check` clean on both changed files, using the rule set from the root `pyproject.toml`
- [x] `pylint` rated 10.00/10 on the changed handler, identical to the unmodified file from `main`

- [ ] Yes. - Link to PR:
- [x] No.

# Checklist:

- [x] Changelogs have been updated — `.changelog/5077.fixed`

---

Disclosure, per [AGENTS.md](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/AGENTS.md): this change was prepared with AI assistance. The behaviour was reproduced and the fix verified locally against the real instrumentor and handler before opening, and the reasoning above reflects the source as it stands on `main`.


instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/handler.py
```diff
@@ -139,6 +149,14 @@ def __init__(
     def _get_attributes(self, record: logging.LogRecord) -> tuple[dict[str, AnyValue], str | None]:
         attributes = {k: v for k, v in vars(record).items() if k not in _RESERVED_ATTRS}
 
+        # The LoggingInstrumentor injects the trace context onto every stdlib
+        # LogRecord so formatters can render it. Formatters still need it there,
+        # but the emitted LogRecord already carries the same context in its
+        # first-class trace_id, span_id and trace_flags fields, and the service
+        # name in the resource, so keeping the copies here only duplicates them.
+        for _injected in _TRACE_CONTEXT_ATTRS:
```
herin049, line 157: Nit: instead of popping these off after construction, we should just include this condition in construction on line 150.

instrumentation/opentelemetry-instrumentation-logging/tests/test_handler.py
```diff
@@ -561,6 +561,94 @@ def test_logging_handler_without_env_var_uses_default_limit(self):
...
+        LoggingInstrumentor().instrument(inject_trace_context=True)
+        processor, logger, handler = set_up_test_logging(logging.WARNING)
+        try:
+            tracer = trace.TracerProvider().get_tracer(__name__)
+            with self.assertLogs(level=logging.WARNING):
+                with tracer.start_as_current_span("test"):
+                    logger.warning("Warning message")
+
+            record = processor.get_log_record(0)
+            for name in (
```
herin049, line 581: nit: just use `_TRACE_CONTEXT_ATTRS` directly here.

instrumentation/opentelemetry-instrumentation-logging/tests/test_handler.py
```diff
@@ -561,6 +561,94 @@ def test_logging_handler_without_env_var_uses_default_limit(self):
...
+        processor, logger, handler = set_up_test_logging(logging.WARNING)
+        try:
+            tracer = trace.TracerProvider().get_tracer(__name__)
+            with self.assertLogs(level=logging.WARNING):
+                with tracer.start_as_current_span("test"):
+                    logger.warning("Warning message", extra={"http.status_code": 200})
+
+            record = processor.get_log_record(0)
+            self.assertEqual(record.log_record.attributes["http.status_code"], 200)
+            self.assertNotIn("otelSpanID", record.log_record.attributes)
```
herin049, line 647: If we're adding this assertion, we should check all attributes in `_TRACE_CONTEXT_ATTRS` for completeness.

instrumentation/opentelemetry-instrumentation-logging/tests/test_handler.py
```diff
@@ -561,6 +561,94 @@ def test_logging_handler_without_env_var_uses_default_limit(self):
...
+        logger.setLevel(logging.WARNING)
+        capture = CaptureHandler()
+        logger.addHandler(capture)
+        try:
+            tracer = trace.TracerProvider().get_tracer(__name__)
+            with tracer.start_as_current_span("test"):
+                logger.warning("Warning message")
+
+            for name in ("otelSpanID", "otelTraceID", "otelTraceSampled"):
+                self.assertIn(name, seen)
```
herin049, line 630: Can we make this assertion also validate that the correct span id, trace id, etc. are in the log output?

RichardoMrMu:
Thanks for the review — all four addressed, and the CLA is now signed so the check is green.

**Construction instead of popping.** Folded the condition into the comprehension on the line you pointed at, so there is one dict built once and no post-hoc mutation:

```python
attributes = {
    k: v
    for k, v in vars(record).items()
    if k not in _RESERVED_ATTRS and k not in _TRACE_CONTEXT_ATTRS
}
```

**`_TRACE_CONTEXT_ATTRS` directly.** Both the duplication test and the unrelated-extras test now iterate the tuple instead of repeating the names, so adding a fifth injected attribute updates all of them at once.

**All attributes, for completeness.** `test_unrelated_extra_attributes_unaffected` now asserts every name in the tuple is absent, not just `otelSpanID`.

**Validating the values.** `test_injected_attributes_remain_on_the_stdlib_record` now checks the ids match the active span rather than merely being present:

```python
self.assertEqual(seen["otelSpanID"], format(span_context.span_id, "016x"))
self.assertEqual(seen["otelTraceID"], format(span_context.trace_id, "032x"))
self.assertEqual(seen["otelTraceSampled"], span_context.trace_flags.sampled)
```

That assertion earned its place immediately: I first wrote the service-name expectation as `"unknown_service"` and it failed with `'' != 'unknown_service'`. The instrumentor reads `service.name` off the provider's resource and falls back to `""` when there is none (`__init__.py` L178-182), and this test's `TracerProvider()` has no resource — which is exactly the expectation `test_logging.py` already encodes at L55 and L262. So the test now asserts `""` with a comment explaining why, rather than freezing the observed value.

I also took Copilot's point about `instrument()` leaking the monkey-patched `logging.basicConfig`/`dictConfig`/`fileConfig` functions, since `uninstrument()` does not restore them. All four tests now pass `enable_log_auto_instrumentation=False`, which is what `test_logging.py` in this same package already does — they only need the `LogRecordFactory` injection. `otelServiceName` is covered as it suggested, via the shared tuple.

Verified: 4 passed for the issue-5071 tests, 46 passed for `test_handler.py`, 75 passed across the package's tests, `ruff check` and `ruff format --check` clean.

instrumentation/opentelemetry-instrumentation-logging/src/opentelemetry/instrumentation/logging/handler.py
RichardoMrMu: Done in 57b4996 — the condition is now part of the comprehension itself (`handler.py` line 156):

```python
attributes = {
    k: v for k, v in vars(record).items()
    if k not in _RESERVED_ATTRS and k not in _TRACE_CONTEXT_ATTRS
}
```

The separate removal loop is gone.

instrumentation/opentelemetry-instrumentation-logging/tests/test_handler.py
RichardoMrMu: Done in ee140be — all three call sites now iterate `_TRACE_CONTEXT_ATTRS` (test lines 582, 626 and 654); the literal tuple is no longer repeated anywhere in the file.

instrumentation/opentelemetry-instrumentation-logging/tests/test_handler.py
RichardoMrMu: Done in ee140be — both exclusion assertions now loop over the whole tuple rather than a subset:

```python
for name in _TRACE_CONTEXT_ATTRS:
    self.assertNotIn(name, record.log_record.attributes)
```

at lines 582 and 654, and the presence check at 626 uses the same loop, so adding a name to the tuple extends all three.

instrumentation/opentelemetry-instrumentation-logging/tests/test_handler.py
RichardoMrMu: Done in ee140be. Two layers now check values rather than presence.

On the emitted record, the first-class fields are compared against the active span (lines 600-601):

```python
self.assertEqual(record.log_record.trace_id, span_context.trace_id)
self.assertEqual(record.log_record.span_id, span_context.span_id)
```

On the stdlib record, the injected copies are compared in their rendered form (lines 631-633), since presence alone would pass even if a formatter rendered placeholder context:

```python
self.assertEqual(seen["otelSpanID"], format(span_context.span_id, "016x"))
self.assertEqual(seen["otelTraceID"], format(span_context.trace_id, "032x"))
self.assertEqual(seen["otelTraceSampled"], span_context.trace_flags.sampled)
```

One note on the fourth name: `otelServiceName` is asserted as `""` rather than a service name, because this tracer provider carries no resource. That matches the existing expectation in `test_logging.py` for a provider without one, so I kept it consistent instead of introducing a resource just for this test.

instrumentation/opentelemetry-instrumentation-logging/tests/test_handler.py
```diff
@@ -561,6 +561,94 @@ def test_logging_handler_without_env_var_uses_default_limit(self):
...
 
+    # --- trace context duplication tests (issue #5071) ---
+
+    def test_injected_trace_context_not_duplicated_in_attributes(self):
+        """The injected otel* attributes stay off the emitted LogRecord.
+
+        The context is already carried by the first-class trace_id, span_id and
+        trace_flags fields, so exporting the copies duplicates it.
+        """
+        LoggingInstrumentor().instrument(inject_trace_context=True)
```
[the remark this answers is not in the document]
RichardoMrMu: Each of the four tests already restores the global state: the body runs inside
`try` and the `finally` calls `LoggingInstrumentor().uninstrument()` — lines 584, 602, 640
and 658 on the current branch. So the root-logger handler and the `basicConfig` patch are
both undone even when an assertion fails, and nothing leaks into the tests that follow.

That is also why the four tests can each call `instrument(...)` with different arguments:
`test_injected_trace_context_not_duplicated_in_attributes` and
`test_first_class_trace_context_still_populated` use `inject_trace_context=True`, while
`test_unrelated_extra_attributes_unaffected` needs its own instance, and none of them can
observe another's handler.

instrumentation/opentelemetry-instrumentation-logging/tests/test_handler.py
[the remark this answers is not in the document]
RichardoMrMu: `otelServiceName` is asserted, at line 637:

```python
self.assertEqual(seen["otelServiceName"], "")
```

It is checked for its real value rather than only for presence. The empty string is not an
oversight: with no `Resource` on the provider the attribute falls back to `""`, and that is
the value `test_logging.py` already expects at its lines 55 and 262. I ran a probe before
writing the assertion because my first guess was `"unknown_service"`, which the code does
not produce.

The surrounding test asserts all three injected attributes with their actual values rather
than just their keys — `otelSpanID` and `otelTraceID` are compared against the ids formatted
from the active span, so a change in the formatting would fail the test too.

PR #5081, opened 2026-09-20
fix(asgi): record response size per request instead of per middleware instance

RichardoMrMu:
### Description

`OpenTelemetryMiddleware` stored the response `Content-Length` on the
middleware instance as `self.content_length_header`, but one middleware
instance serves every request. The value is written in `otel_send`
(`__init__.py` L970 on main) and read back in `__call__`'s `finally` block
(L811), so a request that reaches the read without having written picks up
whatever an earlier request left behind.

Two consequences, both reproduced against unmodified `main`:

**Responses with no `Content-Length`.** Streaming, chunked and 204 responses
never write, but the `finally` block still reads. Two sequential requests, the
first with `Content-Length: 1234` and the second streaming, produce a
`http.server.response.size` total of `2468` — the second request reports the
first one's size. A fresh middleware given only the streaming request records
nothing, which is the correct result.

**Concurrency.** With two requests in flight, a request that yields after
sending its response (background work, cleanup, any remaining `await`) has its
value overwritten before its own `finally` runs. In that interleaving a request
that sent `Content-Length: 100` recorded `9999`, the other request's size.

### Fix

The size is now held in a dict created per request in `__call__` and passed
into `_get_otel_send`. This is the idiom already used in that same function
for `expecting_trailers`, and it matches how the mirror-image *request* size
is handled twenty lines away as a plain local.

`self.content_length_header` is deliberately left in place and still
initialised to `None`. It has no leading underscore, so removing it would be
an unrelated breaking change inside a bug fix; it is simply no longer read for
the metric.

### Type of change

### How Has This Been Tested?

- Added `test_response_size_not_carried_into_next_request` to
  `tests/test_asgi_middleware.py`, using the existing `AsyncAsgiTestBase`
  harness and a path-dispatching app so one middleware serves both requests,
  exactly as `starlette`/`fastapi` install it.
- Against unmodified source that test fails with `AssertionError: 2 != 1` and
  **all 93 pre-existing tests still pass** — the failure is confined to the
  defect path.
- With the fix: `94 passed, 11 subtests passed`.
- Ran the `starlette` and `fastapi` suites both before and after the change
  and compared the failure sets: **no test changes state except the new one**.
  (Those two suites have some failures in my local environment from missing
  entry-point installation; they are identical before and after, so they are
  environmental rather than caused by this change.)
- `ruff format --diff` reports both files already formatted; `ruff check`
  passes with the repository configuration.

### Does This PR Require a Contrib Repo Change?

- [x] No.

### Checklist:

- [x] Changelog has been updated


instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py
```diff
@@ -967,7 +971,7 @@ async def otel_send(message: dict[str, Any]):
             content_length = asgi_getter.get(message, "content-length")
             if content_length:
                 try:
-                    self.content_length_header = int(content_length[0])
+                    response_size_holder["content_length"] = int(content_length[0])
```
herin049, line 973: Can we instead use a solution that relies on context vars instead?

RichardoMrMu: Good question — I considered context vars, but I think a request-local holder is the lighter fit here, for a couple of reasons:

- **The state never leaves this coroutine.** `response_size_holder` is created in `__call__`, written inside the `otel_send` callback, and read in the same `__call__`'s `finally`. All of that runs in one ASGI call frame, so the dict is already per-request-isolated by virtue of being a local. Context vars buy us cross-frame / cross-task propagation, which this path doesn't need — and in return we'd have to manage `set`/`reset` (token) lifetimes and make sure the reset happens on every exit path, which is more moving parts than the bug warrants.
- **Symmetry with request size.** The request-size branch a few lines up already does exactly this with a plain local (`request_size_amount`), read inline in `__call__`. Response size only lived on the instance because it's written from the `otel_send` closure rather than inline — that indirection is precisely what caused the leak. Moving it to a per-request local brings the two sides back to the same, simple scoping model.

If you'd still prefer a `ContextVar` for consistency with how the SDK carries other per-request state, I'm happy to switch — just let me know and I'll push that version.

Separately, I pushed a small follow-up: since the fix makes `self.content_length_header` dead, I removed its initializer so no stale per-request state remains on the shared instance.

PR #5083, opened 2026-09-22
Fix asyncio docs: correct metric name and use asyncio.run in future example

bourbonkk:
# Description

Docs-only fix for the `opentelemetry-instrumentation-asyncio` module docstring (also the published module docs):

- The "asyncio metric types" section listed `asyncio.process.count (count)`, but the counter actually created in `_instrument` is `asyncio.process.created` with unit `{process}`. Dashboards built from the docs query a metric that does not exist.
- The future example used `asyncio.get_event_loop()` without a running loop, which emits a `DeprecationWarning` since Python 3.12. It now uses `asyncio.run()` like the other two examples.

No code change. A changelog fragment is included so the required check passes; happy to drop it in favour of the `Skip Changelog` label if you prefer to keep docs fixes out of the changelog.

## Type of change

# How Has This Been Tested?

- Compared the docstring against the instruments created in `AsyncioInstrumentor._instrument`.
- Ran the updated future example with `python -W error::DeprecationWarning`; it completes without warnings.

- [x] No.

# Checklist:

- [ ] Unit tests have been added (docs-only)


instrumentation/opentelemetry-instrumentation-asyncio/src/opentelemetry/instrumentation/asyncio/__init__.py
```diff
@@ -35,12 +35,12 @@ async def main():
 
     AsyncioInstrumentor().instrument()
 
-    loop = asyncio.get_event_loop()
```
lzchen, line 38: Could you make the same fix to the [readme](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-asyncio/README.rst)?

bourbonkk: Good catch, the README had the same two issues. Fixed in dbf4cb14

lzchen (approved):

xrmx (approved):

